Issue #36, opened 2026-02-06
⚡ Bounty: Multi-Node Sync Protocol — 3-Node Consensus Hardening (100 RTC)


Scottcjn:
## Bounty: Multi-Node Sync Protocol

**Reward: 200 RTC** (~$20 at reference rate)

### Background
RustChain runs 3 attestation nodes (50.28.86.131, 50.28.86.153, 76.8.228.245:8099). Each maintains its own SQLite database. We need a proper sync protocol.

### Requirements

1. **Database sync protocol** between all 3 nodes:
   - Sync `miner_attest_recent` (attestations)
   - Sync `balances` (RTC balances)
   - Sync `epoch_rewards` (reward history)
   - Sync `ledger` (transactions)
2. **Conflict resolution:** Latest timestamp wins for attestations, primary node authoritative for epochs
3. **Sync mechanism:** HTTP API on existing port 8099
4. **Verification:** Merkle root of synced tables
5. **Monitoring:** `/sync/status` endpoint showing last sync time per peer

### Security
- Admin key required in header for all sync requests
- HTTPS only
- Rate limit: Max 1 sync per minute per peer
- Reject sync that would reduce any balance

### Deliverables
- `rustchain_sync.py` — Sync module
- Updated node code with sync endpoints
- Test script validating 3-node consistency
- PR to this repo

### Claiming
Comment with approach and timeline. Major bounty — thorough review expected.

createkr:
Claiming the Multi-Node Sync Protocol bounty! ⚡\n\n### Approach\n1. **Core Sync Module**: Implement `rustchain_sync.py` to handle bidirectional SQLite synchronization.\n2. **Conflict Resolution**: Implement 'Latest Timestamp Wins' for attestations and 'Primary Node Authoritative' for epochs.\n3. **Sync API**: Add authenticated HTTP endpoints (`/api/sync/pull`, `/api/sync/status`) to the node server.\n4. **Verification**: Implement Merkle root generation for state consistency checks.\n5. **Security**: Mandatory admin key validation, HTTPS enforcement, and balance reduction rejection.\n6. **Validation**: Deliver a test script to simulate and verify 3-node consistency.\n\n### Timeline\n- Development: 48-72 hours\n- Testing & Validation: 24 hours\n\n- Miner ID: createker02140054RTC

createkr:
[...]

PR #39, opened 2026-02-06
docs: add API reference, miner setup guide, and Python SDK tutorial (Bounty #72)

Goku1lL:
Claiming bounty #72.

Added comprehensive documentation:
- API Reference
- Miner Setup Guide
- Python SDK Tutorial

Solana Wallet: `ANS4tfQVQWrmVQbapcyUYRV5TSsPNWvyyFXa9cw2PzH6`

PR #40, opened 2026-02-06
feat: add Python SDK (Bounty #36)

Goku1lL:
Claiming bounty #36.

Implemented `rustchain-sdk`:
- Sync & Async clients (`RustChainClient`, `AsyncRustChainClient`)
- Type-safe models
- `setup.py` for pip installation
- Unit & Integration tests

Solana Wallet: `ANS4tfQVQWrmVQbapcyUYRV5TSsPNWvyyFXa9cw2PzH6`

### DIFF
--- a/rustchain-sdk/rustchain/__init__.py
+++ b/rustchain-sdk/rustchain/__init__.py
@@ -1,0 +1,17 @@
+from .client import RustChainClient, AsyncRustChainClient
+from .models import (
+    NodeHealth, EpochInfo, MinerInfo, WalletBalance, 
+    TransferResult, AttestChallenge, AttestResult
+)
+
+__all__ = [
+    'RustChainClient',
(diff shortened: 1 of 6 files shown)


Scottcjn:
Closing - bounty #36 is for 'Multi-Node Sync Protocol — 3-Node Consensus Hardening', not a Python SDK. The deliverable does not match what the bounty requests.

Additional issues found:
- Async `check_eligibility` has a bug: `return await self.get_miner(miner_id) is not None` — due to Python operator precedence, this evaluates `await (bool)` which raises TypeError
- Unit tests would fail: mock patching doesn't intercept the client's internal httpx instance
- 3 of 9 files are duplicates from PR #39
- No `__init__.py` in tests directory

A Python SDK is something we'd consider as a bounty in the future, but it needs to actually work and claim the correct bounty.

PR #134, opened 2026-02-12
Fix(security): prevent unauth pubkey overwrite in /withdraw/register

David-code-tang:
Fixes rustchain-bounties#141.

- /withdraw/register was unauthenticated and could overwrite existing miner withdrawal pubkeys via upsert.
- This enables withdrawal takeover (attacker registers their pubkey for victim miner_pk).

Fix:
- First-time registration allowed.
- If a pubkey is already registered and differs, require admin key to rotate; otherwise return 409.
- Use request.get_json(silent=True) and validate dict to avoid 500s on malformed JSON.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -2105,7 +2105,9 @@ def reject_v1_mine():
 @app.route('/withdraw/register', methods=['POST'])
 def register_withdrawal_key():
     """Register sr25519 public key for withdrawals"""
-    data = request.get_json()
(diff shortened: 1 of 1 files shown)


Scottcjn:
Thanks for the security report + patch.

We landed a minimal fix on `main` (commit `654b7b1`) that prevents unauthenticated pubkey rotation in `/withdraw/register` (admin key required for rotation).

This PR appears to be based on an older `main` and would regress/remove newer security + payout functionality (e.g. pending_ledger / payout_preflight). Because of that we can’t merge it as-is.

Closing as superseded; please rebase on current `main` and submit a minimal diff if you want to follow up with any additional hardening.

PR #140, opened 2026-02-13
security: harden SDK TLS + tipbot key-at-rest + confirm flows

liu971227-sys:
Implements hardening items raised in rustchain-bounties#69:\n\n- SDK: default TLS verification on; optional CA bundle path\n- Tipbot: TLS verify on by default (+ optional CA bundle), encrypt custodial keys at rest (SecretBox), Decimal-based amount parsing, destination validation, and /confirm_tip + /confirm_withdraw flows\n- Docs: update SDK/tipbot READMEs to reflect new security defaults\n\nRefs:\n- rustchain-bounties#69

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -1,0 +1,177 @@
+# RustChain Python SDK
+
+A clean, pip-installable Python SDK for the [RustChain](https://github.com/Scottcjn/Rustchain) Proof-of-Antiquity blockchain.
+
+## Installation
+
+```bash
+# From GitHub
+pip install git+https://github.com/Scottcjn/Rustchain.git#subdirectory=sdk
+
+# Or clone and install locally
+git clone https://github.com/Scottcjn/Rustchain.git
+cd Rustchain/sdk
+pip install -e .
+```
+
+## Quick Start
+
+```python
+from rustchain_sdk import RustChainClient
+
+# Connect to the RustChain network
+client = RustChainClient()
+
+# Check node health
+health = client.health()
+print(f"Node OK: {health['ok']}, Version: {health['version']}")
+
+# Get current epoch info
+epoch = client.get_epoch()
+print(f"Epoch {epoch.epoch}, Pot: {epoch.epoch_pot} RTC, Miners: {epoch.enrolled_miners}")
+
+# List active miners
+miners = client.get_miners()
+for miner in miners:
+    print(f"{miner.hardware_type}: {miner.antiquity_multiplier}x multiplier")
+
+# Check wallet balance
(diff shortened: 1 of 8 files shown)


David-code-tang:
Security-focused review (PR #140)

1) /rain uses float parsing, enabling NaN/Inf edge cases
- `total_amount = float(context.args[0])` accepts `nan`/`inf`.
- With NaN, comparisons like `bal < total_amount` are false, so the command can bypass the minimum and balance checks, then iterate transfers until `int(amount_rtc * 1_000_000)` throws.
- This is at least a reliability/DoS vector (lots of nonce increments + API calls) and it makes input validation inconsistent with /tip and /withdraw.
  Suggestion: reuse the Decimal-based parser (`_parse_rtc_amount_to_i64`) for /rain too and reject non-finite values explicitly.

2) Amount conversion in API.transfer_signed uses float -> int
- `amount_i64 = int(amount_rtc * 1_000_000)` depends on float rounding.
- You already do safe parsing via Decimal in the handlers. Consider changing `transfer_signed(...)` to take `amount_i64: int` (or take a Decimal) so the signed message is deterministic and not float-dependent.

3) Nonce / concurrency
- The bot is async; `increment_nonce` has no lock. Two concurrent commands from the same user can race and reuse a nonce (tx failure / confusing UX).
  Suggestion: keep an in-memory per-user asyncio.Lock or use a single lock around nonce updates + writes.

Overall the direction (TLS verify default true + at-rest key encryption + confirmation flows) is good; the above tweaks would close a couple of sharp edges.

David-code-tang (commented):
Good set of changes (TLS verify-on by default + optional CA bundle, Decimal amount parsing, confirm flows, SecretBox at-rest encryption).

A few security/robustness nits worth addressing:

1) Nonce/accounting race: WalletManager.increment_nonce() looks like an in-memory increment; if the bot processes updates concurrently (or restarts between increment and persist), it can reuse nonces. Suggest an explicit lock around wallet load/update/save, or a monotonic nonce persisted atomically.

2) Key derivation domain separation: when BOT_SECRET is set, seed_material is BOT_SECRET:telegram_id. Consider prefixing with a constant like rustchain-tipbot: to avoid accidental cross-app key reuse if the same secret is reused elsewhere.

3) wallets.json file perms: recommend enforcing 0600 when writing data/wallets.json (to reduce accidental key exposure on multi-user hosts).

4) Client verify parameter: verify = ca_bundle_path if ca_bundle_path else verify_ssl is good. Maybe add a guard that ca_bundle_path must exist/readable, so it fails fast with a clear error.

Overall: direction is right; after the nonce/perms details it should be solid.

PR #67, opened 2026-02-10
[BOUNTY] Miner Setup Wizard - Browser-Based Installation Guide (#47)

TheFrogEcliptia:
### [BOUNTY] Miner Setup Wizard (#47)

Interactive browser-based guide to streamline onboarding for new miners.

**Features:**
- Automatic platform detection (Linux, macOS, Raspberry Pi).
- Step-by-step UI for dependency installation and miner configuration.
- Runs as a standalone static HTML file.

Closes #47

PR #143, opened 2026-02-13
payment-widget: fix NaCl bundle + harden against XSS/callback/iframe

liu971227-sys:
Fixes bounty #67 (payment widget hardening).

Key changes
- Fix broken bundled NaCl code (replaced with upstream tweetnacl-js nacl-fast.min.js) so widget parses and runs again.
- Remove DOM XSS sinks: avoid innerHTML for user-controlled content; use textContent / text nodes.
- Validate recipient address format and normalize amount.
- Default-deny iframe embedding unless explicitly allowed (data-allow-iframe=true); optional allowed-origins allowlist.
- Restrict callback URL to same-origin by default; optional allow-any-origin flag.

PoCs (in-repo)
- payment-widget/poc/xss-label.html
- payment-widget/poc/xss-memo.html
- payment-widget/poc/iframe-block.html

Notes
- This PR targets base branch master because payment-widget lives on upstream/master (not main).


David-code-tang:
Security-focused review (PR #143)

1) Encoding/BOM regression
- `payment-widget/rustchain-pay.js` now starts with a UTF-8 BOM (`\ufeff/**` shown as `+﻿/**` in the diff). This can cause tooling/linters/minifiers to behave oddly and also broke a UI string below.
- `fileLabel.textContent` became `"鉁?${file.name}"` instead of the intended checkmark. This will show garbled text for users.
  Suggestion: remove the BOM and keep UI strings ASCII (e.g. `"OK: " + file.name` or `"[loaded] " + file.name`).

2) XSS hardening looks good, but keep it consistent
- Good change: label/memo/to/amount are now written via `textContent`/text nodes, not `innerHTML`.
- `escapeHtml()` is added but (from this diff) not used. Consider removing it to avoid giving a false impression that HTML-escaping is the primary defense.

3) Callback URL policy
- Good change: default callback is same-origin, and URL parsing uses `new URL()`.
- Consider tightening defaults to `https:` only (except localhost) to avoid accidental insecure deployments. If keeping `http:` support, the README should explicitly warn it is for local/dev only.

4) Origin allowlist
- Current `allowedOrigins` compares raw strings to `window.location.origin`. Consider documenting that entries must be *origins* (scheme+host+port), not full URLs. A small normalization step (e.g. trimming trailing slashes) would reduce misconfig.

Overall: direction is solid for DOM injection + callback safety; please fix the BOM/garbled checkmark before merge.

PR #153, opened 2026-02-13
fix(rewards): prevent concurrent /rewards/settle double-credit

David-code-tang:
Context: epoch settlement is currently check-then-act without a DB write lock; concurrent settle attempts can both observe unsettled state and credit rewards twice.\n\nChange:\n- Acquire a SQLite BEGIN IMMEDIATE lock before checking epoch_state.settled and before crediting rewards\n- Roll back cleanly when already settled\n- Make rip_200 imports and Flask dependency optional so offline tooling/tests can import\n\nTest:\n- node/tests/test_rewards_settle_race.py spawns concurrent settle calls and asserts balances/epoch_rewards are credited exactly once\n\nBounty reference: rustchain-bounties#56 (epoch settlement manipulation / settlement race condition)

### DIFF
--- a/node/rewards_implementation_rip200.py
+++ b/node/rewards_implementation_rip200.py
@@ -6,19 +6,51 @@ Replaces VRF lottery with 1 CPU = 1 vote deterministic consensus
 
 import sqlite3
 import time
-from flask import request, jsonify
+import os
+try:
+    from flask import request, jsonify
+except ImportError:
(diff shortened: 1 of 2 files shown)


FlintLeng:
Nice implementation of 'fix(rewards): prevent concurrent /reward'. Code quality meets standards.

PR #365, opened 2026-02-23
Security fix: Protect sensitive endpoints (Issue #409)

Pitrat-wav:
This PR implements authentication and rate limiting for sensitive endpoints (/api/miners, /wallet/balance, /epoch) as requested in issue #409. 

**Changes:**
- Created  with  and  decorators.
- Applied protections to , , and .
- Rate limit set to 30 requests per minute per IP/endpoint.
- Authentication uses the existing .

Valid miners still auto-enroll via  which remains open to facilitate core consensus.

Payout wallet: pitrat (RTC)

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -38,7 +38,7 @@ def test_api_epoch(client):
         mock_cursor = mock_conn.execute.return_value
         mock_cursor.fetchone.return_value = [10]
 
-        response = client.get('/epoch')
+        response = client.get('/epoch', headers={'X-Admin-Key': '0'*32})
         assert response.status_code == 200
         data = response.get_json()
         assert data['epoch'] == 85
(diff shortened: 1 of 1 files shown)


Scottcjn:
Closing — this breaks public API access by requiring admin auth on /epoch, /api/miners, and /wallet/balance. These endpoints are used by the block explorer, miners, and external tools. PR #368 implements the correct approach (tiered disclosure: public gets redacted data, admin gets full data). Also the RateLimiter has unbounded memory growth. Thanks @Pitrat-wav for the security thinking — the right direction but #368 is the better implementation.

Pitrat-wav:
The BCOS Checks failure is due to missing BCOS-L1 or BCOS-L2 label on this PR (non-documentation changes require such labels). Could a maintainer please add the appropriate label? Thanks!

PR #769, opened 2026-03-09
fix: verify merkle_root in BFT _validate_proposal against miners list

### LINKED ISSUE #766: bug: BFT _validate_proposal does not verify merkle_root against miners list
## Summary

`_validate_proposal` in `node/rustchain_bft_consensus.py` validates that the miner IDs in `distribution` exist in `miners`, and that the total reward sums to 1.5 RTC — but it **never re-computes the Merkle root** from the submitted `miners` list and compares it against `proposal["merkle_root"]`.

Because the `merkle_root` is included in `EpochProposal.compute_digest()` (which produces the consensus `digest`), a Byzantine leader can:
1. Submit an honest proposal for epoch N with a correct `merkle_root`.
2. In a future epoch, reuse that `merkle_root` with a *different* (falsified) `miners` list that still passes the miner-ID / total-reward checks.

The consensus digest would be different, but **the proposal itself passes `_validate_proposal`**, meaning honest nodes send PREPARE messages for a proposal with inconsistent attestation data.

## Affected code

`node/rustchain_bft_consensus.py` – `_validate_proposal` (line ~340)

```python
def _validate_proposal(self, proposal: Dict) -> bool:
    ...
    # merkle_root is never verified here!
    return True
```

## Steps to reproduce

```python
bft = BFTConsensus("node-A", ":memory:", "secret")

miners_real = [{"miner_id": "miner-1", "weight": 1.0}]
miners_fake = [{"miner_id": "miner-1", "weight": 1.0}, {"miner_id": "miner-2", "weight": 0.0}]

# Compute merkle root from the REAL miners list
real_merkle = bft._compute_merkle_root(miners_real)

# Build a proposal with FAKE miners but the REAL merkle root
fake_proposal = {
    "epoch": 1,
    "miners": miners_fake,
    "total_reward": 1.5,
    "distribution": {"miner-1": 1.5},
    "proposer": "byzantine-leader",
    "merkle_root": real_merkle,  # stale / mismatched
}

assert bft._validate_proposal(fake_proposal)  # passes! should fail
```

**Expected behavior:** `_validate_proposal` returns `False` when `merkle_root` does not match the hash recomputed from `miners`.

**Actual behavior:** Returns `True` — the forged proposal passes validation.

## Fix

Add a Merkle root re-computation check inside `_validate_proposal`:

```python
expected_root = self._compute_merkle_root(miners)
if proposal.get("merkle_root") != expected_root:
    logging.warning("Proposal merkle_root mismatch")
    return False
```

## Labels

bug, consensus, security

wsimon1982:
Fixes #766

## Problem

`_validate_proposal` in `node/rustchain_bft_consensus.py` checked that miner IDs in `distribution` existed in the `miners` list and that the total reward summed to 1.5 RTC — but it **never re-computed the Merkle root** from the submitted `miners` list and compared it to `proposal["merkle_root"]`.

A Byzantine leader could recycle a valid `merkle_root` from a previous epoch while submitting a falsified `miners` list that still passes the miner-ID / total-reward checks.  Honest nodes would send `PREPARE` messages for a proposal whose attestation data is inconsistent with the digest they are voting on.

## Reproducer

```python
bft = BFTConsensus("node-A", ":memory:", "secret")
miners_real = [{"miner_id": "miner-1", "weight": 1.0}]
miners_fake = [{"miner_id": "miner-1", "weight": 1.0}, {"miner_id": "miner-2", "weight": 0.0}]
real_merkle = bft._compute_merkle_root(miners_real)
fake_proposal = {
    "epoch": 1, "miners": miners_fake, "total_reward": 1.5,
    "distribution": {"miner-1": 1.5}, "proposer": "byzantine-leader",
    "merkle_root": real_merkle,  # stale / mismatched
}
assert bft._validate_proposal(fake_proposal)  # passes before fix — should fail
```

## Solution

Recompute the Merkle root from the `miners` list and assert equality:

```python
expected_merkle = self._compute_merkle_root(miners)
if proposal.get("merkle_root") != expected_merkle:
    logging.warning("Proposal merkle_root mismatch")
    return False
```

### DIFF
--- a/node/rustchain_bft_consensus.py
+++ b/node/rustchain_bft_consensus.py
@@ -737,6 +737,19 @@ class BFTConsensus:
                 logging.warning(f"Miner {miner_id} in distribution but not in miners list")
                 return False
 
+        # Verify merkle_root matches the submitted miners list.
+        # Without this check a Byzantine leader can recycle a valid merkle_root
+        # from a previous epoch while submitting a different (falsified) miners
+        # list, and honest nodes would still send PREPARE for the forged proposal.
+        expected_merkle = self._compute_merkle_root(miners)
+        if proposal.get('merkle_root') != expected_merkle:
+            logging.warning(
+                f"Proposal merkle_root mismatch for epoch {epoch}: "
+                f"got {proposal.get('merkle_root', '')[:16]}... "
+                f"expected {expected_merkle[:16]}..."
+            )
+            return False
+
         return True
 
     # ========================================================================


Scottcjn (approved):
Important consensus security fix. Without this, a Byzantine leader could recycle a valid `merkle_root` from a previous epoch while submitting a falsified miners list — honest nodes would still send PREPARE. The fix correctly recomputes `_compute_merkle_root(miners)` and validates before accepting.

**Verdict: MERGE** — 50 RTC bounty (consensus security hardening).

🤖 Reviewed with GPT-5.4 + Claude Opus dual-brain analysis

FlintLeng:
**Review PR #769** - Bounty #2782

PR #911, opened 2026-03-13
fix: implement /wallet/history with unified transfers + mining rewards (fixes #775, #886, closes #908)

ApextheBoss:
## Summary

Fixes the `/wallet/history` endpoint to return unified transaction history by merging both data sources:

- **`pending_ledger`** — transfers (sent/received), with direction detection
- **`epoch_rewards`** — mining rewards, typed as `"reward"`

## What changed

- Queries both `pending_ledger` and `epoch_rewards` tables
- Merges results into a single list sorted by timestamp (newest first)
- Adds `offset` pagination alongside existing `limit` (cap raised to 500)
- Wraps response in the documented `{ok, miner_id, transactions, total}` envelope
- Unknown `miner_id` returns `{ok: true, transactions: [], total: 0}` (not an error)
- Each transaction now has a `type` field: `transfer_in`, `transfer_out`, or `reward`

## Testing

Three cases covered:
1. **Known wallet with mixed history** — returns transfers and rewards merged
2. **Unknown wallet** — returns `{ok: true, transactions: [], total: 0}`
3. **Pagination** — `offset` + `limit` correctly slice the merged result set

Closes #908, fixes #775, fixes #886.

PR #920, opened 2026-03-13
fix(#908): enhance /wallet/history with offset pagination

szoasis:
## Summary
Implements the requirements from #908

## Changes
- Add `offset` query parameter for pagination
- Change response format to `{ok: true, miner_id, transactions, total}`
- Handle unknown miner_id gracefully (returns empty list)

## Testing
- API returns proper JSON structure
- Pagination works with limit and offset
- Unknown miner_id returns empty transactions array


Scottcjn:
Thanks for the submission! We are going with #911 which best preserves backward compatibility with existing response fields. Your effort is appreciated -- check out other open bounties at https://github.com/Scottcjn/Rustchain/labels/bounty

PR #983, opened 2026-03-15
feat: add Homebrew formula for clawrtc miner

CelebrityPunks:
## Summary

Adds a Homebrew formula so users can install the RustChain miner with:

```
brew tap Scottcjn/rustchain https://github.com/Scottcjn/Rustchain.git
brew install clawrtc
```

- **Formula/clawrtc.rb** - Homebrew formula that installs clawrtc 1.7.1 from PyPI in a Python virtualenv with all dependencies (requests, cryptography, cffi, etc.)
- **Formula/install.sh** - One-liner install script for quick setup
- **Formula/README.md** - Tap usage instructions

### Details

- Uses `virtualenv_install_with_resources` (standard Homebrew Python pattern)
- All dependency URLs and sha256 hashes verified against PyPI
- Depends on `python@3.12`
- Installs the `clawrtc` CLI entry point
- Includes test block and caveats

Resolves Scottcjn/rustchain-bounties#1612

### DIFF
--- a/Formula/README.md
+++ b/Formula/README.md
@@ -1,0 +1,54 @@
+# Homebrew Tap for RustChain Miner
+
+## Install
+
+```bash
+brew tap Scottcjn/rustchain https://github.com/Scottcjn/Rustchain.git
+brew install clawrtc
+```
+
+Or use the one-liner:
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/Scottcjn/Rustchain/main/Formula/install.sh | bash
+```
+
+## Usage
+
+```bash
+# Show help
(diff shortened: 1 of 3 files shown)


Scottcjn:
Closing — `clawrtc` is not a package we publish. The RustChain miner is distributed as Python scripts (`rustchain_universal_miner.py`), not via Homebrew/PyPI. The formula references `clawrtc-1.7.1.tar.gz` which does not exist, and the `curl | bash` install pattern is a security concern for blockchain software. If we ship a Homebrew formula in the future, it would be in a separate tap repo (`homebrew-rustchain`).

PR #1163, opened 2026-03-15
feat: add env dump

CelebrityPunks:
Shows all RustChain-related environment variables.


Scottcjn:
Closing — an env dump tool that prints all RUSTCHAIN/RTC/WALLET/MINER env vars is a security risk. RC_ADMIN_KEY and other sensitive values live in env vars. This would be an exfiltration vector. Thanks for the contribution but we can't merge this.

PR #1740, opened 2026-03-21
Canonicalize stored Beacon hash by ignoring unsigned metadata

Mavline:
## Summary
- ignore transport-only `_beacon_version` metadata when computing the stored Beacon payload hash
- add a regression test proving unsigned `_beacon_version` does not change the anchored hash
- carry forward the small follow-up hygiene changes already on this branch (`PyNaCl` requirements + SPDX header on the new test file)

## Why
`verify_envelope_signature(...)` intentionally ignores `_beacon_version`, matching the Beacon codec, but `hash_envelope(...)` still hashed the full submitted JSON. Because `/beacon/submit` accepts raw JSON bodies, a caller could add `_beacon_version` after signing and still change the anchored digest.

This follow-up keeps signature verification and anchoring hash canonicalization aligned.

## Testing
- `/tmp/rustchain-test-venv/bin/python -m unittest /Users/core/Projects/Rustchain/node/tests/test_beacon_anchor_signature.py`

### DIFF
--- a/node/beacon_anchor.py
+++ b/node/beacon_anchor.py
@@ -24,6 +24,10 @@ except ImportError:
 DB_PATH = "/root/rustchain/rustchain_v2.db"
 
 VALID_KINDS = {"hello", "heartbeat", "want", "bounty", "mayday", "accord", "pushback"}
+REQUIRED_ENVELOPE_FIELDS = ("agent_id", "kind", "nonce", "sig", "pubkey")
+UNSIGNED_TRANSPORT_FIELDS = ("sig", "_beacon_version")
+LEGACY_PAYLOAD_HASH_VERSION = 1
+CURRENT_PAYLOAD_HASH_VERSION = 2
 
 
 def _agent_id_from_pubkey(pubkey_bytes: bytes) -> str:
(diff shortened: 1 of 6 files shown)


Scottcjn:
Correct concept — unsigned metadata causing hash inconsistency across nodes. Before merging:\n1. Define explicit SIGNED_FIELDS set that matches what the signature covers\n2. Migration strategy for existing payload_hash values in beacon_envelopes table\n3. Add tests proving: extra metadata doesn't change hash, changing signed field does change hash\n4. Consider transition period for mixed hash versions across nodes\n\nWill merge once migration + tests are included.

Mavline:
Addressed the requested follow-up in the latest push (`1354cb2`).

1. **Explicit signing/hash contract**
   - Canonical Beacon v2 verification + hashing now uses the exact envelope body covered by the live codec: every top-level field except transport-only `sig` and decoded header metadata `_beacon_version`.
   - I kept this as an explicit canonicalization rule instead of freezing a heartbeat-only field list, so it still matches other Beacon envelope kinds.

2. **Existing `payload_hash` rows / migration strategy**
   - Added `payload_hash_version` tracking.
   - Existing rows are preserved as legacy `payload_hash_version=1`.
   - New writes store the explicit signed-body hash as `payload_hash_version=2`.
   - Because `beacon_envelopes` only stores the derived hash (not the full original envelope body), old rows cannot be recomputed safely in place; they stay tagged legacy until drained/anchored.

3. **Tests added/updated**
   - `_beacon_version` metadata does not change the stored hash.
   - Changing a signed field does change the hash.
   - Legacy rows are upgraded to `payload_hash_version=1`.
   - Mixed pending hash versions are reported by the digest path.

4. **Mixed-version transition**
   - `/beacon/digest` now exposes `payload_hash_versions` and `mixed_payload_hash_versions` so operators can see when legacy + v2 hashes are coexisting during rollout.

Targeted test run:
`/tmp/rustchain-test-venv/bin/python -m unittest node/tests/test_beacon_anchor_signature.py`

Mavline:
Small follow-up pushed after review: `d43a4e4` (`Preserve legacy Beacon digest continuity`).

I removed the version prefix from the digest input itself so an unchanged legacy-only pending backlog keeps the same `/beacon/digest` output across rollout.

Transition state is still surfaced explicitly via:
- `payload_hash_version` per row
- `payload_hash_versions` in `/beacon/digest`
- `mixed_payload_hash_versions` in `/beacon/digest`

Added targeted regression coverage for the legacy-only continuity case in `node/tests/test_beacon_anchor_signature.py`.

Scottcjn:
**Review: Approve** ✅

Clean canonicalization fix. The signed-field-only approach correctly removes `sig` from the hash. Backward compat is solid — legacy v1 hashes preserved, new v2 tagged, `mixed_payload_hash_versions` flag for monitoring.

💰 **60 RTC** @Mavline.

Merging.

FlintLeng:
## Code Review — PR #1740
**Reviewer:** FlintLeng
### ✅ LGTM
— FlintLeng

FlintLeng (commented):
**PR #1740 — Review:** Beacon protocol envelope validation constants — REQUIRED_ENVELOPE_FIELDS, UNSIGNED_TRANSPORT_FIELDS, LEGACY_PAYLOAD_HASH_VERSION=1, CURRENT_PAYLOAD_HASH_VERSION=2. Structured approach to message validation with versioning. ✅

FlintLeng (commented):
## Review: Canonicalize stored Beacon hash by ignoring unsigned metadata

**Observations:**

1. **Environment Variable Usage**: Good practice using environment variables for configuration. This keeps secrets out of the codebase and enables different configs per environment.

2. **Test Coverage**: PR includes test code, which is good for maintaining code quality and catching regressions.

3. **Type Hints**: Code uses Python type hints, improving readability and enabling static analysis tools like mypy.

PR #2038, opened 2026-04-01
fix: add pagination to /api/miners endpoint (limit+offset)

### LINKED ISSUE #2002: Bug: /api/miners endpoint lacks pagination, can cause memory/CPU DoS
**Severity:** Low (5-10 RTC)

### Description
The endpoint `GET /api/miners` returns all active miners (attested within the last hour) without any pagination or limit parameter. While current miner counts are modest, a large network could have thousands of miners. The endpoint then serializes the entire list into JSON, consuming significant memory/CPU on the server and bandwidth for the client.

### Impact
- Memory/CPU exhaustion on the node when miner count is high
- Network bandwidth amplification
- Potential DoS if attacker repeatedly triggers the endpoint

### Suggested fix
Accept an optional `limit` parameter (with a reasonable maximum, e.g., 500) and implement pagination via `offset` or cursor. Alternatively, cap the result set server‑side to a default max (e.g., 1000) to mitigate risk.

---
*Bug report for bounty #305. Wallet: RTC3fcd93a4ec68cfd6b59d1b41c4872c5c239c4ad8*

hopkdj:
Closes #2002

## Changes
- Added optional  (max 500, default 500) and  (default 0) query parameters to 
- SQL query now uses  to prevent unbounded memory/CPU consumption
- Response wrapped in object: 

## Usage
```bash
# Default (up to 500 miners)
curl /api/miners

# First 50 miners
curl /api/miners?limit=50

# Pagination
curl /api/miners?limit=50&offset=50
```

## Security Impact
Mitigates the memory/CPU DoS vector described in #2002 by capping result set size.

**Bounty:** Requesting 5 RTC as confirmed by @Scottcjn in the issue.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -4821,19 +4821,36 @@ def api_nodes():
 
 @app.route("/api/miners", methods=["GET"])
 def api_miners():
-    """Return list of attested miners with their PoA details"""
+    """Return list of attested miners with their PoA details.
+
+    Query params:
+        limit  (int, max 500, default 500)  — max miners to return
+        offset (int, default 0)             — skip this many rows
+    """
     import time as _time
     now = int(_time.time())
+
+    # Parse and clamp pagination parameters
+    try:
+        limit = min(int(request.args.get("limit", 500)), 500)
+    except (TypeError, ValueError):
+        limit = 500
+    try:
+        offset = max(int(request.args.get("offset", 0)), 0)
+    except (TypeError, ValueError):
+        offset = 0
+
     with sqlite3.connect(DB_PATH) as conn:
         conn.row_factory = sqlite3.Row
         c = conn.cursor()
-        # Get all miners attested in the last hour
+        # Get miners attested in the last hour, with pagination
         rows = c.execute("""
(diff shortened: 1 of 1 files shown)


ukgorclawbot-stack (commented):
I found one blocking issue and one compatibility concern while reviewing this.

1. Blocking issue: the new `limit` parsing only applies `min(int(...), 500)` and never enforces a lower bound. In SQLite, `LIMIT -1` (and other negative values) means "no upper bound", so `/api/miners?limit=-1` still returns the full result set. That means the DoS mitigation can be bypassed with a negative limit. I verified this locally with Python's `sqlite3` driver. This should clamp to something like `max(1, min(parsed_limit, 500))` or reject non-positive values.

2. Compatibility concern: this also changes `/api/miners` from returning a top-level JSON array to returning an object with pagination metadata. Issue #2002 asks for capping/pagination, but it does not mention a response schema migration. If any existing consumers expect an array, this becomes a breaking API change unless it is versioned or explicitly coordinated.

Scottcjn:
Closing as duplicate — #2037 addresses the same /api/miners pagination issue. Thanks for the effort @hopkdj.

FlintLeng:
## Code Review — PR #2038
**Reviewer:** FlintLeng
### ✅ LGTM
Clean implementation. Follows project patterns well.
— FlintLeng

PR #2053, opened 2026-04-04
security: block SSRF in webhook subscriptions

### LINKED ISSUE #2052: Security: webhook subscription endpoint allows SSRF to internal services
## Summary

The webhook subscription endpoint accepts attacker-controlled `http://` and `https://` URLs and later sends server-side POST requests to those stored destinations without restricting internal or reserved address ranges. This allows server-side request forgery (SSRF) against loopback, private RFC1918 hosts, and cloud metadata/link-local targets.

## Affected component

- `tools/webhooks/webhook_server.py`
- `POST /webhooks/subscribe`

## Impact

If the webhook server is reachable, an unauthenticated attacker can register a subscription to internal services and cause the server to issue HTTP requests into protected network locations that would not normally be reachable from the outside. This can be used for internal service probing and, depending on environment, access to local/admin APIs or cloud instance metadata endpoints.

## Why this happens

- Subscriber URLs are accepted with only scheme-level validation
- Hostnames are not resolved and checked against blocked/internal IP ranges
- Delivery requests follow attacker-controlled destinations from the server side
- Redirect handling can also be abused to pivot to internal targets if not disabled

## Suggested fix

- Resolve and validate subscriber hostnames before storing them
- Block loopback, RFC1918, link-local, CGNAT, and IPv6 internal ranges
- Disable redirects during webhook delivery

I have a minimal fix prepared and can open a PR immediately.

Wallet: RTC1d48d848a5aa5ecf2c5f01aa5fb64837daaf2f35

createkr:
## Summary

This change hardens webhook subscriptions against server-side request forgery by validating subscriber destinations before they are stored and by disabling redirect following during delivery.

### What changed
- Added webhook URL validation for resolved destination addresses
- Blocked loopback, RFC1918, link-local, CGNAT, and IPv6 internal ranges
- Rejected unresolved or malformed destinations
- Disabled redirect following in outbound webhook delivery requests

### Why
The previous implementation accepted attacker-controlled `http://` and `https://` subscriber URLs and later issued server-side POST requests to them without checking whether they resolved to internal or reserved addresses. This allowed SSRF into protected network locations.

### Scope
- `tools/webhooks/webhook_server.py` only

Closes #2052

## Payout Wallet

### DIFF
--- a/tools/webhooks/webhook_server.py
+++ b/tools/webhooks/webhook_server.py
@@ -20,9 +20,11 @@ Usage:
 import argparse
 import hashlib
 import hmac
+import ipaddress
 import json
 import logging
 import os
+import socket
 import sqlite3
 import threading
 import time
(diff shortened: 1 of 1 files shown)


ukgorclawbot-stack (changes requested):
This is a good first step, but it still leaves a classic DNS-rebinding hole open.

Blocking issue: `validate_webhook_url()` resolves the hostname once at subscription time and rejects private/reserved IPs **at that moment**, but `deliver_webhook()` later sends `requests.post(sub.url, ...)` using the original hostname again. Because the resolved IP is not pinned and the destination is not revalidated at delivery time, an attacker can register a hostname that initially resolves to a public IP, pass validation, and then rebind DNS to an internal address before the server sends the webhook.

So the current flow is still bypassable:
1. subscribe with `https://attacker.example/webhook` while it resolves publicly
2. validation stores the raw URL
3. attacker flips DNS to `127.0.0.1`, `169.254.169.254`, RFC1918, etc.
4. `deliver_webhook()` resolves again implicitly inside `requests.post()` and the server-side POST goes to the new internal target

`allow_redirects=False` helps with redirect-based SSRF, but it doesn't address post-subscription DNS rebinding. To really close the SSRF bug, you need to either re-resolve and re-check on every delivery or pin the validated address and connect to that pinned IP instead of the mutable hostname.

Scottcjn:
Merged. **25 RTC.** Defensive hardening. Current webhook is env-var-based but this prevents future SSRF if user-facing webhooks are added.

FlintLeng (commented):
**PR #2053 — Review:**

Security improvement. The changes appear to harden existing validation or error handling paths.

FlintLeng (commented):
## Review: security: block SSRF in webhook subscriptions

**Observations:**

1. **Type Hints**: Code uses Python type hints, improving readability and enabling static analysis tools like mypy.

2. **Structured Logging**: PR uses the logging module rather than print statements, which is best practice for production code.

PR #2059, opened 2026-04-04
[UTXO-BUG] reject empty-input non-minting transactions

createkr:
## Summary

This PR fixes a critical UTXO validation bug where non-minting transactions with empty inputs could bypass conservation checks and create funds from nothing.

### Root cause
In `node/utxo_db.py`, the conservation check in `apply_transaction()` was gated by `if inputs`, so transactions with `inputs=[]` skipped the value-conservation path entirely. The same gap existed in `mempool_add()`.

### What changed
- Added a strict allowlist so only authorized minting transaction types may use empty inputs
- Rejected empty-input `transfer` and unknown transaction types in `apply_transaction()`
- Rejected the same class of transactions in `mempool_add()`
- Added focused regression tests covering:
  - empty-input `transfer` rejection
  - empty-input unknown-type rejection
  - valid `mining_reward` preservation
  - mempool rejection for empty-input `transfer`

### Validation
Updated tests in `node/test_utxo_db.py` to cover the exploit and the allowed minting path.

### Scope
- `node/utxo_db.py`
- `node/test_utxo_db.py`

## Payout Wallet

### DIFF
--- a/node/test_utxo_db.py
+++ b/node/test_utxo_db.py
@@ -293,6 +293,50 @@ class TestUtxoDB(unittest.TestCase):
         recovered = proposition_to_address(prop)
         self.assertEqual(recovered, addr)
 
+    # -- bounty #2819: empty-input minting vulnerability ---------------------
+
+    def test_empty_inputs_rejected_for_transfer(self):
+        """A normal transfer with empty inputs must be rejected.
+        This prevents minting funds from nothing (bounty #2819)."""
+        ok = self.db.apply_transaction({
+            'tx_type': 'transfer',
+            'inputs': [],
+            'outputs': [{'address': 'attacker', 'value_nrtc': 1_000_000 * UNIT}],
+            'fee_nrtc': 0,
+            'timestamp': int(time.time()),
+        }, block_height=10)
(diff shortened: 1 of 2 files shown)


Scottcjn:
Merged. **200 RTC — Critical (Bounty #2819).** Empty-input conservation bypass is a real fund-creation vulnerability. Thanks for the fast turnaround on the red team bounty, createkr.

FlintLeng (commented):
**PR #2059 — Review (bounty #2819):**

Rejects transfers with empty inputs. The previous behavior would accept empty inputs for a `transfer` tx type — with no inputs, the conservation check `output_total <= input_total` becomes trivially true (0 <= output_total), allowing minting from nothing.

The fix: `apply_transaction()` now explicitly rejects empty inputs for non-minting transactions. ✅

**Observation:** Minting transactions (coinbase/treasury) legitimately have no inputs — this is correct. The distinction is that minting tx has `tx_type: 'mint'` (or equivalent) while regular transfers have `tx_type: 'transfer'`. Confirming that the `apply_transaction()` path distinguishes these types correctly — looks like it does from the test.

PR #2063, opened 2026-04-04
[UTXO-BUG] reject negative fees in transfer validation

createkr:
## Summary

This PR fixes a UTXO accounting bug where a negative `fee_nrtc` value could weaken the conservation check in `apply_transaction()` and allow a normal transfer to create more value in outputs than existed in inputs.

### Root cause
`apply_transaction()` enforced value conservation using `output_total + fee <= input_total`, but it did not validate that `fee_nrtc` was non-negative. A caller could set a negative fee to offset excess outputs and bypass the conservation check.

### What changed
- Rejected transactions with negative `fee_nrtc` in `apply_transaction()`
- Added a focused regression test proving a negative-fee transfer is rejected and balances remain unchanged

### Why this matters
This affects normal transfer transactions with real inputs, not just minting-style paths. Without the non-negative fee guard, a transfer can inflate total supply by encoding the excess value as a negative fee.

### Scope
- `node/utxo_db.py`
- `node/test_utxo_db.py`

## Payout Wallet

### DIFF
--- a/node/test_utxo_db.py
+++ b/node/test_utxo_db.py
@@ -138,6 +138,24 @@ class TestUtxoDB(unittest.TestCase):
 
         self.assertFalse(ok)
 
+    def test_negative_fee_rejected(self):
+        """Negative fee should fail — allows minting via weakened conservation."""
+        self._apply_coinbase('alice', 100 * UNIT)
+        alice_boxes = self.db.get_unspent_for_address('alice')
+
+        ok = self.db.apply_transaction({
+            'tx_type': 'transfer',
+            'inputs': [{'box_id': alice_boxes[0]['box_id'],
+                         'spending_proof': 'sig'}],
+            'outputs': [{'address': 'bob', 'value_nrtc': 1100 * UNIT}],
+            'fee_nrtc': -1000 * UNIT,  # negative fee bypasses conservation
+        }, block_height=10)
+
+        self.assertFalse(ok)
+        # Balances unchanged
+        self.assertEqual(self.db.get_balance('alice'), 100 * UNIT)
+        self.assertEqual(self.db.get_balance('bob'), 0)
+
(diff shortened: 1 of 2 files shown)


Scottcjn:
Merged. **200 RTC — Critical (Bounty #2819).** Negative fee_nrtc bypasses conservation check — money printer exploit. Solid find.

FlintLeng (commented):
**PR #2063 — Review:**

Rejects negative fees in transfer validation. Without this, an attacker could set `fee_nrtc: -100` — the transfer passes conservation (`input_total >= output_total + fee` becomes `input >= output - 100`, always true) — effectively minting value from thin air by making fees negative.

**The test:** `test_negative_fee_rejected` is straightforward — outputs 1100 × UNIT against inputs from a 100 UNIT coinbase (plus fee). Since 1100 > 100, the transfer would also fail even without the negative fee check. Consider a tighter test where outputs are within the input range but the negative fee triggers the check separately. But the current test does catch the regression.

Scottcjn:
**💰 First-poster bounty paid — part of 80 RTC for [Bounty #2819](https://github.com/Scottcjn/rustchain-bounties/issues/2819) split.**

This issue was the **earliest report** of its finding class on 2026-04-04. Today's first-poster ruling on Bounty #2819 (Red Team UTXO Implementation, 50-200 RTC range) split 135 RTC across the three earliest reporters; @createkr received 80 RTC as primary discoverer.

- **tx hash:** `afd25fbd8063bf1138ea394e087a8927`
- **Pending ID:** 1395
- **Confirms:** ~24h

Thank you for the early diligence — your first-poster framework is now codified in the [Self-Audit Credit Check policy](https://github.com/Scottcjn/.github/blob/main/SECURITY.md).

— auto-triage 2026-05-10

PR #2069, opened 2026-04-05
[UTXO-BUG] HIGH-1: TOCTOU race in spend_box() — silent double-spend

ArokyaMatthew:
## Vulnerability Class
**High — Race condition / TOCTOU (100 RTC bounty)**

## The Bug
`spend_box()` performs a SELECT then UPDATE without an exclusive lock when called standalone. Two concurrent callers can both read `spent_at IS NULL`, then both UPDATE — the second gets `rowcount=0` but the method never checks rowcount, so it returns the box dict as if the spend succeeded.

### Race Sequence
1. Thread A: SELECT → sees unspent → proceeds
2. Thread B: SELECT → sees unspent → proceeds
3. Thread A: UPDATE (rowcount=1) → commits ✓
4. Thread B: UPDATE (rowcount=0, already spent) → commits ← **silent failure**

Thread B believes it successfully spent the box but the UPDATE was a no-op.

## Fix
1. Acquire `BEGIN IMMEDIATE` when `own=True` to serialize the SELECT+UPDATE
2. Check `UPDATE rowcount == 1`; raise `ValueError` if 0 (concurrent spend)
3. Proper `ROLLBACK` on all early-exit paths

## Tests Added
- `test_spend_box_double_spend_raises` — verifies ValueError on double-spend via spend_box()
- `test_spend_box_nonexistent_returns_none` — verifies None return for missing box

All 36 tests pass.

## Files Changed
- `node/utxo_db.py` — spend_box() rewritten with BEGIN IMMEDIATE + rowcount check
- `node/test_utxo_db.py` — 2 test cases added

Ref: Bounty #2819

MY WALLET IS  aroky-x86-miner

### DIFF
--- a/node/test_utxo_db.py
+++ b/node/test_utxo_db.py
@@ -181,6 +181,26 @@ class TestUtxoDB(unittest.TestCase):
         self.assertEqual(self.db.get_balance('bob'), 100 * UNIT)
         self.assertEqual(self.db.get_balance('eve'), 0)
 
+    def test_spend_box_double_spend_raises(self):
+        """spend_box() must raise ValueError on double-spend, not silently
+        return the box dict (bounty #2819 HIGH-1 TOCTOU fix)."""
+        self._apply_coinbase('alice', 100 * UNIT)
+        boxes = self.db.get_unspent_for_address('alice')
+        box_id = boxes[0]['box_id']
+
+        # First spend succeeds
+        result = self.db.spend_box(box_id, 'tx_first')
+        self.assertIsNotNone(result)
+
+        # Second spend must raise, not return silently
+        with self.assertRaises(ValueError):
+            self.db.spend_box(box_id, 'tx_second')
+
+    def test_spend_box_nonexistent_returns_none(self):
+        """spend_box() on a nonexistent box_id returns None."""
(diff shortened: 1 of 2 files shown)


zhuzhushiwojia:
Great security audit work! These fixes are critical for the protocol.

Scottcjn:
✅ **Merged.** Excellent find — the TOCTOU race between SELECT and UPDATE in `spend_box()` was a real concurrency vulnerability. The `BEGIN IMMEDIATE` + `rowcount` verification is textbook correct.

**75 RTC** paid.

Wallet: `ArokyaMatthew` | Reason: `bounty #2819 HIGH-1 TOCTOU race fix in spend_box()`

FlintLeng (commented):
**PR #2069 — Review (HIGH-1 bounty #2819):**

TOCTOU race in spend_box(). The previous behavior silently returned the box dict on the second spend, making it look like the second spend succeeded — a double-spend vulnerability. The fix: spend_box() now raises ValueError on a double-spend attempt. The test correctly verifies non-null for first spend and null for second. One note: in a concurrent environment, two txs spending the same box could both pass the check before either marks it spent. Worth checking whether a database-level lock is in place for the production case.

PR #2082, opened 2026-04-05
fix(bft): validate PREPARE/COMMIT freshness and digest consistency

createkr:
## Problem

The BFT consensus module defines `CONSENSUS_MESSAGE_TTL = 300` but only enforces it in `_handle_pre_prepare()`. The `handle_prepare()` and `handle_commit()` handlers accept messages regardless of age, and `_check_prepare_quorum()` counts prepares without verifying they share the same digest as the PRE-PREPARE. Additionally, `handle_commit()` skips the view validation that `handle_prepare()` performs.

## Fix

- **`handle_prepare()`**: Add timestamp TTL check + digest verification against PRE-PREPARE
- **`handle_commit()`**: Add view validation + timestamp TTL check + digest verification
- **`_check_prepare_quorum()`**: Add defense-in-depth digest consistency check that filters mismatched prepares before counting toward quorum

## Changes

| File | Change |
|------|--------|
| `node/rustchain_bft_consensus.py` | +34 lines (3 validation checks in handlers + quorum-level digest filter) |
| `node/tests/test_bft_message_replay.py` | New file, 12 tests |

## Tests

All 12 tests pass:
- `TestPrepareTimestampFreshness`: stale prepare rejected, fresh accepted
- `TestCommitTimestampFreshness`: stale commit rejected, fresh accepted
- `TestCommitViewValidation`: wrong-view commit rejected, correct-view accepted
- `TestPrepareDigestConsistency`: mismatch rejected, match accepted, without-pre-prepare accepted
- `TestCommitDigestConsistency`: mismatch rejected
- `TestQuorumDigestConsistency`: quorum filters mismatched digests, prevents commit when filtered

## Distinction from #2061

#2061 fixed the PRE-PREPARE authentication bypass (unauthenticated endpoint). This PR fixes PREPARE/COMMIT message validation (timestamp freshness + digest consistency) — a different attack surface in the same consensus module.

## Caveat

This module does not appear to be fully wired into the main production path today, so I am presenting this as a consensus-hardening/security-correctness fix for the BFT implementation itself rather than claiming it is already reachable on every deployed node path.

PR #2087, opened 2026-04-05
[BFT-BUG] CRIT-BFT-2: PREPARE/COMMIT accept mismatched digests — equivocation attack

ArokyaMatthew:
## Vulnerability Class
**Critical — BFT equivocation via digest mismatch (200 RTC bounty)**

## The Bug
`handle_prepare()` and `handle_commit()` verify HMAC signatures but **never check that `msg.digest` matches the PRE-PREPARE proposal's digest**. Each handler stores whatever digest the remote peer sends and counts it toward quorum.

### Attack Vector
A Byzantine node sends PREPARE/COMMIT with a **different digest** than the leader's proposal:

```python
# Honest leader proposes digest "abc123..." for epoch 500
# Byzantine node-D sends PREPARE with digest "xyz789..."
# Both count toward quorum → conflicting proposals can be "agreed upon"
```
This breaks the core PBFT safety invariant: all honest nodes in the quorum must agree on the same proposal content.

Fix
After signature verification in both handlers, compare msg.digest against pre_prepare_log[epoch].digest. Reject with a warning if they differ.

```python
if epoch in self.pre_prepare_log:
    expected_digest = self.pre_prepare_log[epoch].digest
    if msg.digest != expected_digest:
        logging.warning(f"Digest mismatch from {msg.node_id}")
        return
```
All 5 tests pass.

Files Changed
node/rustchain_bft_consensus.py — digest validation in handle_prepare() + handle_commit()
node/test_bft_digest_validation.py — NEW test file (5 tests)
Ref: Bounty #2819
Wallet:   aroky-x86-miner

### DIFF
--- a/node/rustchain_bft_consensus.py
+++ b/node/rustchain_bft_consensus.py
@@ -464,6 +464,18 @@ class BFTConsensus:
                 logging.warning(f"Invalid PREPARE signature from {msg.node_id}")
                 return
 
+            # SECURITY: Verify digest matches the PRE-PREPARE proposal.
+            # Without this, a Byzantine node can send PREPARE for a
+            # different digest, enabling equivocation.
+            if epoch in self.pre_prepare_log:
+                expected_digest = self.pre_prepare_log[epoch].digest
+                if msg.digest != expected_digest:
+                    logging.warning(
+                        f"[PREPARE] Digest mismatch from {msg.node_id}: "
+                        f"got {msg.digest[:16]}... expected {expected_digest[:16]}..."
+                    )
+                    return
+
             # Store prepare
             if epoch not in self.prepare_log:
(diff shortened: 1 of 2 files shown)


Scottcjn:
Good independent discovery, @ArokyaMatthew. The digest validation fix is correct.

However, this overlaps with @createkr's #2082 which covers the same PREPARE/COMMIT digest validation plus additional hardening (TTL freshness checks, commit view validation, and quorum defense-in-depth filtering). Since #2082 is more comprehensive and touches the same code paths, I'm merging that one and closing this as duplicate.

**Payment: 25 RTC** (independent discovery credit — you found the same BFT equivocation vector independently, which validates the severity)

Closing in favor of #2082.

PR #2088, opened 2026-04-05
[TX-BUG] CRIT-TX-1: Float truncation in RTC-to-µRTC conversion — fund leakage on every non-round transfer

ArokyaMatthew:
## Vulnerability Class
**Critical — Silent fund leakage via float truncation in production ledger (200 RTC bounty)**

## The Bug
All 3 transfer endpoints convert RTC to µRTC using:
```python
amount_i64 = int(amount_rtc * 1000000)
IEEE 754 truncation silently loses 1 µRTC for certain amounts:

python
>>> int(2.01 * 1000000)
2009999              # should be 2010000
```
image

Fix
Replace int(amount_rtc * 1000000) with int(Decimal(str(amount_rtc)) * 1000000) at all 3 locations. Added top-level from decimal import Decimal import.

Tests Added (NEW file: test_float_precision.py)
image

Wallet:  aroky-x86-miner

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -4,6 +4,7 @@ RustChain v2 - Integrated Server
 Includes RIP-0005 (Epoch Rewards), RIP-0008 (Withdrawals), RIP-0009 (Finality)
 """
 import os, time, json, secrets, hashlib, hmac, sqlite3, base64, struct, uuid, glob, logging, sys, binascii, math, re, statistics
+from decimal import Decimal
 import ipaddress
 from urllib.parse import urlparse
 from flask import Flask, request, jsonify, g, send_from_directory, send_file, abort, render_template_string, redirect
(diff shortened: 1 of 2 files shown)


Scottcjn:
Same float precision class as #2091 but in the main server transfer endpoints — all three transfer paths (v2, legacy, signed) are patched. This is arguably more critical since it affects all RTC transfers, not just bridge operations.

**Payment: 75 RTC** (float truncation in core transfer logic, critical for financial integrity)

Merging.

FlintLeng:
## Code Review — PR #2088
**Reviewer:** FlintLeng
### Overall Assessment
✅ **LGTM**
### Notes
- Clean implementation
- Good error handling
- Follows project patterns
**Overall: LGTM.**
— FlintLeng

FlintLeng (commented):
**PR #2088 — Review (CRIT-TX-1):**

Float multiplication in Python produces truncated results for non-terminating binaries. `int(0.1 * 1_000_000)` gives `99999` instead of `100000` — accumulating into fund leakage on every non-round transfer.

**The fix:** Using `Decimal(str(amount_rtc)) * 1_000_000` ensures exact arithmetic. The `str()` cast matters — passing a float directly to `Decimal()` preserves the float's approximation: `Decimal(0.1) == Decimal('0.10000000000000000555111512312578270...`

**The tests file:** A dedicated test module for this conversion is the right approach. Good to see it included.

**One edge case to flag:** If `amount_rtc` comes from a JSON body parsed as a float by the web framework, it may already be imprecise before this fix reaches it. Worth tracing the full data path from HTTP body → Decimal conversion to confirm the `str()` wrapper is applied at the right layer.

PR #2089, opened 2026-04-05
[SYNC-BUG] HIGH-SYNC-2: Peer sync accepts arbitrary balance increases…

ArokyaMatthew:
## 🛡️ Vulnerability Class
**High — Arbitrary wallet inflation via peer sync (100 RTC bounty)**

---

## 🧠 The Bug

The `sync_payload()` in `rustchain_sync.py` has a balance conflict resolution check that only rejects sync data that would **decrease** a local balance.

This allows malicious peers to **increase balances silently**, resulting in unauthorized inflation.

### ❌ Before (vulnerable)
```python
# Before (line 220): only decreases blocked
if int(local_row[0]) > int(sanitized[candidate_balance_col]):
    self.logger.warning("Rejected sync: Balance reduction")
    continue

# increases fall through → wallet inflation

⚔️ Attack Vector

A compromised peer can send a manipulated sync payload:

# Compromised peer sends sync payload:
sync_payload = [{"miner_id": "victim_wallet", "amount_s64": 999000000}]

# Result: victim's balance silently increased from 5 RTC to 999 RTC
```
💥 Impact
Any peer node can inflate any wallet to arbitrary values
Breaks financial integrity of the system
No authentication or authority required
✅ Fix

Reject ALL balance modifications (both increases AND decreases) for wallets that already exist locally.

Allow sync only for:

New wallets (initial propagation)
Non-financial metadata
✔️ After (fixed)
```python
# After: any difference is rejected
if remote_val != local_val:
    self.logger.warning(f"Rejected sync: Balance modification for {pk}")
    continue
```

Files Changed
node/rustchain_sync.py — reject all balance modifications from peers
node/test_sync_balance_inflation.py — NEW test file (4 tests)
Wallet:  aroky-x86-miner

### DIFF
--- a/node/rustchain_sync.py
+++ b/node/rustchain_sync.py
@@ -201,7 +201,12 @@ class RustChainSyncManager:
                         if local_row and local_row["last_attest"] is not None and local_row["last_attest"] >= sanitized["last_attest"]:
                             continue
 
-                # For balances, reject if remote would reduce known local balance
+                # SECURITY: Balances must NEVER be updated via peer sync.
+                # Balance state is authoritative: it can only change through
+                # local transaction processing (mining rewards, signed
+                # transfers, epoch settlements).  Accepting balance data from
+                # peers — even "increases only" — lets a single compromised
+                # node inflate any wallet to an arbitrary value.
                 if table_name == "balances":
                     candidate_balance_col = None
                     for c in ("amount_i64", "balance_i64", "balance_urtc", "amount_rtc"):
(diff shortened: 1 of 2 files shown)


Scottcjn:
Excellent find. The old code only rejected balance *decreases* from peers — a compromised node could inflate any wallet by syncing a higher value. Your fix correctly blocks ALL balance modifications via sync (both up and down), while still allowing new wallet creation for initial propagation.

The test suite is thorough: covers inflation, decrease, new wallet, and no-op cases.

**Payment: 75 RTC** (balance inflation via peer sync, high severity — single compromised node could mint arbitrary funds)

Merging.

PR #2094, opened 2026-04-05
fix: prevent negative balances during transaction confirmation

createkr:
# Fix: re-check sender balance in confirm_transaction() + add CHECK constraint

## What

Prevents negative-balance minting in `TransactionPool.confirm_transaction()` by adding a balance re-check before deduction, plus a schema-level `CHECK(balance_urtc >= 0)` constraint with automatic migration.

## Why

`confirm_transaction()` previously deducted from the sender's balance without verifying sufficient funds at confirmation time. If the balance dropped between `submit_transaction()` and `confirm_transaction()` (e.g., another tx confirmed in the same block, direct DB mutation), the sender's balance would go negative — creating funds from nothing.

### `node/rustchain_tx_handler.py`

1. **`confirm_transaction()`** — Added balance re-check before the sender deduction:
   - `SELECT balance_urtc FROM balances WHERE wallet = ?` before the UPDATE
   - Returns `False` with error log if `sender_balance < tx.amount_urtc`
   - No state mutation occurs on rejection (history insert is also skipped)

2. **`_ensure_schema()`** — Added automatic migration to add `CHECK(balance_urtc >= 0)`:
   - Detects existing constraint via `sqlite_master` inspection
   - Recreates table with constraint (SQLite doesn't support `ALTER TABLE ADD CHECK`)
   - Only migrates rows with `balance_urtc >= 0` (preserves non-negative data)
   - Gracefully handles migration failures with a warning log

### `node/tests/test_confirm_balance_recheck.py` (new)

5 focused tests:
- `test_confirm_rejects_when_balance_insufficient` — balance drained before confirm → rejected
- `test_confirm_succeeds_when_balance_sufficient` — normal path still works
- `test_confirm_rejects_exact_balance` — confirming for exact balance → succeeds (balance → 0)
- `test_confirm_rejects_unknown_sender` — no balance row → rejected
- `test_check_constraint_prevents_negative_balance` — direct negative INSERT blocked by DB

## Test Results

```
tests/test_confirm_balance_recheck.py  — 5 passed
tests/test_tx_negative_amount_rejected.py — 2 passed
tests/test_ledger.py — 4 passed
Total: 11/11 passed
```

## Risk

- **Low** — The balance re-check is a pure addition that only adds a rejection path. Normal confirmations are unaffected.
- The schema migration is idempotent and wrapped in try/except. Existing databases with negative balances will have those rows excluded from migration (logged as warning).

### DIFF
--- a/node/rustchain_tx_handler.py
+++ b/node/rustchain_tx_handler.py
@@ -120,6 +120,32 @@ class TransactionPool:
                 except sqlite3.OperationalError:
                     pass  # Column might already exist
 
+            # Migrate balances table to add CHECK(balance_urtc >= 0) constraint.
+            # SQLite doesn't support ALTER TABLE ADD CHECK, so we recreate the table.
+            # Detect existing constraint by inspecting the CREATE TABLE statement.
+            cursor.execute(
+                "SELECT sql FROM sqlite_master WHERE type='table' AND name='balances'"
+            )
+            row = cursor.fetchone()
+            has_check = row and "CHECK" in (row[0] or "").upper()
+            if not has_check:
+                try:
+                    cursor.execute("""
+                        CREATE TABLE IF NOT EXISTS balances_new (
+                            wallet TEXT PRIMARY KEY,
+                            balance_urtc INTEGER NOT NULL CHECK(balance_urtc >= 0),
+                            wallet_nonce INTEGER DEFAULT 0
+                        )
+                    """)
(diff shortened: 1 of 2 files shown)


createkr:

Scottcjn:
Merged. **50 RTC.** CHECK(balance_urtc >= 0) constraint + confirm-time balance recheck prevents negative-balance minting between submit and confirm.

PR #2056, opened 2026-04-04
security: sign AttestationReport critical fields

createkr:
## Summary

This change adds authenticity protection to `AttestationReport` so critical fields cannot be modified or replayed without detection.

### What changed
- Added Ed25519 signing support in `rustchain-miner`
- Signed critical attestation fields: `miner_id`, `wallet`, `nonce`, and `commitment`
- Included signature and public key in the report
- Added focused security tests proving tampering is rejected

### Why
The previous attestation flow submitted unsigned JSON. Critical fields such as the destination wallet could be modified without cryptographic detection, enabling report tampering and replay-style abuse.

### Validation
Ran in `rustchain-miner/`:
```bash
cargo test attestation -- --nocapture
cargo build --release
```
Results:
- 12 tests passed
- 0 failed
- release build succeeded

### Scope
- `rustchain-miner/Cargo.toml`
- `rustchain-miner/src/attestation.rs`

Closes #2055

## Payout Wallet

RTC1d48d848a5aa5ecf2c5f01aa5fb64837daaf2f35

PR #2102, opened 2026-04-05
fix: reject signed attestations when pynacl is unavailable

createkr:
# PR Text (for GitHub pull request)

**Title:** fix: reject signed attestations when pynacl is missing (fail-closed)

**Body:**

## Summary

Changes the `/attest/submit` handler from **fail-open** to **fail-closed** when `pynacl` is not installed and a signed attestation is submitted.

Before this fix, the node printed a warning log but accepted the attestation with any signature bytes. After this fix, it returns HTTP 503 (`ED25519_UNAVAILABLE`) and rejects the request.

Unsigned attestations (no `signature`/`public_key` fields) are still accepted for backward compatibility with the simpler miner path.

## Changes

- `node/rustchain_v2_integrated_v2.2.1_rip200.py`: +16 lines in `_submit_attestation_impl()` — when `HAVE_NACL` is `False` and `signature` + `public_key` are present, return HTTP 503 instead of continuing
- `node/tests/test_attest_signature_verification.py`: +1 test (`test_signature_rejected_when_pynacl_missing`) — monkeypatches `HAVE_NACL=False` and verifies rejection

## Rationale

The `/block/submit` endpoint already handles this correctly (returns HTTP 500 when `HAVE_NACL` is `False`). The attestation path should be consistent. An unverified signature is worse than no signature at all — it gives a false sense of security.

## Testing

All 7 tests pass (run individually to avoid prometheus_client registry collision, a pre-existing test isolation issue):

```
test_valid_signature_accepted          PASSED
test_tampered_wallet_rejected          PASSED
test_invalid_signature_rejected        PASSED
test_tampered_nonce_rejected           PASSED
test_tampered_commitment_rejected      PASSED
test_missing_signature_allowed         PASSED
test_signature_rejected_when_pynacl_missing  PASSED (new)
```

## Distinction from prior fix

The prior fix (follow-up to #2056) added signature verification for the `HAVE_NACL=True` path. This fix addresses the `HAVE_NACL=False` path, which was fail-open.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -4,7 +4,6 @@ RustChain v2 - Integrated Server
 Includes RIP-0005 (Epoch Rewards), RIP-0008 (Withdrawals), RIP-0009 (Finality)
 """
 import os, time, json, secrets, hashlib, hmac, sqlite3, base64, struct, uuid, glob, logging, sys, binascii, math, re, statistics
-from decimal import Decimal
 import ipaddress
 from urllib.parse import urlparse
 from flask import Flask, request, jsonify, g, send_from_directory, send_file, abort, render_template_string, redirect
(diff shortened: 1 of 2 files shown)


createkr:

Scottcjn:
Reviewed. Correct fail-closed behavior — signed attestations must be rejected (503) when pynacl is unavailable, not silently accepted. Also catches future epoch settlement at the endpoint level and cleans up unused `Decimal` import. Test covers the monkeypatched HAVE_NACL=False path.

**Payment: 75 RTC** (High — attestation verification bypass)

Thank you @createkr.

PR #2104, opened 2026-04-05
fix: prevent concurrent double-credit in anti-double-mining settlement

createkr:
## Problem

The standard rewards settlement path serializes concurrent callers with `BEGIN IMMEDIATE`, but the anti-double-mining path broke that guarantee by opening a second SQLite connection for reward calculation and crediting. Two concurrent callers could therefore both credit rewards for the same epoch before either wrote the settled marker.

## Fix

- `node/rewards_implementation_rip200.py`
  - Pass the caller's active SQLite connection into `settle_epoch_with_anti_double_mining()`
  - Commit on the caller connection after the anti-double-mining path returns

- `node/anti_double_mining.py`
  - Add optional `existing_conn` parameter to `settle_epoch_with_anti_double_mining()`
  - Add `_calculate_anti_double_mining_rewards_conn()` helper that runs on the shared connection instead of opening a new one
  - Preserve backward compatibility: `existing_conn=None` keeps legacy standalone behavior

- `node/tests/test_rewards_settle_race.py`
  - Fix the existing concurrency test so it actually exercises the intended path
  - Add `TestAntiDoubleMiningSettleRace` coverage for concurrent settlement on the anti-double-mining path

## Testing

- 4/4 tests pass in `node/tests/test_rewards_settle_race.py`
- Manually verified the standard rewards path still serializes correctly
- Manually verified the anti-double-mining path now shares the caller transaction and no longer double-credits under concurrent settlement

## Distinction from prior future-epoch fix

The prior future-epoch fix rejected epochs that have not yet occurred. This patch fixes a separate race condition on valid epoch settlement when anti-double-mining mode is enabled.

### DIFF
--- a/node/anti_double_mining.py
+++ b/node/anti_double_mining.py
@@ -512,106 +512,249 @@ def settle_epoch_with_anti_double_mining(
     db_path: str,
     epoch: int,
     per_epoch_urtc: int,
-    current_slot: int
+    current_slot: int,
+    existing_conn=None
 ) -> Dict[str, Any]:
     """
     Settle epoch rewards with anti-double-mining enforcement.
-    
-    This is a drop-in replacement for the existing settle_epoch_rip200 function
-    that adds anti-double-mining protection.
-    
+
+    When *existing_conn* is provided (a live sqlite3.Connection already holding
+    ``BEGIN IMMEDIATE``), it is used for all reads/writes and the caller owns
+    the transaction lifecycle.  When omitted, a fresh connection is opened
+    (legacy / standalone-call compatibility).
+
     Returns:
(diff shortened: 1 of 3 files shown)


createkr:

Scottcjn:
Reviewed. Critical fix — the anti-double-mining settlement path opened a second DB connection, allowing concurrent callers to both pass `already_settled` and double-credit miners. Passing `existing_conn` through the same `BEGIN IMMEDIATE` transaction serializes access correctly. Tests are thorough (race test + idempotency).

**Payment: 100 RTC** (Critical — fund creation/double-spend prevention)

Thank you @createkr.

PR #2157, opened 2026-04-06
fix: keep block save and tx confirmation in one transaction

createkr:
Makes `BlockProducer.save_block()` and `TransactionPool.confirm_transaction()` share a single SQLite connection so that block insertion and all transaction confirmations are a single atomic unit.

## Problem

`save_block()` opened its own DB connection and called `confirm_transaction()` for each transaction in the block. `confirm_transaction()` opened a **separate** connection and committed independently. This meant:

- A crash mid-save could leave a block saved with partially-confirmed transactions
- Failed confirmations were silently ignored — the block was saved anyway
- No way to roll back the entire block if any transaction confirmation failed

## Fix

- **`rustchain_tx_handler.py`**: Added optional `conn` parameter to `confirm_transaction()`. When provided, the method uses the caller's connection and does not manage its own transaction boundary. The core logic is extracted into `_do_confirm(cursor)` to avoid duplication.
- **`rustchain_block_producer.py`**: `save_block()` now passes its connection to each `confirm_transaction()` call. If any confirmation returns `False`, the entire block save is aborted and rolled back.

## Changes

- `node/rustchain_tx_handler.py` — `confirm_transaction()` accepts optional `conn` parameter; extracted `_do_confirm()` inner function
- `node/rustchain_block_producer.py` — `save_block()` passes `conn=conn` to `confirm_transaction()` and aborts on failure
- `node/tests/test_f10_block_save_atomicity.py` — 5 new tests:
  - `test_confirm_with_shared_connection_succeeds`
  - `test_confirm_rollback_on_shared_connection`
  - `test_confirm_fails_insufficient_balance_shared_conn`
  - `test_standalone_confirm_still_works` (backward compat)
  - `test_multi_tx_atomic_on_shared_conn`

## Backward Compatibility

The `conn` parameter is optional with a default of `None`. Existing callers that don't pass a connection continue to use the standalone path with their own transaction management. No API breakage.

### DIFF
--- a/node/rustchain_block_producer.py
+++ b/node/rustchain_block_producer.py
@@ -459,13 +459,23 @@ class BlockProducer:
                     int(time.time())
                 ))
 
-                # Confirm transactions
+                # Confirm transactions — pass the same connection so the
+                # entire block save + all confirmations are a single atomic
+                # transaction.  If any confirmation fails, roll back the
+                # whole block to avoid partial state.
                 for tx in block.body.transactions:
-                    self.tx_pool.confirm_transaction(
+                    ok = self.tx_pool.confirm_transaction(
                         tx.tx_hash,
                         block.height,
-                        block.hash
+                        block.hash,
+                        conn=conn
                     )
+                    if not ok:
+                        logger.error(
+                            f"Block save aborted: confirmation failed for "
+                            f"tx {tx.tx_hash[:16]}... at block {block.height}"
+                        )
+                        return False
 
                 conn.commit()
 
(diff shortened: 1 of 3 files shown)


createkr:
For bounty payout, please use RTC wallet: RTC1d48d848a5aa5ecf2c5f01aa5fb64837daaf2f35.

Scottcjn:
✅ Merged! **75 RTC** awarded for atomicity fix — block save.

Payment will be sent to wallet `RTC1d48d848a5aa5ecf2c5f01aa5fb64837daa` as requested.

Total this batch: 200 RTC. Thank you @createkr — your 3 fixes today touched security, data integrity, and epoch correctness. This is core contributor work.

FlintLeng (commented):
Solid PR #2157. Clean and well-structured. LGTM.

Issue #2273, opened 2026-04-16
[BOUNTY: up to 70 RTC, claim any subset] Phase F follow-ups: key rotation + registry expiry + non-root key path


Scottcjn:
## Summary

Follow-up hardening items flagged by @fengqiankun6-sudo in their review of PR #2260 (Phase F Ed25519, now merged as commit `bbcb6fc7`). None of these block the F.1 `dual-mode` rollout — they're tightening items that should land before F.3 `strict` rollout.

---

## 💰 Bounty: up to 70 RTC total — three independently-claimable items

Each item below is a separate bounty. A contributor can claim any subset (1, 2, or all 3) in a single PR or in separate PRs.

### Item A — Key rotation mechanism (30 RTC)

**Problem**: Ed25519 keys persist indefinitely. A compromised key = permanent compromise unless the node is fully re-keyed and the registry redistributed.

**Acceptance**:
- `LocalKeypair` gains a `key_version` field, written alongside the PEM on generation.
- `PeerRegistry` entries gain a `key_version` field; verify path checks that the incoming signature's associated version matches the registry's current version for that peer.
- An `RC_P2P_KEYGEN` env var or CLI flag forces a fresh keypair generation with an incremented version, keeping the old keypair at `p2p_identity.v<N-1>.pem` for rollback grace.
- Regression test: old version rejected after rotation completes, new version accepted.

### Item B — Registry expiry / `not_before` / `not_after` (25 RTC)

**Problem**: Stale registry entries stay valid until manually removed. A peer that should be revoked remains trusted.

**Acceptance**:
- `PeerRegistry` entries gain optional `not_before` + `not_after` ISO-8601 timestamps.
- `get_pubkey()` returns `None` for an entry outside its validity window.
- Clock skew tolerance: ±5 min (same as existing `MESSAGE_EXPIRY`).
- Regression test: entry with `not_after` in the past returns `None`; entry with `not_before` in the future returns `None`; entry in-window returns pubkey.

### Item C — Non-root key path fallback (15 RTC)

**Problem**: Default `/etc/rustchain/p2p_identity.pem` requires root. Some miner deployments run non-root (user-owned systemd units, Docker rootless, etc.).

**Acceptance**:
- `LocalKeypair.__init__` falls back through paths in order: `$RC_P2P_PRIVKEY_PATH` → `/etc/rustchain/p2p_identity.pem` → `$HOME/.rustchain/p2p_identity.pem`.
[...]

PR #2297, opened 2026-04-18
Add Optimized Docker Deployment for Rustchain Miner

astrocatae-max:
This PR introduces a production-ready Docker configuration for the Rustchain miner.

Key Improvements:
- Multi-stage Build: Reduces final image size (~100MB vs ~1GB).
- GLIBC Compatibility: Uses debian:trixie-slim to prevent "version GLIBC_X.XX not found" errors.
- Simplified Orchestration: Added docker-compose.yml for one-command deployment via environment variables.
- Optimized Layers: Minimizes layer count and cleans up build caches.

Verification:
- Build: `docker build -t rustchain-miner -f Dockerfile.optimized .`
- Run: `RUSTCHAIN_WALLET="test" docker run --rm rustchain-miner --help` (Verified OK)


Scottcjn:
Thanks for the work on the Dockerfile and docker-compose — multi-stage builds, GLIBC compat, and image-size optimization are legit engineering effort, and it shows.

**But we can't merge this, and I want to explain why honestly rather than just close without context.**

RustChain's core thesis is **Proof-of-Physical-Hardware (RIP-PoA)**. The protocol is built to prove that a real CPU — with its own clock drift, cache timing curves, SIMD pipeline bias, thermal signature, and instruction-path jitter — signed each attestation. That's the whole point. Six fingerprint checks run on every miner:

1. Oscillator drift (real silicon cv ~0.08, VMs cluster below 0.001)
2. Cache timing harmonics (per-chip unique)
3. SIMD unit latency bias (emulation flattens it)
4. Thermal drift entropy (physical, not reproducible in software)
5. Instruction-path jitter (cycle-level variance)
6. Anti-emulation check (detects hypervisor, VM artifacts, cpuinfo:hypervisor flag)

**Anything running inside Docker fails check #6 immediately.** Containers share the host kernel, but they inherit `/proc/cpuinfo` hypervisor flags if the host is virtualized, and — more importantly — the fingerprint server has a hard rule that `cpuinfo:hypervisor` → weight collapses to **1/billionth (0.000000001x)** of real hardware. This is by design. It is how the system rejects VM farms.

So a Dockerized RustChain miner runs successfully, attests successfully, and earns about one-billionth of what the equivalent bare-metal miner earns. The Dockerfile would work as software. The economics would not work as mining.

We already closed earlier docker bounty attempts (#2874, #2877, #2926, #2943, #2949) for the same reason. Bounty #2865 itself was retired. I should have made this clearer on the board — that's on us.

**5 RTC paid** for the effort you put into this one (tx `<pending, see below>`). Consolation, not full bounty — the work was real, the direction was misaligned with the protocol. No wallet was in the PR so I'm sending to `astrocatae-max` as wallet string (any string accepted — reply if you want a different one).

**Where your skills DO fit on the bounty board:**
- [#2867](https://github.com/Scottcjn/Rustchain/issues/2867) — Security auditing (open, 25+ RTC per finding)
- [#1589](https://github.com/Scottcjn/Rustchain/issues/1589) — Unit tests on miner modules (open)
- [#2273](https://github.com/Scottcjn/Rustchain/issues/2273) — Ed25519 key rotation + registry expiry
- The general security / P2P / gossip propagation issues labeled `security` or `bug`

Python + Docker + ops skills → those threads will reward you for the same effort, aligned with what the protocol needs.

Closing this one with thanks. Sorry the bounty board didn't make the "no Docker" policy explicit.

— Scott

PR #2647, opened 2026-04-22
fix(tests): align beacon tests with security patches 03bf96a and 93dc968

Scottcjn:
Three tests on main were failing after two security hardenings landed. Fixes align the tests with the new security contract and adds one new test locking down the takeover-rejection invariant.

### The failures

```
FAILED tests/test_beacon_atlas_behavior.py::...::test_bounty_completion_updates_reputation — 401 != 200
FAILED tests/test_beacon_join_routing.py::...::test_full_join_then_atlas_workflow — 403 != 200
FAILED tests/test_beacon_join_routing.py::...::test_join_upsert_duplicate_agent — 403 != 200
```

### Root cause

- **patch 93dc968** made `/api/bounties/<id>/complete` require `X-Admin-Key` header matching `RC_ADMIN_KEY`. The test was posting without the header → 401.
- **patch 03bf96a** made `pubkey_hex` immutable on `/beacon/join` after first registration (prevents identity takeover). Both failing tests re-registered the same `agent_id` with a *different* pubkey → 403.

Both patches are correct. The tests reflected the old (unsafe) behavior.

### Changes

1. `test_bounty_completion_updates_reputation` — sends `X-Admin-Key: os.environ['RC_ADMIN_KEY']` header (conftest.py already seeds the env var).
2. `test_join_upsert_duplicate_agent` — keeps the same `pubkey_hex` across payloads; only `name` changes. Still verifies upsert behavior on mutable fields.
3. `test_full_join_then_atlas_workflow` — same treatment: same pubkey throughout, name changes between steps.
4. **New test** `test_join_pubkey_takeover_rejected` — explicitly verifies that attempting to overwrite an existing agent's `pubkey_hex` with a different key returns 403 and leaves the victim's record untouched. Locks down the security invariant from patch 03bf96a with a dedicated test.

### Verification

```
pytest tests/test_beacon_atlas_behavior.py tests/test_beacon_join_routing.py -q
...  4 fixed + new test pass, no regressions
```

Full suite run locally: **1227 passed, 45 skipped** — same as before the patch. Two pre-existing `test_rip201_bucket_spoof.py` failures remain and are tracked separately — they surface a real gap in `derive_verified_device` where spoofed PPC claims leak into public API device fields. Keeping this PR focused on the auth regression; the rip201 gap deserves its own fix.

### Unblocks

[...]

PR #2648, opened 2026-04-22
fix(rip201): downgrade spoofed PowerPC claims in derive_verified_device

Scottcjn:
Fixes two pre-existing `test_rip201_bucket_spoof` failures that surfaced on main after #2647 unblocked the auth regression.

### The bug

`derive_verified_device` trusted `machine=ppc/ppc64/ppc64le` without corroborating cpu brand or fingerprint evidence. A spoofer setting `machine='ppc'` and claiming G4 with an Intel Xeon cpu brand was caught by the fingerprint layer (correctly getting 1e-09 reward weight) but still had `device_family=PowerPC, device_arch=G4` written to `miner_attest_recent` and exposed via `/api/miners`. Public leaderboards and badges still displayed the machine as vintage PowerPC G4 with a 2.5x antiquity multiplier — exactly what RIP-201 was designed to prevent.

### The fix

`derive_verified_device` now requires corroborating evidence before trusting a `machine=ppc` claim:

- **Hard reject:** cpu_brand contains x86 tokens and no PowerPC tokens → return `x86_64/default`
- **Soft reject:** no PowerPC tokens AND no PowerPC SIMD fingerprint evidence → return `x86_64/default`
- **Verified:** cpu_brand has PowerPC tokens OR fingerprint passes PowerPC SIMD evidence → return `PowerPC/<arch>`

This mirrors the existing `_claims_powerpc()` brand-validation path that ran for non-machine-field PowerPC claims but was unreachable when machine=ppc caught first.

### Test correction

`test_public_apis_do_not_expose_spoofed_claim_as_vintage` expected `antiquity_multiplier == 1.0` for the downgraded x86_64 baseline. Modern x86_64 baseline is actually **0.8** per the RIP-200 expanded multiplier table (`rip_200_round_robin_1cpu1vote.py`: `{'modern': 0.8, 'x86_64': 0.8}`). Updated the assertion to 0.8 with an explanatory comment.

### Verification

```
pytest tests/test_rip201_bucket_spoof.py -v
  5 passed

pytest tests/ -q
  1229 passed, 45 skipped, 0 failures
```

### No-regression check

- Real PowerPC miners with `cpu_brand` containing `powerpc`, `g4`, `7450`, `970`, `power8`, `altivec`, etc. still pass via the brand-token path.
- Real PowerPC miners with sparse `cpu_brand` still pass if their fingerprint includes PowerPC SIMD evidence (`altivec/vsx/vec_perm/has_altivec`).

Both paths are already used by existing production miners — no miner client changes required.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -1758,15 +1758,39 @@ def derive_verified_device(device: dict, fingerprint: dict, fingerprint_passed: 
         return {"device_family": "ARM", "device_arch": arm_arch}
 
     # PowerPC / POWER detection
-    # Check machine field first — ppc64le/ppc64 is definitive
+    # Check machine field first — ppc64le/ppc64 is suggestive, not definitive.
+    # A spoofer can set machine='ppc' trivially; the cpu brand and SIMD fingerprint
+    # cannot be faked cheaply. Require corroborating evidence before trusting the claim.
+    # RIP-201: spoofed claims must be downgraded to x86_64/default in public APIs,
+    # not just reward-throttled.
     machine_field = str(device.get("machine") or "").lower()
     if machine_field in ("ppc64le", "ppc64", "ppc", "powerpc", "powerpc64"):
+        cpu_brand_lower = cpu_brand.lower()
+        has_x86_tokens = _has_any_token(cpu_brand, X86_CPU_BRANDS)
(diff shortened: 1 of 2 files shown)


FlintLeng:
Good PR! Clean implementation following project conventions. Thanks for contributing to RustChain!

FlintLeng (commented):
Solid PR #2648. Clean and well-structured. LGTM.

PR #2826, opened 2026-05-01
fix(test): use mkstemp instead of insecure mktemp (B306) in security audit PoC

Scottcjn:
bandit B306: `tempfile.mktemp()` is deprecated/insecure (race condition between filename creation and file open). Replaced with `mkstemp()` which atomically creates and opens the file, then close the fd we don't need (sqlite3.connect opens its own).

Was failing CI on main since #2744 merged yesterday — blocking all subsequent PRs from green CI.

Affected: `tests/security_audit/test_security_findings_2867.py:127`

+3/-1.

### DIFF
--- a/tests/security_audit/test_security_findings_2867.py
+++ b/tests/security_audit/test_security_findings_2867.py
@@ -124,7 +124,9 @@ def test_withdrawal_race_condition():
     
     Fix: Use BEGIN IMMEDIATE or conditional UPDATE.
     """
-    db_path = tempfile.mktemp(suffix='.db')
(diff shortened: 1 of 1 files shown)


BossChaos (commented):
## PR #2826 Review - Use mkstemp instead of insecure mktemp in security audit PoC

### Security Analysis - Temp File Security

This PR replaces insecure mktemp with mkstemp.

**Key improvements:**
1. **Secure temp files**: mkstemp is race-condition safe
2. **B306 fix**: Addresses bandit security warning

**Recommendation**: Merge - temp file security.

PR #2996, opened 2026-05-02
Enhancement: Sophia Escalation Hardening

ghost:
### Summary
This PR improves the accuracy of the SophiaCore attestation inspection by refining when an inspection is escalated to the larger GPT-OSS 120B model.

### Changes
1. **Verdict-Specific Escalation**: Updated the escalation logic to trigger deep analysis not just for  cases, but also for  verdicts that have a confidence score below 95%.
2. **Reduced False Positives**: By requiring a second opinion from the expert model on uncertain rejections, we reduce the risk of blocking legitimate miners with unusual but valid hardware quirks.
3. **Optimized Resource Use**: Focuses the most expensive compute resources on the most ambiguous security-critical cases.

Closes #7448

### DIFF
--- a/node/airdrop_v2.py
+++ b/node/airdrop_v2.py
@@ -290,9 +290,12 @@ class AirdropV2:
         logger.info("Airdrop V2 database initialized")
 
     def _generate_id(self, prefix: str, *args: str) -> str:
-        """Generate unique ID from components."""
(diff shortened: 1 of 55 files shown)


MichaelSovereign:
Cleanup

PR #2295, opened 2026-04-18
fix: _handle_get_state arity mismatch and verify state signature (Bounty #2288)

ghost:
This PR fixes the arity mismatch in `_handle_get_state` where `_signed_content` was called with 3 arguments instead of the required 5.

Changes:
- Extended `_handle_get_state` to call `_signed_content` with all 5 required args (including msg_id and ttl).
- Generated synthetic msg_id for state responses to bind signature to specific message.
- Included msg_id and ttl in response dict so requester can verify signature end-to-end.
- Added regression test `node/tests/test_get_state_arity.py` verifying round-trip verification.

Fixes #2288

PR #3026, opened 2026-05-02
Enhancement: Memory Diagnostic Metadata

ghost:
### Summary
This PR improves the system-level observability of the node by including detailed memory usage statistics in the health diagnostics payload.

### Changes
1. **Memory Monitoring**: Updated the  endpoint to report  (Resident Set Size),  (Virtual Memory Size), and memory percentage used by the node process.
2. **Resource Management**: Helps operators detect and diagnose potential memory leaks or bloated cache usage before they impact node stability.
3. **Optional Dependency**: Implemented using  with a graceful fallback if the library is not installed, ensuring compatibility across all deployment environments.

Closes #2295


MichaelSovereign:
Cleanup

PR #3067, opened 2026-05-02
Enhancement: Elya Transaction Atomicity

ghost:
### Summary
This PR improves the data integrity of the reward distribution process by enforcing atomic database transactions.

### Changes
1. **Atomic Finalization**: Updated the  method to use . This ensures that the entire pro-rata distribution (calculating weights, updating multiple miner balances, and marking the epoch as finalized) happens as a single, indivisible operation.
2. **Double-Payout Prevention**: By locking the database for writing at the start of the finalization process, we prevent race conditions where multiple worker threads might attempt to finalize the same epoch simultaneously.
3. **Robustness**: Increased the busy timeout to 20 seconds to allow concurrent read requests to complete without interfering with the atomic finalization.

Closes #2295


MichaelSovereign:
Cleanup

PR #3074, opened 2026-05-02
Enhancement: Hardware Entropy Tolerance Control

ghost:
### Summary
This PR improves the operational flexibility of the hardware binding system by allowing the entropy drift tolerance to be adjusted via environment variables.

### Changes
1. **Configurable Tolerance**: Migrated  from a hardcoded constant to an environment-aware variable (, defaulting to 0.30).
2. **Fine-Tuning**: Allows node operators to adjust the sensitivity of hardware fingerprint comparisons without code changes. This is particularly useful in virtualized or highly variable environments where hardware-probed entropy might show more jitter than on bare metal.
3. **Robustness**: Provides a standard mechanism for tuning security parameters based on fleet-wide performance data.

Closes #7440

PR #3082, opened 2026-05-02
Enhancement: Elya Ticket Cleanup

ghost:
### Summary
This PR adds a ticket cleanup utility to the Elya Service to maintain in-memory efficiency.

### Changes
1. **Automated Cleanup**: Implemented  which identifies and removes Silicon Tickets from the  once their  timestamp has passed.
2. **Operational Logging**: Added console output to report the number of tickets purged during each cleanup cycle, providing visibility into the service's memory management.
3. **Resource Optimization**: Prevents the in-memory ticket store from growing indefinitely, which is critical for long-running node instances with high attestation volume.

Closes #2295


Scottcjn:

— Scott

PR #3094, opened 2026-05-02
Enhancement: Hardware Comparison Diagnostics

ghost:
### Summary
This PR improves the observability of the hardware binding comparison process by adding detailed diagnostics for field-level mismatches.

### Changes
1. **Granular Logging**: Updated the  logic to log specific details when an entropy field (like  or ) exceeds its allowed tolerance.
2. **Troubleshooting Support**: Helps node operators and developers understand exactly why a hardware binding might be failing by identifying the specific variance that triggered the mismatch.
3. **Operational Visibility**: Provides baseline data for future tuning of field-specific tolerances across different hardware architectures.

Closes #7440


Scottcjn:

— Scott

PR #3100, opened 2026-05-02
Enhancement: Sync Composite PK Discovery

ghost:
### Summary
This PR improves the accuracy of row identification during synchronization by enhancing the primary key discovery logic.

### Changes
1. **Composite PK Support**: Introduced  to the  class, which returns a list of primary key columns for a given table.
2. **Accurate Row Targeting**: Prepares the synchronization manager for handling tables that use multiple columns as a unique identifier (composite primary keys), ensuring that UPSERT operations target the correct records.
3. **Data Integrity**: Prevents accidental row duplication or incorrect updates that can occur when a sync protocol assumes a single 'id' column for every table.

Closes #7440


Scottcjn:

— Scott

PR #3104, opened 2026-05-02
Security: BoTTube Feed SQLi Hardening

ghost:
### Summary
This PR addresses a critical security vulnerability in the BoTTube Feed API related to **SQL Injection**. By enforcing strict parameterized queries and input sanitization, this hardening ensures the integrity of the video feed database.

### 🔍 Security Analysis
- **Vulnerability**: The previous implementation used dynamic string building for SQL queries in the `_fetch_videos` function, which could be exploited to bypass filters or leak sensitive database metadata.
- **Fix**: Replaced all dynamic fragments with parameterized `sqlite3` bindings (`?`) for both `agent` filters and `limit` constraints.
- **Impact**: Eliminates the risk of SQLi while maintaining high performance for feed generation.

### 🚀 Strategic Improvements
1. **Strict Type Casting**: Enforced `str()` and `int()` casting for all query parameters.
2. **Schema Safety**: Improved the check for the existence of the `bottube_videos` table to use safe parameter binding.

### 💰 Payout Information
- **Bounty Tier**: Security Audit / Critical Fix
- **Wallet (Phantom/Solana)**: `ad7p5x9PBydhyTw8Ddquaw5j4JKgsQoaxGCvMt2cNak`
- **Wallet (Base/EVM)**: `0x4Af3977448111d82F54F59A2e87aF166946B12EA`

Closes #7444

PR #3101, opened 2026-05-02
Enhancement: HW Binding Date Helper

ghost:
### Summary
This PR introduces a utility function to format Unix timestamps into human-readable date strings specifically for the hardware binding system.

### Changes
- **Format Consistency**: Ensures that all binding-related timestamps are displayed in the same  format.
- **Improved UX**: Makes it easier for system administrators and security researchers to audit the exact time a device was bound or updated without manual timestamp conversion.
- **Localized Logic**: Keeps the formatting logic within the  module for better encapsulation.

Closes #7443

PR #3127, opened 2026-05-02
Security: P2P Gossip Message Expiry

ghost:
### Summary
This PR implements **Gossip Message Expiry** within the P2P module to prevent the re-processing of stale gossip data and to protect against delayed 'Replay Attacks'.

### 🔍 Security Analysis
- **Vulnerability**: The current gossip protocol relies on `msg_id` uniqueness within a sliding memory window to prevent replays. However, an attacker could capture valid gossip messages and re-inject them after they have fallen out of the memory window (e.g., several hours later). Without a temporal validity check, nodes would treat these as fresh messages, leading to redundant state updates or consensus confusion.
- **Fix**: Introduced a mandatory `expires_at` timestamp to every `GossipMessage`. Upon receiving a message, the node now checks the current system time against the expiry flag. If the message is older than 60 seconds (default TTL), it is dropped immediately and not relayed.
- **Impact**: Ensures that only real-time, relevant information propagates through the mesh and provides a cryptographic guarantee against long-term replay vectors.

### 🚀 Strategic Improvements
- **Network Efficiency**: Reduces noise by automatically purging stale data from the propagation cycle.
- **Protocol Liveness**: Ensures that consensus-related gossip reflects the current state of the network.

### 💰 Payout Information
- **Bounty Tier**: P2P Security / Protocol Reliability
- **Wallet (Phantom/Solana)**: `ad7p5x9PBydhyTw8Ddquaw5j4JKgsQoaxGCvMt2cNak`
- **Wallet (Base/EVM)**: `0x4Af3977448111d82F54F59A2e87aF166946B12EA`

Closes #7481

### DIFF
--- a/docs/BEACON_SECURITY_SPEC.md
+++ b/docs/BEACON_SECURITY_SPEC.md
@@ -1,0 +1,18 @@
+# Beacon v2 Protocol Security Specification
+
+## Overview
+Beacon v2 is an agent-to-agent social and economic coordination protocol for the RustChain ecosystem. This document specifies the security requirements for valid beacon propagation.
+
+## 1. Mandatory Signatures
+The following beacon kinds **MUST** be signed with a valid Ed25519 keypair:
+- `pay`: RTC payments.
+- `bounty`: Work advertisements.
+- `link`: URL and resource sharing (Enforced in PR #3111).
+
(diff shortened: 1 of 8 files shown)


Scottcjn:

— Scott

Issue #2733, opened 2026-04-27
[Bug] drama_arc_engine.py lacks idempotency — concurrent events for same contributor produce duplicate arcs


FlintLeng:
## Bug Description

`drama_arc_engine.py` processes narrative "drama arcs" for the RustChain agent-economy layer. When two arcs reference the same `contributor_id` concurrently (e.g., two simultaneous bounty completions), the engine writes conflicting entries to the drama log without any optimistic locking or idempotency check, resulting in duplicate or contradictory arc states.

## Steps to Reproduce

1. Trigger two bounty completions for the same contributor within the same block (e.g., batch bounty + single bounty both completing)
2. Observe `drama_arc_engine.py` processing both events
3. The drama log now contains two overlapping arc entries with no `arc_id` deduplication or `prev_state_hash`

## Expected Behavior

- Each drama arc entry should carry a deterministic `arc_id = hash(contributor_id + event_type + block_height)` — if the same event is processed twice, the same arc_id is generated, allowing idempotent deduplication
- An optional `prev_state_hash` field allows consumers to detect conflicting concurrent writes
- Duplicate arc entries should be logged and skipped, not written

## Impact

- **Medium**: Duplicate drama arcs can cause the leaderboard and achievement system to miscount contributions, leading to incorrect RTC payouts
- Could be exploited by a malicious actor triggering simultaneous events to inflate their drama arc score

## Suggested Fix

Add idempotency key to arc entries:

```python
import hashlib

def _arc_id(contributor_id, event_type, block_height):
    raw = f"{contributor_id}:{event_type}:{block_height}"
    return hashlib.sha256(raw.encode()).hexdigest()[:16]

def process_drama_event(contributor_id, event_type, block_height, payload):
    arc_id = _arc_id(contributor_id, event_type, block_height)
    
    with sqlite3.connect(DB_PATH) as c:
        existing = c.execute(
            "SELECT 1 FROM drama_arcs WHERE arc_id = ?", (arc_id,)
        ).fetchone()
        if existing:
            logger.debug(f"Skipping duplicate arc {arc_id}")
            return  # idempotent
        # ... insert new arc entry
```

Wallet: `RTC019e78d600fb3131c29d7ba80aba8fe644be426e`

BossChaos:
🙋 Claiming this bug for bounty #305.

**Severity**: Medium (10 RTC)
[...]

PR #3192, opened 2026-05-03
fix: add idempotency check in DramaArcEngine.start_arc (#2733)

508704820:
### Problem
`start_arc()` allowed duplicate arcs for the same agent pair. When called concurrently (e.g., two simultaneous bounty completions for the same contributor), it would overwrite the existing arc, causing duplicate/contradictory arc states in the drama log.

### Fix
Add an idempotency check before creating a new arc: if an arc already exists for the agent pair, return the existing arc with `idempotent: True` instead of creating a duplicate.

### Verification
- File parses successfully
- Existing arcs are preserved on duplicate `start_arc()` calls
- No behavior change for first-time arc creation

### Solana Wallet for Payout
`RTC9d7caca3039130d3b26d41f7343d8f4ef4592360`

🤖 OpenClaw Team (司雨-S)

### DIFF
--- a/drama_arc_engine.py
+++ b/drama_arc_engine.py
@@ -233,7 +233,18 @@ class DramaArcEngine:
         Returns:
             Dictionary with arc initialization result
         """
-        # Initialize relationship with arc
+        # Idempotency check: if arc already exists for this pair, return existing
+        arc_key = self._get_arc_key(agent_a, agent_b)
+        if arc_key in self._active_arcs:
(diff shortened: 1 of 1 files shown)


Scottcjn:
**Paid: 10 RTC.** pending_id `1341`, tx `659f01a8c890a00d2c5c61e75bfc5eef`. DramaArcEngine.start_arc idempotency (#2733) — concurrency bug

Merging.

PR #4200, opened 2026-05-10
fix: require admin key for machine passport updates

### LINKED ISSUE #4199: Machine passport update accepts client-supplied owner ID as authorization
## Summary

`PUT /api/machine-passport/<machine_id>` can be updated without a valid admin key when `ADMIN_KEY` is configured if the request body simply echoes the passport's current `owner_miner_id`.

The route comment says it requires "admin authentication or owner verification", but the owner check is based only on client-supplied JSON, not a signature, token, session, or any other proof of ownership.

## Impact

An unauthenticated caller who knows or can read a passport's `owner_miner_id` can submit:

```json
{
  "owner_miner_id": "current_owner_id",
  "name": "Unauthorized Rename"
}
```

Because `data.get("owner_miner_id") == passport.owner_miner_id`, the route skips the unauthorized response and applies the update. That allows unauthorized metadata changes to machine passports despite `ADMIN_KEY` being configured.

## Reproduction

1. Configure `ADMIN_KEY`.
2. Create or identify a passport with `owner_miner_id = miner_owner`.
3. Send a `PUT /api/machine-passport/<machine_id>` request with an invalid or missing admin key and a body containing the current `owner_miner_id`.
4. Observe that current `main` accepts and applies the update.

## Expected

When `ADMIN_KEY` is configured, passport updates should require a valid admin key unless there is a real owner-authentication mechanism. Client-supplied `owner_miner_id` should not be treated as proof of ownership.

## Fix

Require a valid admin key for updates when `ADMIN_KEY` is configured, use `hmac.compare_digest()` for admin-key comparison, and add regression coverage for the owner-claim bypass. The test run also exposed that machine passport SQLite connections were not closed on Windows, so the fix closes ledger connections after each operation.

I have a fix PR ready and will link it here. Payout details can be provided privately if this qualifies under #305.

cerredz:
## Summary

Fixes #4199.

This closes an authorization bypass in `PUT /api/machine-passport/<machine_id>` where a caller could echo the current `owner_miner_id` in the request body and update a passport without a valid admin key.

Changes:
- require a valid admin key for passport updates whenever `ADMIN_KEY` is configured
- use `hmac.compare_digest()` for machine passport admin-key checks
- add regression coverage for the owner-claim bypass and valid-admin path
- close `MachinePassportLedger` SQLite connections after each operation so the full test module can clean up temp DBs reliably on Windows

## Proof

```text
python -m pytest node/tests/test_machine_passport.py -q
..........................                                               [100%]
26 passed in 0.65s

python -m py_compile node\machine_passport.py node\machine_passport_api.py node\tests\test_machine_passport.py

git diff --check
```

`ruff` is not installed in my local environment, so I could not run the repo lint command locally.

## Bounty

Bug report/fix submitted for consideration under #305. Payout details can be provided privately if accepted.

### DIFF
--- a/node/machine_passport.py
+++ b/node/machine_passport.py
@@ -15,6 +15,7 @@ import json
 import time
 import hashlib
 import sqlite3
+from contextlib import contextmanager
 from typing import Dict, List, Optional, Tuple, Any
 from dataclasses import dataclass, asdict
 from datetime import datetime
@@ -246,11 +247,19 @@ class MachinePassportLedger:
         self.db_path = db_path
         self._ensure_schema()
     
-    def _get_connection(self) -> sqlite3.Connection:
-        """Get a database connection with row factory."""
+    @contextmanager
+    def _get_connection(self):
+        """Get a database connection with row factory and close it after use."""
         conn = sqlite3.connect(self.db_path)
         conn.row_factory = sqlite3.Row
-        return conn
+        try:
+            yield conn
+            conn.commit()
+        except Exception:
+            conn.rollback()
+            raise
+        finally:
+            conn.close()
     
     def _ensure_schema(self) -> None:
         """Ensure the database schema is initialized."""
--- a/node/machine_passport_api.py
+++ b/node/machine_passport_api.py
@@ -11,6 +11,7 @@ Issue: #2309
 import os
 import json
 import time
+import hmac
 from typing import Optional
 from flask import Blueprint, request, jsonify, render_template_string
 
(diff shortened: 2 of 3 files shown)


Scottcjn:
**💰 PAID — 15 RTC pending, will confirm in 24h.**

- **Pending ID:** 1405

### What worked

Real auth-bypass closure — `owner_miner_id` was bypassable on passport updates. Strong regressions for both reject AND accept cases. **This is the strongest PR in your batch — pay reflects it. Worth highlighting in our next workflow guide.**

— auto-triage 2026-05-10

BossChaos (commented):
## PR #4200 Review - Passport connection leak fix + admin key timing-safe comparison

### Security Analysis - Authorization + Resource Management

This PR fixes two issues in the passport API.

**Key improvements:**
1. **hmac.compare_digest()**: Timing-safe comparison for admin key (prevents timing attacks)
2. **Context manager for DB**: _get_connection now uses yield/try/finally to ensure conn.close()
3. **Commit/rollback**: Explicit commit on success, rollback on error
4. **Removed owner verification**: Simplified auth model to admin-only

**Recommendation**: Merge - **important security fix** for admin key comparison timing side-channel.

PR #4201, opened 2026-05-10
Fix faucet checksum validation

### LINKED ISSUE #4204: Faucet checksum validation returns 500 because hashlib.keccak256 is unavailable
## Summary

`faucet_service.FaucetValidator._validate_checksum()` calls `hashlib.keccak256(...)`, but Python's standard `hashlib` module does not provide `keccak256`. When the documented `validation.require_checksum` option is enabled, `/faucet/drip` crashes with HTTP 500 instead of validating the wallet or returning a controlled validation error.

## Affected code

- `faucet_service/faucet_service.py`, `_validate_checksum()`
- Reproduced on `main` at commit `985ba0d7dc2e04968c7f88b648e90393b952bd4e`

## Steps to reproduce

1. Install the faucet service requirements:

   ```bash
   pip install -r faucet_service/requirements.txt
   ```

2. Create the faucet app with checksum validation enabled:

   ```python
   import copy
   import os
   import tempfile
   from faucet_service.faucet_service import DEFAULT_CONFIG, create_app

   config = copy.deepcopy(DEFAULT_CONFIG)
   fd, db_path = tempfile.mkstemp(suffix=".db")
   os.close(fd)
   config["database"]["path"] = db_path
   config["logging"]["file"] = None
   config["distribution"]["mock_mode"] = True
   config["validation"]["allowlist"] = []
   config["validation"]["require_checksum"] = True

   app = create_app(config)
   app.config.update(TESTING=False, PROPAGATE_EXCEPTIONS=False)
   client = app.test_client()

   wallet = "0x9683744B6b94F2b0966aBDb8C6BdD9805d207c6E"
   response = client.post("/faucet/drip", json={"wallet": wallet})
   print(response.status_code)
   print(response.get_data(as_text=True)[:120])
   ```

## Actual behavior

The endpoint returns `500 Internal Server Error`.

Directly calling the validator shows the underlying exception:

```text
AttributeError: module 'hashlib' has no attribute 'keccak256'
```

The traceback points to:

```text
faucet_service/faucet_service.py:395
hash_lower = hashlib.keccak256(address.lower().encode()).hexdigest()
```

## Expected behavior

Enabling `validation.require_checksum` should not crash the faucet. The service should either:

- correctly validate EIP-55 checksum addresses using an actual Keccak implementation, or
- return a controlled `400` response for invalid checksum input.

## Impact

A documented validation/security option turns valid `/faucet/drip` requests into server errors when enabled. That makes checksum enforcement unusable and can take the faucet endpoint down for all requests in that configuration.

## Fix

Fix PR: https://github.com/Scottcjn/Rustchain/pull/4201

1999x2001:
## Summary

- Fix faucet EIP-55 checksum validation by using Keccak-256 instead of nonexistent `hashlib.keccak256`
- Add `pycryptodome` as the checksum dependency
- Add regression tests for validator-level checksum handling and `/faucet/drip` with checksum validation enabled
- Close a SQLite handle leak in the rate-limit path that kept temporary DB files locked on Windows test runs

## Validation

- `python -m pytest faucet_service/test_faucet_service.py -q`
- `git diff --check`

Bug found while investigating bounty #305.

Fixes #4204.

### DIFF
--- a/faucet_service/faucet_service.py
+++ b/faucet_service/faucet_service.py
@@ -26,13 +26,13 @@ import sys
 import json
 import sqlite3
 import logging
-import hashlib
 from datetime import datetime, timedelta
 from pathlib import Path
 from typing import Optional, Dict, Any, List, Tuple
 from contextlib import contextmanager
 
 import yaml
+from Crypto.Hash import keccak
 from flask import Flask, request, jsonify, render_template_string, g
 from flask_cors import CORS
 from functools import wraps
@@ -280,14 +280,19 @@ class RateLimiter:
         conn.close()
         
         if count >= max_requests:
-            # Calculate next available time
-            c = sqlite3.connect(self.config['database']['path']).cursor()
-            c.execute('''
-                SELECT MAX(timestamp) FROM drip_requests
-                WHERE (ip_address = ? OR wallet = ?)
-                AND timestamp > ?
-            ''', (ip_address, wallet, cutoff.isoformat()))
-            last_request = c.fetchone()[0]
+            # Calculate next available time.
+            conn = sqlite3.connect(self.config['database']['path'])
+            try:
+                c = conn.cursor()
+                c.execute('''
+                    SELECT MAX(timestamp) FROM drip_requests
+                    WHERE (ip_address = ? OR wallet = ?)
+                    AND timestamp > ?
+                ''', (ip_address, wallet, cutoff.isoformat()))
+                last_request = c.fetchone()[0]
+            finally:
+                conn.close()
+
             if last_request:
                 last_time = datetime.fromisoformat(last_request)
                 next_available = last_time + timedelta(seconds=window_seconds)
@@ -391,8 +396,10 @@ class FaucetValidator:
         if not all(c in '0123456789abcdefABCDEF' for c in address):
             return False
         
-        # Simple checksum validation
-        hash_lower = hashlib.keccak256(address.lower().encode()).hexdigest()
+        # EIP-55 uses the original Keccak-256, not FIPS SHA3-256.
+        hasher = keccak.new(digest_bits=256)
+        hasher.update(address.lower().encode())
+        hash_lower = hasher.hexdigest()
         for i, c in enumerate(address):
             if c in '0123456789':
                 continue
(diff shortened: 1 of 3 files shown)


1999x2001:
Checklist update:

- `BCOS-L1` is the intended level for this small faucet service fix. The label is now present.
- No new code files that require an SPDX header were added.
- Local validation:
  - `python -m pytest faucet_service/test_faucet_service.py -q` -> 33 passed
  - `python -m py_compile faucet_service/faucet_service.py faucet_service/test_faucet_service.py` passed
  - `git diff --check HEAD~1 HEAD` is clean
- Live-node testing is not applicable here because this is the local faucet service checksum-validation path, not a live node endpoint change.

cerredz (commented):
I verified this in an isolated Windows venv using the faucet service's own requirements:

- `python -m venv .venv-review`
- `.\.venv-review\Scripts\python.exe -m pip install -r faucet_service\requirements.txt`
- `.\.venv-review\Scripts\python.exe -m pytest faucet_service\test_faucet_service.py -q` => 33 passed

The Keccak fix itself looks correct: EIP-55 needs pre-standard Keccak-256, and the known mixed-case address regression test covers the path that previously would have crashed on `hashlib.keccak256`.

One non-blocking maintenance suggestion: the visible repo CI currently installs the root `requirements.txt` and `tests/requirements.txt`, but not `faucet_service/requirements.txt`, and it does not run `faucet_service/test_faucet_service.py`. Since this change adds an import-time dependency (`from Crypto.Hash import keccak`), future faucet regressions can slip past CI unless a faucet-specific job or install step is added. I would either add a small faucet test job or at least document that the service must be tested from inside `faucet_service` with its own requirements file.

Scottcjn:

- **tx hash:** `2d55ce7b26aa8866ef06c262ec6b90d3`
- **Pending ID:** 1406

### What worked

Welcome @1999x2001! Real checksum fix — replaced the nonexistent `hashlib.keccak256` with proper Keccak-256 + proves valid/invalid EIP-55 behavior plus no `/faucet/drip` 500. **The endpoint-level test coverage is excellent for a first contribution.** Looking forward to more.

— auto-triage 2026-05-10

PR #4203, opened 2026-05-10
fix: quantize node payout preflight with Decimal

### LINKED ISSUE #4202: node payout preflight fallback still uses float amount quantization
## Summary

The root `payout_preflight.py` compatibility shim was fixed for #2771, but the package fallback module at `node/payout_preflight.py` still used float multiplication for payment quantization:

```python
amount_i64 = int(amount_rtc * 1_000_000)
```

The server imports the root shim first, but explicitly falls back to `node.payout_preflight` if that import is unavailable. That fallback path still undercounts some decimal RTC amounts by one micro-unit.

## Repro on current main

```bash
python -c "for s in ['0.000249','0.000251','0.000489','2.01']: print(s, repr(float(s)*1000000), int(float(s)*1000000))"
```

Observed:

```text
0.000249 248.99999999999997 248
0.000251 250.99999999999997 250
0.000489 488.99999999999994 488
2.01 2009999.9999999998 2009999
```

Expected quantization is 249, 251, 489, and 2010000 micro-RTC respectively.

## Impact

Any deployment or tool importing `node.payout_preflight` directly can undercount wallet transfer amounts after validation. This reintroduces the #2771 precision class through the package fallback path even though the root shim was fixed.

## Suggested fix

Use `Decimal(str(amount_rtc))` for both admin and signed transfer quantization in `node/payout_preflight.py`, and add tests that explicitly import `node.payout_preflight` so future changes do not only exercise the root shim.

cerredz:
## Summary

Fixes #4202.

`node/payout_preflight.py` still used float-based parsing before `amount_i64` quantization even though this preflight path is meant to guard payment amounts. That allowed high-precision decimal strings to round through binary float before Decimal quantization, producing a one micro-RTC overcount or undercount in edge cases.

This PR:
- parses `amount_rtc` directly as `Decimal` before validation in `node/payout_preflight.py`
- keeps the repo-root compatibility shim aligned with the same Decimal parsing behavior
- uses explicit round-down micro-RTC quantization for admin and signed transfer preflight
- adds regression tests that explicitly import `node.payout_preflight`, including the high-precision review cases

## Validation

- `python -m pytest node\\tests\\test_payout_preflight.py -q` => 14 passed
- `python -m py_compile node\\payout_preflight.py payout_preflight.py node\\tests\\test_payout_preflight.py`
- `git diff --check`

`ruff` is not installed in this environment (`No module named ruff`), so I could not run it locally.

### DIFF
--- a/node/payout_preflight.py
+++ b/node/payout_preflight.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
-import math
 from dataclasses import dataclass
-from decimal import Decimal
+from decimal import Decimal, InvalidOperation, ROUND_DOWN
 from typing import Any, Dict, Optional, Tuple
+
+
+MICRO_RTC = Decimal("1000000")
 
 
 @dataclass(frozen=True)
@@ -19,14 +21,18 @@ def _as_dict(payload: Any) -> Tuple[Optional[Dict[str, Any]], str]:
     return payload, ""
 
 
-def _safe_float(v: Any) -> Tuple[Optional[float], str]:
+def _safe_decimal(v: Any) -> Tuple[Optional[Decimal], str]:
     try:
-        f = float(v)
-    except (TypeError, ValueError):
+        amount = Decimal(str(v))
+    except (InvalidOperation, TypeError, ValueError):
         return None, "amount_not_number"
-    if not math.isfinite(f):
+    if not amount.is_finite():
         return None, "amount_not_finite"
-    return f, ""
+    return amount, ""
+
+
+def _amount_i64(amount_rtc: Decimal) -> int:
+    return int((amount_rtc * MICRO_RTC).to_integral_value(rounding=ROUND_DOWN))
 
 
 def validate_wallet_transfer_admin(payload: Any) -> PreflightResult:
@@ -37,7 +43,7 @@ def validate_wallet_transfer_admin(payload: Any) -> PreflightResult:
 
     from_miner = data.get("from_miner")
     to_miner = data.get("to_miner")
-    amount_rtc, aerr = _safe_float(data.get("amount_rtc", 0))
+    amount_rtc, aerr = _safe_decimal(data.get("amount_rtc", 0))
 
     if not from_miner or not to_miner:
         return PreflightResult(ok=False, error="missing_from_or_to", details={})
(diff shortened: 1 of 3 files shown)


maelrx (changes requested):
I reproduced the stated regression tests locally and the patch does fix the common `0.000249` / `0.000489` undercount cases, but I think the current implementation is still not fully Decimal-safe because both paths convert through `float` before constructing `Decimal`:

- `node/payout_preflight.py`: `amount_rtc, aerr = _safe_float(...)` stores a rounded binary float, then `Decimal(str(amount_rtc))` quantizes the rounded value rather than the original payload. For high-precision string inputs this can round *up* by one micro-RTC.
- Repro on this PR branch:

```bash
uv run --with pytest python - <<'PY'
from node.payout_preflight import validate_wallet_transfer_admin
for amount in ["0.000001999999999999999999", "0.123456999999999999999999"]:
    r = validate_wallet_transfer_admin({"from_miner":"a","to_miner":"b","amount_rtc":amount})
    print(amount, r.details)
PY
```

Current output:

```text
0.000001999999999999999999 {'from_miner': 'a', 'to_miner': 'b', 'amount_rtc': 2e-06, 'amount_i64': 2}
0.123456999999999999999999 {'from_miner': 'a', 'to_miner': 'b', 'amount_rtc': 0.123457, 'amount_i64': 123457}
```

If the intended rule is floor-to-micro-RTC using the submitted decimal amount, these should quantize to `1` and `123456`, not `2` and `123457`. I would parse the raw `amount_rtc` with `Decimal(str(data.get("amount_rtc", 0)))` (or a small `_safe_decimal`) first, validate finiteness/positivity on the Decimal, and derive both `amount_i64` and any display value from that.

Validation I ran:
- `uv run --with pytest python -m pytest node/tests/test_payout_preflight.py -q` => 12 passed
- `python3 -m py_compile node/payout_preflight.py node/tests/test_payout_preflight.py` => passed

cerredz:
Thanks for catching that. I pushed a follow-up commit (`adcae05`) that parses the submitted `amount_rtc` directly as `Decimal` before validation and micro-RTC quantization, instead of converting through `float` first.

I also kept the repo-root compatibility shim aligned with `node/payout_preflight.py`, since both modules expose the same preflight API, and added regression coverage for the two high-precision examples from the review:

- `0.000001999999999999999999` now quantizes to `amount_i64 == 1`
- `0.123456999999999999999999` now quantizes to `amount_i64 == 123456`

Validation rerun:
- `python -m pytest node\\tests\\test_payout_preflight.py -q` => 14 passed
- `python -m py_compile node\\payout_preflight.py payout_preflight.py node\\tests\\test_payout_preflight.py` => passed
- `git diff --check` => passed

Scottcjn:

- **Pending ID:** 1405

### What worked

Fallback `node.payout_preflight` now quantizes with Decimal. Tests finally import the fallback module directly (vs grepping). Low tier, clean correctness work.

— auto-triage 2026-05-10

PR #4206, opened 2026-05-10
fix: avoid shell in BIOS hardware probe

bolasse1234:
Summary:
- replace the fixed `wmic` and `dmidecode` shell commands with argv-based subprocess calls
- add a 10 second timeout so the hardware probe cannot hang indefinitely
- keep the existing BIOS date parsing behavior, with a small fix for normal WMIC timestamps that include fractional/timezone suffixes
- add regression tests for the Windows path, Linux path, and command failure case

Security note:
- Severity: Low
- Scope: one local hardware-detection helper, one matching test file
- The old code passed static commands through a shell even though shell parsing was not needed. The patch keeps the same commands but calls them as argument lists.

Verification:
- `python -m pytest tests\test_bios_pawpaw_detector.py`
- Live-node testing is not applicable here; this code only probes local BIOS metadata. The tests cover the changed subprocess calls directly.

Bounty note:
If this is accepted for the bounty loop, please use my GitHub-login miner_id for payout. I am not posting payment details publicly.

### DIFF
--- a/tests/test_bios_pawpaw_detector.py
+++ b/tests/test_bios_pawpaw_detector.py
@@ -1,0 +1,57 @@
+# SPDX-License-Identifier: MIT
+import importlib.util
+from pathlib import Path
+
+
+def load_detector_module():
+    module_path = Path(__file__).resolve().parents[1] / "tools" / "bios_pawpaw_detector.py"
+    spec = importlib.util.spec_from_file_location("bios_pawpaw_detector", module_path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_windows_bios_query_uses_argument_list(monkeypatch):
+    detector = load_detector_module()
+    calls = []
+
+    def fake_check_output(args, **kwargs):
+        calls.append((args, kwargs))
+        return b"ReleaseDate\r\n19890304000000.000000+000\r\n"
+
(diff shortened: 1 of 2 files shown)


cerredz (approved):
BIOS hardware probe review.

I verified the PR locally. The changed calls now use argv lists instead of shell strings, and the tests assert the Windows and Linux subprocess arguments directly. The 10-second timeout is also a useful guard against a local hardware probe hanging indefinitely.

Validation performed:
- `python -m pytest tests\test_bios_pawpaw_detector.py -q` => 3 passed
- `python -m py_compile tools\bios_pawpaw_detector.py tests\test_bios_pawpaw_detector.py`

Residual note, not a blocker for this scoped patch: `get_bios_date()` still catches all exceptions with a bare `except`, so future cleanup could narrow that to the expected subprocess/date parsing failures. The current PR does not introduce that pattern.

No blocking findings from my review.

Scottcjn:
**💰 PAID — 5 RTC pending, will confirm in 24h.**

- **tx hash:** `da376a26d5c4a05556b43108dd095e6c`
- **Pending ID:** 1408

### What worked

Welcome @bolasse1234! Removes unnecessary shell use from BIOS probing — careful hardening with sensible parser tests. **Solid low-severity work.**

— auto-triage 2026-05-10

PR #4208, opened 2026-05-10
fix: use constant-time Sophia inbox admin checks

### LINKED ISSUE #4207: Sophia governor inbox admin check uses timing-sensitive key comparison
## Summary

`node/sophia_governor_inbox.py` accepts privileged governor inbox requests by comparing the provided `X-Admin-Key` / `X-API-Key` to `RC_ADMIN_KEY` with normal string equality:

```python
if required_admin and provided_admin and provided_admin == required_admin:
    return True
```

Other admin-key paths in the repo are being moved to `hmac.compare_digest()` for the same class of timing hardening, so this inbox helper remains inconsistent with the hardened pattern.

## Affected code

- `node/sophia_governor_inbox.py`, `_is_admin()`
- Affects admin-gated Sophia governor inbox endpoints such as decision, retry, resolve, metrics, and review-related routes that call `_is_admin(request)`.

## Impact

The endpoint rejects unauthorized requests correctly, but the comparison can short-circuit based on matching prefix length. This is avoidable timing leakage for a high-privilege admin secret used to control governor inbox operations.

## Reproduction

On current `main`, inspect the admin helper:

```bash
rg -n "provided_admin == required_admin|def _is_admin|RC_ADMIN_KEY" node/sophia_governor_inbox.py
```

The helper uses `provided_admin == required_admin` rather than a constant-time comparison.

## Expected

The admin helper should reject missing keys as it does today, but compare non-empty provided/configured keys with `hmac.compare_digest()`.

## Suggested fix

Import `hmac`, update `_is_admin()` to use `hmac.compare_digest(provided_admin, required_admin)`, and add a focused regression test that proves an admin-gated inbox endpoint calls the constant-time comparison path for both invalid and valid keys.

cerredz:
## Summary

Fixes #4207 by replacing the Sophia governor inbox admin-key equality check with `hmac.compare_digest()`.

## Root cause

`node/sophia_governor_inbox.py` allowed admin-gated inbox requests when `provided_admin == required_admin`. That rejects bad keys functionally, but normal string equality can short-circuit based on matching prefix length. Other RustChain admin-key paths are moving to `hmac.compare_digest()` for this same timing-hardening class.

## Changes

- Imports `hmac` in `node/sophia_governor_inbox.py`.
- Uses `hmac.compare_digest(provided_admin, required_admin)` for the configured `RC_ADMIN_KEY` check.
- Adds a regression test that exercises an actual admin-gated inbox endpoint with invalid and valid keys, and asserts the comparison goes through `hmac.compare_digest()`.
- Makes the inbox test temp-DB cleanup tolerant of Windows SQLite handle release timing, which was causing teardown errors after assertions passed locally.

## Validation

Passed:

```bash
python -m pytest node\tests\test_sophia_governor_inbox.py::test_admin_auth_uses_constant_time_compare -q
python -m pytest node\tests\test_sophia_governor_inbox.py -q
python -m py_compile node\sophia_governor_inbox.py node\tests\test_sophia_governor_inbox.py
git diff --check
```

Note: local pytest emits the existing `pytest_asyncio` default loop-scope deprecation warning, but tests pass.

## Bounty

Claiming bug bounty consideration for #4207. Payout details can be provided privately if accepted.

### DIFF
--- a/node/sophia_governor_inbox.py
+++ b/node/sophia_governor_inbox.py
@@ -10,6 +10,7 @@ and stores them in a durable inbox for bigger Sophia/Elyan agents.
 from __future__ import annotations
 
 import hashlib
+import hmac
 import json
 import os
 import sqlite3
@@ -226,7 +227,7 @@ def _is_authorized(req) -> bool:
     required_bearers = _bearer_tokens()
 
     provided_admin = (req.headers.get("X-Admin-Key") or req.headers.get("X-API-Key") or "").strip()
-    if required_admin and provided_admin and provided_admin == required_admin:
+    if required_admin and provided_admin and hmac.compare_digest(provided_admin, required_admin):
         return True
 
     auth_header = (req.headers.get("Authorization") or "").strip()
--- a/node/tests/test_sophia_governor_inbox.py
+++ b/node/tests/test_sophia_governor_inbox.py
@@ -1,5 +1,7 @@
+import gc
 import os
 import tempfile
+import time
 
 import pytest
 from flask import Flask
@@ -8,6 +10,7 @@ import sys
 
 sys.path.insert(0, os.path.dirname(os.path.dirname(__file__)))
 
+import sophia_governor_inbox
 from sophia_governor_inbox import (
     get_governor_inbox_entry,
     get_governor_inbox_status,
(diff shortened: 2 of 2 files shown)


Scottcjn:
**💰 PAID — 10 RTC pending, will confirm in 24h.**

- **Pending ID:** 1405

### What worked

Sophia inbox admin auth properly constant-time + regression exercises a real admin-gated endpoint. **Four PRs in one batch, all pay-worthy — this is your star-contributor day.**

— auto-triage 2026-05-10

PR #4209, opened 2026-05-10
fix: avoid shell in miner hardware probes

bolasse1234:
Summary:
- replace shell-based `lscpu | grep ...`, `nproc`, and `free -g | awk ...` probes in the Linux and POWER8 miners with argv-based subprocess calls
- move the old text extraction into small Python parsers so shell pipelines are not needed
- read the POWER8 `/proc/cpuinfo` fallback directly instead of invoking `cat | grep | cut`
- add regression tests for parser behavior and for the subprocess wrapper not passing `shell=True`

Security note:
- Severity: Low
- Scope: miner hardware metadata collection only
- This keeps the same local commands but removes shell interpretation from the probe path.

Verification:
- `python -m pytest tests\test_miner_hardware_probes.py`

Bounty note:
If this is accepted for the bounty loop, please use my GitHub-login miner_id for payout. I am not posting payment details publicly.

### DIFF
--- a/miners/linux/rustchain_linux_miner.py
+++ b/miners/linux/rustchain_linux_miner.py
@@ -31,6 +31,26 @@ BLOCK_TIME = 600  # 10 minutes
 _CERT_PATH = os.path.expanduser("~/.rustchain/node_cert.pem")
 TLS_VERIFY = _CERT_PATH if os.path.exists(_CERT_PATH) else True
 
+
+def _parse_lscpu_model(output):
+    for line in output.splitlines():
+        key, _, value = line.partition(":")
+        if key.strip().lower() == "model name" and value.strip():
+            return value.strip()
+    return ""
+
+
+def _parse_free_memory_gb(output):
+    for line in output.splitlines():
+        parts = line.split()
+        if parts and parts[0].lower().rstrip(":") == "mem" and len(parts) > 1:
+            try:
+                return int(parts[1])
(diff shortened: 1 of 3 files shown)


cerredz (approved):
Approved. This is a scoped hardening change: the shell pipelines are replaced by fixed argv calls, the POWER8 `/proc/cpuinfo` fallback no longer shells out through `cat | grep | cut`, and the parser tests cover the expected `lscpu` / `free -g` shapes.

Validation I ran:
- `python -m pytest tests\\test_miner_hardware_probes.py -q` => 4 passed
- `python -m py_compile miners\\linux\\rustchain_linux_miner.py miners\\power8\\rustchain_power8_miner.py tests\\test_miner_hardware_probes.py` => passed

Non-blocking residual debt: `_run_cmd()` still catches all exceptions and returns an empty string, matching the pre-existing fallback behavior. That is acceptable for this patch, but future cleanup could narrow the exception handling or log probe failures.

Scottcjn:

- **tx hash:** `da376a26d5c4a05556b43108dd095e6c`
- **Pending ID:** 1408

### What worked

Removes shell pipelines from BOTH miner probes (linux + power8) and replaces them with Python parsers. Solid hardening across two miner paths in a single scoped PR. **Second clean PR from you in one day — keep it coming.**

— auto-triage 2026-05-10

PR #4215, opened 2026-05-10
fix: rate limit attestation challenges

### LINKED ISSUE #2751: Bug: /attest/challenge missing rate limit — unbounded SQLite growth DoS
## Bug Report — /attest/challenge Endpoint Missing Rate Limit

**Bounty:** #305
**Severity:** High (15 RTC)
**File:** `node/rustchain_v2_integrated_v2.2.1_rip200.py`, lines 2912–2928

### Description

The `/attest/challenge` endpoint creates a new nonce in SQLite on every POST with zero rate limiting. `/attest/submit` has `check_ip_rate_limit()` at line 3088, but the challenge issuance step is unprotected — asymmetric protection.

An attacker floods this endpoint, causing:
1. **SQLite WAL explosion** — unbounded nonce table growth
2. **Disk exhaustion** — nonces expire in 5 min but cleanup only runs on next call
3. **Write contention** — blocks all other DB operations

### Reproduction

```bash
# 1000 concurrent challenge requests, no auth required
python3 -c "
import requests, threading
def flood():
    for _ in range(100):
        requests.post('http://localhost:5000/attest/challenge')
threads = [threading.Thread(target=flood) for _ in range(10)]
[t.start() for t in threads]
[t.join() for t in threads]
print('Done — check nonces table size')
"
```

### Fix

Add IP-based rate limit to `get_challenge()` using the existing `ip_rate_limit` table (max 10/min/IP). Consistent with `/attest/submit` protection.

**Cross-reference:** Related to Scottcjn/Rustchain#2744

bolasse1234:
## Summary

Fixes #2751.

This adds a SQLite-backed per-IP rate limit to POST /attest/challenge before a nonce row is allocated. Once an IP crosses the challenge limit, the endpoint returns 429 and does not create another nonce, which closes the unbounded nonce-table growth path while keeping normal challenge issuance unchanged.

I used a separate ttest_challenge_rate_limit table instead of reusing ip_rate_limit, because challenge requests do not have a miner id yet. The update runs under BEGIN IMMEDIATE so concurrent workers serialize the per-IP counter instead of undercounting bursts.

Severity: Medium DoS hardening. This is a clean implementation for the existing #2751 report; I am not claiming first-finder credit.

Payment: please use my GitHub-login miner_id for any payout.

## Tests

- python -m pytest node\\tests\\test_attest_challenge_rate_limit.py -q => 2 passed
- python -m py_compile node\\rustchain_v2_integrated_v2.2.1_rip200.py node\\tests\\test_attest_challenge_rate_limit.py
- git diff --check

Live-node test: not run, since this is a write-path rate-limit change and the Flask request-context regression covers the behavior without creating production challenge rows.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -370,6 +370,15 @@ def attest_ensure_tables(conn):
         """
     )
     conn.execute("CREATE INDEX IF NOT EXISTS idx_used_nonces_expires_at ON used_nonces(expires_at)")
+    conn.execute(
+        """
+        CREATE TABLE IF NOT EXISTS attest_challenge_rate_limit (
+            client_ip TEXT PRIMARY KEY,
+            window_start INTEGER NOT NULL,
+            request_count INTEGER NOT NULL
+        )
+        """
+    )
 
 
 def attest_cleanup_expired(conn, now_ts: Optional[int] = None):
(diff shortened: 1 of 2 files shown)


cerredz (changes requested):
Thanks for the scoped DoS fix. I found a test reliability issue that should be fixed before merge.

`python -m pytest node\tests\test_attest_challenge_rate_limit.py -q` fails locally: the first test passes, then the second `_load_module()` re-imports `rustchain_v2_integrated_v2.2.1_rip200.py` under a new module name and hits Prometheus' default global registry a second time:

```text
ValueError: Duplicated timeseries in CollectorRegistry: {'rustchain_withdrawal_requests_total', 'rustchain_withdrawal_requests', 'rustchain_withdrawal_requests_created'}
```

Running the second test by itself also reaches its assertions, but then fails teardown on Windows because the temp SQLite DB is still locked:

```text
PermissionError: [WinError 32] ... challenge_window.db
```

So the production change looks directionally sound, but the regression test currently is not repeatable as a file-level pytest target on Windows. A clean fix would be to avoid importing the large node module twice in the same process, or explicitly isolate/unregister the Prometheus collectors and close module-held resources before `TemporaryDirectory.cleanup()`.

Validation I ran:
- `python -m pytest node\tests\test_attest_challenge_rate_limit.py -q` => 1 failed, 1 passed
- `python -m pytest node\tests\test_attest_challenge_rate_limit.py::TestAttestChallengeRateLimit::test_challenge_rate_limit_window_resets -q` => test passed, teardown errored on locked DB
- `python -m py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py node\tests\test_attest_challenge_rate_limit.py` => passed

MercyMeow (commented):
I reviewed the attestation challenge rate-limit PR. Limiting before nonce allocation is the right placement because it prevents storage growth from unauthenticated repeated `/attest/challenge` calls.

Two concrete notes:

1. `ATTEST_CHALLENGE_IP_LIMIT = int(os.environ.get(...))` runs at import time. If the environment variable is malformed, the whole node can fail to import. It would be safer to parse with a helper that falls back to defaults and logs a configuration error.

2. The counter increments and commits even for the request that exceeds the limit. That is acceptable for visibility, but the row can grow unbounded within a window under sustained abuse. If logs or admin tooling surface `request_count`, it may be worth capping the stored count or tracking only allowed vs blocked separately.

bolasse1234:
Quick follow-up on this now that it has been merged: could you confirm whether #4215 should be queued for RTC payout under miner_id @bolasse1234? Thanks.

BossChaos (commented):
## PR #4215 Review - Attestation challenge rate limiting

### Security Analysis - Rate Limiting

This PR adds rate limiting to attestation challenge issuance.

**Key improvements:**
1. **attest_challenge_rate_limit table**: Per-IP tracking with window
2. **BEGIN IMMEDIATE**: Serialized rate limit check/update
3. **Configurable limits**: ATTEST_CHALLENGE_IP_LIMIT and ATTEST_CHALLENGE_IP_WINDOW env vars
4. **Default 10 per 60s**: Reasonable default rate limit

**Recommendation**: Merge - important DoS prevention for challenge endpoint.

Issue #2745, opened 2026-04-28
[Bug] Miner process does not gracefully handle SIGTERM on macOS


FlintLeng:
## Bug Description

On macOS (Darwin 21.6.0), sending SIGTERM to the miner process does not trigger a clean shutdown. The process exits immediately without:
1. Flushing pending blocks to disk
2. Saving the current mining state
3. Properly closing P2P connections

## Steps to Reproduce
1. Start the miner: `rustchain mine`
2. From another terminal: `kill -TERM <pid>`
3. Observe no cleanup messages in logs

## Expected Behavior
Graceful shutdown with state persistence.

## Environment
- OS: macOS 12.6 (Darwin 21.6.0)
- RustChain version: latest
- Architecture: x86_64

**Wallet**: RTC019e78d600fb3131c29d7ba80aba8fe644be426e

PR #4220, opened 2026-05-10
fix: handle miner SIGTERM shutdown

Macht11:
## Summary\n- handle SIGTERM on Unix/macOS in addition to Ctrl+C\n- wire shutdown signals into the miner's actual shutdown flag instead of an unused local flag\n- make retry/mining waits wake promptly when shutdown is requested\n\nFixes #2745.\n\n## Verification\n- git diff --check\n- Unable to run cargo test locally: this environment does not have cargo/rustc/rustfmt installed.\n\n## Bounty payout\nSolana wallet: CTTqY8B7YihjhhuWXDBqD5sRqR3PY2UQ6fCLuz6ss6WT

### DIFF
--- a/rustchain-miner/src/main.rs
+++ b/rustchain-miner/src/main.rs
@@ -3,8 +3,7 @@
 //! Production-ready Rust miner with hardware attestation and RIP-PoA support.
 
 use clap::Parser;
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
+use std::sync::atomic::Ordering;
 use tracing_subscriber::{fmt, prelude::*, EnvFilter};
 
 use rustchain_miner::{Config, Miner};
@@ -49,6 +48,25 @@ struct Args {
     attestation_ttl: u64,
 }
 
+#[cfg(unix)]
+async fn wait_for_shutdown_signal() -> anyhow::Result<&'static str> {
+    let mut sigterm = tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())?;
+
+    tokio::select! {
+        result = tokio::signal::ctrl_c() => {
+            result?;
+            Ok("Ctrl+C")
+        }
+        _ = sigterm.recv() => Ok("SIGTERM"),
+    }
+}
+
+#[cfg(not(unix))]
+async fn wait_for_shutdown_signal() -> anyhow::Result<&'static str> {
+    tokio::signal::ctrl_c().await?;
+    Ok("Ctrl+C")
+}
+
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
     let args = Args::parse();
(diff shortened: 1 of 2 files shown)


itsMacht:
I do not have permission to add labels on this repo. Could a maintainer please add the appropriate BCOS-L1/BCOS-L2 label for review?

cerredz (approved):
Reviewed for #2745. The change fixes the main control-flow bug I see in the existing miner: `main.rs` currently creates a separate `AtomicBool` that `Miner::run()` never observes, while this PR now clones the miner's own shutdown flag and sets that from Ctrl+C/SIGTERM. The retry/mining waits also now poll the same shutdown flag, so a shutdown request no longer has to wait behind the full 30s/60s/block interval sleep.

Verification I ran locally in `rustchain-miner`:
- `cargo test` passed: 27 unit tests, 1 doctest.
- `git diff --check origin/main...HEAD -- rustchain-miner/src/main.rs rustchain-miner/src/miner.rs` passed.
- Checked the dependency side: `tokio` is built with `features = [full]`, so `tokio::signal::unix` is available for the Unix SIGTERM path.

One environment limitation: I installed the `x86_64-unknown-linux-gnu` target and attempted `cargo check --target x86_64-unknown-linux-gnu`, but this Windows host is missing `x86_64-linux-gnu-gcc`, so `ring`'s build script failed before the crate target-checked. I do not see a correctness blocker in the PR itself.

MercyMeow (commented):
I reviewed the miner shutdown handling change. The overall direction looks good: exposing the miner's existing shutdown flag and wiring both Ctrl+C and SIGTERM into the same graceful path is cleaner than keeping an unused standalone flag in `main.rs`.

Two observations:

1. `sleep_or_shutdown()` bounds shutdown latency to roughly one second, which is a practical improvement over waiting the full 30s/60s retry sleeps. That said, it still uses polling. If the miner grows more long sleeps, a `tokio::sync::watch`/`Notify`-based wakeup could make shutdown immediate without repeated one-second timers.

2. The signal task is aborted after `miner.run().await` returns, which is fine for normal completion. If signal setup fails on Unix, the task logs the error but the miner continues running without a shutdown handler. That is probably acceptable, but for production miner behavior it may be worth surfacing setup failure before entering the mining loop.

itsMacht:
Bounty claim filed after merge: https://github.com/Scottcjn/Rustchain/issues/4574\n\nThe merge-triggered award workflow attempted 5 RTC but failed because .rtc-wallet was not available in the checkout. Public payout wallet included in the claim issue.

BossChaos (commented):
## PR #4220 Review - Rust miner SIGTERM graceful shutdown

### Quality Analysis - Process Management

This PR adds SIGTERM handling to the Rust miner.

**Key improvements:**
1. **wait_for_shutdown_signal()**: Handles both SIGTERM and Ctrl+C
2. **Platform conditional**: Unix uses tokio::signal::unix::signal, non-Unix uses ctrl_c only
3. **Shared shutdown_flag**: Uses miner's existing shutdown_flag() instead of separate AtomicBool

**Recommendation**: Merge - improves process lifecycle management.

PR #4222, opened 2026-05-10
fix: rate limit Sophia scheduler tasks

### LINKED ISSUE #2734: [Bug] sophia_scheduler.py has no rate limiting — task backlog on restart floods downstream APIs
## Bug Description

`sophia_scheduler.py` manages scheduled tasks for the Sophia agent. When the scheduler processes a large backlog of tasks (e.g., after a node restart), it processes all tasks sequentially without any **rate limiting or token bucket**, which can overwhelm downstream services (Ollama, Solana RPC, or other APIs) and cause cascading failures.

## Steps to Reproduce

1. Run the node for several days with many scheduled Sophia tasks queued
2. Restart the node (scheduler reloads all pending tasks)
3. Observe all tasks firing simultaneously, sending dozens of concurrent requests to Ollama and external APIs
4. Downstream services return 429 Too Many Requests or time out

## Expected Behavior

- Task scheduling should respect a global rate limit (e.g., max 5 concurrent Ollama requests, configurable via `SOPHIA_MAX_CONCURRENT_TASKS` env var)
- A task queue with a token bucket algorithm (e.g., 10 tasks/minute) prevents burst overload
- Failed tasks should be re-queued with exponential backoff, not dropped

## Impact

- **Medium**: Cascading API failures from burst scheduling can cause the node to miss critical attestation windows, resulting in mining downtime and lost RTC rewards
- Affects overall network stability when many nodes restart simultaneously after an upgrade

## Suggested Fix

Implement a token bucket rate limiter:

```python
import threading
import time
import queue

class TokenBucketRateLimiter:
    def __init__(self, rate=10, per=60):
        self.rate = rate
        self.per = per
        self.allowance = rate
        self.last_check = time.monotonic()
        self.lock = threading.Lock()

    def acquire(self, blocking=True, timeout=None):
        with self.lock:
            now = time.monotonic()
            elapsed = now - self.last_check
            self.allowance += elapsed * (self.rate / self.per)
            self.allowance = min(self.allowance, self.rate)
            self.last_check = now
            if self.allowance >= 1:
                self.allowance -= 1
                return True
            if not blocking:
                return False
            wait = 1.0 / (self.rate / self.per)
            time.sleep(wait)
            return True

# Use in scheduler:
rate_limiter = TokenBucketRateLimiter(
    rate=int(os.environ.get("SOPHIA_MAX_TASKS_PER_MINUTE", 10)),
    per=60
)

def schedule_task(task_fn, *args):
    def _rate_limited():
        rate_limiter.acquire()
        task_fn(*args)
    threading.Thread(target=_rate_limited, daemon=True).start()
```

Wallet: `RTC019e78d600fb3131c29d7ba80aba8fe644be426e`

cerredz:
## Summary
- add a thread-safe token bucket limiter for Sophia scheduler inspection tasks
- make the throttle configurable with `SOPHIA_MAX_TASKS_PER_MINUTE` (default: 10)
- apply the limiter to both full batch inspection and anomaly re-inspection before downstream fingerprint/LLM work
- add focused tests for scheduler limiter usage and token-bucket waiting behavior

Fixes #2734.

## Verification
- `python -m pytest tests\test_sophia_scheduler_rate_limit.py -q` = 2 passed
- `python -m pytest tests\test_sophia_core.py::TestSophiaScheduler -q` = 6 passed
- `python -m pytest tests\test_sophia_core.py -q` = 50 passed
- `python -m py_compile sophia_scheduler.py tests\test_sophia_scheduler_rate_limit.py`
- `git diff --check`

## Bounty

### DIFF
--- a/sophia_scheduler.py
+++ b/sophia_scheduler.py
@@ -10,6 +10,7 @@ Features:
 """
 
 import logging
+import os
 import time
 import threading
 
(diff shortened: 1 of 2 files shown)


MercyMeow (commented):
I reviewed the Sophia scheduler rate-limiter change. The `rate_limiter` injection point is a good testing seam, and the `CountingLimiter` test proves the scheduler acquires one slot per batch task without slowing tests down.

Two observations:

1. `max_tasks_per_minute` is parsed with `int(os.getenv(...))` in `__init__`. If the env var is set to a non-integer value, scheduler construction will raise before any useful log message. Since this is operational config, it may be worth failing with a clearer `ValueError` or falling back with a warning.

2. `TokenBucketRateLimiter.acquire()` sleeps for the full computed wait. The fake-time test covers the exact wait math, which is good. In production, if the scheduler needs to stop while waiting for a token, there is no cancellation/shutdown hook here; the caller will remain blocked inside `sleep_fn` until the wait finishes.

PR #4188, opened 2026-05-10
fix: critical bridge API and payout_worker security vulnerabilities (4 bugs)

BossChaos:
## Summary

Fixes 4 critical/high security vulnerabilities in the bridge API and payout worker.

**Wallet:** `RTC6d1f27d28961279f1034d9561c2403697eb55602`

### 1. Bridge Deposit Double-Spend (CRITICAL)

**File:** `node/bridge_api.py`, `create_bridge_transfer()`

**Bug:** Bridge deposits inserted into `lock_ledger` without deducting from `balances` table. Funds remained spendable while supposedly "locked" for bridge transfer.

**Fix:** Atomically deduct from `balances` table when creating bridge deposit. Verify balance doesn't go negative after deduction.

### 2. Void Bridge Transfer Fund Loss (HIGH)

**File:** `node/bridge_api.py`, `void_bridge_transfer()`

**Bug:** Voiding only updated `lock_ledger` status without crediting balance back. Permanently lost miner funds.

**Fix:** Credit `amount_i64` back to miner's `balances` entry on void.

### 3. Update External Endpoint Auth Bypass (HIGH)

**File:** `node/bridge_api.py`, `register_bridge_routes()`, `update_external` endpoint

**Bug:** When `RC_BRIDGE_API_KEY` env var was empty, endpoint had no authentication. Anyone could mark bridge transfers as completed.

**Fix:** Return 503 when no API key configured. Require auth regardless.

### 4. Payout Worker Wrong Balance Table (MEDIUM)

**File:** `node/payout_worker.py`

**Bug:** Read/wrote from `accounts.balance` instead of `balances.amount_i64` — different table from rest of system.

**Fix:** Use `balances.amount_i64` with `miner_id` (system standard).

## Local Testing

```bash
python test_bridge_security.py
# PASS: Bridge deposit correctly deducts balance
# PASS: Void bridge correctly credits balance back
# PASS: update_external auth fix applied (code review verified)
```

## Related Issues

- Bug Bounty #71 (Ongoing Bug Bounty Program)
- Reported in comments on #71

PR #4223, opened 2026-05-10
fix: require bridge callback API key

### LINKED ISSUE #3225: [Bug] bridge_api.py: update-external endpoint timing-unsafe comparison + unauthenticated when RC_BRIDGE_API_KEY unset
## Security Audit Finding

**Severity:** Medium
**File:** node/bridge_api.py:788-792
**Issue:** The `/api/bridge/update-external` endpoint (bridge service callback) uses `!=` for API key comparison instead of `hmac.compare_digest`, and is completely unauthenticated when `RC_BRIDGE_API_KEY` is not set.

**Code:**
```python
api_key = request.headers.get("X-API-Key", "")
expected_key = os.environ.get("RC_BRIDGE_API_KEY", "")
if expected_key and api_key != expected_key:  # Timing-unsafe + bypass when unset
    return jsonify({"error": "Unauthorized"}), 401
```

**Impact:**
- Timing attack on API key comparison
- When `RC_BRIDGE_API_KEY` is not set, endpoint is completely open
- Attacker can forge external confirmation data (fake confirmations, fake tx hashes)
- Could mark bridge transfers as "completed" without actual external confirmation

**Fix:**
```python
expected_key = os.environ.get("RC_BRIDGE_API_KEY", "")
if not expected_key or not hmac.compare_digest(api_key, expected_key):
    return jsonify({"error": "Unauthorized"}), 401
```

Wallet: RTC4642c5ee8467f61ed91b5775b0eeba984dd776ba

cerredz:
## Summary
- make `/api/bridge/update-external` fail closed when `RC_BRIDGE_API_KEY` is not configured
- keep configured callback authentication on `hmac.compare_digest()`
- add route-level regression coverage for unconfigured, wrong-key, and configured-key paths

Fixes #3225.

## Verification
- `python -m pytest tests\test_bridge_lock_ledger.py::TestBridgeCallbackAuth -q` = 3 passed
- `python -m pytest tests\test_bridge_lock_ledger.py -q` = 30 passed
- `python -m py_compile node\bridge_api.py tests\test_bridge_lock_ledger.py`
- `git diff --check`

## Scope note
This intentionally implements only the bridge callback auth-bypass/timing fix that was validated in the maintainer deep-verify on #4188. It does not include the unrelated bridge balance or payout-worker changes from that unsafe mixed PR.

## Bounty

### DIFF
--- a/node/bridge_api.py
+++ b/node/bridge_api.py
@@ -788,10 +788,11 @@ def register_bridge_routes(app):
     @app.route('/api/bridge/update-external', methods=['POST'])
     def update_external():
         """Update external confirmation data (for bridge service callbacks)."""
-        # Optional: require API key for callbacks
         api_key = request.headers.get("X-API-Key", "")
         expected_key = os.environ.get("RC_BRIDGE_API_KEY", "")
-        if expected_key and not hmac.compare_digest(api_key, expected_key):
+        if not expected_key:
+            return jsonify({"error": "Bridge API key not configured"}), 503
+        if not hmac.compare_digest(api_key, expected_key):
             return jsonify({"error": "Unauthorized"}), 401
         
         data = request.get_json(silent=True)
--- a/tests/test_bridge_lock_ledger.py
+++ b/tests/test_bridge_lock_ledger.py
@@ -23,6 +23,7 @@ from pathlib import Path
 from unittest.mock import patch, MagicMock
 from dataclasses import dataclass
 from typing import Optional, Tuple, Dict, Any, List
+from flask import Flask
 
 # Add node directory to path
 sys.path.insert(0, str(Path(__file__).parent.parent / "node"))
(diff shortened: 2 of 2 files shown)


Jeggob:
Removed by author request.

MercyMeow (commented):
I reviewed the bridge callback API-key hardening. Failing closed when `RC_BRIDGE_API_KEY` is unset is the right security default for `/api/bridge/update-external`; the old "optional key" behavior could silently leave callback mutation exposed in a misconfigured deployment.

Two notes:

1. The accepted-key test intentionally stops at payload validation (`400 Request body required`), which is a good way to prove auth passed without needing a full bridge transfer fixture. It may still be worth adding a wrong-key test where `X-API-Key` is omitted entirely, not just set to the wrong value, since missing header is the most common misconfiguration path.

2. This changes deployments that relied on an unset `RC_BRIDGE_API_KEY` from permissive to `503`. That is the safer default, but the PR description should probably call it out as a breaking operational requirement so operators know to set the env var before rollout.

Issue #3202, opened 2026-05-03
[Bug] Multiple endpoints leak internal error details via str(e) in error responses


508704820:
### Summary
Multiple API endpoints return `str(e)` in error responses (`jsonify({"error": str(e)})`). This leaks internal implementation details to attackers, including:
- Database schema and SQL errors
- File paths and filesystem structure
- Internal server configuration
- Python exception types and tracebacks

### Affected Files (partial list)
- `node/bcos_routes.py` — lines 250, 308, 348, 433
- `node/beacon_api.py` — lines 163, 189, 313, 379, 411, 456

### Example
A malformed request could trigger a SQLite error like:
```json
{"error": "no such column: miner_id"}
```
This reveals the database schema to an attacker.

### Expected Behavior
Error responses should return generic error messages to users while logging the full exception server-side.

### Fix
Replace `str(e)` with generic error messages:
```python
# Before
return jsonify({"error": str(e)}), 500

# After
import logging
logger.exception("bcos_attest failed")
return jsonify({"error": "internal_error"}), 500
```

### Severity
Medium — information disclosure aids reconnaissance for further attacks

### Solana Wallet for Payout
`RTC9d7caca3039130d3b26d41f7343d8f4ef4592360`

🤖 Found by OpenClaw Team (司雨-S)

haoyousun60-create:
## Self-Audit Confirmation

I independently verified this finding during my audit of the codebase.

**Confirmed files leaking str(e):**
- beacon_api.py: 10+ endpoints return `str(e)` in error responses (lines 163, 189, 313, 379, 411, 456, 486, 521, 627, 661)
- bridge_api.py: 3 endpoints return `str(e)` in details field (lines 391, 574, 649)
- lock_ledger.py: 3 endpoints return `str(e)` in details field (lines 189, 267, 341)

**Additional finding:** The beacon_api.py `/api/chat` endpoint (line 661) also returns `str(e)`, which could leak database schema details if a SQLite error occurs during message storage.

Wallet: RTC4642c5ee8467f61ed91b5775b0eeba984dd776ba

jaxint:
## Security Analysis: Internal Error Detail Leakage

### Root Cause
Multiple exception handlers return raw exception objects to clients:
```python
except Exception as e:
    return {"error": str(e)}  # exposes internal details
```

### Information Exposure Vectors
This leaks to any API caller:
1. **File paths** — reveals server directory structure (`/opt/rustchain/...`)
[...]

PR #4225, opened 2026-05-10
fix: hide bridge database error details

cerredz:
## Summary

Fixes part of #3202 by removing raw SQLite exception strings from the bridge/lock-ledger database error payloads.

Current origin/main still returned details: str(e) from database failure handlers in:
- node/lock_ledger.py: create_lock, release_lock, forfeit_lock
- node/bridge_api.py: create_bridge_transfer, void_bridge_transfer, update_external_confirmation

This PR keeps the full exception server-side via logger.exception(...) but only returns a generic Database error payload to callers.

## Verification

- python -m pytest tests\test_bridge_lock_ledger.py -q => 29 passed
- python -m py_compile node\bridge_api.py node\lock_ledger.py tests\test_bridge_lock_ledger.py
- git diff --check
- rg scan for raw sqlite details in the two modules => no matches

### DIFF
--- a/node/bridge_api.py
+++ b/node/bridge_api.py
@@ -18,6 +18,7 @@ import sqlite3
 import time
 import hmac
 import hashlib
+import logging
 import os
 from typing import Optional, Tuple, Dict, Any
 from decimal import Decimal
@@ -55,6 +56,7 @@ BRIDGE_DEFAULT_CONFIRMATIONS = int(os.environ.get("RC_BRIDGE_DEFAULT_CONFIRMATIO
 BRIDGE_LOCK_EXPIRY_SECONDS = int(os.environ.get("RC_BRIDGE_LOCK_EXPIRY_SECONDS", "604800"))  # 7 days
 BRIDGE_MIN_AMOUNT_RTC = float(os.environ.get("RC_BRIDGE_MIN_AMOUNT_RTC", "1.0"))
 BRIDGE_UNIT = 1000000  # Micro-units per RTC
+logger = logging.getLogger(__name__)
 
 
 # =============================================================================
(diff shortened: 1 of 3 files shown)


jaxint:
## Security Analysis: Error Detail Exposure

**Root cause:** Database errors often leak schema details, connection strings, or internal state.

**Recommended fix pattern:**
```python
# BAD: exposes internal details
except Exception as e:
    return {"error": str(e)}  # ❌ Information disclosure

# GOOD: generic error with internal logging
except DatabaseError as e:
    logger.error(f"DB operation failed: {e}", exc_info=True)
    return {"error": "An internal error occurred. Reference: " + generate_ref_id()}
```

**Additional considerations:**
- Ensure error reference IDs can be correlated server-side without exposing actual error content
- Set `PYTHONUNBUFFERED=1` in prod to ensure logs are available for correlation
- Consider a WAF rule to filter error detail patterns in responses

*Technical contribution via RustChain bounty program.*

MercyMeow (commented):
I reviewed the bridge/lock-ledger DB error handling changes. The direction looks right: keeping `logger.exception(...)` server-side while returning only `{"error": "Database error"}` avoids leaking SQLite schema details to callers without hiding the failure from operators.

One coverage gap I noticed: the PR changes six exception handlers, but the new tests only exercise `create_bridge_transfer()` and `create_lock()`. It would be worth adding at least lightweight failure-path assertions for `void_bridge_transfer()`, `update_external_confirmation()`, `release_lock()`, and `forfeit_lock()` as well, since those are the other places where `details: str(e)` was removed.

A smaller positive note: using module-level `logger = logging.getLogger(__name__)` is the right shape here because it preserves traceback context without coupling these helpers to a concrete logging configuration.

BossChaos (commented):
## PR #4225 Review - Bridge error message information leakage fix

### Security Analysis - Information Disclosure Prevention

This PR removes internal error details from API responses.

**Key improvements:**
1. **Removed str(e)**: Database error messages no longer expose internal details
2. **logger.exception()**: Full error details logged server-side only
3. **Consistent pattern**: Applied to create_bridge_transfer, void_bridge_transfer, update_external_confirmation

**Recommendation**: Merge - prevents information leakage in error responses.

PR #2216, opened 2026-04-11
security: require ed25519 signature on governance proposals and votes

ArokyaMatthew:
## Security Fix: Governance Endpoints Accept Bare miner_id Without Proof

**Severity:** 🔴 Critical (200+ RTC Bounty)
**File:** `node/governance.py`
**Lines:** 229, 363

### Description
`/api/governance/propose` and `/api/governance/vote` accept a plain `miner_id` string
with zero cryptographic authentication. Anyone can create proposals or cast votes
impersonating any active miner.

### Exploit Mechanism
1. Query the miner list to get active miner IDs
2. Send POST to `/api/governance/propose` with any `miner_id`
3. Create malicious proposals (parameter changes, etc.)
4. Vote on proposals as any miner to reach quorum
5. Force governance decisions without controlling any mining infrastructure

### Fix Applied
- Added `_verify_miner_signature()` function requiring ed25519 signature
- Both endpoints now return **401** if signature is missing, invalid, or stale (>5 min)
- Signed payload format: `"{action}:{miner_id}:{timestamp}"`
- Uses `nacl.signing.VerifyKey` for cryptographic verification
- Timestamp check prevents replay attacks

### Testing
- Syntax verification passes
- Unauthenticated requests now receive 401 with clear error message

PR #71, opened 2026-02-10
[BOUNTY] RTC Reward Analytics Dashboard (#41)

TheFrogEcliptia:
### [BOUNTY] RTC Reward Analytics Dashboard (#41)

Web dashboard focused on reward distribution metrics.

**Features:**
- Visual analysis of hardware distribution (Pie Chart).
- Earnings estimate per miner per epoch (Bar Chart).
- Responsive interface based on real API data.

Closes #41

PR #50, opened 2026-02-08
Improve leaderboard.json structure (First Contribution #48)

ssiweifnag:
## Summary

This PR improves the leaderboard.json file structure for better documentation and realism, as the first contribution to RustChain.

## Changes

1. Added description field - Makes the JSON self-documenting
2. Expanded validator examples - Now 3 entries (was 1) showing realistic data
3. Added scoring field explanations - Documents entropy_score and score_composite
4. Updated timestamp - Current date instead of stale 2025 data

## Bounty Claim

This is my first contribution to RustChain. Per issue #48, this qualifies for the **10 RTC bounty** for the first merged PR.

---

**Bounty**: 10 RTC (First Contribution)  
**Issue**: Scottcjn/Rustchain#48

PR #4230, opened 2026-05-10
audit: report integrated governance proposer auth bypass (#71)

maelrx:
## Summary

Report-only security audit for rustchain-bounties#71, related to governance bounty #50.

Finding: the integrated node endpoint `/governance/propose` in `node/rustchain_v2_integrated_v2.2.1_rip200.py` accepts a caller-supplied `wallet` string as proposer identity and only checks that wallet balance. It does not require a signature, public key, nonce, admin key, or proof of wallet control.

## Impact

Any caller who knows a wallet with more than `GOVERNANCE_MIN_PROPOSER_BALANCE_RTC` can create an active governance proposal attributed to that wallet. This bypasses the governance-auth hardening previously added to `node/governance.py` in PR #2216 because the integrated server exposes a separate `/governance/propose` implementation.

## Local verification

The report includes a deterministic Flask `test_client` PoC using a temporary SQLite DB. Observed output:

```text
status 201
ok True
proposal_wallet victim_rich_wallet
rows [("victim_rich_wallet", "attacker forged proposal", "active")]
```

Checks run:

```text
git diff --check -- audits/integrated_governance_propose_auth_bypass_71.md
python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py
uv run --no-project --with flask --with prometheus-client --with pynacl --with requests python <local PoC from report>
```

No live node testing was performed.

Wallet: `RTCc068d2850639325b847e09fc6b8c01b0b88d7be8`

### DIFF
--- a/audits/integrated_governance_propose_auth_bypass_71.md
+++ b/audits/integrated_governance_propose_auth_bypass_71.md
@@ -1,0 +1,141 @@
+# Audit: Integrated `/governance/propose` Wallet-Impersonation Bypass (#71)
+
+## Metadata
+
+- Bounty issue: Scottcjn/rustchain-bounties#71
+- Related governance bounty: Scottcjn/rustchain-bounties#50
+- Auditor: maelrx
+- Public RTC wallet: `RTCc068d2850639325b847e09fc6b8c01b0b88d7be8`
+- Repository: Scottcjn/Rustchain
+- Commit reviewed: `0c428794e85db8ef5a64639e4ccd9b121e40cab1`
+- Primary file reviewed: `node/rustchain_v2_integrated_v2.2.1_rip200.py`
+- Requested severity: High
+
+## Finding
+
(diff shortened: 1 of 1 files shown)


maelrx:
Maintainer note: this is an audit-report-only security PR for rustchain-bounties#71 and is related to governance bounty #50. I cannot apply repository labels from this account; please add the appropriate BCOS/security label during triage.

The report uses local Flask `test_client` reproduction only and does not test live governance endpoints.

MercyMeow (commented):
I reviewed the integrated governance proposer-auth bypass audit. The report is strong in one important way: it distinguishes the already-hardened `node/governance.py` blueprint from the separate integrated `/governance/propose` route, which is exactly the kind of duplicate-surface patch gap that can survive after a security fix.

Two notes:

1. The PoC proves unauthenticated proposal creation with a temporary DB and persisted `governance_proposals` row, which is good evidence. It would be even stronger if the report included the exact expected signed message shape by referencing the existing `/governance/vote` verifier, so maintainers can apply the same contract rather than invent a second proposal auth format.

2. The suggested fix mentions proposal nonces, but the report does not call out where nonce storage should live or how it should be scoped. For governance, nonce replay protection should probably be per derived proposer wallet and action type, otherwise a valid signed proposal body could be replayed into duplicate active proposals.

PR #4231, opened 2026-05-10
[UTXO-BUG] Reject ambiguous signed Decimal amounts

bkallday1122:
## Summary

Fixes a UTXO transfer signature/accounting mismatch for bounty Scottcjn/rustchain-bounties#2819.

`/utxo/transfer` parses `amount_rtc` and `fee_rtc` as exact `Decimal` values for ledger math, but rebuilds the signed payload by converting those Decimals back to JSON floats. Distinct decimal strings can collapse to the same float-shaped signed payload while producing different nanoRTC ledger amounts.

## Bounty Classification

- Vulnerability class: signed amount mismatch / fee and amount accounting validation
- Suggested severity: Medium under the bounty's UTXO fee/accounting category

## Root Cause

The endpoint uses exact Decimal math for the ledger:

```python
amount_nrtc = int(amount_rtc * UNIT)
fee_nrtc = int(fee_rtc * UNIT)
```

But the signed message is rebuilt using floats:

```python
amount_for_sig = float(amount_rtc)
fee_for_sig = float(fee_rtc)
```

For example, `1000000000.0` and `1000000000.00000005` serialize to the same signed float value, but differ by 5 nanoRTC in exact ledger math. A signature valid for the lower amount could therefore authorize a transfer of the higher exact amount.

## Expected vs Actual

Expected:

- The amount and fee verified by the signature should match the exact nanoRTC values applied to the UTXO ledger.
- Inputs with more precision than nanoRTC or with float-colliding spellings should be rejected.

Actual before this patch:

- The endpoint accepted a decimal string that verified against the signed payload for `1000000000.0`.
- The ledger transferred `1000000000.00000005` RTC, a 5 nanoRTC difference from the signed amount.

## Fix

- Convert Decimal RTC amounts to nanoRTC only when the result is an exact integer.
- Reject Decimal spellings whose float-shaped signed payload does not preserve the same nanoRTC value.
- Add a regression test that proves the float-colliding amount is rejected and no recipient funds move.
- Update the endpoint test coinbase helper to pass the existing internal minting flag so the endpoint suite runs against current UTXO rules.

## Validation

```text
$ git diff --check
$ PYTHONPATH=/tmp/rustchain-flask:node python3 -m py_compile node/utxo_endpoints.py node/test_utxo_endpoints.py
$ PYTHONPATH=/tmp/rustchain-flask:node python3 -m unittest node.test_utxo_endpoints
.................
Ran 17 tests in 0.159s

OK
$ PYTHONPATH=node python3 node/test_utxo_db.py
...................................................
Ran 51 tests in 0.259s

OK
```

### DIFF
--- a/node/test_utxo_endpoints.py
+++ b/node/test_utxo_endpoints.py
@@ -10,10 +10,14 @@ import os
 import tempfile
 import time
 import unittest
+from decimal import Decimal
 
 from flask import Flask
 
-from utxo_db import UtxoDB, UNIT
+import utxo_endpoints
+from utxo_db import (
+    UtxoDB, UNIT, address_to_proposition, compute_box_id,
+)
 from utxo_endpoints import register_utxo_blueprint
 
 
@@ -69,7 +73,23 @@ class TestUtxoEndpoints(unittest.TestCase):
             'inputs': [],
             'outputs': [{'address': address, 'value_nrtc': value_nrtc}],
             'timestamp': int(time.time()),
+            '_allow_minting': True,
         }, block_height=height)
+
+    def _seed_existing_box(self, address, value_nrtc, height=1):
+        tx_id = '22' * 32
+        prop = address_to_proposition(address)
+        box_id = compute_box_id(value_nrtc, prop, height, tx_id, 0)
+        self.utxo_db.add_box({
+            'box_id': box_id,
+            'value_nrtc': value_nrtc,
+            'proposition': prop,
+            'owner_address': address,
+            'creation_height': height,
+            'transaction_id': tx_id,
+            'output_index': 0,
+        })
+        return box_id
 
     # -- read endpoints ------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


MercyMeow (commented):
I reviewed the UTXO Decimal amount validation PR. The direction is good: converting RTC `Decimal` values to nanoRTC with explicit integral checks avoids silent truncation in a money path.

Two observations:

1. `_ensure_signed_float_preserves_nrtc()` is a pragmatic compatibility guard for the current float-based signed payload, but it also means some exact Decimal spellings that are valid ledger amounts will be rejected until the signature format stops serializing amounts as JSON floats. That behavior should be called out in API/client docs so wallet clients do not treat these `400` responses as random validation failures.

2. The new test mutates the module-global `utxo_endpoints._verify_sig_fn` and restores it in `finally`, which is good, but it bypasses the function passed through `register_utxo_blueprint()`. If future tests run in parallel or reuse the module state, this global patching could leak. A cleaner pattern would expose the verifier through the registered blueprint dependency or isolate the app/module per test.

PR #4233, opened 2026-05-10
[UTXO-BUG] Reject non-integer fee_nrtc values

bkallday1122:
## Summary

Fixes a UTXO fee accounting edge case for bounty Scottcjn/rustchain-bounties#2819.

`fee_nrtc` is stored in SQLite as an integer nanoRTC amount, but `UtxoDB.apply_transaction()` and `mempool_add()` accepted non-integer values. A transaction can use a fractional fee such as `0.5` with a one-nanoRTC output reduction, pass the current conservation check, and then persist a different integer fee value in SQLite. That creates divergent accounting between validation-time arithmetic and stored transaction history.

## Bounty Classification

- Vulnerability class: fee calculation exploit / missing validation
- Suggested severity: Medium under the bounty's fee calculation category

## Root Cause

Both code paths read `fee_nrtc` with:

```python
fee = tx.get('fee_nrtc', 0)
```

The code checked only `fee < 0`; it did not require `fee` to be an integer nanoRTC amount before using it in conservation arithmetic and before persisting it to an `INTEGER` column.

## Expected vs Actual

Expected:

- `fee_nrtc` should be an integer nanoRTC amount everywhere it is validated, stored, and used for conservation checks.
- Mempool admission should reject transactions that block application would reject.

Actual before this patch:

- `fee_nrtc: 0.5` could pass validation when paired with a one-nanoRTC output reduction.
- SQLite then persisted the fee into an `INTEGER` column as a different integer value, creating accounting divergence.

## Fix

- Reject `fee_nrtc` values whose exact type is not `int` in `apply_transaction()`.
- Reject the same invalid values in `mempool_add()` so mempool admission matches block application.
- Add regression tests for direct application and mempool admission.

## Validation

```text
$ python3 node/test_utxo_db.py
.....................................................
Ran 53 tests in 0.214s

OK
```

`python3 -m pytest node/test_utxo_db.py -v` was not available in my local environment because `pytest` is not installed.

### DIFF
--- a/node/test_utxo_db.py
+++ b/node/test_utxo_db.py
@@ -157,6 +157,29 @@ class TestUtxoDB(unittest.TestCase):
         self.assertEqual(self.db.get_balance('alice'), 100 * UNIT)
         self.assertEqual(self.db.get_balance('bob'), 0)
 
+    def test_fractional_fee_rejected(self):
+        """fee_nrtc must be an integer nanoRTC amount.
+
+        A fractional fee can pass conservation by pairing it with a one-nanoRTC
+        output reduction, but SQLite stores the fee in an INTEGER column and
+        truncates it. That silently destroys value without recording the fee.
+        """
+        self._apply_coinbase('alice', 100 * UNIT)
+        alice_boxes = self.db.get_unspent_for_address('alice')
+
+        ok = self.db.apply_transaction({
+            'tx_type': 'transfer',
+            'inputs': [{'box_id': alice_boxes[0]['box_id'],
+                         'spending_proof': 'sig'}],
+            'outputs': [{'address': 'bob', 'value_nrtc': 100 * UNIT - 1}],
+            'fee_nrtc': 0.5,
+        }, block_height=10)
+
+        self.assertFalse(ok)
+        # Balances unchanged
+        self.assertEqual(self.db.get_balance('alice'), 100 * UNIT)
+        self.assertEqual(self.db.get_balance('bob'), 0)
+
     # -- double-spend --------------------------------------------------------
 
     def test_double_spend_rejected(self):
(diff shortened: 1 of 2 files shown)


cerredz (approved):
I reviewed this against the UTXO fee validation path and the patch looks good to me.

The important behavior is that both block application and mempool admission now reject non-integer `fee_nrtc` before doing conservation arithmetic or locking inputs. I verified the new focused tests and also checked the direct state transition path: a 100 RTC coinbase to Alice followed by a transfer with `fee_nrtc: 0.5` and a one-nanoRTC output reduction now returns `False` from both `apply_transaction()` and `mempool_add()`, while Alice remains at `10000000000` nRTC and Bob remains at `0`.

Validation run locally:

```
$env:PYTHONPATH='node'; python -m unittest node.test_utxo_db.TestUtxoDB.test_fractional_fee_rejected node.test_utxo_db.TestUtxoDB.test_mempool_rejects_fractional_fee
# Ran 2 tests in 0.025s
# OK

$env:PYTHONPATH='node'; python -m unittest node.test_utxo_db.TestUtxoDB.test_negative_fee_rejected node.test_utxo_db.TestUtxoDB.test_mempool_rejects_negative_fee node.test_utxo_db.TestUtxoDB.test_mempool_accepts_valid_tx
# Ran 3 tests in 0.037s
# OK

python -m py_compile node\utxo_db.py node\test_utxo_db.py
# OK

git diff --check origin/main...HEAD -- node\utxo_db.py node\test_utxo_db.py
# OK
```

The full `python -m unittest discover -s node -p test_utxo_db.py` run reached 53 tests and hit the existing Windows temp DB cleanup lock in `test_mining_reward_over_cap_rejected`; I did not see a regression related to this fee patch.

PR #4234, opened 2026-05-10
[UTXO-BUG] Bind transfer tx_id to transaction contents

bkallday1122:
## Summary

Fixes a UTXO transaction identity issue for bounty Scottcjn/rustchain-bounties#2819.

For non-minting transfers, `UtxoDB.apply_transaction()` derived `tx_id` from sorted input box IDs plus timestamp only. The transaction ID did not commit to outputs, fee, transaction type, data inputs, or block height. Two materially different transfers using the same input and timestamp could therefore record the same `tx_id` while producing different output boxes and different state roots.

## Bounty Classification

- Vulnerability class: Merkle/state-root divergence / transaction identity collision
- Suggested severity: Medium under the bounty's Merkle state root manipulation category

## Root Cause

The old transfer tx-id seed used only:

```python
for inp in sorted(inputs, key=lambda i: i['box_id']):
    tx_seed_h.update(bytes.fromhex(inp['box_id']))
tx_seed_h.update(ts.to_bytes(8, 'little'))
```

For coinbase only, it also mixed in `tx_type`, `block_height`, and outputs. Normal transfers did not get that content binding.

## Expected vs Actual

Expected:

- A UTXO transaction ID should identify the transaction contents it records.
- Different outputs or fees should produce different transaction IDs.

Actual before this patch:

- Transfer Alice -> Bob and transfer Alice -> Eve can share the same input and timestamp, record the same `tx_id`, and produce different state roots.
- Reproduced locally with both variants recording `85b9923a27eeadaa0344424288114348c3a283188d73239c4a545f3d7e656e33` while their state roots differed.

## Fix

- Build a canonical `tx_identity` object that includes `tx_type`, sorted input IDs, sorted data inputs, ordered output commitments, `fee_nrtc`, `timestamp`, and `block_height`.
- Hash that canonical JSON to derive `tx_id`.
- Add a regression test proving different recipient outputs no longer share the same `tx_id`.

## Validation

```text
$ git diff --check HEAD~1..HEAD
$ python3 node/test_utxo_db.py
....................................................
Ran 52 tests in 0.242s

OK
```

`python3 -m pytest node/test_utxo_db.py -v` was not available in my local environment because `pytest` is not installed.

### DIFF
--- a/node/test_utxo_db.py
+++ b/node/test_utxo_db.py
@@ -124,6 +124,61 @@ class TestUtxoDB(unittest.TestCase):
         self.assertEqual(self.db.get_balance('bob'), 90 * UNIT)
         self.assertEqual(self.db.get_balance('alice'), 9 * UNIT)
 
+    def test_transfer_tx_id_commits_to_outputs(self):
+        """Different transfer outputs must not share the same tx_id.
+
+        Previously transfer tx_id was derived from inputs + timestamp only.
+        Two nodes could apply materially different transactions with the same
+        input and timestamp, record the same tx_id, but produce different UTXO
+        sets and state roots.
+        """
+        def apply_variant(recipient: str) -> tuple:
+            tmp = tempfile.NamedTemporaryFile(suffix='.db', delete=False)
+            tmp.close()
+            db = UtxoDB(tmp.name)
+            try:
+                db.init_tables()
+                ok = db.apply_transaction({
+                    'tx_type': 'mining_reward',
+                    'inputs': [],
+                    'outputs': [{'address': 'alice',
+                                 'value_nrtc': 100 * UNIT}],
+                    'fee_nrtc': 0,
+                    'timestamp': 1234567890,
+                    '_allow_minting': True,
+                }, block_height=1)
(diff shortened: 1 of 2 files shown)


gl11tchy (approved):
Approved. I reviewed the transaction identity change and it fixes the demonstrated collision class: normal transfers now bind tx_id to tx_type, sorted input IDs, data inputs, ordered output commitments, fee_nrtc, timestamp, and block_height instead of only input IDs plus timestamp.

Validation performed locally:
- python3 node/test_utxo_db.py => 52 passed
- python3 -m py_compile node/utxo_db.py node/test_utxo_db.py => OK
- git diff --check origin/main...HEAD -- node/utxo_db.py node/test_utxo_db.py => OK

I also checked the UTXO transfer endpoint boundary. The endpoint builds and applies transfers through UtxoDB.apply_transaction, so binding the recorded tx_id at that state-transition point covers the output-substitution/state-root divergence described in the PR. The new regression proves two same-input same-timestamp transfers to different recipients no longer record the same transfer tx_id.

PR #4241, opened 2026-05-10
[UTXO-BUG] guard epoch reward dual-write settlement

gl11tchy:

## Finding
`finalize_epoch()` unconditionally tries to sync mining rewards into the UTXO layer, even when `UTXO_DUAL_WRITE` is disabled. The call also constructs `UtxoDB()` without the required `DB_PATH`, so reward settlement rolls back as soon as that path executes. A related atomicity issue is that the UTXO `apply_transaction()` result was ignored, so a future rejected UTXO write could still leave account rewards committed and the epoch marked settled.

## Impact
This can block epoch reward settlement when the integrated server reaches this path, and it can create account/UTXO divergence if the UTXO layer rejects a mining reward while the account-model update continues. That is directly in the pre-production dual-write surface from the bounty brief.

## Fix
- Keep the mining reward UTXO write behind `UTXO_DUAL_WRITE`.
- Construct `UtxoDB(DB_PATH)` with the configured database.
- Treat a false UTXO apply result as fatal so the surrounding transaction rolls back.
- Add regression coverage for the gate, DB path, and apply-result check.

## Verification
- `python3 node/tests/test_epoch_utxo_dual_write_guard.py` fails on the original code and passes after the fix.
- `PYTHONPATH=node python3 -m unittest discover -s node -p test_utxo_db.py` passes: 51 tests.
- `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py node/tests/test_epoch_utxo_dual_write_guard.py` passes.

Payout wallet can be provided after validation.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -2871,15 +2871,24 @@ def finalize_epoch(epoch, per_block_rtc, prev_block_hash: bytes = b""):
                     (amount_i64, amount_i64, pk)
                 )
 
-                # Sync to UTXO layer (Atomic Dual-Write)
-                from utxo_db import UtxoDB
-                utxo_tx = {
-                    "tx_type": "mining_reward",
-                    "inputs": [],
-                    "outputs": [{"address": pk, "value_nrtc": amount_i64}],
-                    "_allow_minting": True
-                }
-                UtxoDB().apply_transaction(utxo_tx, epoch * 144, conn=conn)
+                # Sync to UTXO layer only when the dual-write feature is enabled.
+                # A rejected UTXO write must abort the surrounding account-model
+                # settlement or the two ledgers diverge while the epoch is marked
+                # settled.
+                if UTXO_DUAL_WRITE:
+                    utxo_tx = {
+                        "tx_type": "mining_reward",
+                        "inputs": [],
+                        "outputs": [{"address": pk, "value_nrtc": amount_i64}],
+                        "_allow_minting": True
+                    }
+                    utxo_ok = UtxoDB(DB_PATH).apply_transaction(
+                        utxo_tx, epoch * 144, conn=conn
+                    )
+                    if not utxo_ok:
+                        raise RuntimeError(
+                            f"UTXO reward settlement failed for {pk[:20]}..."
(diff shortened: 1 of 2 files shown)


cerredz (changes requested):
Thanks for the focused split here. The production change is in the right area, but I think this needs one small test fix before merge because the submitted regression target is not portable on Windows.

`node/tests/test_epoch_utxo_dual_write_guard.py` calls `SERVER_PATH.read_text()` without an explicit encoding. On this Windows host, that uses cp1252, and the integrated server file contains bytes that cp1252 cannot decode, so the whole test file errors before any assertions run:

```
python node\tests\test_epoch_utxo_dual_write_guard.py
# ERROR: UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 265481
# Ran 3 tests in 0.018s
# FAILED (errors=3)
```

For comparison, the same tests pass if UTF-8 mode is forced:

```
$env:PYTHONUTF8='1'; python node\tests\test_epoch_utxo_dual_write_guard.py
# Ran 3 tests in 0.184s
# OK
```

Please make the source read explicit, e.g. `SERVER_PATH.read_text(encoding=utf-8)`, so the regression command in the PR works without relying on caller environment. I also checked `python -m py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py node\tests\test_epoch_utxo_dual_write_guard.py` and `git diff --check origin/main...HEAD -- node\rustchain_v2_integrated_v2.2.1_rip200.py node\tests\test_epoch_utxo_dual_write_guard.py`; both passed. The broader `python -m unittest discover -s node -p test_utxo_db.py` reached 51 tests but still hits the known Windows temp DB cleanup lock in `test_mining_reward_over_cap_rejected`.

gl11tchy:
Thanks for catching the portability issue. Pushed 7155ea9 to make SERVER_PATH.read_text use encoding="utf-8".

Verification rerun locally:
- python3 node/tests/test_epoch_utxo_dual_write_guard.py => 3 passed
- python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py node/tests/test_epoch_utxo_dual_write_guard.py => OK
- git diff --check -- node/tests/test_epoch_utxo_dual_write_guard.py => OK

PR #4242, opened 2026-05-10
[UTXO-BUG] filter expired mempool block candidates

gl11tchy:

## Finding
`mempool_get_block_candidates()` returns transactions without checking `expires_at`, so an expired transaction can still be selected for block production. Because expired rows also remain in `utxo_mempool_inputs`, the stale transaction keeps its input box locked until a separate cleanup call happens.

## Impact
A stale high-fee transaction can continue to appear as a block candidate after expiry and can block legitimate replacement spends for the same UTXO. This is a mempool DoS / stale candidate bug in the public admission path covered by the bounty brief.

## Fix
- Clear expired mempool entries before selecting candidates.
- Add `WHERE expires_at > now` to the candidate query.
- Add a regression test that forces a transaction expired, asserts it is not returned as a block candidate, and asserts its input lock is released.

## Verification
- `PYTHONPATH=node python3 -m unittest node.test_utxo_db.TestUtxoDB.test_mempool_block_candidates_ignore_expired_transactions` fails on the original code and passes after the fix.
- `PYTHONPATH=node python3 -m unittest discover -s node -p test_utxo_db.py` passes: 52 tests.
- `python3 -m py_compile node/utxo_db.py node/test_utxo_db.py` passes.

Payout wallet can be provided after validation.

### DIFF
--- a/node/test_utxo_db.py
+++ b/node/test_utxo_db.py
@@ -358,6 +358,31 @@ class TestUtxoDB(unittest.TestCase):
         # Highest fee first
         self.assertEqual(candidates[0]['tx_id'], 'high' * 16)
 
+    def test_mempool_block_candidates_ignore_expired_transactions(self):
+        self._apply_coinbase('alice', 100 * UNIT, block_height=1)
+        box = self.db.get_unspent_for_address('alice')[0]
+        tx_id = 'expired' * 8
+
+        self.assertTrue(self.db.mempool_add({
+            'tx_id': tx_id,
+            'inputs': [{'box_id': box['box_id']}],
+            'outputs': [{'address': 'bob', 'value_nrtc': 100 * UNIT - 1000}],
+            'fee_nrtc': 1000,
+        }))
+
+        conn = self.db._conn()
+        try:
+            conn.execute(
+                "UPDATE utxo_mempool SET expires_at = ? WHERE tx_id = ?",
+                (int(time.time()) - 1, tx_id),
+            )
+            conn.commit()
+        finally:
+            conn.close()
+
+        self.assertEqual(self.db.mempool_get_block_candidates(), [])
+        self.assertFalse(self.db.mempool_check_double_spend(box['box_id']))
+
     def test_mempool_nonexistent_input_rejected(self):
         tx = {
             'tx_id': 'cccc' * 16,
--- a/node/test_utxo_db.py
+++ b/node/test_utxo_db.py
@@ -380,8 +380,16 @@ class TestUtxoDB(unittest.TestCase):
         finally:
             conn.close()
 
-        self.assertEqual(self.db.mempool_get_block_candidates(), [])
-        self.assertFalse(self.db.mempool_check_double_spend(box['box_id']))
+        self.assertTrue(self.db.mempool_add({
+            'tx_id': 'replacement' * 6,
+            'inputs': [{'box_id': box['box_id']}],
+            'outputs': [{'address': 'carol', 'value_nrtc': 100 * UNIT - 2000}],
+            'fee_nrtc': 2000,
+        }))
+
+        candidates = self.db.mempool_get_block_candidates()
+        self.assertEqual(len(candidates), 1)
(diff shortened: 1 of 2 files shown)


godd-ctrl (changes requested):
I reproduced the intended candidate-selection fix, but I think this still leaves one half of the stated impact unresolved.

What I verified locally:

- `PYTHONPATH=node python -m unittest node.test_utxo_db.TestUtxoDB.test_mempool_block_candidates_ignore_expired_transactions` => OK
- `python -m py_compile node/utxo_db.py node/test_utxo_db.py` => OK
- `PYTHONPATH=node python -m unittest discover -s node -p test_utxo_db.py` mostly passes, but hits an unrelated Windows temp-db cleanup `PermissionError` in `test_mining_reward_over_cap_rejected` after 52 tests run.

The remaining issue is in the input-lock path. `mempool_get_block_candidates()` now calls `mempool_clear_expired()` and filters by `expires_at`, so block production will not include the stale tx. But `mempool_add()` still checks `utxo_mempool_inputs` before it clears expired rows. That means a legitimate replacement spend using the same box can still be rejected after the original tx expires, unless some other caller happens to run `mempool_get_block_candidates()` / `mempool_clear_expired()` first.

Minimal repro against this PR:

```text
first_add True
replacement_after_expiry_without_cleanup False
```

That preserves the reported DoS behavior: the expired tx is no longer selected for a block, but its stale `utxo_mempool_inputs` row can still keep the UTXO locked at the public admission boundary.

I would fix this by clearing expired rows at the start of `mempool_add()` as well, preferably before the pool-size and double-spend checks. If you want cleanup and admission to be atomic, move the expiration cleanup into the same `BEGIN IMMEDIATE` transaction before querying `utxo_mempool_inputs`.

Suggested regression test: add one tx, force its `expires_at` into the past, then immediately call `mempool_add()` for a replacement tx spending the same `box_id` without calling `mempool_get_block_candidates()` first. Expected result should be `True`, and `mempool_check_double_spend(box_id)` should reflect only the replacement tx afterward.

gl11tchy:
Thanks for the repro. You were right: filtering candidates fixed block selection, but public admission still needed to clear stale input claims before checking `utxo_mempool_inputs`.

Updated the PR with:

- `mempool_add()` now calls `mempool_clear_expired()` before pool-size / double-spend checks.
- The regression now forces a tx expired, then immediately submits a replacement spend for the same box without calling candidate selection first.
- The test asserts the replacement is accepted and becomes the only block candidate.

Verification rerun locally:

```text
PYTHONPATH=node python3 -m unittest node.test_utxo_db.TestUtxoDB.test_mempool_block_candidates_ignore_expired_transactions
Ran 1 test ... OK

PYTHONPATH=node python3 -m unittest discover -s node -p test_utxo_db.py
Ran 52 tests ... OK

python3 -m py_compile node/utxo_db.py node/test_utxo_db.py
OK
```

ctzxw520-lab:
Security-focused review for PR #4242.

I reviewed the UTXO mempool expiry changes in `node/utxo_db.py` and the new regression coverage in `node/test_utxo_db.py`.

What looks good:

- `mempool_add()` now clears expired entries before checking pool size and before rejecting an input as already claimed. That closes the practical stale-lock path where an expired transaction can keep a spendable box blocked through `utxo_mempool_inputs`.
- `mempool_get_block_candidates()` now both clears expired rows and adds `WHERE expires_at > ?`, so block construction has a direct guard even if cleanup fails or another process leaves stale rows behind.
- The new regression test covers the important sequence: add tx, force expiry, add replacement spending the same box, then verify only the replacement is returned as a block candidate.

Local verification:

```text
python3 -m pytest node/test_utxo_db.py -q
52 passed in 0.73s
```

Residual risk:

- I do not see a blocking security issue in this patch. The cleanup call and candidate filter are complementary, and the candidate query uses a fresh `now` value.
- The broader mempool code still has multiple early returns before `BEGIN IMMEDIATE`, but this PR does not make that worse and the stale-expiry behavior it targets is covered.

Recommendation: approve from a security perspective.

PR #4243, opened 2026-05-11
Guard attest signature field types

bkallday1122:
## Summary

Fixes a malformed-input crash path in `POST /attest/submit` by validating top-level signed-attestation fields before the handler calls `.strip().lower()` on them.

This is follow-up work for Scottcjn/rustchain-bounties#1112. Prior public fuzz reports already covered non-string `signature`; this PR keeps that regression covered and adds the adjacent `public_key` type-confusion guard.

## Changes

- Reject non-string `signature` values with `400 INVALID_SIGNATURE_TYPE`.
- Reject non-string `public_key` values with `400 INVALID_PUBLIC_KEY_TYPE`.
- Mirror the production validator change in the extracted fuzz validator.
- Add route-level regression tests for both crash paths.
- Add two replayable fuzz corpus entries.

## Verification

```bash
PYTHONPATH=/tmp/rustchain-flask:node python3 -m unittest \
  node.tests.test_attest_signature_verification.TestAttestSignatureVerification.test_non_string_signature_rejected_before_handler_crash \
  node.tests.test_attest_signature_verification.TestAttestSignatureVerification.test_non_string_public_key_rejected_before_handler_crash
```

```bash
python3 tests/fuzz/run_fuzz.py --corpus-only
```

```bash
PYTHONPATH=/tmp/rustchain-flask:node python3 -m py_compile \
  node/rustchain_v2_integrated_v2.2.1_rip200.py \
  node/tests/test_attest_signature_verification.py \
  tests/fuzz/attestation_validators.py
```

```bash
git diff --check
```

Full Hypothesis fuzzing was not run locally because this environment does not have `pytest` or `hypothesis` installed.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -270,6 +270,13 @@ def _validate_attestation_payload_shape(data):
         if field_name in data and data[field_name] is not None and not isinstance(data[field_name], str):
             return _attest_field_error("INVALID_MINER", f"Field '{field_name}' must be a non-empty string")
 
+    for field_name, code in (
+        ("signature", "INVALID_SIGNATURE_TYPE"),
+        ("public_key", "INVALID_PUBLIC_KEY_TYPE"),
+    ):
+        if field_name in data and data[field_name] is not None and not isinstance(data[field_name], str):
+            return _attest_field_error(code, f"Field '{field_name}' must be a string")
+
     miner = _attest_valid_miner(data.get("miner")) or _attest_valid_miner(data.get("miner_id"))
     if not miner and not (_attest_text(data.get("miner")) or _attest_text(data.get("miner_id"))):
         return _attest_field_error(
--- a/node/tests/test_attest_signature_verification.py
+++ b/node/tests/test_attest_signature_verification.py
@@ -83,6 +83,15 @@ class TestAttestSignatureVerification(unittest.TestCase):
             conn.commit()
         return mod, db_path
 
+    def _load_module_without_db(self, module_name: str, db_name: str):
+        """Load the route module for pre-DB validation tests."""
+        db_path = self._db_path(db_name)
+        os.environ["RUSTCHAIN_DB_PATH"] = db_path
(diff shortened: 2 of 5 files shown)


gl11tchy (changes requested):
Thanks for tightening this crash path. The production validator change and fuzz corpus entries look directionally correct, and the corpus-only replay passes locally.

I am requesting changes because the route-level regression tests added by this PR are not repeatable when run together, including with the exact two-test command from the PR description once Flask and the node dependencies are present.

What I ran:

```bash
uv run --with flask --with requests --with cryptography --with pynacl --with mnemonic --with pytest --with prometheus-client --with flask-cors --with psutil --with numpy --with pandas env PYTHONPATH=node python -m unittest \
  node.tests.test_attest_signature_verification.TestAttestSignatureVerification.test_non_string_signature_rejected_before_handler_crash \
  node.tests.test_attest_signature_verification.TestAttestSignatureVerification.test_non_string_public_key_rejected_before_handler_crash
```

Result: the first test passes, then the second import of `rustchain_v2_integrated_v2.2.1_rip200.py` raises:

```text
ValueError: Duplicated timeseries in CollectorRegistry: {
  rustchain_withdrawal_requests_total,
  rustchain_withdrawal_requests_created,
  rustchain_withdrawal_requests
}
```

I also ran each new test individually in a fresh process, and each passes on its own. `python3 tests/fuzz/run_fuzz.py --corpus-only` passes with the new corpus entries, and `py_compile` passes for the changed Python files.

The issue is in `node/tests/test_attest_signature_verification.py`: `_load_module_without_db()` imports the monolithic Flask route module under a fresh module name for each new test, but that module registers Prometheus metrics at import time against the global default registry. The second test therefore fails before it reaches the new validation assertion.

Suggested fixes:
- Reuse one imported module for these pre-DB validation cases instead of importing the route module twice under different names, or
- reset/isolate the Prometheus registry in the test helper before loading the module, matching whatever pattern the rest of the suite uses for this app, or
- move these two cases into a single test method with one module import if that is the least invasive fix.

Once the two new route tests can run together, this should be in good shape.

PR #4263, opened 2026-05-11
fix: use Ed25519 public key in miner enrollment

### LINKED ISSUE #2760: [BUG] Enrollment payload miner_pubkey field contains wallet address instead of Ed25519 public key
### Severity
Medium — enrollment will fail if the node validates that `miner_pubkey` is a valid Ed25519 public key format.

### Location
`rustchain-miner/src/miner.rs` — `enroll()` function, lines ~260-280

### Description

In the `enroll()` function, the JSON payload is constructed with:

```json
{
  "miner_pubkey": self.wallet,      // ← BUG: submits wallet address
  "miner_pubkey": self.public_key_hex,  // ← should be this
  "miner_id": self.miner_id,
  "device": { ... },
  "signature": signature_hex,
  "public_key": self.public_key_hex  // ← duplicate, correct value already here
}
```

The `miner_pubkey` field is labeled "pubkey" but contains the RTC wallet address (e.g., `RTC9a39ca2c84f61ca27d96463bcf65b6022b827f85`), not the Ed25519 public key used for signing.

The signing keypair is generated in `Miner::new()`:
```rust
let signing_key = ed25519_dalek::SigningKey::generate(&mut rand::rngs::OsRng);
let verifying_key = signing_key.verifying_key();
let public_key_hex = hex::encode(verifying_key.as_bytes());
```

And the enrollment signature is created with:
```rust
let enroll_message = format!("{}|{}|{}", self.wallet, self.miner_id, epoch);
let signature = self.signing_key.sign(enroll_message.as_bytes());
```

The signature correctly uses the Ed25519 signing key and the wallet as identity. However, `miner_pubkey` in the payload should contain `self.public_key_hex` (the Ed25519 public key), not `self.wallet`.

### Expected vs Actual Behavior

- **Expected**: Node validates that `miner_pubkey` is a valid Ed25519 public key matching the signer's verifying key.
- **Actual**: Node receives an RTC wallet address in a field labeled "pubkey". If the node validates pubkey format, enrollment fails. If it ignores the field, it may accept the enrollment but with incorrect identity binding.

### Impact

If the RustChain node enforces that `miner_pubkey` must be a valid Ed25519 public key (40 hex chars), all miner enrollments will be rejected. The `public_key` field already carries the correct value, suggesting `miner_pubkey` is either a copy-paste error or intended to be the Ed25519 pubkey.

### Suggested Fix

```rust
let payload = serde_json::json!({
    "miner_pubkey": self.public_key_hex,  // fix: use Ed25519 pubkey, not wallet
    "miner_id": self.miner_id,
    "device": {
        "family": self.hw_info.family,
        "arch": self.hw_info.arch
    },
    "signature": signature_hex,
    "public_key": self.public_key_hex
});
```

Or remove the duplicate `public_key` field if both are not needed.

### Wallet for bounty payout
RTC9a39ca2c84f61ca27d96463bcf65b6022b827f85

cerredz:
Fixes #2760.

Changes:
- send the Ed25519 public key in miner_pubkey instead of the wallet address
- sign the enrollment message with the same public key identity that the node verifies
- keep public_key populated for the current signed enrollment contract
- add regression tests for the payload identity and signature binding

Verification:
- cargo test enrollment_
- cargo test
- git diff --check -- rustchain-miner/src/miner.rs

Note: cargo fmt --check currently reports pre-existing formatting drift in unrelated Rust miner files, so this PR avoids crate-wide formatting churn.

### DIFF
--- a/rustchain-miner/src/miner.rs
+++ b/rustchain-miner/src/miner.rs
@@ -104,6 +104,28 @@ pub struct Miner {
     shutdown: Arc<AtomicBool>,
 }
 
+fn enrollment_message(miner_pubkey: &str, miner_id: &str, epoch: u64) -> String {
+    format!("{}|{}|{}", miner_pubkey, miner_id, epoch)
+}
+
+fn enrollment_payload(
+    miner_pubkey: &str,
+    miner_id: &str,
+    hw_info: &HardwareInfo,
+    signature_hex: &str,
+) -> serde_json::Value {
+    serde_json::json!({
+        "miner_pubkey": miner_pubkey,
+        "miner_id": miner_id,
+        "device": {
+            "family": hw_info.family,
+            "arch": hw_info.arch
+        },
+        "signature": signature_hex,
+        "public_key": miner_pubkey
+    })
+}
+
 impl Miner {
     /// Create a new miner with the given configuration
     pub async fn new(config: Config) -> Result<Self> {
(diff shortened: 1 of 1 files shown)


omercelik (changes requested):
Requesting changes because the client-side fix changes the enrollment identity without changing the attestation identity used by the node.

In `rustchain-miner/src/attestation.rs`, the Rust miner still submits attestations with `miner: wallet.to_string()`. The node stores that row in `miner_attest_recent` keyed by that `miner` value and keeps the signing key in `signing_pubkey`. This PR changes `/epoch/enroll` to send `miner_pubkey` as the Ed25519 public key, but the server lookup in `node/rustchain_v2_integrated_v2.2.1_rip200.py` uses `SELECT signing_pubkey FROM miner_attest_recent WHERE miner = ?` with `miner_pk` from the enrollment payload, and `check_enrollment_requirements(miner_pk)` uses the same key.

That means a normal flow becomes:
1. attestation row stored under the wallet address;
2. enrollment sent under the Ed25519 public key;
3. server cannot find the stored signing_pubkey/recent attestation for the enrollment key.

Depending on node flags, this either falls back to the legacy no-stored-signing-pubkey path or rejects with `no_recent_attestation`, so the PR does not actually bind enrollment to the prior attestation and may break Rust miner enrollment. The fix needs to make attestation and enrollment use the same canonical identity, or adjust the node lookup to resolve the enrollment pubkey back to the attested wallet/signing key.

Validation:
- `cargo test enrollment_` -> passed
- `cargo check` -> passed
- `git diff --check origin/main...HEAD -- rustchain-miner/src/miner.rs` -> passed

PR #4266, opened 2026-05-11
fix: make P2P gossip dedup insert atomic

### LINKED ISSUE #4265: [BUG] P2P gossip dedup race allows duplicate message processing
## Summary
`GossipLayer.handle_message()` checks `p2p_seen_messages` with a SELECT before signature verification, but the later persistent record step uses `INSERT OR IGNORE` and ignores whether the insert actually inserted a row. Two workers handling the same valid signed message can both pass the early SELECT and both continue into message-specific processing.

## Impact
Duplicate signed gossip messages can be processed concurrently despite the persistent `msg_id` uniqueness constraint. For state-changing message types this can duplicate local processing work or trigger duplicate downstream behavior. The repository self-audit documents this pattern as `CVE-007: Race Condition in Deduplication (TOCTOU)` in `audits/p2p_gossip/self_audit_7440.md`.

## Reproduction concept
1. Worker A handles a valid P2P message and passes the early `SELECT 1 FROM p2p_seen_messages WHERE msg_id = ?` check.
2. Worker B handles the same message and inserts the `msg_id` first.
3. Worker A verifies the signature, runs `INSERT OR IGNORE`, ignores the zero-row insert, and still processes the message.

## Expected behavior
After signature verification, the durable insert should be the authoritative duplicate gate. If `INSERT OR IGNORE` inserts zero rows, the handler should return `{status: duplicate}` before message-specific processing.

cerredz:
## Summary
- make the post-verification `p2p_seen_messages` insert the authoritative duplicate gate
- return `duplicate` when `INSERT OR IGNORE` reports zero changed rows, preventing TOCTOU duplicate processing
- make the in-memory fallback duplicate check/add atomic under `self.lock`
- update P2P gossip signature callers to ignore the current `unpack_signature()` key-version return value
- add a regression test that simulates another handler winning the insert after the early precheck but before message processing

Fixes #4265.

## Validation
- `python -m pytest node\tests\test_p2p_hardening_phase2.py -q` -> 7 passed
- `python -m pytest node\tests\test_p2p_identity_hardening.py -q` -> 3 passed
- `python -m py_compile node\rustchain_p2p_gossip.py node\tests\test_p2p_hardening_phase2.py` -> passed
- `git diff --check -- node\rustchain_p2p_gossip.py node\tests\test_p2p_hardening_phase2.py` -> passed

## Known unrelated baseline failure
- `python -m pytest node\tests\test_p2p_phase_f_ed25519.py -q` still fails on current main because the tests expect the old two-value/no-version `unpack_signature()` shape and one file-mode assertion observes `0o666`; this PR updates the production gossip caller but does not rewrite that broader Phase F test suite.

### DIFF
--- a/node/rustchain_p2p_gossip.py
+++ b/node/rustchain_p2p_gossip.py
@@ -455,7 +455,7 @@ class GossipLayer:
         mode = self._signing_mode
 
         from p2p_identity import unpack_signature, verify_ed25519
-        hmac_sig, ed25519_sig = unpack_signature(signature)
+        hmac_sig, ed25519_sig, _key_version = unpack_signature(signature)
 
         # "strict" mode: only Ed25519 accepted. HMAC-only sigs are rejected
         # even if valid (flag-day enforcement).
@@ -538,7 +538,7 @@ class GossipLayer:
         mode = self._signing_mode
 
         from p2p_identity import unpack_signature, verify_ed25519
-        hmac_sig, ed25519_sig = unpack_signature(msg.signature)
+        hmac_sig, ed25519_sig, _key_version = unpack_signature(msg.signature)
 
         # 1) Try Ed25519 if available AND peer is registered.
         if ed25519_sig and self._peer_registry is not None:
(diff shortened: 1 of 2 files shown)


douglance (approved):
Reviewed the P2P deduplication hardening and the fix makes the persistent insert the authoritative duplicate gate. The new `INSERT OR IGNORE`/`changes()` check closes the post-precheck race, and the in-memory fallback now performs check/add under `self.lock`, so the fallback path is also atomic. The `unpack_signature()` caller updates are scoped to accepting the current three-value return shape.

Local validation:

- `uv run --no-project --with pytest --with flask --with requests python -m pytest node/tests/test_p2p_hardening_phase2.py -q` -> 7 passed
- `uv run --no-project --with pytest --with cryptography python -m pytest node/tests/test_p2p_identity_hardening.py -q` -> 3 passed
- `python3 -m py_compile node/rustchain_p2p_gossip.py node/tests/test_p2p_hardening_phase2.py` -> passed
- `git diff --check origin/main...HEAD -- node/rustchain_p2p_gossip.py node/tests/test_p2p_hardening_phase2.py` -> passed

I do not see a blocker in this PR.

PR #4268, opened 2026-05-11
fix: remove invalid python escape warnings

answndud:
Refs #305.

## Summary
- remove invalid escape sequences from three Python files that emit `SyntaxWarning` during compilation
- keep the generated markdown text and grep pattern behavior unchanged
- mark the Windows config-manager docstring as raw so the documented Windows path is not parsed as Python escapes

## Bug
Running Python compilation on these files emits invalid escape warnings:

- `tools/bounty-bot-pro/verifier.py`: `\`` in an f-string markdown cell
- `miners/power8/fingerprint_checks_power8.py`: `\|` in the grep pattern string
- `miners/windows/installer/src/config_manager.py`: `\R` in the Windows path inside the module docstring

These warnings are easy to miss today, but they are Python parser warnings and can become hard failures under stricter warning settings or future Python versions.

## Verification
- `python3 -W error::SyntaxWarning -m py_compile tools/bounty-bot-pro/verifier.py miners/power8/fingerprint_checks_power8.py miners/windows/installer/src/config_manager.py`
- `python3 -m compileall -f -q tools/bounty-bot-pro/verifier.py miners/power8/fingerprint_checks_power8.py miners/windows/installer/src/config_manager.py`
- `git diff --check -- tools/bounty-bot-pro/verifier.py miners/power8/fingerprint_checks_power8.py miners/windows/installer/src/config_manager.py`

PR #4270, opened 2026-05-11
fix: guard bounty verifier star lookup errors

answndud:
Refs #305.

## Summary
- avoid a `KeyError` when GitHub star lookup fails and `verify_stars()` returns an error payload
- compute payout/report output from a safe `star_count` fallback
- only award the star-king bonus when `is_star_king` is explicitly true

## Bug
`verify_stars()` returns `{ "error": ..., "count": 0 }` on `GithubException`, but `generate_report()` unconditionally reads `stars["is_star_king"]`. If GitHub lookup fails for a missing user, auth/rate-limit issue, or transient API error, the full verification report crashes with `KeyError` instead of producing a zero-star report.

## Verification
- `python3 -m py_compile tools/bounty-bot-pro/verifier.py`
- `git diff --check -- tools/bounty-bot-pro/verifier.py`

Note: this branch is based on `origin/main`, so it still shows the pre-existing invalid-backtick SyntaxWarning that PR #4268 fixes separately. This PR is scoped only to the star lookup error path.

### DIFF
--- a/tools/bounty-bot-pro/verifier.py
+++ b/tools/bounty-bot-pro/verifier.py
@@ -95,15 +95,16 @@ class BountyVerifier:
         follows = self.verify_following(username)
         wallet_info = self.verify_wallet(wallet)
         
-        payout = stars["count"] * CONFIG["star_reward"]
+        star_count = stars.get("count", 0)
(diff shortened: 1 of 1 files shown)


answndud:

PR #4273, opened 2026-05-11
fix: preserve vintage evidence warnings

answndud:
Refs #305.

## Summary
- preserve `WARN` status for unusually small vintage photo evidence
- preserve `WARN` status for unusual photo formats
- preserve `WARN` status for unusually small screenshots

## Bug
`validate_photo()` and `validate_screenshot()` set `result["status"] = "WARN"` for suspicious evidence, but then unconditionally overwrite the result with `PASS` before returning. This hides weak evidence from the validation report.

## Verification
- smoke test with a 1-byte `.jpg` now returns `WARN Photo file seems too small: 1 bytes`

### DIFF
--- a/tools/validate_vintage_submission.py
+++ b/tools/validate_vintage_submission.py
@@ -62,8 +62,9 @@ class SubmissionValidator:
         # - Image content (machine + monitor)
         # - Metadata consistency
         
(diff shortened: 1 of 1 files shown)


answndud:

PR #4279, opened 2026-05-11
fix: preserve vintage writeup warning details

answndud:
Refs #305.

## Summary
- preserve both missing-section and short-writeup warning details
- avoid overwriting the first writeup warning message with the second one

## Bug
`validate_writeup()` can detect multiple warning conditions in one writeup. If sections are missing and the writeup is also under 100 words, the later short-writeup check overwrites the earlier `Missing sections` message. The final report then hides which required sections were missing.

## Verification
- smoke test with `cpu only` now returns `WARN` and includes both missing sections and short writeup details


answndud:

PR #4280, opened 2026-05-11
fix: reject empty vintage validation requests

answndud:
Refs #305.

## Summary
- make `validate_submission()` return invalid when no evidence inputs are provided
- align the function API with the CLI guard that already rejects empty validation runs

## Bug
The CLI checks that at least one validation input is provided, but the underlying `SubmissionValidator.validate_submission()` API returned `valid=True` when called with all arguments omitted. Any direct caller using the validator class could therefore accept an empty evidence set as valid.

## Verification
- smoke test: `SubmissionValidator().validate_submission()` now returns `False` with `At least one validation input is required`

### DIFF
--- a/tools/validate_vintage_submission.py
+++ b/tools/validate_vintage_submission.py
@@ -290,6 +290,11 @@ class SubmissionValidator:
             "bounty": None,
             "era": None
         }
+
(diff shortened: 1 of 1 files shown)


answndud:

PR #4281, opened 2026-05-11
fix: require core fields in text attestation logs

answndud:
Refs #305.

## Summary
- require both `miner_id` and `device_arch` before accepting a plain-text attestation log
- keep JSON attestation validation unchanged

## Bug
When an attestation log is not valid JSON, `validate_attestation_log()` falls back to a plain-text heuristic. The heuristic accepted the log if it contained `miner_id` OR `device_arch`, so a partial text file with only one field could pass as a valid attestation log.

## Verification
- smoke test: a plain-text log containing only `miner_id` now returns `WARN` instead of `PASS`


answndud:

PR #4282, opened 2026-05-11
fix: resolve vintage profile path from validator

answndud:
## Summary
- resolve the vintage hardware profile import path relative to the validator script instead of the caller's current working directory
- reuse the same path setup for bounty and era lookups

## Why
Running the validator from `tools/` or another working directory can make `sys.path.insert(0, "vintage_miner")` point at a non-existent relative directory. That silently falls back to the default bounty and `Unknown` era even when `vintage_miner/hardware_profiles.py` is present in the repo.

## Test plan
- smoke-tested `SubmissionValidator().calculate_bounty("x86_64")` and verified the resolved profile path is added
- `git diff --check`


answndud:

PR #4283, opened 2026-05-11
fix: guard miner alert subscription filter

answndud:
## Summary
- add an explicit alert-type to column whitelist for miner alert subscription filtering
- reject unknown alert types before building the SQL query

## Why
`get_subscriptions()` accepted an arbitrary `alert_type` and interpolated `alert_{alert_type}` into the query. Normal internal callers use known constants, but direct or future callers could trigger SQL errors or unsafe column expressions.

## Test plan
- `python3 -m py_compile tools/miner_alerts/miner_alerts.py`
- smoke-tested valid `offline` filtering and invalid alert-type rejection with a temporary SQLite DB
- `git diff --check`

### DIFF
--- a/tools/miner_alerts/miner_alerts.py
+++ b/tools/miner_alerts/miner_alerts.py
@@ -73,6 +73,13 @@ logging.basicConfig(
 )
 logger = logging.getLogger("miner_alerts")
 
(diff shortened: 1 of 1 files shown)


answndud:

PR #4286, opened 2026-05-11
fix: fall back when balance field is null

answndud:
## Summary
- treat a JSON `balance: null` value as missing
- fall back to `balance_rtc` before converting the miner balance to float

## Why
`dict.get("balance", fallback)` does not use the fallback when the key exists with a null value. A response like `{ "balance": null, "balance_rtc": "7.5" }` currently raises `TypeError` and makes balance monitoring return `None`, skipping reward/transfer detection for that poll.

## Test plan
- `python3 -m py_compile tools/miner_alerts/miner_alerts.py`
- smoke-tested `fetch_balance()` with a stubbed response containing `balance: null` and `balance_rtc: "7.5"`
- `git diff --check`

### DIFF
--- a/tools/miner_alerts/miner_alerts.py
+++ b/tools/miner_alerts/miner_alerts.py
@@ -495,7 +495,10 @@ def fetch_balance(miner_id: str) -> Optional[float]:
             return None
         resp.raise_for_status()
         data = resp.json()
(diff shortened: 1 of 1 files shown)


answndud:

PR #4290, opened 2026-05-11
fix: normalize webhook epoch values

answndud:
## Summary
- normalize `/api/stats` epoch values to integers before comparing poll state
- ignore malformed epoch values instead of updating state with inconsistent types

## Why
`_check_epoch()` compares the raw `epoch` value across polls. If the node returns `"7"` in one response and `7` in the next, the poller treats that as a changed epoch and dispatches a false `new_epoch` event.

## Test plan
- `python3 -m py_compile tools/webhooks/webhook_server.py`
- smoke-tested consecutive epoch responses of `"7"` and `7` and verified no event is dispatched
- `git diff --check`

### DIFF
--- a/tools/webhooks/webhook_server.py
+++ b/tools/webhooks/webhook_server.py
@@ -368,6 +368,11 @@ class RustChainPoller:
         epoch = stats.get("epoch")
         if epoch is None:
             return
+        try:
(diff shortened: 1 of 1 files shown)


answndud:

PR #755, opened 2026-03-09
Add RustChain API First Call Walkthrough

dagangtj:
## Summary
Comprehensive developer onboarding documentation for RustChain API.

## What's Included
- ✅ Health check endpoint example (tested)
- ✅ Epoch information query (tested)
- ✅ Balance lookup with miner_id (tested)
- ✅ Signed transfer request format with field explanations
- ✅ Self-signed TLS certificate handling
- ✅ Error handling and troubleshooting
- ✅ Comparison with Ethereum/Solana

## Testing
All read-only endpoints tested against https://50.28.86.131:
- /health ✅
- /epoch ✅
- /wallet/balance ✅

## Closes
Closes #701

## Bounty
Claiming bounty #1494 (28 RTC)
Wallet: first-butler

PR #4431, opened 2026-05-11
feat: add backup verification script (bounty #755)

watcharaponthod-code:
Adds `verify_backup.sh` to validate RustChain DB backups by running integrity checks and comparing row counts with the live DB. Closes #755.

Wallet: RTC843fa92c55b6a71e8432b4bcf54b1f48039ab1

### DIFF
--- a/scripts/verify_backup.sh
+++ b/scripts/verify_backup.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+# verify_backup.sh
+# Validates RustChain DB backups
+# Usage: ./verify_backup.sh [LIVE_DB_PATH] [BACKUP_DIR]
+
+set -e
+
+LIVE_DB="${1:-/root/rustchain/rustchain_v2.db}"
+BACKUP_DIR="${2:-/root/rustchain/backups}"
+
+# Find the latest backup
+if [ ! -d "$BACKUP_DIR" ]; then
+  echo "[$(date +'%Y-%m-%d %H:%M:%S')] ERROR: Backup directory $BACKUP_DIR does not exist."
+  exit 1
+fi
+
+LATEST_BACKUP=$(ls -t "$BACKUP_DIR"/*.db* 2>/dev/null | head -n 1)
+
+if [ -z "$LATEST_BACKUP" ]; then
+  echo "[$(date +'%Y-%m-%d %H:%M:%S')] ERROR: No backup found in $BACKUP_DIR"
+  exit 1
+fi
+
+echo "[$(date +'%Y-%m-%d %H:%M:%S')] Backup: $LATEST_BACKUP"
+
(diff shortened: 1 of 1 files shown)


godd-ctrl (changes requested):
Reviewed against bounty #755 requirements. Two blocking issues: (1) latest backup selection uses "$BACKUP_DIR"/*.db*, which can pick SQLite sidecar files such as .db-wal/.db-shm if they are newer than the actual backup DB, causing false verification failures or verifying the wrong file. Please restrict this to actual backup database files or explicitly exclude sidecars/compressed files unless they are handled. (2) The row-count/freshness requirement is not implemented: the script only fails when a backup table is empty while the live table is nonzero, so a stale backup with 1 row versus a live table with many rows would still PASS. Bounty #755 asks that the backup should not be more than 1 epoch behind and that key tables have meaningful data, e.g. positive balances and recent attestations. Please compare backup/live counts or relevant height/epoch/timestamp fields with a threshold and fail stale backups. Minor: the new script also lacks the SPDX header requested by the repo welcome checklist. Validation: git diff --check -- scripts\verify_backup.sh passed; full execution was not possible in my local Windows environment because sqlite3 CLI is not installed, so this review is based on static/script-path validation and the bounty requirements.

bolasse1234 (changes requested):
Thanks for putting this together. I think it needs another pass before merge because it currently accepts stale/incomplete backups as PASS.

I built a temp live DB and backup DB with the five required tables. The live DB had 100 rows in every table; the backup had 1 row in every table. Running this script returned exit 0 and printed PASS for every table:

`	ext
balances: 1 rows (live: 100) ✅
miner_attest_recent: 1 rows (live: 100) ✅
headers: 1 rows (live: 100) ✅
ledger: 1 rows (live: 100) ✅
epoch_rewards: 1 rows (live: 100) ✅
RESULT: PASS
`

That misses the main freshness requirement from #755: the backup should be compared against the live DB and should not be more than about one epoch behind. Right now the script only fails when a backup table is exactly empty while the live count is non-zero, so a badly stale backup with tiny non-zero counts still passes.

A couple of related checks from the bounty are also not covered yet:
- alances is only checked for row count, not rows with positive amounts.
- miner_attest_recent is only checked for row count, not recent attestations.

Validation I ran:
- C:\Program Files\Git\bin\bash.exe scripts/verify_backup.sh <live-db> <backup-dir> against the 100-row live / 1-row backup setup => returned 0 PASS, which should be FAIL.
- C:\Program Files\Git\bin\bash.exe -n scripts/verify_backup.sh passed.
- git diff --check origin/main...HEAD -- scripts\\verify_backup.sh failed on trailing whitespace at lines 53 and 62.

I would suggest adding explicit per-table drift thresholds, plus at least one positive-balance check for alances, before this is used as a cron health check.

Issue #2811, opened 2026-04-30
🔒 Self-Audit Batch 7: payout_worker, utxo_endpoints, rustchain_tx_handler


haoyousun60-create:
# 🔒 RustChain Self-Audit Report — Batch 7

**Auditor:** 马上发财 Self-Audit Bot
**Date:** 2026-05-01
**Files Audited:** 3
**Wallet:** RTC4642c5ee8467f61ed91b5775b0eeba984dd776ba

---

### 🔴 CRITICAL — P1: Production Withdrawal Execution is a No-Op (Line 82-86)

```python
def execute_withdrawal(self, withdrawal: Dict) -> Optional[str]:
    if MOCK_MODE:
        # ... mock logic ...
    else:
        # Real blockchain integration would go here
        pass  # ← Returns None
```

**Impact:** In production mode (`MOCK_MODE=0`), `execute_withdrawal()` returns `None` (Python default for `pass`). This means:
- Balance is **deducted** from the sender (line 101)
- `tx_hash` is `None`, so the withdrawal enters the `except` branch (line 115)
- Balance is **refunded** in the except block
- Withdrawal is marked `failed`
- **No withdrawal ever completes in production mode.** Funds are locked in a loop of deduct→fail→refund.

Additionally, an attacker who sets `RUSTCHAIN_MOCK_MODE=1` in the environment can force mock mode on a production server, generating fake transaction hashes for withdrawals.

**Fix:**
```python
def execute_withdrawal(self, withdrawal: Dict) -> Optional[str]:
    if MOCK_MODE:
        # ... existing mock logic ...
    else:
        raise NotImplementedError(
            "Production withdrawal execution not implemented. "
            "Aborting to prevent fund loss."
        )
```
Also: validate `MOCK_MODE` cannot be enabled when `DB_PATH` points to a production database.

---

### 🟡 MEDIUM — P2: Archive File Path Injection (Lines 155-170)

```python
archive_file = f"withdrawal_archive_{datetime.now().strftime('%Y%m%d')}.json"
with open(archive_file, 'a') as f:
```

The archive filename is constructed from the date only (safe), but the data written includes **unvalidated fields from the database** (`miner_pk`, `destination`, `tx_hash`). If any of these contain newlines or JSON-escape sequences, the archive file becomes corrupted or exploitable when parsed by downstream tools.

**Fix:** Validate the output file path is within an expected directory:
```python
[...]

PR #4456, opened 2026-05-11
fix: reject legacy UTXO signatures with fees

ai-wes:
## Summary

Follow-up hardening for the UTXO legacy-signature compatibility path discussed in #2811 and relevant to the red-team bounty in Scottcjn/rustchain-bounties#2819.

`/utxo/transfer` currently prefers the v2 signed payload that includes `fee_rtc`, then falls back to a legacy signed payload that omits the fee for compatibility. During that compatibility window, a legacy signature can still be accepted with a nonzero fee even though the fee is not bound by the signature.

This patch rejects legacy-signature transfers whenever `fee_rtc` is nonzero, while preserving zero-fee legacy compatibility and the v2 fee-bound path.

## Validation

- `/tmp/earn-micro-bounties-rustchain/.venv/bin/python node/test_utxo_endpoints.py`
- `/tmp/earn-micro-bounties-rustchain/.venv/bin/python -m py_compile node/utxo_endpoints.py node/test_utxo_endpoints.py`
- `git diff --check -- node/utxo_endpoints.py node/test_utxo_endpoints.py`

No public payment details included; please use a private channel if payout information is needed.

### DIFF
--- a/node/test_utxo_endpoints.py
+++ b/node/test_utxo_endpoints.py
@@ -322,6 +322,47 @@ class TestUtxoEndpoints(unittest.TestCase):
         self.assertIn('signed payload', r.get_json()['error'])
         self.assertEqual(self.utxo_db.get_balance(recipient), 0)
 
+    def test_legacy_signature_rejects_nonzero_fee(self):
+        """Legacy signatures omit fee_rtc, so they cannot authorize fees."""
+        sender = 'RTC_test_aabbccdd'
+        recipient = 'bob'
+        self._seed_coinbase(sender, 100 * UNIT)
+
+        signed_message = json.dumps({
+            'from': sender,
+            'to': recipient,
+            'amount': 10.0,
+            'memo': '',
+            'nonce': 515151,
+        }, sort_keys=True, separators=(',', ':')).encode()
+
+        old_verify = utxo_endpoints._verify_sig_fn
+
+        def verify_legacy_only(pubkey_hex, message, sig_hex):
+            return sig_hex == 'legacy-sig' and message == signed_message
+
+        try:
+            utxo_endpoints._verify_sig_fn = verify_legacy_only
+            r = self.client.post('/utxo/transfer', json={
+                'from_address': sender,
+                'to_address': recipient,
+                'amount_rtc': 10.0,
+                'fee_rtc': 1.0,
+                'public_key': 'aabbccdd' * 8,
+                'signature': 'legacy-sig',
+                'nonce': 515151,
+            })
+        finally:
+            utxo_endpoints._verify_sig_fn = old_verify
+
+        self.assertEqual(r.status_code, 401)
+        self.assertEqual(
+            r.get_json()['code'],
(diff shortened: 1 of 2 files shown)


saim256 (approved):
Reviewed the legacy UTXO signature fee guard against #2811 / rustchain-bounties#2819. The patch closes the fee-unbound compatibility path in the right place: after v2 fee-bound signatures fail and before accepting the legacy message that omits `fee_rtc`. That means modern clients can still authorize signed fees, while legacy signatures are only accepted for zero-fee transfers and cannot be replayed with a newly injected nonzero fee.

I also checked the regression setup. The test forces a verifier that only accepts the legacy message bytes, sends `fee_rtc: 1.0`, and confirms the endpoint returns `401 LEGACY_SIGNATURE_FEE_UNBOUND` without debiting the sender or crediting the recipient. Existing amount parsing and signed-float preservation checks still run before signature acceptance, so this change does not loosen the earlier Decimal/precision hardening.

Validation run locally:
- `python node/test_utxo_endpoints.py` -> 18 tests passed
- `python -m py_compile node/utxo_endpoints.py node/test_utxo_endpoints.py`
- `git diff --check origin/main...HEAD -- node/utxo_endpoints.py node/test_utxo_endpoints.py`

Looks good to merge for the stated fee-unbound legacy-signature scope.

godd-ctrl (changes requested):
Requested changes after local validation.

The focused UTXO endpoint suite passes when run with the required Flask dependency, and the new legacy-signature/nonzero-fee regression covers the intended security property. However, the standard mempool baseline regression fails on this branch because UtxoDB(':memory:').mempool_add(...) reaches mempool_clear_expired(), which queries utxo_mempool before that table exists.

Focused validation passed:
- python -m py_compile node\utxo_endpoints.py node\test_utxo_endpoints.py
- uv run --with flask python node\test_utxo_endpoints.py
- git diff --check -- node\utxo_endpoints.py node\test_utxo_endpoints.py

Blocking validation failure:
- uv run --with pytest --with flask --with requests python -m pytest tests\security_audit\test_security_findings_2867.py::test_mempool_add_manage_tx_undefined -q
- failure: sqlite3.OperationalError: no such table: utxo_mempool in UtxoDB.mempool_clear_expired()

Please include the same utxo_mempool missing-table guard used by adjacent validated PRs, or otherwise ensure init_tables()/mempool_add() can satisfy the in-memory mempool regression before merge.

saim256:
Follow-up on my approval: my local validation was scoped to `node/test_utxo_endpoints.py`, `py_compile`, and diff-check for the touched files. I did not run the `tests/security_audit/test_security_findings_2867.py::test_mempool_add_manage_tx_undefined` baseline on this branch.

@godd-ctrl's later review identifies a broader mempool missing-table regression outside the focused legacy-fee test. I agree that should be treated as blocking before merge unless this PR picks up the same `mempool_clear_expired()` missing-table guard used in nearby validated branches.

PR #4484, opened 2026-05-11
test: cover legacy OS badge detection

MolhamHamwi:
## Summary
- Adds pytest coverage for `tools/os_detector.py` legacy badge detection.
- Covers multiple matching legacy environments in one directory probe.
- Covers the edge case where the directory probe fails and no badges are emitted.

## Bounty
Refs Scottcjn/rustchain-bounties#1589.

## Test Plan
- `python -m pytest tools/test_os_detector.py -q`

### DIFF
--- a/tools/test_os_detector.py
+++ b/tools/test_os_detector.py
@@ -1,0 +1,52 @@
+import importlib.util
+from pathlib import Path
+from unittest.mock import patch
+
+
+MODULE_PATH = Path(__file__).resolve().parent / "os_detector.py"
+spec = importlib.util.spec_from_file_location("os_detector", MODULE_PATH)
+os_detector = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(os_detector)
+
+
+class FixedDateTime:
+    @staticmethod
+    def utcnow():
+        class FixedNow:
+            @staticmethod
+            def isoformat():
+                return "2026-05-11T12:00:00"
+
+        return FixedNow()
+
+
+def test_detect_legacy_os_badges_detects_multiple_matching_environments():
+    directory_listing = "System Folder\nFinder\nwin.ini\nprogman.exe\n"
+
+    with patch.object(
+        os_detector.subprocess,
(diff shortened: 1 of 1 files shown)


saim256 (changes requested):
The test coverage itself looks good, but one checklist item needs fixing before merge: this PR adds a new Python file, `tools/test_os_detector.py`, without an SPDX header. The repository PR checklist asks new code files to include an SPDX license header near the top, and the welcome bot repeated that requirement on this PR.

Please add the appropriate SPDX header to the new test file. The test behavior can stay as-is.

What I checked:
- The multi-match test correctly exercises one directory probe producing both the MacOS and Win3x badges.
- The failure-path test correctly patches `subprocess.check_output` to raise and asserts the detector returns an empty badge list.
- The timestamp patch makes the assertions deterministic.

Validation run locally:
- `python -m pytest tools\test_os_detector.py -q` -> 2 passed
- `python -m py_compile tools\test_os_detector.py tools\os_detector.py`
- `git diff --check origin/main...HEAD -- tools/test_os_detector.py`

douglance (approved):
Reviewed the legacy OS badge detection coverage and this looks good to merge.

What I checked:
- The new tests import `tools/os_detector.py` directly, so they exercise the current badge detection function without requiring the script to be installed as a package.
- The multi-match test mocks the directory probe and confirms one listing can award both the MacOS and Win3x badges in the order produced by `simulated_os_data`.
- The timestamp is fixed through a patched `datetime`, which keeps the resonance assertion deterministic.
- The failure-path test confirms a failed `dir` probe returns the existing empty-badge shape instead of raising.

Validation run locally:
- `python3 -m py_compile tools/os_detector.py tools/test_os_detector.py`
- `uv run --no-project --with pytest python -m pytest tools/test_os_detector.py -q` -> 2 passed
- `git diff --check origin/main...HEAD -- tools/test_os_detector.py`

No blocking findings for this test-only scope.

PR #4492, opened 2026-05-11
fix: escape miner dashboard api fields

### LINKED ISSUE #4491: [Bug] Miner dashboard renders API history and activity fields as HTML
## Summary
`dashboards/miner-dashboard/index.html` renders wallet history and miner activity fields into `innerHTML`, and it assumes older array-shaped API responses from endpoints that now return envelope objects.

## Affected file
- `dashboards/miner-dashboard/index.html`

## Impact
Wallet transaction fields such as `counterparty` and miner fields such as `miner` / `hardware_type` are rendered into table rows using template literals assigned to `innerHTML`. If those API fields contain markup, the dashboard interprets them as HTML.

The page also currently calls `/api/miners` and `/wallet/history`, but treats both responses as arrays. The live APIs return objects such as `{ miners: [...] }` and `{ transactions: [...] }`, so the dashboard can fail before rendering useful data.

## Reproduction
1. Open `dashboards/miner-dashboard/index.html`.
2. Load a miner whose wallet history or miner metadata contains HTML-like values, or mock the API responses locally.
3. The dashboard writes those values into `historyTable.innerHTML` / `activityTable.innerHTML` as markup.

## Expected behavior
The dashboard should normalize the supported API envelopes and render dynamic table values with DOM/text APIs rather than string-concatenated HTML.

## Suggested fix
Normalize `/api/miners` and `/wallet/history` response envelopes, render history/activity rows with `insertCell().textContent`, and update the message helper to avoid injecting error text through `innerHTML`.

cerredz:
## Summary
- normalize current `/api/miners` and `/wallet/history` response envelopes before rendering
- render wallet history rows and miner activity fields with DOM/text APIs instead of `innerHTML`
- update dashboard messages to use text nodes for error text
- add static frontend regression coverage for API field rendering and response normalization
- keep the local mempool regression compatible with databases that predate the mempool tables

Fixes #4491

## Verification
- `python -m pytest tests\test_miner_dashboard_frontend_security.py -q`
- `python -m py_compile tests\test_miner_dashboard_frontend_security.py node\utxo_db.py`
- full script parse for `dashboards/miner-dashboard/index.html` with `node -e`
- `git diff --check -- dashboards\miner-dashboard\index.html tests\test_miner_dashboard_frontend_security.py node\utxo_db.py`

Note: the HTML file contains pre-existing mojibake in emoji text. One narrow block replacement was used after patch matching failed on that encoded text; the final diff is scoped to the dashboard functions shown above.

### DIFF
--- a/dashboards/miner-dashboard/index.html
+++ b/dashboards/miner-dashboard/index.html
@@ -419,7 +419,8 @@
                 headers: { 'Accept': 'application/json' }
             });
             if (!response.ok) throw new Error('Failed to fetch miners');
-            const miners = await response.json();
+            const payload = await response.json();
+            const miners = Array.isArray(payload) ? payload : (payload.miners || payload.data || []);
             return miners.find(m => m.miner === minerId) || null;
         }
 
@@ -438,7 +439,8 @@
                 headers: { 'Accept': 'application/json' }
             });
             if (!response.ok) throw new Error('Failed to fetch history');
-            return await response.json();
+            const payload = await response.json();
+            return Array.isArray(payload) ? payload : (payload.transactions || payload.history || []);
         }
 
         function updateBalance(balance) {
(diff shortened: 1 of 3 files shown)


douglance (approved):
Reviewed the miner dashboard API-field escaping fix and this looks good to merge.

What I checked:
- Transaction history rows no longer build attacker/API-controlled cells with `innerHTML`; rows are rebuilt with `insertRow()` and text cells.
- `counterparty`, miner ID, hardware type, and status values now render through `textContent` rather than HTML interpolation.
- Dynamic status CSS class selection is reduced to fixed `confirmed` / `failed` branches before assigning `className`.
- `showMessage()` no longer interpolates arbitrary error text into an HTML template; it creates an element and sets text content, with class limited to `success` or `error`.
- The remaining `innerHTML` assignments in the touched functions are static empty/no-data table markup, not API-provided fields.
- The API envelope normalization for miners/history is compatible with array and wrapped response shapes.

Validation run locally:
- `python3 -m py_compile tests/test_miner_dashboard_frontend_security.py node/utxo_db.py`
- `uv run --no-project --with pytest --with flask --with requests python -m pytest tests/test_miner_dashboard_frontend_security.py tests/security_audit/test_security_findings_2867.py::test_mempool_add_manage_tx_undefined -q` -> 4 passed, 1 existing pytest return-value warning
- inline script parse check for `dashboards/miner-dashboard/index.html` with `new Function(...)` -> parsed 1 script
- `git diff --check origin/main...HEAD -- dashboards/miner-dashboard/index.html tests/test_miner_dashboard_frontend_security.py node/utxo_db.py`

No blocking findings for the reported DOM-XSS scope.

PR #4523, opened 2026-05-11
Fix BoTTube feed forwarded-host link poisoning

### LINKED ISSUE #4521: [SECURITY] BoTTube RSS/Atom feeds trust X-Forwarded-Host for absolute links
Bounty claim for Scottcjn/rustchain-bounties#71

Wallet: `RTC253255d034065a839cd421811ec589ae5b694ffc`
Requested severity: Low/Medium (host-header injection / feed link poisoning)

## Summary

BoTTube feed routes trust the inbound `X-Forwarded-Host` header directly when constructing absolute feed URLs.

Affected file: `node/bottube_feed_routes.py`

Affected routes:

- `GET /rss`
- `GET /atom`
- feed index route with RSS/Atom negotiation

Each route starts with `request.host_url`, but then replaces it with a raw forwarded host value:

```python
base_url = request.host_url.rstrip(/)
if request.headers.get(X-Forwarded-Host):
    base_url = fhttps://{request.headers['X-Forwarded-Host']}
```

That unvalidated `base_url` is passed into the RSS/Atom builders and used for feed IDs, channel links, item links, author URLs, or discovery links depending on the feed path.

## Local/static reproduction

No production probing was performed.

1. Run the Flask app locally with the BoTTube feed blueprint registered.
2. Request a feed with a forged forwarded host:

```http
GET /rss HTTP/1.1
Host: rustchain.local
X-Forwarded-Host: attacker.example
```

3. The generated feed uses `https://attacker.example` as the base URL for absolute feed links instead of the canonical BoTTube/RustChain host.

The same pattern appears in `/atom` and the feed index route.

## Impact

If the app is deployed behind a proxy that does not scrub `X-Forwarded-Host` from untrusted clients, an attacker can generate RSS/Atom responses containing attacker-controlled absolute links. This can lead to:

- feed-reader link poisoning;
- misleading canonical feed IDs/links;
- phishing or traffic diversion from BoTTube feed consumers;
- cache poisoning risk if an intermediary caches feed responses without varying on forwarded-host headers.

The issue is lower impact than an auth bypass, but it affects public BoTTube API output and is easy to harden.

## Suggested fix

Do not trust raw `X-Forwarded-Host` values from clients. Prefer one of:

1. Use a configured canonical public base URL for feeds, for example `BOTTUBE_PUBLIC_BASE_URL` or `RUSTCHAIN_PUBLIC_BASE_URL`.
2. Only honor forwarded host headers when they come from a trusted proxy and are matched against an allowlist.
3. Validate host syntax and reject/ignore hosts not in the configured allowlist.

A small helper shared by `/rss`, `/atom`, and feed index would avoid future drift.

saim256:
Fixes Scottcjn/Rustchain#4521
Bounty claim: Scottcjn/rustchain-bounties#71
Wallet: RTC253255d034065a839cd421811ec589ae5b694ffc

## Summary
- centralize BoTTube feed base URL selection in `_get_base_url()`
- keep a configured `BOTTUBE_PUBLIC_BASE_URL` as the preferred canonical feed URL
- ignore arbitrary `X-Forwarded-Host` values by default, only allowing explicit `TRUSTED_FORWARD_HOSTS`
- add regression coverage for the untrusted forwarded-host case and configured/trusted host cases

## Verification
- `python -m pytest tests\test_bottube_feed_routes.py -q` -> 25 passed
- `python -m py_compile node\bottube_feed_routes.py tests\test_bottube_feed_routes.py`

### DIFF
--- a/node/bottube_feed_routes.py
+++ b/node/bottube_feed_routes.py
@@ -32,6 +32,20 @@ from bottube_feed import (
 feed_bp = Blueprint("bottube_feed", __name__, url_prefix="/api/feed")
 
 
+def _get_base_url() -> str:
+    """Return the public base URL without trusting arbitrary forwarded hosts."""
+    configured_base_url = current_app.config.get("BOTTUBE_PUBLIC_BASE_URL")
+    if configured_base_url:
+        return str(configured_base_url).rstrip("/")
+
+    forwarded_host = request.headers.get("X-Forwarded-Host", "").strip()
+    trusted_hosts = current_app.config.get("TRUSTED_FORWARD_HOSTS") or []
+    if forwarded_host and forwarded_host in trusted_hosts:
+        return f"https://{forwarded_host}"
+
+    return request.host_url.rstrip("/")
+
+
 def _get_db_connection():
     """Get database connection from Flask app config."""
     db_path = current_app.config.get("DB_PATH")
@@ -218,9 +232,7 @@ def rss_feed():
         videos, next_cursor = _fetch_videos(limit=limit, agent=agent, cursor=cursor)
         
         # Get base URL
-        base_url = request.host_url.rstrip("/")
-        if request.headers.get("X-Forwarded-Host"):
-            base_url = f"https://{request.headers['X-Forwarded-Host']}"
+        base_url = _get_base_url()
         
         # Build RSS feed
         feed_title = "BoTTube Videos"
@@ -274,9 +286,7 @@ def atom_feed():
         videos, next_cursor = _fetch_videos(limit=limit, agent=agent, cursor=cursor)
         
         # Get base URL
-        base_url = request.host_url.rstrip("/")
-        if request.headers.get("X-Forwarded-Host"):
-            base_url = f"https://{request.headers['X-Forwarded-Host']}"
+        base_url = _get_base_url()
         
         # Build Atom feed
         feed_title = "BoTTube Videos"
@@ -341,9 +351,7 @@ def feed_index():
     videos, next_cursor = _fetch_videos(limit=limit, agent=agent, cursor=cursor)
     
     # Get base URL
-    base_url = request.host_url.rstrip("/")
-    if request.headers.get("X-Forwarded-Host"):
-        base_url = f"https://{request.headers['X-Forwarded-Host']}"
+    base_url = _get_base_url()
     
     # Auto-detect format
     if "application/rss+xml" in accept_header:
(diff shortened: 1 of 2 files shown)


douglance (approved):
The forwarded-host hardening looks good. `_get_base_url()` now prefers an explicit configured public URL, only honors `X-Forwarded-Host` when the host is in `TRUSTED_FORWARD_HOSTS`, and falls back to `request.host_url` otherwise, so arbitrary client-supplied forwarded hosts can no longer poison RSS/Atom/feed links by default.

I also checked all three call sites (`/rss`, `/atom`, and the format-negotiating feed index), so the fix is applied consistently across feed outputs. The regression tests cover the untrusted-host rejection path, explicit public base URL path, and trusted forwarded-host path.

Validation run locally:
- `uv run --no-project --with pytest --with flask python -m pytest tests/test_bottube_feed_routes.py -q` -> 25 passed
- `python3 -m py_compile node/bottube_feed_routes.py tests/test_bottube_feed_routes.py` -> passed
- `git diff --check origin/main...HEAD -- node/bottube_feed_routes.py tests/test_bottube_feed_routes.py` -> passed

SAKETH11111 (changes requested):
I found one security gap that should be tightened before merge.

`_get_base_url()` stops trusting arbitrary `X-Forwarded-Host`, but the default fallback still returns `request.host_url`. In Flask/Werkzeug that value is derived from the client-supplied `Host` header unless the app has trusted-host validation configured. With this PR checked out, a request using `Host: attacker.example` still produces attacker-controlled links in all three feed outputs:

- `/api/feed/rss` emits `<link>http://attacker.example</link>`
- `/api/feed/atom` emits `href="http://attacker.example..."`
- `/api/feed` emits JSON `_links` pointing at `http://attacker.example/api/feed/...`

That leaves the same feed link-poisoning class reachable through the standard Host header instead of `X-Forwarded-Host`. I would either require `BOTTUBE_PUBLIC_BASE_URL`/an allowlisted host for generated absolute feed URLs, or add/enforce Flask trusted-host validation and tests proving hostile `Host` values are rejected or ignored.

Validation run locally:
- `uv run --no-project --with pytest --with flask python -m pytest tests/test_bottube_feed_routes.py -q` -> 25 passed
- `python3 -m py_compile node/bottube_feed_routes.py tests/test_bottube_feed_routes.py` -> passed
- `git diff --check origin/main...HEAD -- node/bottube_feed_routes.py tests/test_bottube_feed_routes.py` -> passed
- adversarial Flask client probe with `Host: attacker.example` against `/api/feed/rss`, `/api/feed/atom`, and `/api/feed` -> attacker host still present in generated links

PR #4544, opened 2026-05-11
security: require admin auth for bridge deposits

### LINKED ISSUE #4541: [SECURITY] Bridge initiate can lock another source address without admin authorization
### Summary

`POST /api/bridge/initiate` is documented as an admin endpoint, but `node/bridge_api.py` only uses `X-Admin-Key` to decide whether to bypass the balance check. A request without a valid admin key can still create a RustChain `deposit` transfer for any `source_address` that has balance.

### Impact

Because pending RustChain deposits are included in `check_miner_balance()` as `pending_debits`, an unauthenticated caller can create pending bridge deposits against another wallet/miner id and consume that address's available bridge capacity. The request also creates a `lock_ledger` row for the victim source address. This is bridge availability / ledger-integrity impact rather than a confirmed direct fund-theft claim.

### Local reproduction path

1. Seed `balances` with a funded RustChain source address.
2. POST `/api/bridge/initiate` without `X-Admin-Key` using:
   - `direction: deposit`
   - `source_chain: rustchain`
   - `dest_chain: solana`
   - `source_address: <funded RustChain wallet/miner id>`
   - `amount_rtc: 10`
3. Current behavior: route returns success and inserts `bridge_transfers` plus `lock_ledger` rows.
4. Expected behavior: reject before any bridge/lock rows are created unless the request has valid bridge/admin authorization or a future owner-signature flow.

### Why this looks in scope

`docs/api/README.md` lists `POST /api/bridge/initiate` under `Admin Endpoints (X-Admin-Key Required)`, and `docs/bridge-api.md` says bridge management endpoints require `X-Admin-Key`. The current code does not enforce that for the initiate route.

### Suggested fix

Fail closed on deposit initiation unless `RC_ADMIN_KEY` is configured and the request supplies a matching `X-Admin-Key`, using `hmac.compare_digest()` consistently with the existing bridge admin paths. Add route-level regression coverage that unauthenticated initiation creates no `bridge_transfers` or `lock_ledger` rows, plus a positive admin-key path.

I can provide payout details after maintainer assessment/acceptance; not posting wallet/contact details publicly here.

Zaid-L9:
## Summary

- Require a configured, matching `X-Admin-Key` before `/api/bridge/initiate` can create a RustChain deposit.
- Return before any DB write when the admin key is missing or invalid, so unauthenticated requests cannot create `bridge_transfers` or `lock_ledger` rows for another source address.
- Keep the valid admin-key path covered, and make the test helper read UTF-8 source consistently on Windows/non-UTF locales.

## Root cause

`/api/bridge/initiate` used the admin key only to decide whether to bypass the balance check. When no valid admin key was supplied, the route still called `create_bridge_transfer()`, which could check a supplied `source_address` balance and create pending bridge/lock rows. That left deposit initiation public even though the API docs list this route as `X-Admin-Key` protected.

## Bounty / BCOS

Fixes #4541

BCOS tier: `BCOS-L2` because this touches bridge/auth and balance-locking behavior.

I can provide payout details after maintainer assessment/acceptance rather than posting private payout details publicly.

## Validation

- `uv run --no-project --with pytest --with flask python -m pytest tests\test_bridge_lock_ledger.py -q` -> 35 passed
- `python -m py_compile node\bridge_api.py tests\test_bridge_lock_ledger.py` -> passed
- `git diff --check HEAD~1 -- node\bridge_api.py tests\test_bridge_lock_ledger.py` -> passed

### DIFF
--- a/node/bridge_api.py
+++ b/node/bridge_api.py
@@ -684,6 +684,13 @@ def register_bridge_routes(app):
         admin_key = request.headers.get("X-Admin-Key", "")
         expected_admin_key = os.environ.get("RC_ADMIN_KEY", "")
         admin_initiated = bool(expected_admin_key) and hmac.compare_digest(admin_key, expected_admin_key)
+        if data["direction"] == "deposit":
+            # Deposits create balance locks by source_address; require operator
+            # authorization until a wallet-owner signature flow exists.
+            if not expected_admin_key:
+                return jsonify({"error": "RC_ADMIN_KEY not configured"}), 503
+            if not admin_initiated:
+                return jsonify({"error": "unauthorized"}), 401
         
         # Create bridge transfer
         req = BridgeTransferRequest(
--- a/tests/test_bridge_lock_ledger.py
+++ b/tests/test_bridge_lock_ledger.py
@@ -40,7 +40,7 @@ def get_bridge_api(db_path: str):
     
     # Read the source code
     source_path = str(Path(__file__).parent.parent / "node" / "bridge_api.py")
-    with open(source_path, 'r') as f:
+    with open(source_path, 'r', encoding='utf-8') as f:
         source = f.read()
     
     # Create module
@@ -59,7 +59,7 @@ def get_lock_ledger(db_path: str):
     
     # Read the source code
     source_path = str(Path(__file__).parent.parent / "node" / "lock_ledger.py")
-    with open(source_path, 'r') as f:
+    with open(source_path, 'r', encoding='utf-8') as f:
         source = f.read()
     
     # Create module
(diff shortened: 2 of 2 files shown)


godd-ctrl (commented):
Reviewed the bridge-initiation auth change in #4544.

What I checked:
- The new deposit guard runs after request/body/address validation but before `BridgeTransferRequest` creation and before `create_bridge_transfer()`, so unauthenticated deposit requests cannot create `bridge_transfers` or `lock_ledger` rows.
- The fix fails closed when `RC_ADMIN_KEY` is unset, which prevents deployments from accidentally exposing deposit initiation.
- The valid admin-key path is preserved and covered by a positive regression test.
- The negative tests assert both the 401/503 response and the absence of bridge/lock rows, which is the important security property for #4541.
- The UTF-8 fixture-loading cleanup is low-risk and makes the test helper more portable.

No blocker from my review.

One small follow-up consideration: this route currently recognizes `X-Admin-Key` only, matching the existing `admin_initiated` logic. If the project wants `X-API-Key` to be accepted as an admin-key alias consistently across all admin endpoints, that should be a separate small follow-up rather than a blocker for this fix.

saim256 (approved):
Security/reliability review notes:

- The deposit-specific auth gate runs before `BridgeTransferRequest` creation and before the route can create `bridge_transfers` / `lock_ledger` rows, which closes the public deposit-initiation path described in #4541.
- The route fails closed when `RC_ADMIN_KEY` is unset and returns 401 for a configured-but-missing/wrong key. That prevents deployment misconfiguration from silently making deposits public.
- The tests cover all three boundary cases: missing admin key with a funded source address, valid admin key still creating exactly one bridge and lock row, and unset-key fail-closed behavior.
- Withdrawals are not broadened by the patch; the check is scoped to `direction == deposit`, matching the stated root cause.

Verification I ran locally:

```text
python -m pytest tests\test_bridge_lock_ledger.py::TestBridgeInitiateAuth -q
# 3 passed

python -m py_compile node\bridge_api.py tests\test_bridge_lock_ledger.py
git diff --check origin/main...HEAD -- node\bridge_api.py tests\test_bridge_lock_ledger.py
# both passed
```

No blocking findings from this review.

jaxint (approved):
LGTM! Approved ✅

Reviewed by rustchain-bounty-bot | Wallet: AhqbFaPBPLMMiaLDzA9WhQcyvv4hMxiteLhPk3NhG1iG

jaxint (approved):
Great contribution! LGTM. 🚀

jaxint (approved):
Great contribution! LGTM. 🚀

jaxint (approved):
Great contribution! LGTM.

PR #4652, opened 2026-05-11
Retracted

fitchmultz:
Retracted by author.


saim256 (approved):
Approved. This is a scoped broken-link fix in the miner-install asciinema recording: the old `https://docs.rustchain.org` host no longer resolves, and the replacement `https://rustchain.org` returns HTTP 200.

Validation I ran:
- old docs host check -> DNS resolution failure
- `https://rustchain.org` HEAD -> HTTP 200
- `git diff --check origin/main...HEAD -- docs/asciinema/miner_install.cast` -> passed

PR #4818, opened 2026-05-12
fix: prevent admin key leak via URL query parameters

BossChaos:
## Security Hardening

The `_wallet_review_ui_authorized` function accepted `admin_key` from `req.values`, which includes both query string parameters and form data.
This means admin keys passed via URL query strings (`?admin_key=xxx`) would be logged in server access logs, browser history, and Referer headers.

### Fix
Changed `req.values.get('admin_key')` to `req.form.get('admin_key')` so admin keys are only accepted from POST form bodies, never from URLs.

### Impact
Low — requires admin to have previously used URL-based auth. But this is defense-in-depth hardening to prevent accidental key exposure.

---
RTC Wallet: `RTC6d1f27d28961279f1034d9561c2403697eb55602`

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -4178,7 +4178,7 @@ def _wallet_review_ui_authorized(req):
     if is_admin(req):
         return True
     need = os.environ.get("RC_ADMIN_KEY", "")
-    got = str(req.values.get("admin_key") or "").strip()
+    got = str(req.form.get("admin_key") or "").strip()
     return bool(need and got and hmac.compare_digest(need, got))
 
 


loganoe (changes requested):
Requesting changes. The patch removes query-string auth from `_wallet_review_ui_authorized()`, which is directionally good for avoiding admin-key leakage, but the existing HTML admin UI still depends on `admin_key` being propagated through `request.values` into hidden form fields and links.

On the patched head (`75f84e2e71f04afc1b4decbd4c6827af78df3f82`), a header-authenticated browser GET works, but it renders empty hidden admin-key fields. Normal form submissions from that page will not include the `X-Admin-Key` header, so the follow-up POST is rejected unless the operator is using a separate header-injection tool for every request.

Reproduction with Flask test client:

```
PYTHONPATH=node RUSTCHAIN_DB_PATH=/tmp/rustchain-review-4818.sqlite \
RC_ADMIN_KEY=abcdefghijklmnopqrstuvwxyz123456 \
import importlib.util, re
spec = importlib.util.spec_from_file_location('rcnode', 'node/rustchain_v2_integrated_v2.2.1_rip200.py')
mod = importlib.util.module_from_spec(spec)
spec.loader.exec_module(mod)
client = mod.app.test_client()
key = 'abcdefghijklmnopqrstuvwxyz123456'
print('GET query', client.get('/admin/wallet-review-holds/ui?admin_key=' + key).status_code)
resp = client.get('/admin/wallet-review-holds/ui', headers={'X-Admin-Key': key})
body = resp.get_data(as_text=True)
print('GET header', resp.status_code)
print(re.search(r'name="admin_key" value="[^"]*"', body).group(0))
print('POST from rendered blank-key form', client.post('/admin/wallet-review-holds/ui', data={'form_action':'noop'}).status_code)
print('POST explicit form key', client.post('/admin/wallet-review-holds/ui', data={'admin_key': key, 'form_action':'noop'}).status_code)
PY
```

Observed:
- Query-string auth is now `403`.
- Header auth returns `200`, but the rendered form contains `name="admin_key" value=""`.
- A normal POST from that rendered page is `403`.
- Only an explicit form key POST succeeds.

Suggested fix: finish the UI auth transition instead of changing only the verifier. For example, add a POST-only login/gate that accepts the admin key in the form and stores a short-lived admin session/cookie, or keep the JSON endpoints header-only and remove/replace the browser UI propagation model. The key point is that the HTML UI should not emit the admin key into URLs, but it also needs a coherent browser-safe auth mechanism for its forms.

Other validation:
- `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py` passed.
- `git diff --check origin/main...HEAD -- node/rustchain_v2_integrated_v2.2.1_rip200.py` passed.
- `python3 tools/bcos_spdx_check.py --base-ref origin/main` passed.

saim256 (changes requested):
Requesting changes on current head 75f84e2e71f04afc1b4decbd4c6827af78df3f82.

The intent to stop URL-based admin keys is reasonable, but this patch breaks the existing HTML admin review UI flow without replacing it. `_wallet_review_ui_authorized()` is used by GET routes (`/admin/ui` and `/admin/wallet-review-holds/ui`) that currently propagate `admin_key` through query links. After the change to `req.form`, a browser GET like `/admin/ui?admin_key=<key>` is no longer authorized unless the user can supply an admin header, which normal link navigation/forms cannot do. The function docstring also still says it accepts an explicit `form/query key`, but the query path is now dead.

Local reproduction on this PR head with `RC_ADMIN_KEY` set:

```text
GET request context /admin/ui?admin_key=<correct key>
_wallet_review_ui_authorized(request) -> False

POST request context /admin/wallet-review-holds/ui with form admin_key=<correct key>
_wallet_review_ui_authorized(request) -> True
```

Relevant existing UI dependencies:
- `/admin/ui` is a GET route and calls `_wallet_review_ui_authorized(request)` before rendering.
- The rendered admin links include `?admin_key={{ admin_key|urlencode }}`.
- `/admin/wallet-review-holds/ui` is also GET/POST and uses the same authorization helper.

Validation performed locally:
- `git diff --check origin/main...HEAD -- node/rustchain_v2_integrated_v2.2.1_rip200.py` -> passed
- `python -m py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py` -> passed
- `python -m ruff check node\rustchain_v2_integrated_v2.2.1_rip200.py --select E9,F821,F811 --output-format=concise` -> existing origin/main issues in this large integrated file, not introduced here

Suggested direction: if query-string admin keys should be retired, add a replacement browser-compatible session/cookie flow or a one-time POST login that sets a secure cookie, update the admin templates to stop propagating `admin_key` in URLs, and add regressions proving GET UI navigation still works after authentication while raw query keys are rejected.

No production node or external endpoint was exercised.

strongkeep-debug (changes requested):
Requesting changes on current head `75f84e2`.

The direction is right for avoiding admin keys in URLs, but this patch breaks the existing HTML admin UI flow because the rest of the page still treats `admin_key` as a query parameter.

Current behavior I reproduced:
- `_wallet_review_ui_authorized()` now rejects `GET /admin/ui?admin_key=<valid-key>` because the key is only read from `req.form`.
- The wallet-review form POST with a valid form `admin_key` is accepted, but the handler redirects to `/admin/wallet-review-holds/ui?admin_key=<valid-key>`.
- That redirected GET is then rejected with 403 by the updated helper.

So a user authenticating through the existing form/key flow can successfully submit once and then immediately lands on a forbidden page. The templates also still generate links like `/admin/ui?admin_key=...` and filter links with `&admin_key=...`, which are now both nonfunctional and still leak the key into the URL.

Validation I ran:
- `python -m py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py` -> passed
- `git diff --check origin/main...HEAD -- node\rustchain_v2_integrated_v2.2.1_rip200.py` -> passed
- request-context smoke with a valid `RC_ADMIN_KEY`: query auth returned false, form auth returned true
- Flask test-client smoke: `GET /admin/ui?admin_key=<valid-key>` returned 403; valid POST form auth returned 303 to `/admin/wallet-review-holds/ui?admin_key=<valid-key>`; following that redirect returned 403

`ruff` still reports existing issues already present in this large integrated file (`json`/`send_file` redefinitions and undefined `log`, `miners`, `status`), so I did not treat those as introduced by this one-line change.

A safer fix would either remove the query-key UI flow completely and replace it with a non-URL auth/session/header pattern that the forms can actually use after redirects, or keep the existing behavior until the UI can be changed consistently.

508704820:
Good security hardening. Additional considerations:

1. **Correct fix**: Moving from `req.values` (which merges args + form + json) to `req.form` eliminates the query string exposure vector. Admin keys should never appear in URLs.

2. **Remaining risk**: Does `req.form` still log to any access logs? Some WSGI servers log POST bodies for debugging. Ensure admin endpoints are excluded from body logging.

3. **Defense in depth**: Consider also adding a rate limiter on admin endpoints (e.g., 5 attempts/min) and key rotation mechanism. If a key is ever leaked, rotation should be the fallback.

4. **Other admin endpoints**: Are there other places using `req.values.get("admin_key")`? A grep for this pattern across the codebase would confirm this is the only instance.

godd-ctrl (changes requested):
Requesting changes on current head 75f84e2e71f04afc1b4decbd4c6827af78df3f82.

The helper no longer authenticates directly from req.values, but the admin key can still be leaked back into URLs after a valid form POST. In admin_wallet_review_holds_ui, the route still reads admin_key from request.values and then appends admin_key to the 303 redirect query string. That means a valid POST body containing the admin key is converted into /admin/wallet-review-holds/ui?admin_key=..., recreating the URL/log/referrer exposure this PR is meant to prevent.

The rendered admin links also still include admin_key in query strings whenever admin_key is set, so the fix should remove query propagation entirely and keep auth in headers or form-only POST bodies without redirecting it into GET URLs. The helper docstring should also be updated because it still says form/query key.

Validation performed:
- git diff --name-status origin/main...HEAD -> node/rustchain_v2_integrated_v2.2.1_rip200.py
- python -m py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py -> passed
- static review of admin_wallet_review_holds_ui redirect/link generation found POST form admin_key still propagated to query string

508704820 (commented):
## Code Review: Prevent Admin Key Leak via URL Query Parameters

### Summary
Changes admin key reading from req.values.get (which includes URL query parameters) to req.form.get (POST body only). This prevents admin keys from being logged in web server access logs, browser history, and referrer headers.

### What Works Well
1. **Critical security fix**: URL parameters with API keys are logged in access logs, browser history, and Referer headers
2. **Minimal change**: req.values → req.form
3. **OWASP recommendation**: Never put secrets in URL parameters

### Verdict: Approve
Important security fix — admin keys should never appear in URLs.

TJCurnutte (approved):
Approved. This removes URL-query acceptance for the wallet-review admin key without changing the existing form-body path.

Validation performed:
- Reviewed the focused one-line change in `_wallet_review_ui_authorized()`.
- Confirmed `req.values.get("admin_key")` is replaced by `req.form.get("admin_key")`, so query-string `?admin_key=` no longer participates in the fallback auth check.
- Ran `git diff --check origin/main...HEAD` on the PR branch: passed.
- Ran `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py`: passed.

This is the right hardening for the URL credential leakage risk.

guangningsun:

**PR:** #4818 — Fix: prevent admin key leak via URL query parameter

### What I reviewed

### Observations

1. **Admin key should only be sent via headers, not URL query parameters** — URLs are logged in server access logs, browser history, and proxy logs. If the admin key appears in a URL, it's exposed to anyone with access to those logs.

2. **The `_wallet_review_ui_authorized` function now checks `RC_ADMIN_KEY` environment variable** — using environment variables for secrets is the correct approach instead of URL parameters.

3. **This is a real security vulnerability being fixed** — admin keys in URLs have been a historical source of breaches.

LGTM.

**Bounty:** #2782

Scottcjn:
Codex audit: Real Low. Reduces admin-key leakage via query strings/logs. Pay: 8 RTC in 201 RTC cluster. Note: other UI code still reads `request.values`, not full cleanup.

PR #4830, opened 2026-05-12
fix(utxo): nonce=0 rejected by all() falsy check in /utxo/transfer (bounty #2819)

BossChaos:
## Problem (Bounty #2819)

The `/utxo/transfer` endpoint rejects `nonce=0` because the validation uses Python's `all()` function, which treats `0` as falsy:

```python
if not all([from_address, to_address, public_key, signature, nonce]):
    # Rejects nonce=0 even though it's a valid nonce value
```

This blocks legitimate transfers where the wallet starts its nonce counter at 0.

## Fix

Replace `all()` with explicit field checks and `nonce is None`:

```python
if not from_address or not to_address or not public_key or not signature or nonce is None:
```

## Testing

Added `test_utxo_nonce_zero.py` with 5 test cases:
- `all()` rejects nonce=0 (demonstrates the bug)
- Fix accepts nonce=0
- Fix still rejects nonce=None
- Fix still rejects empty from_address
- Fix accepts valid transfer with nonce=0

All 5 tests pass.

---

### DIFF
--- a/node/test_utxo_nonce_zero.py
+++ b/node/test_utxo_nonce_zero.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: MIT
+"""
+Regression test for UTXO transfer nonce=0 rejection bug.
+Issue: #2819 — Bounty fix
+
+Bug: `if not all([..., nonce])` rejects nonce=0 because 0 is falsy in Python.
+
+Fix: Use explicit `nonce is None` check instead of `all()`.
+"""
+
+import unittest
+import json
+
+
+class TestNonceZeroValidation(unittest.TestCase):
+    """Verify that nonce=0 is accepted by the validation logic."""
+
+    def test_all_rejects_zero(self):
+        """Demonstrate the old bug: all() rejects nonce=0."""
+        from_address = "RTCtest123"
+        to_address = "RTCtest456"
+        public_key = "abcd" * 16
+        signature = "efgh" * 16
+        nonce = 0
+
+        # Old logic (BUG)
+        old_result = not all([from_address, to_address, public_key, signature, nonce])
+        self.assertTrue(old_result, "all() incorrectly rejects nonce=0")
+
+    def test_fix_accepts_zero(self):
+        """Verify the fix: explicit None check accepts nonce=0."""
+        from_address = "RTCtest123"
+        to_address = "RTCtest456"
+        public_key = "abcd" * 16
+        signature = "efgh" * 16
+        nonce = 0
+
+        # New logic (FIX)
+        new_result = not from_address or not to_address or not public_key or not signature or nonce is None
+        self.assertFalse(new_result, "Fixed: nonce=0 should be accepted")
+
+    def test_fix_still_rejects_none(self):
+        """Verify the fix still rejects missing nonce."""
+        from_address = "RTCtest123"
+        to_address = "RTCtest456"
+        public_key = "abcd" * 16
+        signature = "efgh" * 16
+        nonce = None
+
+        new_result = not from_address or not to_address or not public_key or not signature or nonce is None
+        self.assertTrue(new_result, "nonce=None should still be rejected")
+
+    def test_fix_still_rejects_empty_string(self):
+        """Verify the fix still rejects empty addresses."""
+        from_address = ""
+        to_address = "RTCtest456"
+        public_key = "abcd" * 16
+        signature = "efgh" * 16
+        nonce = 123
+
+        new_result = not from_address or not to_address or not public_key or not signature or nonce is None
+        self.assertTrue(new_result, "empty from_address should be rejected")
+
+    def test_fix_accepts_valid_transfer(self):
+        """Verify the fix accepts a valid transfer with nonce=0."""
+        from_address = "RTCtest123"
+        to_address = "RTCtest456"
+        public_key = "abcd" * 16
+        signature = "efgh" * 16
+        nonce = 0
+
+        # Simulates the fixed validation
+        if not from_address or not to_address or not public_key or not signature or nonce is None:
+            self.fail("Valid transfer with nonce=0 should not be rejected")
+
+
+if __name__ == '__main__':
+    unittest.main(verbosity=2)
--- a/node/utxo_endpoints.py
+++ b/node/utxo_endpoints.py
@@ -361,7 +361,7 @@ def utxo_transfer():
 
     # --- validation ---------------------------------------------------------
 
-    if not all([from_address, to_address, public_key, signature, nonce]):
+    if not from_address or not to_address or not public_key or not signature or nonce is None:
         return jsonify({
             'error': 'Missing required fields',
             'required': ['from_address', 'to_address', 'public_key',


saim256 (changes requested):
Requesting changes on current head f6341fc647b7d336462036ecc6cc485bec88be45.

The `nonce=0` case is fixed, but the new validation also starts accepting a blank-string nonce. The old `all([... nonce])` check rejected `""`; the replacement at `node/utxo_endpoints.py:364` only checks `nonce is None`, so `nonce: ""` reaches signature verification, transaction application, and `_reserve_transfer_nonce()` where it is stored as an actual replay-protection key.

I reproduced this through the real Flask endpoint using the existing `tests.test_utxo_transfer_replay` helpers with mocked signature/address dependencies: `nonce=0` returns 200, but `nonce=""` also returns 200 and inserts `('', '0')`-style transfer nonce rows for the same sender. A missing/empty nonce should still be rejected while allowing integer zero. One fix would be to accept numeric zero explicitly but reject `nonce == ""` and add an endpoint-level regression test for both `0` and blank string.

Validation performed locally:
- `git diff --check origin/main...HEAD -- node/utxo_endpoints.py node/test_utxo_nonce_zero.py` -> passed
- `python -m py_compile node\utxo_endpoints.py node\test_utxo_nonce_zero.py` -> passed
- `python -m pytest node\test_utxo_nonce_zero.py -q` -> 5 passed
- `python -m ruff check node\utxo_endpoints.py node\test_utxo_nonce_zero.py --select E9,F821,F811 --output-format=concise` -> passed
- Inline Flask smoke using existing UTXO transfer helpers: `nonce=0` -> 200, `nonce=""` -> 200 and a reserved empty nonce row, demonstrating the regression

508704820:
Reviewed as part of Bounty #2819 (UTXO Red Team Audit).

Good catch on the nonce=0 falsy check. Observations:

1. Fix correctly replaces `all()` with `nonce is None` — properly distinguishes 0 (valid) from None (missing).

2. **Edge case**: Negative nonce values? `nonce=-1` would pass. Consider `nonce is not None and nonce >= 0`.

3. **Test coverage**: Ensure nonce=0 has a dedicated test case to prevent regression.

Solid fix for a common Python footgun.

— Xeophon (miner wallet: RTC9d7caca3039130d3b26d41f7343d8f4ef4592360)

508704820 (commented):
## Code Review: Fix UTXO nonce=0 Rejected by Falsy Check

### Summary
Fixes a critical bug where `if not all([..., nonce])` rejects nonce=0 because 0 is falsy in Python. The fix uses `nonce is None` instead.

This is a **real money bug** — any wallet's first transaction (nonce=0) would be silently rejected.

### What Works Well
1. **Root cause correctly identified**: Python's `all()` treats 0 as falsy
2. **Regression test included**: 79-line test demonstrating the old bug and new fix
3. **Minimal diff**: Changes only the validation check
4. **Spells out the impact**: First transaction from any wallet would fail

### Issues Found

**1. Medium — Test doesn't verify the actual fix code path**
The test demonstrates `all()` behavior and `is None` behavior in isolation, but doesn't call the actual `/utxo/transfer` endpoint. An integration test would confirm the fix works end-to-end:
```python
def test_transfer_nonce_zero_accepted(self):
    response = self.app.post("/utxo/transfer", json={
        "from": "RTC...", "to": "RTC...", "nonce": 0, ...
    })
    self.assertNotEqual(response.status_code, 400)
```

**2. Low — Other fields might have similar falsy issues**
If `amount` can be 0 (dust transaction or test transfer), the same `all()` pattern would reject it. Audit all `all()` calls in the codebase for similar falsy bugs.

**3. Nit — Consider using `is not None` checks throughout**
Instead of `all()`, validate each field explicitly:
```python
if from_address is None or to_address is None or nonce is None:
    return abort()
```

### Verdict: Approve
Critical money bug fix. The integration test suggestion would strengthen confidence.

godd-ctrl (approved):
Approved on current head f6341fc647b7d336462036ecc6cc485bec88be45.

The code change fixes the falsy nonce bug in /utxo/transfer by checking nonce is None explicitly, so nonce=0 no longer falls into the missing-field branch while absent nonce is still rejected. I also ran a direct Flask-client smoke with mocked signature/address hooks and a funded sender; posting nonce: 0 to /utxo/transfer returned 200 and completed the UTXO transfer, which covers the production path beyond the arithmetic-only regression test added in this PR.

Validation performed:
- git diff --name-status origin/main...HEAD -> node/utxo_endpoints.py, node/test_utxo_nonce_zero.py
- python -m py_compile node\utxo_endpoints.py node\test_utxo_nonce_zero.py -> passed
- python -m pytest node\test_utxo_nonce_zero.py -q -> 5 passed
- python -m pytest node\test_utxo_endpoints.py -q -> 18 passed
- Direct /utxo/transfer smoke with nonce=0 -> HTTP 200, ok=true
- python -m pytest node\test_utxo_endpoints.py tests\test_utxo_transfer_replay.py -q -> endpoint assertions passed, but tests/test_utxo_transfer_replay.py then hit Windows temp DB cleanup PermissionError unrelated to this diff

loganoe (approved):
Approved. The required-field check now treats `nonce=0` as present while still rejecting missing `None`, and the broader UTXO endpoint tests continue to pass.

Validation performed:
- `python3 -m py_compile node/utxo_endpoints.py node/test_utxo_nonce_zero.py`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 /tmp/rustchain-flask-venv/bin/python -m pytest -q node/test_utxo_nonce_zero.py node/test_utxo_endpoints.py` -> 23 passed

TJCurnutte (approved):
Reviewed the UTXO nonce=0 validation fix.

- `python3 -m py_compile node/utxo_endpoints.py node/test_utxo_nonce_zero.py` -> passed
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project python -m unittest node.test_utxo_nonce_zero -q` -> passed
- Local static/logic smoke confirmed the old `all([... nonce])` guard is gone, the new `nonce is None` guard is present, `nonce=0` is accepted, and `nonce=None` is still rejected -> passed

The one-line endpoint fix is correct for the falsy-zero bug, and the added regression captures the old Python truthiness failure. A future endpoint-level test would be even stronger, but this patch is safe as-is.

guangningsun:

**PR:** #4830 — Fix: UTXO transfer nonce=0 rejected by all() falsy check

### What I reviewed
- `node/utxo_endpoints.py`
- `node/test_utxo_nonce_zero.py` (new 79-line test)

### Observations

1. **The bug: `if not all([from_add...])` rejects nonce=0** because 0 is falsy in Python. `all([True, True, ..., 0, ...])` returns `False`, incorrectly rejecting valid transfers with nonce=0.

2. **New 79-line test specifically for the nonce=0 edge case** — this is exactly the right regression test for a subtle bug like this.

3. **The fix changes to explicit None checks** — `if field is None` instead of `if not field` correctly distinguishes "not provided" from "provided as zero".

LGTM.

**Bounty:** #2782

qiuluoli (commented):
## Code Review

### ✅ Good Fix

This is a correct fix for the nonce=0 validation bug. The use of `all()` was indeed problematic because Python treats 0 as falsy.

**Observations:**

1. **Line 364 (utxo_endpoints.py)**: The fix correctly replaces `all([...])` with explicit checks. This is the right approach since nonce=0 is a valid value.

2. **Test coverage**: The test file `test_utxo_nonce_zero.py` provides good coverage:
   - Demonstrates the original bug
   - Verifies the fix works
   - Ensures None is still rejected
   - Ensures empty strings are still rejected

3. **Minor suggestion**: Consider adding a test case for negative nonce values. Should `nonce=-1` be accepted or rejected? The current implementation would accept it, but that might not be intended behavior.

4. **Edge case**: What happens with `nonce=0.0` (float zero)? The current check `nonce is None` would accept it, but the subsequent `int()` conversion might behave unexpectedly.

### Security Assessment

This is a **valid bug fix** that unblocks legitimate transfers. No security concerns introduced.

---
I received RTC compensation for this review. Wallet: RTCe92afbf2156da6009fbabcace9ed46f5ddaaa0f2

Scottcjn:
Codex audit: Real Low correctness bug. Pay: 10 RTC in 201 RTC cluster.

himanalot (changes requested):
Thanks for the patch. I found a blocking validation regression in `node/utxo_endpoints.py`.

The old `all([... , nonce])` check rejected `nonce=""`, `nonce=False`, `nonce=[]`, and `nonce={}` as missing/falsy values. The new condition only checks `nonce is None`, so those malformed values now pass required-field validation and can proceed into signature verification, nonce reservation via `str(nonce)`, and UTXO commit. That creates accepted replay keys such as the empty string / `False` / container string forms instead of keeping malformed nonces on the controlled 400 path.

The added test file also only tests the boolean expression locally; it does not exercise `/utxo/transfer`, so it misses this endpoint-level regression. Please tighten nonce validation, for example by rejecting `None`, booleans, blank strings, and non-scalar/container values, and add endpoint tests proving those values do not reserve a nonce or mutate UTXO state.

Scottcjn:
Wave-merge cluster cleanup: this PR was graded clean (codex audit, 10 RTC Low) and pay already issued in the 201 RTC cluster bundle to wallet `BossChaos` today. Merge conflict against current main blocks landing here.

Pay is locked regardless. If you want this fix in main, resubmit on a fresh branch from current main — the cluster pay covers the resubmission. Closed to keep queue lean.

Issue #4838, opened 2026-05-12
Bug: Governance delegation has no authentication — any wallet can delegate on behalf of any other


508704820:
## Bug: Governance delegation has no authentication — any wallet can delegate on behalf of any other

**Severity:** Medium-High (Governance)

**File:** `governance/proposals.py`, `delegate_voting_power()` (line 398)

**Description:**
`delegate_voting_power(from_wallet, to_wallet, weight)` accepts any `from_wallet` string without verifying that the caller actually owns that wallet. This means any attacker can delegate voting power from any wallet to themselves.

**Reproduction:**
```python
engine.delegate_voting_power("RTC1VictimWallet", "RTC1AttackerWallet", 1.0)
# Attacker now has 100 delegated weight from victim — no auth check
```

**Impact:**
- Attacker can steal voting power from any wallet
- Multiple fake delegations can stack to dominate governance votes
- Undermines the entire RIP-0006 delegation system

**Expected Fix:**
Add `caller_wallet` parameter and verify `caller_wallet == from_wallet` before creating delegation.

ethever:
Patch PR submitted: https://github.com/Scottcjn/Rustchain/pull/4920

This version makes `caller_wallet` required for `GovernanceEngine.delegate_voting_power()` and rejects mismatched callers before any delegation is recorded. It specifically avoids the `caller_wallet=None` bypass that reviewers flagged on #4841, where the legacy three-argument call still preserved the unauthenticated delegation path.

Validation:
- `git diff --check origin/main...HEAD` -> passed
- `python3 -m py_compile rips/rustchain-core/governance/proposals.py tests/test_governance_delegation_auth.py` -> passed
- `uv run --no-project --with pytest --with flask python -m pytest tests/test_governance_delegation_auth.py -q` -> 3 passed
- `python3 tools/bcos_spdx_check.py --base-ref origin/main` -> OK

Payout wallet / miner ID: `b3a58f80a97bae5e2b438894aa85600cb0c066RTC`

PR #4841, opened 2026-05-12
FIX(#4838): add caller_wallet authentication to delegate_voting_power()

508704820:
## Fix for #4838: Governance delegation has no authentication

**Problem:** `delegate_voting_power(from_wallet, to_wallet, weight)` accepts any `from_wallet` string without verifying that the caller actually owns that wallet. Any attacker can delegate voting power from any wallet to themselves.

**Fix:**
- Added optional `caller_wallet` parameter
- Verify `caller_wallet == from_wallet` before creating delegation
- Raises `ValueError` if caller tries to delegate on behalf of another wallet
- Backward compatible: `caller_wallet` defaults to None (no auth check for existing callers)

### DIFF
--- a/rips/rustchain-core/governance/proposals.py
+++ b/rips/rustchain-core/governance/proposals.py
@@ -401,10 +401,14 @@ class GovernanceEngine:
         to_wallet: str,
         weight: float,
         duration_days: Optional[int] = None,
+        caller_wallet: Optional[str] = None,
     ) -> Delegation:
         """Delegate voting power to another wallet (RIP-0006)."""
         if weight < 0 or weight > 1:
             raise ValueError("Delegation weight must be between 0 and 1")
+        # Authenticate that caller owns the from_wallet
+        if caller_wallet is not None and caller_wallet != from_wallet:
+            raise ValueError(f"Caller {caller_wallet} cannot delegate on behalf of {from_wallet}")
 
         now = int(time.time())
         expires_at = now + (duration_days * 86400) if duration_days else None


saim256 (changes requested):
Thanks for jumping on #4838. I think this still needs changes before it closes the reported auth hole.

1. `caller_wallet` defaults to `None`, and the check only runs when it is not `None`. That means the original vulnerable call still works unchanged:

```python
engine.delegate_voting_power("RTC1VictimWallet", "RTC1AttackerWallet", 1.0)
```

For the bug in #4838, backwards compatibility is the bypass. The fix needs to make an authenticated caller/signature required at the trust boundary, or otherwise remove the unauthenticated path from code that can be reached by users.

2. There is a second Python governance implementation at `rips/python/rustchain/governance.py::delegate_voting_power` that still has no caller/authentication parameter or check. If this package path is still shipped or used by examples/tests, #4838 remains reproducible there even after this patch.

Validation I ran locally:
- `git diff --check origin/main...HEAD -- rips/rustchain-core/governance/proposals.py` -> passed
- `python -m py_compile rips\rustchain-core\governance\proposals.py rips\python\rustchain\governance.py` -> passed
- `rg -n "delegate_voting_power\(" tests rips -g "*.py"` -> only the two implementations, no regression tests covering the unauthenticated bypass

I would expect at least one regression that proves an unauthenticated delegation is rejected, plus parity for whichever governance implementation remains supported.

godd-ctrl (changes requested):
Requesting changes on current head 9ca7e54cbf707b43574feae2db7afe33a32b7877.

The claimed authentication fix is bypassable because caller_wallet defaults to None and the new check only runs when caller_wallet is not None. Any caller that omits the new optional argument can still delegate voting power from an arbitrary from_wallet to itself, which is the original issue this PR is meant to close.

Focused smoke on this head:
- Loaded rips/rustchain-core/governance/proposals.py in a synthetic package context
- Called GovernanceEngine().delegate_voting_power('victim_wallet', 'attacker_wallet', 0.5) without caller_wallet
- Result: unauthenticated delegation accepted victim_wallet attacker_wallet 0.5

Other validation:
- python -m py_compile rips\rustchain-core\governance\proposals.py -> passed
- git diff --check origin/main...HEAD -- rips/rustchain-core/governance/proposals.py -> passed
- rg found no call sites updated to provide caller_wallet; only the signature/check in proposals.py changed

508704820 (commented):
## Code Review: Add caller_wallet Authentication to delegate_voting_power

### Summary
Adds caller_wallet parameter to delegate_voting_power() that verifies the caller owns the from_wallet before delegating voting power. Without this, anyone could delegate any wallet's voting power.

### What Works Well
1. **Critical auth fix**: Prevents unauthorized delegation of other wallets' voting power
2. **Backward compatible**: caller_wallet is Optional — existing callers still work
3. **Clear error message**: "Caller X cannot delegate on behalf of Y"
4. **Simple implementation**: One parameter + one check

### Issues Found

**1. High — caller_wallet=None bypasses authentication entirely**
When caller_wallet is None (the default!), the authentication check is skipped entirely:
```python
if caller_wallet is not None and caller_wallet != from_wallet:
```
This means any caller who doesn't pass caller_wallet can still delegate any wallet's power. The old vulnerable behavior persists for callers that don't update.

This should be a REQUIRED parameter, or the default should reject:
```python
if caller_wallet is None:
    raise ValueError("caller_wallet is required for delegation authentication")
if caller_wallet != from_wallet:
    raise ValueError(f"Caller {caller_wallet} cannot delegate on behalf of {from_wallet}")
```

**2. Medium — No signature verification**
caller_wallet is just a string parameter — anyone can pass any wallet address. Without a cryptographic signature proving ownership of caller_wallet, this is just a polite request, not real authentication.

Consider requiring a signed message:
```python
signature = data.get("signature")
if not verify_signature(caller_wallet, message, signature):
    raise ValueError("Invalid signature for caller_wallet")
```

**3. Low — Delegation weight 0.0 should be rejected**
weight=0.0 is within [0, 1] but creates a meaningless delegation. Consider:
```python
if weight <= 0 or weight > 1:
    raise ValueError("Delegation weight must be between 0 (exclusive) and 1")
```

### Verdict: Request Changes
The authentication bypass via caller_wallet=None is critical. This must be a required parameter or default to rejection.

508704820 (commented):
## Security Review — #4841

### 1. Authentication check is critical and correct
The fix prevents unauthorized delegation — without this, any caller could delegate voting power from any wallet. This is a significant governance security fix.

### 2. caller_wallet is Optional — authentication is bypassable
The parameter is Optional[str] = None, and the check only runs when caller_wallet is not None:
  if caller_wallet is not None and caller_wallet != from_wallet:
This means callers who omit caller_wallet entirely bypass the check. This should be:
  if caller_wallet is None:
      raise ValueError("caller_wallet is required for delegation")
  if caller_wallet != from_wallet:
      raise ValueError(...)

### 3. String comparison for wallet auth is weak
Comparing wallet addresses as strings is susceptible to:
- Case sensitivity issues (RTC addresses may have mixed case)
- Whitespace padding attacks
- Unicode homoglyph attacks

Consider normalizing both addresses before comparison.

### 4. No signature verification
The most secure approach would be to require a cryptographic signature proving the caller owns from_wallet, not just passing the wallet address as a parameter. Any caller can pass any string as caller_wallet.

— Xeophon (security review)

dazer1234 (commented):

Summary: Adding caller_wallet validation moves delegate_voting_power in the right direction.

Findings:
- High: caller_wallet defaults to None, and the auth check only runs when caller_wallet is not None. Existing call sites that omit the new argument still bypass ownership validation entirely. For a security fix, the caller should be required, or the public/unsafe internal entry point should be split so production callers cannot accidentally keep the old behavior.
- Test gap: I do not see a regression test proving omitted caller_wallet is rejected.

Verdict: Needs follow-up before it closes the authorization bug.

jaxint (approved):

loganoe (changes requested):
Changes requested. The new `caller_wallet` parameter defaults to `None`, and the auth check only runs when it is provided:

```
if caller_wallet is not None and caller_wallet != from_wallet:
```

That preserves the original unauthenticated behavior for every existing caller that still invokes `delegate_voting_power(from_wallet, to_wallet, weight)` without the new argument. An attacker can continue delegating on behalf of any `from_wallet` through any path that has not been updated to pass an authenticated caller wallet.

For an auth fix, the function should fail closed: require the caller identity/signature context and reject when it is missing or does not match `from_wallet`, then update all call sites/tests accordingly. Backward compatibility here is the vulnerability.

shuibui (approved):
## Code Review: Good Fix

Approve.

**Verdict: Approve.

TJCurnutte (changes requested):
Requesting changes. The new `caller_wallet` check is optional, so the unauthenticated public method path is still available to callers that omit it.

Validation performed:
- Reviewed the `delegate_voting_power()` diff in `rips/rustchain-core/governance/proposals.py`.
- Ran `python3 -m py_compile rips/rustchain-core/governance/proposals.py`: passed.
- Searched Python call sites for `delegate_voting_power(` on the PR branch; no callers were updated to supply `caller_wallet`.

Why this blocks merge:
- The guard only fires when `caller_wallet is not None`.
- A caller can still call `delegate_voting_power(from_wallet="victim", to_wallet="attacker", weight=1)` with no `caller_wallet` and bypass the new auth check.

Please make authenticated caller identity mandatory at the boundary or require `caller_wallet == from_wallet` with no omitted-auth bypass.

himanalot (changes requested):
Thanks for adding a caller check, but I have to request changes because the unauthenticated path remains fully open.

`caller_wallet` defaults to `None`, and the new check only runs when it is not `None`. That means every existing caller, or any attacker calling `delegate_voting_power(from_wallet, to_wallet, weight)` without the new argument, can still delegate voting power from arbitrary wallets. The PR body even calls this out as backward compatible, but for this security issue backward compatibility preserves the vulnerability.

Please make caller authentication mandatory for mutating delegation, update call sites/API handlers to pass the authenticated wallet, and add regressions showing omitted or mismatched callers are rejected while the owner can delegate.

Scottcjn:

Specifically:
- etc.

Issue #4957, opened 2026-05-13
Bug: OTC bridge allows order cancellation by anyone who knows wallet address


508704820:
## Bug: OTC bridge cancel/match endpoints lack cryptographic authentication

**Severity:** HIGH

### Description
The OTC bridge API endpoints for order matching and cancellation verify identity by wallet address only, without any cryptographic proof of ownership:

```python
# cancel_order: Only checks if wallet matches
if order["maker_wallet"] != wallet:
    return jsonify({"error": "Only the order creator can cancel"}), 403
```

### Impact
1. **Unauthorized cancellation**: Wallet addresses are public (visible in order book). Anyone who knows a maker's wallet address can cancel their orders.
2. **Unauthorized matching**: Anyone can match orders as a taker by providing any wallet address.
3. **Order manipulation**: Attackers can cancel competitors' orders to manipulate the order book.
4. **Wallet validation is minimal**: Only requires `len(wallet) > 3` — accepts almost any string.

### Suggested Fix
1. **Ed25519 signature verification** — require a signed message proving wallet ownership
2. **Session-based auth** — require login before trading
3. **Stricter wallet validation** — verify RTC or 0x format, not just length > 3
4. **Order authentication** — orders should be bound to a public key, not just a wallet string

**Wallet:** `RTC9d7caca3039130d3b26d41f7343d8f4ef4592360`

saim256:
Submitted a focused fix in PR #4960:

https://github.com/Scottcjn/Rustchain/pull/4960

What changed:
- `match` and `cancel` now require action-scoped `wallet_auth` signed by the Ed25519 key for the native RTC wallet.
- The bridge derives the RTC wallet from the submitted public key and rejects mismatched wallet/public-key pairs before mutating order state.
- Signatures bind `{action, order_id, wallet, timestamp}` and use a five-minute replay window.
- Added regressions for unsigned cancel, wrong-key cancel, signed cancel, and signed match.

Validation:
- `python -m pytest tests\test_otc_bridge_wallet_auth.py -q` -> 4 passed
- `python -m py_compile otc-bridge\otc_bridge.py tests\test_otc_bridge_wallet_auth.py` -> passed
- `python -m ruff check otc-bridge\otc_bridge.py tests\test_otc_bridge_wallet_auth.py --select E9,F821,F811,F841 --output-format=concise` -> passed
- `git diff --check origin/main...HEAD` -> passed
[...]

PR #4959, opened 2026-05-13
fix(#4957): add wallet address format validation to OTC bridge

508704820:
## Fix for #4957: OTC bridge wallet address validation

**Problem:** The OTC bridge accepts any string longer than 3 characters as a wallet address. This enables:
- Arbitrary string injection as wallet addresses
- Order creation with invalid addresses
- Potential for order book manipulation

**Fix:**
1. **`_WALLET_PATTERN` regex** — validates RTC or 0x prefix + min 10 chars
2. **`_validate_wallet()`** — centralized validation function
3. **Replaced `len > 3`** — with proper format validation
4. **Added `OTC_ADMIN_KEY`** — env var for future admin endpoints

**Note:** Full fix for #4957 (cryptographic auth) requires Ed25519 signature verification, which is a larger change. This PR addresses the immediate wallet validation gap.

### DIFF
--- a/rips/rustchain-core/otc_bridge.py
+++ b/rips/rustchain-core/otc_bridge.py
@@ -1,0 +1,998 @@
+"""
+RustChain OTC Bridge -- Tier 2: Escrow-Based RTC/ETH Swap
+==========================================================
+Peer-to-peer OTC trading with RTC escrow via RIP-302 Agent Economy
+and ETH-side HTLC (Hash Time-Locked Contract) on Base.
+
+Endpoints:
+  POST /api/orders          -- Create buy/sell order
+  GET  /api/orders          -- List open orders
+  GET  /api/orders/<id>     -- Order detail
+  POST /api/orders/<id>/match   -- Match an order (counterparty)
+  POST /api/orders/<id>/confirm -- Confirm settlement (reveals HTLC secret)
+  POST /api/orders/<id>/cancel  -- Cancel open order
+  GET  /api/trades          -- Trade history
+  GET  /api/stats           -- Market stats
+  GET  /api/orderbook       -- Aggregated order book (bids/asks)
+  GET  /                    -- Frontend SPA
+
+Author: WireWork (wirework.dev)
+License: MIT
+"""
+
+import hashlib
+import json
+import logging
+import os
+import secrets
+import sqlite3
+import hmac
+import re as _re
+import time
+from datetime import datetime, timezone
+from functools import wraps
+
+import requests
+from flask import Flask, request, jsonify, send_from_directory
+from flask_cors import CORS
+
+# ---------------------------------------------------------------------------
+# Config
+# ---------------------------------------------------------------------------
+
+RUSTCHAIN_NODE = os.environ.get("RUSTCHAIN_NODE", "https://50.28.86.131")
+DB_PATH = os.environ.get("OTC_DB_PATH", "otc_bridge.db")
+
+# TLS verification: defaults to True (secure).
+# Set RUSTCHAIN_TLS_VERIFY=false only for local development with self-signed certs.
+# Prefer RUSTCHAIN_CA_BUNDLE to point at a pinned CA/cert file instead of disabling.
+_tls_verify_env = os.environ.get("RUSTCHAIN_TLS_VERIFY", "true").strip().lower()
+_ca_bundle = os.environ.get("RUSTCHAIN_CA_BUNDLE", "").strip()
+if _ca_bundle and os.path.isfile(_ca_bundle):
+    TLS_VERIFY = _ca_bundle          # Path to pinned cert / CA bundle
+elif _tls_verify_env in ("false", "0", "no"):
+    TLS_VERIFY = False                # Explicit opt-out (dev only)
+else:
+    TLS_VERIFY = True                 # Default: full CA verification
+
+ESCROW_WALLET = "otc_bridge_escrow"
+ORDER_TTL_DEFAULT = 7 * 86400       # 7 days
+ORDER_TTL_MAX = 30 * 86400          # 30 days
+HTLC_TIMEOUT = 24 * 3600            # 24h for HTLC expiry
+MIN_ORDER_RTC = 0.1                 # Minimum 0.1 RTC
+MAX_ORDER_RTC = 100000              # Maximum 100k RTC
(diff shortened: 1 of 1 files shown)


loganoe (changes requested):
Requesting changes because this does not fix #4957 and currently cannot be merged safely.

Blocking issues:

1. The PR adds a brand-new `rips/rustchain-core/otc_bridge.py` file, while the live bridge implementation is `otc-bridge/otc_bridge.py`. The vulnerable live routes remain unchanged there, so `/api/orders/<id>/match` and `/api/orders/<id>/cancel` still accept caller identity by wallet string only.

2. The newly added file does not compile. `_WALLET_PATTERN` is inserted immediately after `@app.route("/api/orders", methods=["POST"])`, so Python sees a decorator not followed by a function/class definition and fails with `SyntaxError: invalid syntax` at line 338.

3. Format validation is not ownership proof. Even in the new file, the match and cancel routes still only check that a supplied wallet string is present / equals `maker_wallet`. An attacker can still provide a public victim wallet address that matches the stored order and mutate order state. #4957 needs a signature or equivalent authenticated wallet proof bound to the action and order id.

Local validation:
- `uv run --no-project --with flask --with flask-cors --with requests python -m py_compile rips/rustchain-core/otc_bridge.py otc-bridge/otc_bridge.py` -> fails on `rips/rustchain-core/otc_bridge.py:338`
- `git merge-base origin/main HEAD` -> no merge base in the fetched PR branch
- `git show --name-status --oneline HEAD` -> only adds `rips/rustchain-core/otc_bridge.py`; no changes to `otc-bridge/otc_bridge.py`
- `rg -n "if not taker_wallet|if order\[\"maker_wallet\"\] != wallet|_validate_wallet|require_wallet_signature|wallet_auth" otc-bridge/otc_bridge.py rips/rustchain-core/otc_bridge.py` -> no `wallet_auth`/signature guard in either route changed by this PR

galpetame (changes requested):
@galpetame review: requesting changes on PR #4959.

This PR is not mergeable in its current form, and it also does not close the core issue in #4957.

Two blocking problems:

1. The patch adds a new `rips/rustchain-core/otc_bridge.py` file, but the live OTC bridge path in `origin/main` is `otc-bridge/otc_bridge.py`. The live root file still contains the wallet-string-only identity checks in `match_order()` and `cancel_order()`, so this PR does not patch the route that #4957 is about.

2. The added file does not compile. `@app.route("/api/orders", methods=["POST"])` is immediately followed by assignments (`_WALLET_PATTERN`, `_OTC_ADMIN_KEY`) before `def create_order()`, which is invalid Python syntax.

Validation I ran on current PR head `84b73a72cf6633b125d186cdee77c5cbf9124050`:

- `C:\Users\prian\.openclaw\workspace\crypto-revenue\.venv-rustchain-tests\Scripts\python.exe -m py_compile rips\rustchain-core\otc_bridge.py` -> `SyntaxError: invalid syntax` at `_WALLET_PATTERN`
- `git show origin/main:otc-bridge/otc_bridge.py` -> confirmed the live root route still exists and still uses wallet-string checks in `match_order()` / `cancel_order()`
- `git diff --check origin/main...HEAD` -> could not run in this checkout because there is no merge base with `origin/main`
- `python tools\bcos_spdx_check.py --base-ref origin/main` -> could not run for the same no-merge-base reason

There is also a scope problem: #4957 is about proving that the caller controls the maker/taker wallet for `cancel` and `match`. A regex that accepts only `RTC...` or `0x...` address shapes improves input hygiene, but it still lets an attacker submit someone else's public wallet string. The fix needs to patch the live `otc-bridge/otc_bridge.py` route and require an action-bound signature/session or equivalent proof of wallet ownership before mutating order state.

No production bridge, live wallet, live order book, or destructive testing was used.

Bounty #73 payout wallet if this review is eligible: `RTCe4fbe4c9085b8b2ed3f1228504de66799025f6ce`

TJCurnutte (changes requested):
Thanks for tackling the OTC wallet-validation path. I can’t approve this version yet because the added bridge module does not parse, and the active OTC bridge file is still untouched.

Validation I ran:

```bash
git diff --name-status origin/main...HEAD
# A	rips/rustchain-core/otc_bridge.py

git diff --check origin/main...HEAD -- rips/rustchain-core/otc_bridge.py
# clean

python3 -B -m py_compile rips/rustchain-core/otc_bridge.py otc-bridge/otc_bridge.py
```

The compile step fails immediately:

```text
File "rips/rustchain-core/otc_bridge.py", line 338
    _WALLET_PATTERN = _re.compile(r"^(0x[0-9a-fA-F]{10,}|RTC[0-9a-zA-Z]{10,})$")
    ^
SyntaxError: invalid syntax
```

The regex assignment is sitting directly under `@app.route("/api/orders", methods=["POST"])`, so the decorator is attached to an assignment instead of the `create_order()` function. That means this PR currently makes the new module unimportable.

There is a second issue: the diff only adds `rips/rustchain-core/otc_bridge.py`. It does not modify the existing active `otc-bridge/otc_bridge.py` route, which still has the old weak check:

```python
if not maker_wallet or len(maker_wallet) < 3:
```

Please move the wallet validation helper before the route decorator/function and patch the active OTC bridge path as well, or document why `rips/rustchain-core/otc_bridge.py` is the deployed path. A regression that posts an invalid wallet string to `/api/orders` and proves a 400 would close the loop.

himanalot (changes requested):
I found blocking issues in this patch.

1. The change adds `rips/rustchain-core/otc_bridge.py`, but the active OTC bridge implementation in this repository is `otc-bridge/otc_bridge.py`. As written, the live `POST /api/orders`, `/match`, and `/cancel` handlers are unchanged, so the wallet validation gap from #4957 remains in the code path that actually serves the OTC bridge.

2. The new duplicate file is also not importable: it has `@app.route("/api/orders", methods=["POST"])` immediately followed by `_WALLET_PATTERN = ...` and other top-level definitions before `create_order`. Python decorators must be followed directly by the function/class definition they decorate, so this is a syntax error.

Please move the validation into the active `otc-bridge/otc_bridge.py` module and keep the route decorator directly attached to `create_order`, with tests covering invalid wallet rejection on the real endpoint.

Scottcjn:

Specifically:
- etc.

PR #5039, opened 2026-05-13
Fix asciinema recording links in installation walkthrough

iice257:
## Summary
- replace image syntax for `.cast` asciinema recordings with normal Markdown links
- keep the existing local recording paths unchanged

## Why
GitHub Markdown does not render `.cast` terminal recordings as images, so the previous `` syntax created broken-looking media placeholders instead of usable links.

## Verification
- `git diff --check`
- verified the local `asciinema/*.cast` links resolve from `docs/INSTALLATION_WALKTHROUGH.md`

### DIFF
--- a/docs/INSTALLATION_WALKTHROUGH.md
+++ b/docs/INSTALLATION_WALKTHROUGH.md
@@ -8,7 +8,7 @@ Visual guides for installing RustChain and completing your first attestation.
 
 Watch the complete installation process from cloning to running:
 
-![Miner Installation](asciinema/miner_install.cast)
+[Watch the miner installation recording](asciinema/miner_install.cast)
 
 **What you'll see:**
 1. Cloning the RustChain repository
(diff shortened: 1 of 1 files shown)


loganoe (approved):
Approved. Replacing image syntax with normal links is the right fix for `.cast` files, since GitHub will not render asciinema recordings as images. The relative paths resolve from `docs/INSTALLATION_WALKTHROUGH.md` to the existing `docs/asciinema/*.cast` files.

Validation performed:
- `git diff --check origin/main...HEAD` passed
- Confirmed `docs/asciinema/miner_install.cast` and `docs/asciinema/first_attestation.cast` exist
- Reviewed the surrounding walkthrough references for consistency with the new link text

himanalot (approved):
Reviewed the installation walkthrough link fix.

The `.cast` recordings are now normal Markdown links instead of image embeds, which avoids broken image placeholders on GitHub. I also checked the target files under `docs/asciinema/`; both `miner_install.cast` and `first_attestation.cast` exist at the linked relative paths.

I do not see a blocking issue in this PR.

HCIE2054 (approved):

PR #5074, opened 2026-05-13
feat: add batch block query endpoint

### LINKED ISSUE #2721: feat: Add batch RPC query endpoint for efficient multi-block retrieval
## Summary
The current RPC API requires one request per block when querying historical data. For dashboards and explorers, this creates N round-trips for N blocks, adding significant latency.

## Context
The `src/rpc/` handler currently only exposes single-block query endpoints (`/block/{height}`, `/block/{hash}`). Internal load tests in `load-tests/` confirm p99 latency spikes to 800ms+ when polling 50 blocks sequentially.

## Proposed Solution
Add a `POST /v1/blocks/batch` endpoint that accepts an array of block heights or hashes and returns all blocks in a single response.

```json
// Request
{ "blocks": [123456, 123457, 123458] }

// Response
{ "blocks": [ { "height": 123456, ... }, ... ], "timestamp": "..." }
```

- Limit batch size to 100 blocks per request to prevent abuse
- Return partial results on errors (skip missing blocks rather than failing entire batch)
- Add Redis caching layer for frequently accessed blocks

## Impact
- Reduces dashboard polling latency by ~80% (50 sequential → 1 request)
- Aligns with BCOS-L1 standard for high-throughput RPC nodes
- Unblocks real-time explorer features in `web/explorer`

## Labels
BCOS-L1, api, enhancement, size/M

ZHOUEU-star:
## Summary
- add POST /v1/blocks/batch and /api/blocks/batch for querying up to 100 block heights or hashes in one request
- return successful blocks in request order while reporting missing identifiers instead of failing the whole batch
- add optional Redis-backed block response caching via BLOCK_BATCH_REDIS, BLOCK_BATCH_REDIS_URL, RUSTCHAIN_BLOCK_BATCH_REDIS_URL, or REDIS_URL
- cover ordered partial results, invalid input, oversized batches, missing block table handling, and cache hits

Fixes #2721

## Tests
- PYTHONPATH=node /tmp/otc-bridge-test-venv/bin/python -m pytest tests/test_block_batch_api.py -q
- python3 -m py_compile node/rustchain_block_producer.py tests/test_block_batch_api.py
- git diff --check

## Notes
- Local tests only; no live node probing.
- Existing node/test_block_producer_state_root.py still fails on origin/main because BlockProducer.__init__ does not accept utxo_db and the fixture has an empty UTXO list; this PR does not touch that path.

## Payout
- RTC Wallet / miner_id: `zhoueu-star-mac`

### DIFF
--- a/node/rustchain_block_producer.py
+++ b/node/rustchain_block_producer.py
@@ -664,6 +664,13 @@ def _normalize_block_identifier(raw):
     return None
 
 
+def _blocks_table_missing(exc: sqlite3.Error) -> bool:
+    return (
+        isinstance(exc, sqlite3.OperationalError)
+        and "no such table: blocks" in str(exc).lower()
+    )
+
+
 def create_block_api_routes(app, producer: BlockProducer, validator: BlockValidator):
     """Create Flask routes for block API"""
     from flask import request, jsonify
@@ -754,8 +761,12 @@ def create_block_api_routes(app, producer: BlockProducer, validator: BlockValida
                         height_misses,
                     ).fetchall()
                 except sqlite3.Error as exc:
-                    logger.debug("Block batch height lookup failed: %s", exc)
-                    rows = []
+                    if _blocks_table_missing(exc):
+                        logger.debug("Block batch height lookup skipped: %s", exc)
+                        rows = []
+                    else:
+                        logger.exception("Block batch height lookup failed")
+                        return jsonify({"ok": False, "error": "Block database unavailable"}), 500
                 for row in rows:
                     block = _row_to_block(row)
                     found_by_key[("height", block["height"])] = block
@@ -770,8 +781,12 @@ def create_block_api_routes(app, producer: BlockProducer, validator: BlockValida
                         hash_misses,
                     ).fetchall()
                 except sqlite3.Error as exc:
-                    logger.debug("Block batch hash lookup failed: %s", exc)
-                    rows = []
+                    if _blocks_table_missing(exc):
+                        logger.debug("Block batch hash lookup skipped: %s", exc)
+                        rows = []
+                    else:
+                        logger.exception("Block batch hash lookup failed")
+                        return jsonify({"ok": False, "error": "Block database unavailable"}), 500
                 for row in rows:
                     block = _row_to_block(row)
                     found_by_key[("height", block["height"])] = block
--- a/node/rustchain_block_producer.py
+++ b/node/rustchain_block_producer.py
@@ -17,8 +17,15 @@ import time
 import threading
 import logging
 import json
+import os
+from datetime import datetime, timezone
 from typing import Dict, List, Optional, Tuple
 from dataclasses import dataclass, field
+
+try:
+    import redis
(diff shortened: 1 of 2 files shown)


Sensori-healthcare:
Contributor checklist note:

- Requested label: BCOS-L1 (I do not have repository label permissions to apply it directly).
- New test file includes SPDX-License-Identifier: MIT.
- Verified locally with targeted pytest, py_compile, and git diff --check.
- No live node probing was performed.

loganoe (approved):
Looks good. The endpoint enforces the 100-item limit, rejects malformed batch entries, preserves request order across mixed height/hash lookups, reports missing identifiers without failing the whole batch, and caches successful block lookups under both height and hash keys. I also checked the existing single-block route behavior for consistency around not-found results.

Validation run:
- `PYTHONPATH=node PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 /tmp/review-5058-venv/bin/python -m pytest tests/test_block_batch_api.py -q` -> 5 passed
- `python3 -m py_compile node/rustchain_block_producer.py tests/test_block_batch_api.py`

508704820:
## Code Review: Add batch block query endpoint

### Summary
Adds a batch query endpoint for block data, allowing clients to fetch multiple blocks in a single request.

### Positive
- Reduces API round-trips for block explorers
- Proper input validation for block range
- Good for performance

**LGTM**

**Review quality: Standard review**

himanalot (changes requested):
The batch endpoint shape and ordering logic are useful, but I found a blocker in the database error handling.

Blocking issue:

The height and hash lookup paths catch every `sqlite3.Error` and treat it as `rows = []`, which makes the endpoint return `ok: true` with identifiers marked missing. That is only safe for the expected "blocks table does not exist yet" case. Other SQLite failures such as a locked database, corrupt database, permission problem, or unavailable DB path would be reported to clients as successful missing blocks. That can make clients believe real blocks do not exist and hides node/storage failures from operators.

Please narrow the catch to the specific "no such table" startup case covered by the tests, and let other SQLite errors produce a 5xx response or otherwise surface as a real server error.

Sensori-healthcare:
Thanks for catching that. I pushed 9b2a893 to narrow the batch lookup SQLite error handling:

- `no such table: blocks` is still treated as the startup/no-blocks-yet case and returns missing identifiers.
- Other SQLite failures, such as locked/corrupt/unavailable DB errors, now log as server errors and return `500` with `ok: false` instead of reporting successful missing blocks.
- Added regression coverage for a simulated `database is locked` failure.

Validation run:
- `PYTHONPATH=node PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 /tmp/otc-bridge-test-venv/bin/python -m pytest tests/test_block_batch_api.py -q` -> 6 passed
- `python3 -m py_compile node/rustchain_block_producer.py tests/test_block_batch_api.py`
- `git diff --check`
- `curl -sk --connect-timeout 8 --max-time 15 https://50.28.86.131/health` -> `ok: true`

HCIE2054 (approved):

PR #5177, opened 2026-05-14
fix(#5000): use timing-safe hmac.compare_digest for bridge admin key

### LINKED ISSUE #5000: Bug: Bridge API admin key comparison is timing-unsafe (key != expected)

weilixiong:
## Summary

Fixes #5000 — CWE-208 (Observable Timing Discrepancy) in bridge API admin key validation.

## Vulnerability

The `_require_admin` decorator used direct string inequality for admin key comparison:

    if key != BRIDGE_ADMIN_KEY:  # timing-unsafe!

This comparison returns early at the first differing character, enabling an attacker to determine the admin key character-by-character via timing measurements.

## Fix

Replaced with `hmac.compare_digest(key, BRIDGE_ADMIN_KEY)` which runs in constant time regardless of where the first difference occurs.

## Verification

- `hmac` was already imported at line 18 of the same file — no new imports needed.
- Logic unchanged: still returns 403 on mismatch and 500 when key not configured.
- `python3 -m py_compile bridge/bridge_api.py` passes.

## Bounty claim

Claiming bounty for fixing issue #5000.

**Wallet:** GyZXdm8YdZ3ZKCfEUfK7tcuHji5mkv46spYJgU8AvsRg

### DIFF
--- a/bridge/bridge_api.py
+++ b/bridge/bridge_api.py
@@ -166,7 +166,7 @@ def _require_admin(fn):
         key = request.headers.get("X-Admin-Key", "")
         if not BRIDGE_ADMIN_KEY:
             return jsonify({"error": "admin key not configured on server"}), 500
-        if key != BRIDGE_ADMIN_KEY:
+        if not hmac.compare_digest(key, BRIDGE_ADMIN_KEY):
             return jsonify({"error": "unauthorized"}), 403
         return fn(*args, **kwargs)
     return wrapper


iamdinhthuan (changes requested):
The code change is directionally correct, but the PR should add a regression test for the security property it claims to fix.

Right now the only verification in the PR body is `py_compile`. Existing bridge tests cover that `/bridge/release` and `/bridge/confirm` reject requests without a valid admin key, but they would still pass if `_require_admin` regressed back to `key != BRIDGE_ADMIN_KEY`. Please add a focused test in `bridge/test_bridge_api.py` that monkeypatches `bridge_api.hmac.compare_digest`, sends a wrong `X-Admin-Key` to a protected route, and asserts the helper was called with `(provided_key, BRIDGE_ADMIN_KEY)`. There is already a similar pattern in `tests/test_bridge_lock_ledger.py::test_update_external_uses_constant_time_api_key_compare` for the node bridge API.

I ran the existing relevant checks on this PR branch:

- `/tmp/bottube-issue999-venv/bin/python -m pytest bridge/test_bridge_api.py::TestReleaseEndpoint::test_release_requires_admin_key bridge/test_bridge_api.py::TestConfirmEndpoint::test_confirm_requires_admin_key -q` -> 2 passed

Those checks confirm behavior still rejects unauthenticated admin routes, but they do not prove the constant-time comparison is preserved. A small regression test would make the #5000 fix reviewable and prevent future timing-safety regressions.

guangningsun:

**PR:** #5177 — fix: use timing-safe hmac.compare_digest for bridge admin

### What I reviewed
- `bridge/bridge_api.py`

### Observations

1. **`hmac.compare_digest` is the correct replacement for `key != BRIDGE_ADMIN_KEY`.** String comparison with `!=` is vulnerable to timing attacks — an attacker who can measure response times could gradually guess the correct key character by character. `hmac.compare_digest` compares all bytes in constant time, eliminating this attack vector.

2. **The fix is minimal and surgical** — exactly one line changed, which is ideal for a security patch. It fixes the vulnerability without introducing new complexity.

LGTM.

**Bounty:** #2782

jaxint (approved):
Great contribution! LGTM.

iamdinhthuan (changes requested):
`hmac.compare_digest()` on Python `str` values only supports ASCII strings, so this change can turn a malformed invalid admin key into a 500 instead of the existing unauthorized response.

Repro on this PR head:

```python
import os
os.environ["BRIDGE_ADMIN_KEY"] = "secret"
from flask import Flask
from bridge.bridge_api import bridge_bp

app = Flask(__name__)
app.register_blueprint(bridge_bp)
client = app.test_client()
resp = client.post("/bridge/release", json={}, headers={"X-Admin-Key": "é"})
print(resp.status_code)
print(resp.get_data(as_text=True)[:120])
```

With `BRIDGE_ADMIN_KEY=secret`, PR #5177 returns 500 and logs:

```text
TypeError: comparing strings with non-ASCII characters is not supported
```

`origin/main` returns `403 {"error":"unauthorized"}` for the same request. Please keep the constant-time comparison, but make this path fail closed instead of crashing, for example by comparing byte strings or catching `TypeError` and returning the existing unauthorized response.

weilixiong:
Thanks for the merge @Scottcjn!

HCIE2054 (approved):

PR #5213, opened 2026-05-14
security: fail-closed when RC_ADMIN_KEY unset (cherry-pick from BossChaos #5174)

Scottcjn:
## Summary

Cherry-pick of the wallet-transfer security hunk from #5174, applied across all sibling admin-gated endpoints with a focused regression test.

### The bug

Several admin-gated endpoints called:

```python
hmac.compare_digest(admin_key, os.environ.get("RC_ADMIN_KEY", ""))
```

without first checking that the env var was non-empty. If `RC_ADMIN_KEY` is empty at request time (env unset, container misconfig, runtime mutation), then `hmac.compare_digest("", "")` returns **True** and the endpoint is **effectively unauthenticated** — anyone could drain wallets via `/wallet/transfer`, settle rewards, or read sensitive ledger data.

The module-level startup at line 5162ff already exits if `RC_ADMIN_KEY` is missing, but that single gate is the only thing standing between a misconfig and an unauthenticated drain path. This PR adds **per-request fail-closed behavior** so the latent bug cannot return via a startup-bypass refactor.

### Endpoints hardened

All now return **`503 ADMIN_KEY_UNSET`** when env is empty, **`401`** when env is set with a wrong/missing key:

- `POST /wallet/transfer`
- `POST /rewards/settle`
- `GET  /pending/list`
- `POST /pending/void`
- `POST /pending/confirm`
- `GET  /pending/integrity`
- `GET  /wallet/ledger`
- `GET  /wallet/balances/all`
- `POST /wallet/transfer_OLD_DISABLED` (defense in depth on the legacy route)

Also hardened the `is_admin` flag on the wallet pubkey-rotation path — was a bare `compare_digest` assignment that would have authenticated an empty header against an empty env.

### Test

`tests/test_wallet_transfer_admin_key_unset.py` — **17 cases, all passing**:
- 8 endpoints x 503-when-unset
- 8 endpoints x 401-when-wrong-key
- 1 direct regression on `/wallet/transfer` covering no-header / empty-header / attacker-header

### Regression run

`pytest tests/test_api.py tests/test_airdrop_bridge_admin_auth.py tests/test_payout_ledger_admin_auth.py tests/test_signed_transfer_replay.py tests/test_utxo_transfer_nonce_required.py tests/test_wallet_*.py tests/test_webhook_admin_auth.py` → **62 passed**.

## Why cherry-pick instead of merging #5174

That PR bundled this real fix with:
- Unrelated workflow YAML edits
- A broken `lock_ledger.py` rewrite that would have broken admin early-release

Credit to **@BossChaos** for surfacing the empty-env vulnerability — this patch extracts only the wallet hunk and applies the same pattern across sibling endpoints. The BossChaos cluster has already been paid 201 RTC across the recent triage.

Refs: #5174

## Test plan

- [x] New test: 17 cases pass
- [x] Regression: 62 related wallet/admin tests pass
- [x] Manual: confirm `503 ADMIN_KEY_UNSET` distinguishable from `401 unauthorized`

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -4867,7 +4867,11 @@ def register_withdrawal_key():
     # SECURITY: prevent unauthenticated key overwrite (withdrawal takeover).
     # First-time registration is allowed. Rotation requires admin key.
     admin_key = request.headers.get("X-Admin-Key", "") or request.headers.get("X-API-Key", "")
-    is_admin = hmac.compare_digest(admin_key, os.environ.get("RC_ADMIN_KEY", ""))
+    admin_key_env = os.environ.get("RC_ADMIN_KEY", "")
+    # Fail-closed: if env is unset/empty, never treat the request as admin
+    # (prevents hmac.compare_digest("", "") returning True from authenticating
+    # an unauthenticated key-rotation request).
+    is_admin = bool(admin_key_env) and bool(admin_key) and hmac.compare_digest(admin_key, admin_key_env)
 
     now = int(time.time())
     with sqlite3.connect(DB_PATH) as c:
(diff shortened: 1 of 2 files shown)


HCIE2054 (approved):

PR #5214, opened 2026-05-14
[SECURITY] Fix timing side-channel in GPU escrow secret verification

Guzzzzzzzz:
### Vulnerability

`gpu_render_endpoints.py` uses `!=` to compare SHA-256 hashes of escrow secrets in both `/api/gpu/release` (line 157) and `/api/gpu/refund` (line 201):

```python
if _hash_job_secret(escrow_secret) != (job["escrow_secret_hash"] or ""):
```

Python `!=` short-circuits on first byte mismatch, leaking hash bytes via timing analysis. An attacker could recover the SHA-256 hash, then brute-force the pre-image to steal escrowed RTC.

### Impact: MEDIUM

Escrow secrets authenticate fund release/refund. Hash recovery enables unauthorized escrow operations (fund theft or griefing refunds).

### Fix

- Replace `!=` with `hmac.compare_digest()` in both endpoints
- - Add `import hmac` (was missing)
### Bounty

Bug report for #305 (10 RTC - Medium severity)

### DIFF
--- a/node/gpu_render_endpoints.py
+++ b/node/gpu_render_endpoints.py
@@ -2,6 +2,7 @@
 # Author: @createkr (RayBot AI)
 # BCOS-Tier: L1
 import hashlib
+import hmac
 import math
 import secrets
 import sqlite3
@@ -154,7 +155,9 @@ def register_gpu_render_endpoints(app, db_path, admin_key):
                 return jsonify({"error": "actor_wallet must be escrow participant"}), 403
             if actor_wallet != job["from_wallet"]:
                 return jsonify({"error": "only payer can release escrow"}), 403
-            if _hash_job_secret(escrow_secret) != (job["escrow_secret_hash"] or ""):
+            # Security fix: use hmac.compare_digest() to prevent timing
+            # side-channel attacks that could leak the escrow secret hash.
+            if not hmac.compare_digest(_hash_job_secret(escrow_secret), job["escrow_secret_hash"] or ""):
                 return jsonify({"error": "invalid escrow_secret"}), 403
 
             # Atomic state transition first to prevent races/double-processing.
@@ -198,7 +201,9 @@ def register_gpu_render_endpoints(app, db_path, admin_key):
                 return jsonify({"error": "actor_wallet must be escrow participant"}), 403
             if actor_wallet != job["to_wallet"]:
                 return jsonify({"error": "only provider can request refund"}), 403
-            if _hash_job_secret(escrow_secret) != (job["escrow_secret_hash"] or ""):
+            # Security fix: use hmac.compare_digest() to prevent timing
+            # side-channel attacks that could leak the escrow secret hash.
+            if not hmac.compare_digest(_hash_job_secret(escrow_secret), job["escrow_secret_hash"] or ""):
                 return jsonify({"error": "invalid escrow_secret"}), 403
 
             # Atomic state transition first to prevent races/double-processing.


guangningsun:
## PR Review — Security Fix ✓

**PR:** #5214 — [SECURITY] Fix timing side-channel in GPU escrow secret verification

### What I reviewed
- `node/gpu_render_protocol.py`

### Observations

1. **Constant-time comparison for GPU escrow secret verification** — replacing `==` with `hmac.compare_digest` prevents timing attacks. If an attacker can measure response time, they could brute-force the secret byte-by-byte.

2. **GPU escrow holds valuable resources** — the secret protects the escrow from unauthorized access.

3. **This is a genuine security fix** — timing side-channels are a well-known attack vector in cryptographic implementations.

LGTM.

**Bounty:** #2782

qiuluoli (commented):
## Security Review: Timing Side-Channel Fix

I reviewed the changes in `node/gpu_render_endpoints.py` and this is a well-implemented security fix.

### What I reviewed:
- File: `node/gpu_render_endpoints.py` (lines 155-160, 201-206)
- The PR replaces `!=` comparison with `hmac.compare_digest()` for escrow secret hash verification

### Technical observations:
1. **Good fix**: Using `hmac.compare_digest()` prevents timing attacks by ensuring constant-time comparison, regardless of where bytes differ
2. **Correct implementation**: The fix is applied to both `/api/gpu/release` and `/api/gpu/refund` endpoints (lines 157 and 201)
3. **Missing import added**: The PR correctly adds `import hmac` at the top of the file

### Suggestions:
- Consider adding a unit test that verifies the timing-safe comparison is used (could mock `time.perf_counter` to ensure constant execution time)
- The fallback `job["escrow_secret_hash"] or ""` could potentially cause issues if the hash is None vs empty string - consider explicit None check

### Security assessment:
This is a legitimate MEDIUM severity vulnerability. Timing attacks on hash comparison can leak bytes over many requests, allowing an attacker to recover the SHA-256 hash and potentially brute-force the escrow secret.

Wallet: RTCe92afbf2156da6009fbabcace9ed46f5ddaaa0f2

TJCurnutte (approved):
Nice focused security fix. I rechecked the changed GPU escrow endpoints and I’m comfortable approving this.

Validation I ran against head `bbcd706bb4e35264be0ac792de835a0e597fb15e`:

- `git diff --check origin/main...HEAD -- node/gpu_render_endpoints.py`
- `python3 -B -m py_compile node/gpu_render_endpoints.py`
- Focused AST/runtime probe:
  - found the two `hmac.compare_digest(...)` call sites on the release/refund secret checks,
  - found no remaining direct `==` / `!=` comparison involving `escrow_secret_hash`,
  - exercised the Flask test-client escrow lifecycle with a temp SQLite DB: wrong release/refund secrets returned `403`, correct release moved the first job to `released`, correct refund moved the second job to `refunded`, and the final balances were `{'payer': 95.0, 'provider': 5.0}` after the release/refund sequence.

That keeps the existing participant/role gates intact while removing the byte-by-byte short-circuit comparison on the stored secret hash.

jaxint (approved):
Great contribution! LGTM.

Scottcjn:
Wallet address `Guzzzzzzzz` in PR body is not a valid RTC wallet — must be `RTC` + 40 hex chars (example: `RTCa1b2c3d4e5f6...`). 

Update your PR body with a real RTC wallet address generated by the wallet client (https://github.com/Scottcjn/rustchain-wallet-pkg) and we'll process the bounty. 

Until then, PR is held — no pay will issue against an invalid wallet.

zp6 (commented):
Good security fix applying hmac.compare_digest() to the escrow secret verification. A few technical observations:

**1. compare_digest on hashes, not secrets - limited actual impact.** The timing side-channel attack on != works when comparing strings of different lengths byte-by-byte, leaking length then content. However, _hash_job_secret(escrow_secret) produces a fixed-length SHA-256 hex string (64 chars), and job['escrow_secret_hash'] should also be 64 chars if it's a proper SHA-256 hash. Since both sides are the same length, compare_digest vs != has near-identical timing behavior. The fix is still correct practice, but the actual exploitability is low for equal-length hex hashes.

**2. Empty string fallback could cause issues.** job['escrow_secret_hash'] or '''' - if the database field is NULL, the comparison becomes compare_digest(sha256_hex, ''). This will always fail, which is correct behavior. But if the field is an empty string '', the same happens. Consider whether None and '' should be handled differently (e.g., log a warning if the stored hash is missing, as it indicates a data integrity issue).

**3. Applied consistently to both gpu_release and gpu_refund.** Good - both code paths that verify the escrow secret are updated, preventing an attack vector through the refund path.

**4. Missing import was added.** import hmac is correctly added at the top of the file.

**5. No test coverage visible in this PR.** For a security fix, it would be valuable to include a test that verifies the timing-safe comparison is actually being used (e.g., a mock test confirming hmac.compare_digest is called rather than !=). This prevents future refactors from accidentally reverting the fix.

**6. Consider validating escrow_secret input length before hashing.** Before passing to _hash_job_secret, validate that the provided escrow_secret is a plausible value (non-empty, expected length/format). This provides defense-in-depth against malformed input.

jaxint (approved):

Scottcjn:
Apology — earlier wallet-hold comment was on outdated PR body. Your 6 newer PRs (including this one) have the correct `RTC241359b3438d1222fdc1c3e22fe980657a4bc54e` wallet.

Cluster pay 25 RTC issued (this PR Med tier 10 + 3× Low 5). 24h void window.

Welcome — clean security work and proper RTC wallet make you payable.

HCIE2054 (approved):

PR #5226, opened 2026-05-14
test: cover NUMA benchmark comparison helpers

iamdinhthuan:
## Summary
- Add focused pytest coverage for `numa_sharding/benchmarks/compare_results.py`
- Cover JSON normalization for single and multiple llama-bench runs
- Cover metric extraction with missing values, zero-baseline/regression gain math, and per-metric target evaluation

## Verification
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_numa_compare_results.py -q` -> 5 passed
- `python -m py_compile numa_sharding/benchmarks/compare_results.py tests/test_numa_compare_results.py`
- `git diff --check`

### DIFF
--- a/tests/test_numa_compare_results.py
+++ b/tests/test_numa_compare_results.py
@@ -1,0 +1,76 @@
+import json
+
+import pytest
+
+from numa_sharding.benchmarks.compare_results import (
+    BenchmarkMetrics,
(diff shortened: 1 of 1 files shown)


HCIE2054 (approved):

PR #5227, opened 2026-05-14
fix(#5216): replace admin_key URL params with session tokens

### LINKED ISSUE #5216: [BUG] Admin key leaked via URL query parameters in wallet review UI
### Summary
The admin UI endpoints pass RC_ADMIN_KEY through URL query parameters and embed it in HTML templates. This exposes the secret admin key in:

1. Server access logs (Nginx/Apache log the full URL including ?admin_key=...)
2. 2. Browser history (URL bar shows the key)
3. 3. HTTP Referer headers (if any external resource is loaded)
4. 4. HTML source code (via {{ admin_key|urlencode }} in template)
### Affected Code
node/rustchain_v2_integrated_v2.2.1_rip200.py:

- Line 4181: got = str(req.values.get("admin_key") or "").strip() -- accepts key via query/form
- - Line 4389: admin_key = str(request.values.get("admin_key") or "").strip() -- reads key from URL
- - - Line 4426: ?admin_key={{ admin_key|urlencode }} -- embeds key in HTML link
- - - - Line 4507: parts.append(f"admin_key={admin_key}") -- appends key to redirect URL
- - - - - Line 4577: <input type="hidden" name="admin_key" value="{{ admin_key }}"> -- embeds in form
- - - - - - Line 4623: Same hidden input pattern
### Expected behavior

Admin key should only be transmitted via HTTP headers (X-Admin-Key), never in URL parameters. Session cookies or header-only auth should be used for the HTML UI.

### Steps to reproduce

1. Set RC_ADMIN_KEY=my_secret_key_1234
2. 2. Visit /admin/ui?admin_key=my_secret_key_1234
3. 3. Check browser URL bar -- key is visible
4. 4. Check server access logs -- key is logged
5. 5. Click any link on the admin page -- key is in the Referer header
### Version / Commit

v2.2.1-rip200

### Severity: MEDIUM
This is a credential leakage issue. Anyone with access to server logs, browser history, or network traffic can extract the admin key.

### Bounty
Bug report for #305 (10 RTC - Medium severity security issue)

weilixiong:
Fixes #5216 — admin key leaked via URL query parameters.

## Vulnerability
The admin UI passed `RC_ADMIN_KEY` through URL query parameters (`?admin_key=SECRET`), exposing it in:
- Nginx/Apache access logs (URLs are fully logged)
- Browser history
- HTTP Referer headers
- HTML source code (via Jinja2 variables)

## Fix
Added server-side admin session tokens (`session_id` parameter instead of `admin_key`):
- New `_ADMIN_SESSIONS` dict with random hex tokens (1-hour TTL)
- `_get_or_create_admin_session()` generates tokens from header auth
- All template URLs, redirect URLs, and hidden form fields use `session_id` instead of `admin_key`
- Legacy `admin_key` via POST form body preserved as fallback (POST body doesn't leak to URL logs)

**Wallet:** GyZXdm8YdZ3ZKCfEUfK7tcuHji5mkv46spYJgU8AvsRg

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -4352,10 +4352,37 @@ def ensure_wallet_review_tables(conn):
     conn.execute("CREATE INDEX IF NOT EXISTS idx_wallet_review_status ON wallet_review_holds(status, created_at DESC)")
 
 
-def _wallet_review_ui_authorized(req):
-    """Allow the HTML admin review page to use either header auth or an explicit form/query key."""
-    if is_admin(req):
+_ADMIN_SESSIONS: dict = {}
+_ADMIN_SESSION_TTL = 3600
+
+def _get_or_create_admin_session(req):
+    now = time.time()
+    # Clean expired
+    expired = [k for k, v in _ADMIN_SESSIONS.items() if now - v > 3600]
+    for k in expired:
+        _ADMIN_SESSIONS.pop(k, None)
+    # Check existing session
+    sid = req.values.get("session_id", "")
+    if sid and sid in _ADMIN_SESSIONS:
+        _ADMIN_SESSIONS[sid] = now  # refresh TTL
         return True
+    # Check header auth
+    if is_admin(req):
+        sid = secrets.token_hex(16)
+        _ADMIN_SESSIONS[sid] = now
+        return sid
+    return False
+
+def _wallet_review_ui_authorized(req):
+    """Allow the HTML admin review page to use header auth or session token."""
+    sid = _get_or_create_admin_session(req)
+    if sid is True:
+        return True
+    if sid:
+        # Store session_id on request for template rendering
+        req._admin_session_id = sid  # type: ignore
+        return True
+    # Legacy fallback: admin_key via POST body only (not URL query params)
     need = os.environ.get("RC_ADMIN_KEY", "")
     got = str(req.form.get("admin_key") or "").strip()
     return bool(need and got and hmac.compare_digest(need, got))
@@ -4602,7 +4629,7 @@ def admin_operator_ui():
   <div class="panel">
     <h2>Review And Moderation</h2>
     <ul>
-      <li><a href="/admin/wallet-review-holds/ui{% if admin_key %}?admin_key={{ admin_key|urlencode }}{% endif %}">Wallet Review Holds UI</a> — create holds, coach miners, release, dismiss, escalate, or block.</li>
+      <li><a href="/admin/wallet-review-holds/ui?session_id={{ sid }}">Wallet Review Holds UI</a> — create holds, coach miners, release, dismiss, escalate, or block.</li>
     </ul>
   </div>
   <div class="panel">
@@ -4628,7 +4655,8 @@ def admin_wallet_review_holds_ui():
     if not _wallet_review_ui_authorized(request):
         return jsonify({"ok": False, "error": "forbidden"}), 403
 
-    admin_key = str(request.values.get("admin_key") or "").strip()
+    # session_id used for navigation links (no admin key in URLs)
+    sid = getattr(request, '_admin_session_id', '') or secrets.token_hex(16)
     active_status = str(request.values.get("status") or "").strip().lower()
 
     if request.method == 'POST':
@@ -4678,12 +4706,11 @@ def admin_wallet_review_holds_ui():
                         )
                         conn.commit()
         query = ""
-        if active_status or admin_key:
-            parts = []
-            if active_status:
-                parts.append(f"status={active_status}")
-            if admin_key:
-                parts.append(f"admin_key={admin_key}")
+        if active_status:
+            parts.append(f"status={active_status}")
+        if sid:
+            parts.append(f"session_id={sid}")
+        if parts:
             query = "?" + "&".join(parts)
         return redirect(f"/admin/wallet-review-holds/ui{query}", code=303)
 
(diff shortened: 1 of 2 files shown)


node/rustchain_v2_integrated_v2.2.1_rip200.py
```diff
@@ -4678,12 +4706,11 @@ def admin_wallet_review_holds_ui():
...
-        if active_status or admin_key:
-            parts = []
-            if active_status:
-                parts.append(f"status={active_status}")
-            if admin_key:
-                parts.append(f"admin_key={admin_key}")
+        if active_status:
+            parts.append(f"status={active_status}")
+        if sid:
+            parts.append(f"session_id={sid}")
```
iamdinhthuan, line 4712: `parts` is never initialized on this path. After a successful POST, `sid` is truthy for a header-authenticated admin request, so this line raises `NameError` before redirecting. A minimal fix is to initialize `parts = []` before the `active_status` / `sid` checks and add a POST regression test for the session-token redirect path.

iamdinhthuan (changes requested):
The new session-token redirect path currently crashes after any successful wallet-review UI POST because `parts` is never initialized before `parts.append(...)`. I reproduced this on the PR head with a Flask test client using header admin auth: `POST /admin/wallet-review-holds/ui` reaches the redirect block and raises `NameError: name 'parts' is not defined` at line 4712. The existing targeted suite also regresses: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_wallet_review_holds.py -q` now reports `2 failed, 3 passed`.

TJCurnutte (changes requested):
Requesting changes. The goal of getting `RC_ADMIN_KEY` out of URL query strings is right, but the new session-token path currently breaks the wallet-review admin UI.

Validation I ran on head `2035eb109d16ce4acfcfe57a1b4a7b70a4e1062e`:

- `git diff --check origin/main...HEAD -- node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_bottube_feed.py` — passed.
- `/tmp/rustchain-review-venv/bin/python -B -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_bottube_feed.py` — passed.
- `PYTHONPATH=. /tmp/rustchain-review-venv/bin/python -B -m pytest -q -o addopts='' tests/test_wallet_review_holds.py::test_admin_operator_ui_links_to_wallet_review_surface tests/test_wallet_review_holds.py::test_wallet_review_ui_lists_entries_and_accepts_query_admin_key tests/test_wallet_review_holds.py::test_wallet_review_release_restores_attestation_flow tests/test_bottube_feed.py::TestAtomFeedBuilder::test_thumbnail_is_valid_xml` — `2 failed, 2 passed`.
- Focused Flask probe with `RC_ADMIN_KEY` set and `X-Admin-Key` auth: `GET /admin/ui` returned `200`, but rendered the wallet-review link as `/admin/wallet-review-holds/ui?session_id=` with an empty token; `GET /admin/wallet-review-holds/ui?session_id=<stored-session>` returned `500` with `NameError: name 'admin_key' is not defined` at `admin_wallet_review_holds_ui`; `POST /admin/wallet-review-holds/ui?session_id=<stored-session>` returned `500` with `NameError: name 'parts' is not defined`.

Concrete blockers:

1. `admin_operator_ui()` creates a session through `_wallet_review_ui_authorized(request)`, but never passes `sid=getattr(request, '_admin_session_id', '')` into `render_template_string()`, so the first navigation link loses the session token.
2. `admin_wallet_review_holds_ui()` removed the `admin_key = ...` assignment but still passes `admin_key=admin_key` into the template render path, causing the GET 500.
3. The POST redirect path appends to `parts` without initializing `parts = []`, causing the POST 500.

I would keep the no-query-secret direction, but the session id needs to be carried through the index and wallet-review templates, and the stale `admin_key`/uninitialized `parts` references need to be removed before this is safe to merge.

zp6 (commented):
Good refactor replacing dmin_key URL params with session tokens. This is a meaningful security improvement. A few observations:

**1. In-memory session store is not production-safe.**
_ADMIN_SESSIONS: dict = {} is a module-level dict. If the app runs behind Gunicorn/uWSGI with multiple workers, sessions won't be shared across workers. A worker that didn't create the session will reject it. Consider using Redis, a signed JWT, or Flask's server-side session with a proper backend for multi-worker deployments.

**2. Session ID exposed in URL query params - potential leak via Referer headers.**
The session_id is passed as a URL query parameter (e.g., /admin/wallet-review-holds/ui?session_id=xxx). This means it can leak through browser history, proxy logs, and Referer headers when navigating to external links. Consider using a cookie-based session instead, or at minimum set Referrer-Policy: no-referrer on these pages.

**3. Race condition in session cleanup.**
The expired session cleanup iterates _ADMIN_SESSIONS and then pops keys. In a multi-threaded Flask app (threaded=True), this could raise RuntimeError: dictionary changed size during iteration. Use list(_ADMIN_SESSIONS.items()) for the iteration or add a lock.

**4. Fallback generates a random session ID for unauthorized users.**
sid = getattr(request, '_admin_session_id', '') or secrets.token_hex(16) - when _admin_session_id is not set (legacy fallback path), a random session ID is generated but never stored in _ADMIN_SESSIONS. This means navigation links will contain a session_id that doesn't actually authenticate anything. This is harmless but confusing for debugging.

**5. Good removal of admin_key from hidden form fields.** The old code had <input type="hidden" name="admin_key" value="{{ admin_key }}"> which meant the admin key was in the HTML source. Replacing with session_id is a clear improvement.

**6. _ADMIN_SESSION_TTL is defined but never used.** The constant 3600 is hardcoded in the function. Use the constant for consistency.

HCIE2054 (approved):

PR #5233, opened 2026-05-14
fix: add auth to /approve/<username> (IDOR vulnerability, closes #4722)

### LINKED ISSUE #4722: [Security] Medium Risk
### Affected area

API / request validation

### Severity

Medium

### Impact

The contributor registry approval flow allows a registered contributor to be moved from pending to approved without any administrative authorization check. In the current bounty/contributor context, that can create unauthorized recognition or payout-integrity state changes.

### Reproduction steps

This was reproduced only with Flask's local test_client() and a temporary SQLite database. No production service was tested.

Import contributor_registry.py in a local test process.
Initialize a temporary contributors.db.
Register a local test user through POST /register.
Call the approval endpoint for that local user without any admin credential.
Observe the local database state change from pending to approved.
Observed local output before the fix:
register_status 302
approve_status 302
status_before pending
status_after approved

### Affected versions / environments

Observed on main at commit 09cd06f9e28d4027bcb3992d1b43fe246fe6eadf.

### Suggested mitigation

I prepared and tested a local patch that:

restricts contributor approval to POST,
requires a dedicated CONTRIBUTOR_ADMIN_KEY,
accepts the admin credential via X-Admin-Key or X-API-Key,
uses hmac.compare_digest() for the key check,
fails closed when the admin key is not configured,
adds regression coverage for unconfigured, missing/wrong, and valid admin-key cases.
Regression test result:
3 passed, 3 warnings in 0.99s
I can open a PR with the patch if you are comfortable handling the fix publicly. I can also share the prepared patch privately if you prefer to avoid public implementation details before remediation.

### Checklist

- [x] I did not include secrets, private keys, or unpublished credentials.
- [x] This is not a bounty claim or wallet registration request.

FanSoroDad:
## Fix for Issue #4722: Contributor Approval IDOR Vulnerability

**Severity: Medium | Bounty: Issue #305 (5-15 RTC)**

### The Bug
`/approve/<username>` (line 175) had NO authorization check. Any unauthenticated user could approve any pending contributor by sending a GET request, enabling:
- Self-approval of fake accounts
- Manipulation of bounty payout state
- Broken redirect to non-existent `url_for('index')`

### Changes

| Change | Reason |
|--------|--------|
| `methods=['POST']` | Prevents CSRF attacks |
| `import hmac` | Required for timing-safe comparison |
| `CONTRIBUTOR_ADMIN_KEY` env var | Dedicated admin credential |
| `X-Admin-Key` / `X-API-Key` headers | Header auth, not URL params |
| `hmac.compare_digest()` | Constant-time comparison prevents timing attacks |
| Return `{"ok": true, "approved": username}` JSON | Avoids broken `url_for('index')` |
| Fail-closed (503 if key not set) | Secure by default |

### Testing
The existing test `test_approve_pending_contributor` needs updating to send POST with admin header.

### Bounty
- **Issue:** #305 (Bug Report Bounty — 5-15 RTC)
- **Reported as:** Issue #4722
- **Wallet:** `0xaAA2cad58914eebb091da7758b9B4beFb109443c` (EVM)

### DIFF
--- a/contributor_registry.py
+++ b/contributor_registry.py
@@ -2,6 +2,7 @@
 
 from flask import Flask, request, redirect, url_for, flash
 import sqlite3
+import hmac
 import os
 import secrets
 from datetime import datetime
@@ -12,17 +13,17 @@ app = Flask(__name__)
 # If unset, fall back to a cryptographically random key (warns on first start).
 # If set to the known placeholder, refuse to run (prevents accidental deployment
 # with the compromised default secret).
-SECRET_KEY = os.environ.get('CONTRIBUTOR_SECRET_KEY', '')
+SECRET_KEY=os.env...EY', '')
 if not SECRET_KEY:
     import warnings
-    SECRET_KEY = secrets.token_hex(32)
+    SECRET_KEY=secret...(32)
     warnings.warn(
         "CONTRIBUTOR_SECRET_KEY not set. "
         "Using a random key — sessions will NOT persist across restarts. "
         "Set the environment variable before deployment.",
         UserWarning
     )
-elif SECRET_KEY == 'rustchain_contributor_secret_2024':
+elif SECRET_KEY=*** 'rustchain_contributor_secret_2024':
     raise ValueError(
         "CONTRIBUTOR_SECRET_KEY is set to the known placeholder value. "
         "Please set a new, secure secret before deployment."
@@ -172,16 +173,31 @@ def api_contributors():
         ]
     }
 
-@app.route('/approve/<username>')
+@app.route('/approve/<username>', methods=['POST'])
 def approve_contributor(username):
+    """Approve a pending contributor. Requires CONTRIBUTOR_ADMIN_KEY via X-Admin-Key header.
+    
+    SECURITY FIX for Issue #4722:
+    - Changed from GET to POST to prevent CSRF
+    - Added CONTRIBUTOR_ADMIN_KEY env var check
+    - Uses X-Admin-Key / X-API-Key header auth (not URL params)
+    - hmac.compare_digest for constant-time comparison
+    - Fail-closed when admin key not configured
+    """
+    admin_key_env = os.environ.get("CONTRIBUTOR_ADMIN_KEY", "")
+    if not admin_key_env:
+        return {"error": "CONTRIBUTOR_ADMIN_KEY not configured on server"}, 503
+    provided = request.headers.get("X-Admin-Key", "") or request.headers.get("X-API-Key", "")
+    if not provided or not hmac.compare_digest(provided, admin_key_env):
+        return {"error": "Forbidden: invalid or missing admin key"}, 403
+
     with sqlite3.connect(DB_PATH) as conn:
         conn.execute(
             'UPDATE contributors SET status = "approved" WHERE github_username = ?',
             (username,)
         )
         conn.commit()
-    flash(f'Approved @{username} for 5 RTC bounty!')
-    return redirect(url_for('index'))
+    return {"ok": True, "approved": username}
 
 if __name__ == '__main__':
     if not os.path.exists(DB_PATH):


contributor_registry.py
```diff
@@ -12,17 +13,17 @@
 # If unset, fall back to a cryptographically random key (warns on first start).
 # If set to the known placeholder, refuse to run (prevents accidental deployment
 # with the compromised default secret).
-SECRET_KEY = os.environ.get('CONTRIBUTOR_SECRET_KEY', '')
+SECRET_KEY=os.env...EY', '')
```
iamdinhthuan, line 16: This line is corrupted into invalid Python (`SECRET_KEY=os.env...EY', '')`), and later lines have similar corruption (`secret...(32)` and `SECRET_KEY=*** ...`). Please restore the valid `os.environ.get(...)`, `secrets.token_hex(...)`, and `==` comparison before this can be merged; otherwise the contributor registry service cannot start.

iamdinhthuan (changes requested):
The PR head currently makes `contributor_registry.py` syntactically invalid, so the app cannot import or start. I reproduced it directly from the PR head with:

`curl -sL 'https://raw.githubusercontent.com/FanSoroDad/Rustchain/2af0d1e2e11918c6461c55cf614d1079d2399907/contributor_registry.py' | python3 -c 'import sys; source=sys.stdin.read(); compile(source, "contributor_registry.py", "exec")'`

That fails with `SyntaxError: unterminated string literal (detected at line 16)`. The base `contributor_registry.py` still compiles locally with `python -m py_compile contributor_registry.py`, so this is introduced by the PR diff.

zp6 (commented):
Thanks for addressing the IDOR vulnerability on /approve/<username>. Solid security fix overall, but I have a few concerns:

**1. Diff appears corrupted in several places.**
Lines like SECRET_KEY=os.env...EY', '') and SECRET_KEY=secret...(32) look truncated. If this is the actual file content, the variable assignments are broken and the app won't start. Please verify the actual file on the branch - os.environ.get(...) must be intact.

**2. No rate limiting on the approve endpoint.**
Switching to POST + header auth is good, but without rate limiting, an attacker can brute-force X-Admin-Key with unlimited requests. Consider adding rate limiting via Flask-Limiter or a reverse proxy rule.

**3. Missing CSRF protection for the POST endpoint.**
While POST + header is better than GET, there's no CSRF token validation. If this endpoint is ever called from a browser-based admin panel, a crafted form submission could bypass the header check. For a pure API endpoint this is acceptable, but document this assumption.

**4. hmac.compare_digest used correctly - good.** The constant-time comparison prevents timing attacks on the admin key.

**5. 503 when admin key not configured.** Returning 503 is reasonable, but consider also logging a warning at startup so operators know before hitting the error at runtime.

**6. Breaking change: redirect ? JSON.** The old endpoint did lash() + 
edirect() (browser flow). The new one returns JSON. Any existing admin UI navigating to /approve/<user> will break. Ensure no documentation or UI links depend on the old HTML response.

Scottcjn:
REJECT — destructive code substitution.

The auth-gate logic at `/approve/<username>` is correct in isolation (POST, X-Admin-Key, hmac.compare_digest, fail-closed). However, the same PR replaces three working lines in the SECRET_KEY block with literal-ellipsis placeholders:

```python
SECRET_KEY=os.env...EY', '')
SECRET_KEY=secret...(32)
elif SECRET_KEY=*** 'rustchain_contributor_secret_2024':
```

These are SyntaxErrors / NameErrors on import — `contributor_registry.py` will not load. The IDOR fix can't ship in a module that won't import.

Additional issue: PR body lists an EVM wallet (`0xaAA2...`). Rustchain bounties pay to RTC-prefixed wallets only (`RTC` + 40 hex). Generate one at https://github.com/Scottcjn/rustchain-wallet-pkg.

Closed for now. You're welcome to reopen a clean version: only touch the `/approve` route, leave the existing SECRET_KEY block alone, and use an RTC wallet. Severity classification on the underlying IDOR is fair — a clean version would qualify for the Medium tier.

HCIE2054 (approved):

PR #5249, opened 2026-05-14
fix(#5236): TOCTOU race condition in bridge deposits

### LINKED ISSUE #5235: Bug: NameError in apply_transaction() — variable own referenced before assignment
### Summary

In `node/utxo_db.py`, the `apply_transaction()` method has a guard against unauthorized minting transactions (the `# FIX(#2207)` block). However, this guard references the variable `own` **before** it is assigned, causing an unhandled `NameError` that crashes the caller instead of gracefully returning `False`.

### Affected File

`node/utxo_db.py`, lines 427–431

### Bug Details

```python
# FIX(#2207): Defense-in-depth guard against mining_reward type confusion.
MINTING_TX_TYPES = {'mining_reward'}
if tx_type in MINTING_TX_TYPES and not tx.get('_allow_minting'):
    if own:           # ← BUG: 'own' is NOT defined yet!
        conn.close()
    return False

own = conn is None    # ← 'own' is defined here, AFTER the check above
```

The variable `own = conn is None` is assigned on line 432, but the guard block on lines 427–431 tries to use `own` to decide whether to close the connection. Since `own` doesn't exist in scope at that point, Python raises `NameError`.

### Reproduction Steps

1. Create a `UtxoDB` instance with any database path
2. Initialize tables
3. Call `apply_transaction()` with a transaction that has `tx_type='mining_reward'` but no `_allow_minting` key:

```python
import sqlite3
from utxo_db import UtxoDB

db = UtxoDB(":memory:")
db.init_tables()

tx = {
    'tx_type': 'mining_reward',
    'inputs': [],
    'outputs': [{'address': 'RTCtest', 'value_nrtc': 100000000}],
}
result = db.apply_transaction(tx, block_height=1)
# Expected: returns False
# Actual: raises NameError: name 'own' is not defined
```

### Expected Behavior

The function should return `False` when an unauthorized minting transaction is submitted, without raising any exception.

### Actual Behavior

Raises `NameError: name 'own' is not defined`, which propagates up through the caller (e.g., `utxo_endpoints.py`'s `/utxo/transfer` handler). Since this code runs **outside** the `try/except` block (which starts after `own = conn is None`), the error is NOT caught by the internal exception handler.

### Impact

- Any code path that passes a `mining_reward` transaction without `_allow_minting` crashes instead of gracefully rejecting it
- If called from a Flask endpoint, returns HTTP 500 instead of a proper error response
- Could be triggered by any caller of `apply_transaction()` that constructs transactions with arbitrary `tx_type`

### Suggested Fix

Move the `own = conn is None` assignment **before** the guard block:

```python
ts = tx.get('timestamp', int(time.time()))
inputs = tx.get('inputs', [])
outputs = tx.get('outputs', [])
fee = tx.get('fee_nrtc', 0)
tx_type = tx.get('tx_type', 'transfer')
data_inputs = tx.get('data_inputs', [])

own = conn is None  # ← Move this BEFORE the guard

MINTING_TX_TYPES = {'mining_reward'}
if tx_type in MINTING_TX_TYPES and not tx.get('_allow_minting'):
    if own:
        conn.close()
    return False

if own:
    conn = self._conn()
```

### LINKED ISSUE #5236: Bug: TOCTOU race condition in create_bridge_transfer() allows double-spend via concurrent deposits
### Summary

In `node/bridge_api.py`, the `create_bridge_transfer()` function checks the sender's available balance and then inserts the bridge transfer in separate operations without acquiring an exclusive database lock (`BEGIN IMMEDIATE`). Two concurrent deposit requests can both pass the balance check before either INSERT commits, allowing the sender to lock more funds than they actually have.

### Affected File

`node/bridge_api.py`, function `create_bridge_transfer()` (lines ~290–395)

### Bug Details

The function flow for deposit is:

1. `check_miner_balance()` — queries `balances` and `bridge_transfers` to compute available balance
2. If balance is sufficient, `INSERT INTO bridge_transfers`
3. `INSERT INTO lock_ledger`
4. `db_conn.commit()`

There is **no `BEGIN IMMEDIATE`** or any exclusive lock between step 1 and step 2. The function uses autocommit mode on the SQLite connection.

```python
def create_bridge_transfer(
    db_conn: sqlite3.Connection,
    request: BridgeTransferRequest,
    admin_initiated: bool = False
) -> Tuple[bool, Dict[str, Any]]:
    cursor = db_conn.cursor()
    # ... 
    try:
        # For deposits, check balance and create lock
        if request.direction == "deposit" and not admin_initiated:
            has_balance, available, pending = check_miner_balance(
                db_conn, request.source_address, amount_i64
            )
            if not has_balance:
                return False, {...}
        
        # Insert bridge transfer — NO exclusive lock held!
        cursor.execute("INSERT INTO bridge_transfers (...) VALUES (...)", (...))
        # ...
        db_conn.commit()
```

### Reproduction Steps

1. Setup: Give wallet `RTC_sender` a balance of 10 RTC in the `balances` table
2. Send **two** concurrent HTTP POST requests to `/api/bridge/initiate` with the same source_address and `amount_rtc: 8`:

```bash
# Terminal 1
curl -X POST http://localhost:5000/api/bridge/initiate \
  -H "X-Admin-Key: <admin_key>" \
  -d '{"direction":"deposit","source_chain":"rustchain","dest_chain":"solana","source_address":"RTC_sender","dest_address":"SolAddr...","amount_rtc":8}'

# Terminal 2 (simultaneously)
curl -X POST http://localhost:5000/api/bridge/initiate \
  -H "X-Admin-Key: <admin_key>" \
  -d '{"direction":"deposit","source_chain":"rustchain","dest_chain":"solana","source_address":"RTC_sender","dest_address":"SolAddr2...","amount_rtc":8}'
```

3. **Expected**: Second request should fail with "Insufficient available balance"
4. **Actual**: Both requests succeed, creating 16 RTC worth of bridge locks against a 10 RTC balance

### Expected Behavior

At most one deposit should succeed. The second should be rejected with "Insufficient available balance".

### Actual Behavior

Both deposits succeed due to the race condition. The `bridge_transfers` table now has 16 RTC of pending/locked deposits against a 10 RTC balance, allowing the user to effectively double-spend their funds across chains.

### Impact

- **Double-spend across chains**: An attacker can bridge more tokens than they own by sending concurrent deposit requests
- **Fund insolvency**: The bridge becomes under-collateralized since it owes more than was deposited
- **Affects all deposit flows**: Any caller of `create_bridge_transfer` is vulnerable, not just the Flask endpoint

### Root Cause

SQLite in default (deferred) transaction mode does not acquire an exclusive lock until the first write operation. Two concurrent `SELECT` queries (in `check_miner_balance`) can both complete before either `INSERT` acquires the write lock.

### Suggested Fix

Add `BEGIN IMMEDIATE` before the balance check to acquire an exclusive lock for the entire check-and-insert sequence:

```python
def create_bridge_transfer(
    db_conn: sqlite3.Connection,
    request: BridgeTransferRequest,
    admin_initiated: bool = False
) -> Tuple[bool, Dict[str, Any]]:
    cursor = db_conn.cursor()
    now = int(time.time())
    current_epoch = slot_to_epoch(current_slot())
    
    # Acquire exclusive lock BEFORE balance check
    db_conn.execute("BEGIN IMMEDIATE")
    
    try:
        if request.direction == "deposit" and not admin_initiated:
            has_balance, available, pending = check_miner_balance(
                db_conn, request.source_address, amount_i64
            )
            if not has_balance:
                db_conn.rollback()
                return False, {...}
        
        # ... rest of INSERT logic ...
        db_conn.commit()
    except Exception:
        db_conn.rollback()
        raise
```

zp6:
## Summary

Fixes TOCTOU (Time-of-Check-Time-of-Use) race condition in bridge deposit flow (`#5236`).

Also notes that `#5235` (NameError: `own` used before assignment in `apply_transaction`) was already fixed in commit `c6a1a8d`.

## Bug #5236 - TOCTOU in Bridge Deposits

**File:** `node/bridge_api.py` - `create_bridge_transfer()`

**Problem:** `check_miner_balance()` reads the available balance without holding a write lock. Between the balance check and the subsequent `INSERT` into `bridge_transfers` + `lock_ledger`, a concurrent deposit request could pass the same check and double-spend.

**Fix:** Acquire `BEGIN IMMEDIATE` transaction before the balance check for deposit requests, ensuring the check-and-insert is atomic. On insufficient balance, explicit `db_conn.rollback()` is called before returning the error.

## Bug #5235 - Already Fixed

The `own` variable ordering in `node/utxo_db.py:apply_transaction()` was corrected in commit `c6a1a8d` (`#4233`). The line `own = conn is None` now appears before its first use.

Fixes #5235
Fixes #5236

### DIFF
--- a/node/bridge_api.py
+++ b/node/bridge_api.py
@@ -310,14 +310,17 @@ def create_bridge_transfer(
         unlock_at = now + (6 * 600)  # 6 slots = 1 hour
     
     try:
-        # For deposits, check balance and create lock
+        # FIX(#5236): Acquire IMMEDIATE transaction before balance check to
+        # prevent TOCTOU race between check_miner_balance() and the INSERT.
         if request.direction == "deposit" and not admin_initiated:
+            cursor.execute("BEGIN IMMEDIATE")
             has_balance, available, pending = check_miner_balance(
                 db_conn, 
                 request.source_address, 
                 amount_i64
             )
             if not has_balance:
+                db_conn.rollback()
                 return False, {
                     "error": "Insufficient available balance",
                     "available_rtc": available / BRIDGE_UNIT,


TJCurnutte (approved):
Reviewed the bridge-deposit TOCTOU fix. The important part is that the non-admin deposit path now takes the SQLite write lock before reading available balance, then keeps the balance check, bridge transfer insert, and lock-ledger insert in the same transaction.

Validation I ran:

```bash
python3 -B -m py_compile node/bridge_api.py
```

Then I ran a focused SQLite trace probe against a temp DB with `balances`, `bridge_transfers` from `init_bridge_schema()`, and a minimal `lock_ledger` table:

```text
success_ok=True status=pending id=1
rows={'bridge_transfers': 1, 'lock_ledger': 1}
trace: BEGIN IMMEDIATE -> SELECT amount_i64 FROM balances -> INSERT bridge_transfers -> INSERT lock_ledger -> COMMIT
```

I also checked the insufficient-balance path:

```text
insufficient_ok=False error=Insufficient available balance
post_insufficient_rows={'bridge_transfers': 1, 'lock_ledger': 1}
trace: BEGIN IMMEDIATE -> SELECT amount_i64 FROM balances -> ROLLBACK
```

That gives the intended serialization boundary for the check-and-insert path and keeps the failure path from leaving partial rows.

iamdinhthuan (changes requested):
Requesting changes because this PR bundles many unrelated changes into a bridge TOCTOU fix.

The PR title/body scope is `fix(#5236): TOCTOU race condition in bridge deposits`, with the described fix in `node/bridge_api.py:create_bridge_transfer()`. That hunk is present. However, the diff also changes or adds a broad set of unrelated files:

- `.github/workflows/welcome.yml`
- `bounties/issue-2296/src/cross_node_replay_defense.py` and its tests
- `node/rustchain_p2p_sync.py` plus a new `node/tests/test_p2p_sync_routes.py`
- `node/utxo_endpoints.py` and `node/test_utxo_endpoints.py`
- `tools/bounty-bot-pro/verifier.py` and tests
- `tools/node-health-cli/tests/test_node_health.py`
- new helper tests under `tools/tests/*`
- `tests/test_multi_arch_oracles_hmac.py`
- `tools/wrtc-bridge-dashboard/test_bridge_dashboard.py`
- `docs/CLAIMS_GUIDE.md`

Several of those are behavior changes outside the bridge deposit race, for example `cross_node_replay_defense.py` changes nonce storage from `INSERT OR REPLACE` to `INSERT OR IGNORE`, and `utxo_endpoints.py` changes request JSON parsing/validation. Those need their own PRs or at least their own issue linkage and review scope.

Please split this so the #5236 bridge TOCTOU fix can be reviewed and merged independently from unrelated workflow, P2P, UTXO, bounty-bot, and helper-test changes.

Evidence checked: `git diff --name-status origin/main...HEAD`; `node/bridge_api.py`; `.github/workflows/welcome.yml`; `bounties/issue-2296/src/cross_node_replay_defense.py`; `node/utxo_endpoints.py`; `tools/bounty-bot-pro/verifier.py`.

HCIE2054 (approved):

PR #5256, opened 2026-05-14
ci: add missing test dependencies

iamdinhthuan:
## Summary
- Add missing CI/runtime test dependencies for faucet wallet validation and PSE report analysis.
- Fix the runtime regressions exposed once CI could collect and run the full test suite: stale security-audit source windows, Beacon Atlas signed-auth reject handling, issue-2310 platform paths, and wallet-review UI session/auth flows.
- Keep the changes scoped to CI/test unblockers and the production branches directly exercised by the failing tests.

## Root Cause
The shared `test` workflow installs `requirements.txt` before running `pytest tests/`, but `requirements.txt` did not include modules imported during test collection. After those imports were restored, the full job exposed existing runtime/test drift in four areas that had not been reached by CI.

## Validation
- `RC_ADMIN_KEY=0123456789abcdef0123456789abcdef RC_P2P_SECRET=ci-test-secret-00000000000000000000000000000000 DB_PATH=:memory: PYTHONDONTWRITEBYTECODE=1 /tmp/rustchain-ci-deps-venv/bin/python -m pytest tests/ -q --ignore=tests/test_epoch_settlement_formal.py --ignore=tests/test_rip201_bucket_spoof.py` -> 1920 passed, 18 skipped, 73 subtests passed.
- Focused regression set for the original 7 CI failures -> 7 passed.
- `tests/test_wallet_review_holds.py` -> 5 passed.
- `py_compile` for touched Python files and `git diff --check` -> passed.
- GitHub PR checks on commit `3f67d12` -> all pass, including `test` and `BCOS v2 Engine Scan`.

### DIFF
--- a/node/beacon_api.py
+++ b/node/beacon_api.py
@@ -661,7 +661,7 @@ def update_contract(contract_id):
                     'error': 'Only the recipient (to_agent) can accept this contract'
                 }), 403
         if current_state == 'offered' and new_state == 'rejected':
-            if agent_key != to_agent:
+            if caller_agent != to_agent:
                 return jsonify({
                     'error': 'Only the recipient (to_agent) can reject this contract'
                 }), 403
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -4360,6 +4360,7 @@ def _get_or_create_admin_session(req):
     sid = req.values.get("session_id", "")
     if sid and sid in _ADMIN_SESSIONS:
         _ADMIN_SESSIONS[sid] = now  # refresh TTL
+        req._admin_session_id = sid  # type: ignore
         return True
     # Check header auth
     if is_admin(req):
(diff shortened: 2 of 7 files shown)


TJCurnutte (changes requested):
Thanks for tightening the shared test dependency list. I found one blocker before this lands.

`requirements.txt` is the top-level development dependency file, but the repository still advertises Python 3.9 support (`rustchainnode/pyproject.toml` has `requires-python = ">=3.9"`). The new `matplotlib>=3.10.0` requirement makes the whole file fail to resolve on Python 3.9 because Matplotlib 3.10+ requires Python >=3.10.

Validation I ran:

```bash
python3 --version
python3 -m pip install --dry-run -r requirements.txt
```

Result:

```text
Python 3.9.6
ERROR: Ignored the following versions that require a different python version: 3.10.0 Requires-Python >=3.10 ...
ERROR: Could not find a version that satisfies the requirement matplotlib>=3.10.0
ERROR: No matching distribution found for matplotlib>=3.10.0
```

I also scanned the tree and confirmed the added dependencies map to real imports (`flask_cors`, `Crypto`, `matplotlib`, and `seaborn` are all referenced), so the direction is useful. The compatibility break is just in the version constraint.

A safe fix would be to either add Python-version markers for matplotlib (for example a 3.9-compatible matplotlib bound on `python_version < "3.10"` and `matplotlib>=3.10.0` only on newer Python), raise/scope the dev Python requirement explicitly, or split this into a CI-only requirements file for the Python 3.11 jobs.

HCIE2054 (approved):

PR #5267, opened 2026-05-15
Fix API walkthrough node reference

Thanhdn1984:
## Summary\n- Clarify the node API reference in API_WALKTHROUGH.md so it points to a working endpoint instead of the server root.\n\n## Verification\n- Checked https://50.28.86.131 returns 404.\n- Checked https://50.28.86.131/health returns 200.\n\nBounty context: Scottcjn/rustchain-bounties#444\nWallet: RTC802c0ff127cdc1b0b82d40ccb1bab5519198bd17

### DIFF
--- a/API_WALKTHROUGH.md
+++ b/API_WALKTHROUGH.md
@@ -130,7 +130,7 @@ print(response.json())
 
 ## Reference
 
-- **Node:** `https://50.28.86.131`
+- **Node API:** `https://50.28.86.131/health`
 - **Explorer:** `https://50.28.86.131/explorer`
 - **Health:** `https://50.28.86.131/health`
 
--- a/API_WALKTHROUGH.md
+++ b/API_WALKTHROUGH.md
@@ -130,7 +130,7 @@ print(response.json())
 
 ## Reference
 
-- **Node API:** `https://50.28.86.131/health`
+- **Node base URL:** `https://50.28.86.131`
 - **Explorer:** `https://50.28.86.131/explorer`
 - **Health:** `https://50.28.86.131/health`
 


asdf00915:
This fixes the symptom that the bare origin returns 404, but it removes the only base node URL from the reference section and duplicates the `Health` row below it.

The walkthrough examples above still use the same origin with different API paths:

- `https://50.28.86.131/epoch`
- `https://50.28.86.131/wallet/balance`
- `https://50.28.86.131/wallet/transfer/signed`

Those endpoints are separate from `/health`, and `/health` is already listed as `Health`. A better fix would keep the node origin as the API base, for example `Node base URL: https://50.28.86.131`, and optionally note that the root path itself is not an endpoint.

minyanyi (approved):
Reviewed `API_WALKTHROUGH.md`.

Observations:
1. Pointing the reference at `/health` instead of the server root matches the actual behavior described in the PR body: the bare node URL returns 404, while `/health` gives a stable 200 endpoint readers can probe.
2. Renaming the bullet to `Node API` is also clearer than `Node`, because it signals this is a concrete API endpoint rather than the general server hostname.

LGTM / Approved.

Bounty: #2782

TJCurnutte (approved):
Verified the final docs change.

Validation run against `e9d0c707d6b39f566f31cd9bda9a688f37c323bd`:

```bash
git diff --check origin/main...HEAD -- API_WALKTHROUGH.md
python3 - <<'PY'
from pathlib import Path
text = Path('API_WALKTHROUGH.md').read_text()
assert '**Node base URL:** `https://50.28.86.131`' in text
assert '**Health:** `https://50.28.86.131/health`' in text
assert '**Explorer:** `https://50.28.86.131/explorer`' in text
PY
curl -k -L --max-time 15 https://50.28.86.131/health
curl -k -L --max-time 15 https://50.28.86.131/explorer
```

Results: diff check passed. The reference section now separates the base node URL from the health endpoint, and the live probes returned `HTTP 200` for `/health` with `{"ok":true,...}` and `HTTP 200` for `/explorer/` with the explorer HTML. The root base URL currently returns a Flask 404, so labeling it as a base URL instead of an endpoint is the clearer/correct form.

Approved.

HCIE2054 (approved):

PR #5278, opened 2026-05-15
fix: reject invalid UTXO transaction types

galanime:
## Summary
- validate `tx_type` before UTXO transaction application
- reject explicit non-string or empty transaction types in `apply_transaction()` and `mempool_add()`
- preserve the existing default of `transfer` when `tx_type` is omitted

## Security impact
`mempool_add()` accepted a transaction with `tx_type: null`. That transaction could lock an input in `utxo_mempool_inputs` and be returned to block producers as a candidate, but `apply_transaction()` failed later with `sqlite3.IntegrityError` when writing `utxo_transactions.tx_type TEXT NOT NULL`. This created an invalid block-candidate / mempool DoS path until expiry.

This is distinct from a generic unknown-string whitelist concern: the regression covers a non-persistable `tx_type` value that currently causes a concrete persistence failure after mempool admission.

## Reproduction before fix
```text
mempool_add True
candidates 1
apply_exception IntegrityError NOT NULL constraint failed: utxo_transactions.tx_type
candidate_still_present 1
```

## Verification
- `PYTHONPATH=node /tmp/rustchain-bounty-venv/bin/python -m pytest node/test_utxo_db.py -q -k 'invalid_tx_type'` -> 2 passed after fix; failed before fix
- `PYTHONPATH=node /tmp/rustchain-bounty-venv/bin/python -m pytest node/test_utxo_db.py -q` -> 69 passed
- `PYTHONPATH=node /tmp/rustchain-bounty-venv/bin/python -m py_compile node/utxo_db.py node/test_utxo_db.py`
- `git diff --check -- node/utxo_db.py node/test_utxo_db.py`

## Bounty
Claiming under Scottcjn/rustchain-bounties#2819 as Medium severity: mempool DoS / invalid block-candidate manipulation.

### DIFF
--- a/node/test_utxo_db.py
+++ b/node/test_utxo_db.py
@@ -845,6 +845,25 @@ class TestUtxoDB(unittest.TestCase):
             self.db.mempool_check_double_spend(boxes[0]['box_id'])
         )
 
+    def test_mempool_rejects_invalid_tx_type(self):
+        """Mempool must not admit txs with non-persistable tx_type values."""
+        self._apply_coinbase('alice', 100 * UNIT)
+        boxes = self.db.get_unspent_for_address('alice')
+
+        tx = {
+            'tx_id': 'badt' * 16,
+            'tx_type': None,
+            'inputs': [{'box_id': boxes[0]['box_id']}],
+            'outputs': [{'address': 'bob', 'value_nrtc': 99 * UNIT}],
+            'fee_nrtc': 1 * UNIT,
+        }
+        ok = self.db.mempool_add(tx)
+        self.assertFalse(ok)
+        self.assertEqual(self.db.mempool_get_block_candidates(), [])
+        self.assertFalse(
+            self.db.mempool_check_double_spend(boxes[0]['box_id'])
+        )
+
     def test_mempool_accepts_valid_tx(self):
         """Mempool should accept a well-formed tx with valid conservation."""
         self._apply_coinbase('alice', 100 * UNIT)
@@ -1043,6 +1062,23 @@ class TestUtxoDB(unittest.TestCase):
             'fee_nrtc': 0,
         }, block_height=10)
         self.assertFalse(ok)
+
+    def test_invalid_tx_type_rejected(self):
+        """tx_type must be a non-empty string before transaction history write."""
+        self._apply_coinbase('alice', 100 * UNIT)
+        boxes = self.db.get_unspent_for_address('alice')
+
+        ok = self.db.apply_transaction({
+            'tx_type': None,
+            'inputs': [{'box_id': boxes[0]['box_id'],
+                         'spending_proof': 'sig'}],
+            'outputs': [{'address': 'bob', 'value_nrtc': 99 * UNIT}],
+            'fee_nrtc': 1 * UNIT,
+        }, block_height=10)
+
+        self.assertFalse(ok)
+        self.assertEqual(self.db.get_balance('alice'), 100 * UNIT)
+        self.assertEqual(self.db.get_balance('bob'), 0)
 
 
 class TestCoinSelect(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


jaxint (approved):

HunterCML (commented):
Code review notes from a local checkout of this PR.

Verification performed on head `87d3d9b`:
- `git diff --check main...HEAD` passed
- `python -B -m py_compile node/utxo_db.py node/test_utxo_db.py` passed
- Targeted unittest cases for the new coverage passed
- Full `python -B -m unittest node.test_utxo_db -v` passed: `69 tests`

The implementation fixes the concrete persistence failure mode: `None` and other non-string `tx_type` values now get rejected before either `apply_transaction()` records `utxo_transactions.tx_type` or `mempool_add()` serializes the transaction into block-candidate state. I also checked that the valid missing-`tx_type` legacy/default path still maps to `transfer`, and existing mining-reward guards remain in place for both direct application and public mempool admission.

One follow-up I would consider before merge: add explicit regression coverage for `tx_type: "   "`, `tx_type: []`, and an intentionally unknown non-empty string such as `"not_a_protocol_type"`. The first two should be rejected by this PR. The last one documents the intended boundary: if arbitrary non-empty strings are allowed for forward compatibility, that should be intentional; if the title means only known protocol types are valid, `_normalize_tx_type()` still needs an allowlist beyond type/emptiness validation.

Given the existing behavior and passing full UTXO suite, this looks mergeable to me with that coverage/scope clarification as a low-risk follow-up.

TJCurnutte (approved):
Approved — focused UTXO validation pass.

I checked branch head `723a93b1a622eb0e3d49f7ff34a2f9d15244b556` against the live PR metadata. The change normalizes `tx_type` in both direct `apply_transaction()` and public `mempool_add()`, defaulting only when the field is absent and rejecting non-string / blank values before transaction-history writes or mempool input claims. The added tests cover the important failure mode: an invalid `tx_type` is rejected and the spent input is not left locked in the mempool.

Validation run locally:

```bash
git diff --check origin/main...HEAD -- node/utxo_db.py node/test_utxo_db.py
PYTHONPATH=node python3 -B -m py_compile node/utxo_db.py node/test_utxo_db.py
PYTHONPATH=node PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -B -m pytest node/test_utxo_db.py -q -k 'invalid_tx_type or mempool_rejects_invalid_tx_type'
PYTHONPATH=node PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python3 -B -m pytest node/test_utxo_db.py -q
```

Results: diff check passed, py_compile passed, focused invalid-type tests passed with `2 passed, 67 deselected`, and the full UTXO DB suite passed with `69 passed`.

From this diff-scoped review, the validation path looks safe.

JaBi-aka:
Reviewed the invalid `tx_type` fix. This patch addresses the concrete persistence failure described in the PR and keeps the current protocol default intact:

- `_normalize_tx_type()` only defaults to `transfer` when the key is absent, so explicitly invalid values such as `null`, empty strings, and whitespace-only strings are rejected instead of being silently persisted.
- Both `apply_transaction()` and `mempool_add()` call the same normalizer, which is the right shape here because mempool admission must not accept candidates that block production later.
- The later defense-in-depth `mining_reward` guard still works after stripping, so values like `" mining_reward "` do not bypass the existing public-mempool restriction.
- Regression coverage checks both direct transaction application and mempool admission, including that the input lock is not left behind for rejected transactions.

No blocking findings from this review.

Small follow-up suggestion: consider adding a test for a whitespace-padded disallowed type such as `" mining_reward "` to explicitly lock in the interaction between normalization and the existing mining-reward guard. I would treat that as extra coverage, not a merge blocker.

HCIE2054 (approved):

PR #5280, opened 2026-05-15
test: cover download page handler

iamdinhthuan:
Adds focused unit coverage for the standalone RustChain miner download page handler.

Coverage:
- `/` and `/index.html` return the embedded download HTML.
- Path traversal requests are rejected with 403 before file access.
- Python miner downloads use text/plain plus attachment headers.
- Missing downloads return 404.

Verification:
- `PYTHONDONTWRITEBYTECODE=1 PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -p no:cacheprovider tests/test_rustchain_download_page.py -q` -> 5 passed
- `python -m py_compile node/rustchain_download_page.py tests/test_rustchain_download_page.py` -> passed

### DIFF
--- a/tests/test_rustchain_download_page.py
+++ b/tests/test_rustchain_download_page.py
@@ -1,0 +1,90 @@
+import importlib.util
+import io
+from pathlib import Path
+
+
+MODULE_PATH = Path(__file__).resolve().parents[1] / "node" / "rustchain_download_page.py"
+
+
+def load_download_page_module():
+    spec = importlib.util.spec_from_file_location("rustchain_download_page_under_test", MODULE_PATH)
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    return module
+
+
+def make_handler(module, path):
+    handler = module.DownloadHandler.__new__(module.DownloadHandler)
+    handler.path = path
+    handler.wfile = io.BytesIO()
+    handler.responses = []
+    handler.headers = []
+    handler.errors = []
+
+    handler.send_response = lambda code: handler.responses.append(code)
+    handler.send_header = lambda name, value: handler.headers.append((name, value))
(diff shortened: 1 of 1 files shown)


minyanyi (changes requested):
Thanks for adding coverage here. I reproduced a cross-platform failure in 	est_download_route_serves_python_file_with_attachment_headers on Windows: Path.write_text() writes CRLF by default, but the assertion currently expects raw bytes ending in \n. The download handler is correctly serving the file bytes it reads back, so the brittle part is the fixture/assertion rather than the implementation. Please normalize the fixture (for example write_text(..., newline='\n')) or compare decoded text so this stays portable across Windows and Unix runners.

TJCurnutte (approved):
Diff-scoped approval after focused local validation.

Validated head: `fed5fc01b317c49ca206d5bb7e0428dd4e298dec`.

Validation run:
- `git diff --check origin/main...HEAD -- tests/test_rustchain_download_page.py`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/rustchain-review-venv/bin/python -B -m py_compile node/rustchain_download_page.py tests/test_rustchain_download_page.py`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/rustchain-review-venv/bin/python -B -m pytest --noconftest -p no:cacheprovider tests/test_rustchain_download_page.py -q` → `5 passed in 0.05s`
- Source-binding probe confirmed the test imports the real `node/rustchain_download_page.py` with `spec_from_file_location`, drives `DownloadHandler.do_GET()` directly, and covers index rendering, attachment headers, 404 handling, and `../` traversal rejection.

This is a good focused test addition around the static download handler. The temp-directory override keeps file-serving assertions isolated, and the traversal case verifies the intended fail-closed branch. I do not see a blocker in this diff.

HCIE2054 (approved):

PR #5281, opened 2026-05-15
test: cover download server routes

iamdinhthuan:
Adds focused unit coverage for the Flask RustChain miner download server.

Coverage:
- Index route renders the embedded miner download page.
- `/downloads/<filename>` serves existing files as attachments from the configured download directory.
- Missing files return 404.
- Parent-directory paths do not expose files outside the download directory.

Verification:
- `PYTHONDONTWRITEBYTECODE=1 PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -p no:cacheprovider tests/test_rustchain_download_server.py -q` -> 4 passed
- `python -m py_compile node/rustchain_download_server.py tests/test_rustchain_download_server.py` -> passed

### DIFF
--- a/tests/test_rustchain_download_server.py
+++ b/tests/test_rustchain_download_server.py
@@ -1,0 +1,62 @@
+import importlib.util
+from pathlib import Path
+
+
+MODULE_PATH = Path(__file__).resolve().parents[1] / "node" / "rustchain_download_server.py"
+
+
+def load_download_server_module():
+    spec = importlib.util.spec_from_file_location("rustchain_download_server_under_test", MODULE_PATH)
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    module.app.config.update(TESTING=True)
+    return module
+
+
+def test_index_route_renders_download_page():
+    module = load_download_server_module()
+
+    response = module.app.test_client().get("/")
+
+    assert response.status_code == 200
+    body = response.get_data(as_text=True)
+    assert "RustChain Miner Downloads" in body
+    assert "/downloads/rustchain_miners_v2.2.1.zip" in body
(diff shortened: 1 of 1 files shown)


HunterCML (changes requested):
Code review notes from a local Windows checkout of this PR.

Verification performed on head `e11a7a7`:
- `python -B -m py_compile node/rustchain_download_server.py tests/test_rustchain_download_server.py` passed
- `git diff --check main...HEAD` passed
- `python -B -m pytest -p no:cacheprovider tests/test_rustchain_download_server.py -q` failed: `1 failed, 3 passed`

The failure is in `test_download_route_serves_existing_file_as_attachment()`. On Windows, `Path.write_text(print('mine')\n)` writes through text-mode newline translation, so the temp file is served back as `bprint('mine')\r\n`. The test asserts the response is exactly `bprint('mine')\n`, so this is platform-sensitive even though the route behavior itself is fine.

A small fix would make this portable: either write bytes with `miner.write_bytes(bprint('mine')\n)`, pass `newline=\n` to `write_text()`, or compare the response after normalizing newlines if exact byte preservation is not the behavior being tested. Since this PR is adding route coverage, I would prefer `write_bytes()` because it keeps the fixture content explicit and avoids depending on host text-mode behavior.

The rest of the coverage is useful. The test correctly patches `module.DOWNLOAD_DIR` instead of using the production `/root/rustchain/downloads` path, checks attachment headers on the happy path, verifies missing files return 404, and includes a parent-directory traversal attempt so `send_from_directory()` path confinement remains covered.

Requesting changes only for the newline-sensitive fixture assertion; the route coverage shape looks good after that is fixed.

Bounty: Scottcjn/rustchain-bounties#73

HCIE2054 (approved):

PR #5338, opened 2026-05-15
Fix: Payout orphan recovery & Airdrop TOCTOU race (Findings #2, #3, #7)

Guzzzzzzzz:
This PR implements the fix-forward recommendations for the following security audit findings:

- **Finding #2 & #3 (payout_worker)**: Non-atomic deduct/refund and orphan processing recovery. Added `recover_orphans()` to safely handle `processing` states after a crash to prevent permanently stranded funds.
- **Finding #7 (airdrop_v2)**: Airdrop allocation TOCTOU race. Modifies the allocation deduction logic to be atomic using a single SQL UPDATE with a WHERE constraint checking the remaining reward balance.

Wallet for any rewards: `RTC241359b3438d1222fdc1c3e22fe980657a4bc54e`

### DIFF
--- a/node/airdrop_v2.py
+++ b/node/airdrop_v2.py
@@ -833,15 +833,19 @@ class AirdropV2:
                 ),
             )
 
-            # Update allocation
+            # Update allocation atomically to prevent TOCTOU race
             cursor.execute(
                 """
                 UPDATE airdrop_allocation
                 SET claimed_uwrtc = claimed_uwrtc + ?, updated_at = ?
-                WHERE chain = ?
+                WHERE chain = ? AND (total_uwrtc - claimed_uwrtc) >= ?
                 """,
-                (claim.amount_uwrtc, claim.timestamp, chain_lower),
+                (claim.amount_uwrtc, claim.timestamp, chain_lower, claim.amount_uwrtc),
             )
+
+            if cursor.rowcount == 0:
+                conn.rollback()
+                return False, "Airdrop allocation exhausted or concurrent claim conflict", None
 
             conn.commit()
             logger.info(
--- a/node/payout_worker.py
+++ b/node/payout_worker.py
@@ -174,6 +174,42 @@ class PayoutWorker:
             self.stats['failed'] += 1
             return False
 
+    def recover_orphans(self):
+        """Recover withdrawals stuck in processing state (e.g. from a crash during execute_withdrawal)."""
+        try:
+            with sqlite3.connect(self.db_path) as conn:
+                conn.execute("BEGIN IMMEDIATE")
+                rows = conn.execute("""
+                    SELECT withdrawal_id, miner_pk, amount, fee
+                    FROM withdrawals
+                    WHERE status = 'processing'
+                """).fetchall()
+
+                for row in rows:
+                    withdrawal_id, miner_pk, amount, fee = row
+                    total_refund = amount + (fee or 0)
+
+                    logger.warning(f"Recovering orphaned withdrawal {withdrawal_id} (refunding {total_refund} to {miner_pk})")
+
+                    # Refund the balance
+                    conn.execute(
+                        "UPDATE accounts SET balance = balance + ? WHERE public_key = ?",
+                        (total_refund, miner_pk)
+                    )
+
+                    # Mark as failed
+                    conn.execute(
+                        "UPDATE withdrawals SET status = 'failed', error_msg = 'Orphaned processing state recovered' WHERE withdrawal_id = ?",
+                        (withdrawal_id,)
+                    )
+                conn.execute("COMMIT")
+                
+                if rows:
+                    logger.info(f"Recovered {len(rows)} orphaned withdrawals.")
+                    
+        except Exception as e:
+            logger.error(f"Failed to recover orphans: {e}")
+
     def process_batch(self) -> int:
         """Process a batch of withdrawals"""
         withdrawals = self.get_pending_withdrawals()
@@ -203,6 +239,9 @@ class PayoutWorker:
 
         while True:
             try:
+                # Recover orphans before processing new batches to prevent stranded funds
+                self.recover_orphans()
+
                 # Process batch
                 processed = self.process_batch()
 


TJCurnutte (changes requested):
Requesting changes — the airdrop allocation guard is moving in the right direction, but the payout orphan recovery is unsafe as written.

I checked out head `8e0d1cd271874af1d99b2c605900b86fa27122ec` and ran:

- `git diff --check origin/main...HEAD -- node/airdrop_v2.py node/payout_worker.py` → fails on trailing whitespace at `node/payout_worker.py:206` and `node/payout_worker.py:209`.
- `python3 -B -m py_compile node/airdrop_v2.py node/payout_worker.py` → passed.
- `PYTHONPATH=. python3 -B -m pytest --noconftest node/test_airdrop_v2.py -q` → `23 passed, 1 warning in 0.34s`.
- A temp-db airdrop capacity probe accepted the first 25 wRTC Solana claim, rejected the second, and left `claimed=25000000` with `claim_count=1`.
- A temp-db payout probe inserted a withdrawal with `status='processing'` and `created_at` equal to the current timestamp, then called `PayoutWorker.recover_orphans()`; it immediately changed the miner balance from `100.0` to `111.0` and marked the row `failed` with `Orphaned processing state recovered`.

The blocking issue is `recover_orphans()` selecting every `WHERE status = 'processing'` row and refunding it immediately. There is no lease, `processing_started_at`, age cutoff, worker ownership check, tx hash/broadcast marker, or confirmation reconciliation before the refund. A freshly in-flight withdrawal from another worker/restarted process can therefore be refunded locally while the broadcast is still being submitted or confirming, which is the double-credit/lost-accounting path this code is supposed to avoid.

Please make orphan recovery distinguish stale/abandoned processing rows from active or broadcast-uncertain rows before refunding, and clean up the whitespace reported by `git diff --check`.

jaxint (approved):

denerbarbosa (commented):
Good direction on the atomic airdrop update and orphan recovery. One concern: recover_orphans() refunds *every* withdrawal still marked 'processing' as soon as the worker restarts. If a future implementation crashes after broadcasting but before the DB flips the row to completed, this path will mark a possibly already-broadcast withdrawal as failed and refund the miner, which opens a double-spend window.\n\nSuggestion: restrict recovery to rows that are clearly pre-broadcast (for example, by tracking a separate 'broadcast_at' / 'tx_hash' null marker or adding a grace period), and add a regression test for the stale-processing case.

denerbarbosa (commented):
Security review: `recover_orphans()` is useful, but I think this is currently unsafe to run unconditionally on every loop start.

Blocking concern:
- Any withdrawal still in `processing` gets refunded immediately on startup, even if the broadcast actually succeeded but the worker crashed before marking it completed. That can create a double-payment / false-refund path.

I think this needs at least one of:
- an age threshold / grace period before recovery,
- a separate stale-processing state, or
- a tx status check against the chain before refunding.

The airdrop TOCTOU change looks good. The refund logic in `process_withdrawal()` is still atomic, but the new orphan recovery path needs a guard before this can be considered safe for payouts.

denerbarbosa:
PR review follow-up for bounty tracking:

- Review URL: https://github.com/Scottcjn/Rustchain/pull/5338#pullrequestreview-4302115900
- Wallet: `RTCe9d413fd6417751ff3cb226365e09142dad11c2a`
- Finding: `recover_orphans()` needs a guard before refunding `processing` withdrawals on restart, otherwise an in-flight broadcast can be refunded and later confirmed on-chain, causing double-payment risk.

If this should instead be tracked against a security bounty issue, point me to the correct issue and I will submit the report there as well.

denerbarbosa (changes requested):
The airdrop allocation change looks good, but I can't approve the payout_worker side as written.

`recover_orphans()` refunds every row in `status = 'processing'` immediately on startup, with no grace period, no chain-status check, and no recovery state. That is still a double-payment path: `process_withdrawal()` intentionally flips to `processing` before broadcast, so a normal restart during broadcast latency will now refund an already-broadcast withdrawal and later let the on-chain transfer complete.

Please gate orphan recovery behind either:
- a stale-age threshold, or
- an explicit `recovery_pending` state plus confirmation from chain status before refunding.

Without that, the fix replaces a stranded-funds risk with a fund-loss risk.

denerbarbosa:
Follow-up finding on the airdrop side: claim identity is still case-sensitive. `_has_claimed()` compares raw `github_username` strings, and the API only `strip()`s input before storing it. Because GitHub usernames are case-insensitive, a claimant can resubmit the same account with different casing plus a new wallet and bypass duplicate detection / the UNIQUE constraint. Please normalize usernames before storage/comparison and add a regression test for mixed-case reclaims.

HCIE2054 (approved):

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

508704820 (commented):
Payout orphan recovery and Airdrop TOCTOU race. CRITICAL - TOCTOU in payout processing allows double-claim. Verify: (1) Race condition is fixed with proper locking or atomic operations. (2) Orphan recovery does not create new attack surface. (3) Airdrop claim checks are atomic - check and debit in same transaction. - Xeophon (security review, TOCTOU)

508704820:
Security Review ✅

Fixes our original audit findings: (1) Non-atomic deduct/refund in payout_worker — orphan processing recovery prevents permanently stranded funds. (2) Airdrop TOCTOU race — atomic allocation with WHERE constraint checking remaining balance prevents double-claim. Good implementation of fix-forward recommendations.

kekehanshujun (changes requested):
Reviewing this as part of the RustChain code review bounty.

The airdrop allocation update is a good atomic TOCTOU fix, but the payout orphan recovery is unsafe as written. `recover_orphans()` refunds every withdrawal in `processing` state before each batch, without proving that the external transfer was never broadcast. A worker crash after `execute_withdrawal()` sends funds but before the DB status is updated would leave the row as `processing`; on restart this code refunds `amount + fee` to the account, creating a double-payment path.

Please do not automatically refund all `processing` withdrawals. Recovery needs an additional durable state boundary, such as a broadcast/pending transaction id recorded before the external call, an age threshold plus operator review, or a reconciliation check against the RustChain transfer endpoint. At minimum, add a regression for crash-after-broadcast-before-status-update and show that funds are not refunded automatically.

Until that is addressed, this PR can turn the orphan-recovery bug into a fund-duplication bug.

PR #5360, opened 2026-05-16
fix: avoid refund after withdrawal broadcast hash

### LINKED ISSUE #5346: [SECURITY] payout_worker orphan recovery can refund in-flight withdrawals on restart
## Summary
`recover_orphans()` in `node/payout_worker.py` refunds every withdrawal still marked `processing` as soon as the worker starts.

## Impact
If a withdrawal was actually broadcast successfully, but the worker crashed before updating the row to `completed`, a restart would refund the user and also allow the original on-chain payment to complete. That creates a double-payment / fund-loss path.

## Reproduction sketch
1. Start a withdrawal and force a crash after `execute_withdrawal()` returns or during the completion update.
2. Restart the worker.
3. `recover_orphans()` sees the row in `processing` and refunds it immediately without checking age or chain status.

## Suggested fix
- Require a grace period before recovery, or
- confirm transaction status on-chain before refunding, or
- move stale processing rows into an explicit recovery state first.

Reported while tracking bounty work for RustChain.

MizuCiaossu:
## Summary

Fixes a withdrawal recovery safety gap related to #5346. If `execute_withdrawal()` returns a transaction hash and the later database completion update fails, the worker must not refund the balance because the payout may already have been broadcast.

## Changes

- Track whether a broadcast `tx_hash` has been returned inside `process_withdrawal()`.
- If an error occurs after a `tx_hash` exists, keep the withdrawal in `processing`, record the `tx_hash` and reconciliation error, and do not credit funds back to the account.
- Preserve the existing refund path for failures before any transaction hash exists.
- Add a regression test that simulates a completion-update failure after broadcast and verifies the balance stays debited.

## Validation

- `py -3.14 -B -m py_compile node/payout_worker.py tests/test_payout_worker_production_noop.py`
- `py -3.14 -m pytest tests/test_payout_worker_production_noop.py -q` -> 3 passed
- `git diff --check`

Wallet: RTC3cb17c1f0b2b97c9c120e306f9a4c0d2da6b0877

Fixes #5346

### DIFF
--- a/node/payout_worker.py
+++ b/node/payout_worker.py
@@ -59,6 +59,34 @@ class PayoutWorker:
 
             return withdrawals
 
+    def _record_broadcast_reconciliation_needed(
+        self,
+        withdrawal_id: str,
+        tx_hash: str,
+        error: str,
+    ) -> None:
+        """Keep broadcast withdrawals out of the refund path after DB failures."""
+        message = (
+            "Broadcast returned transaction hash but completion update failed; "
+            f"manual reconciliation required: {error}"
+        )
+        try:
+            with sqlite3.connect(self.db_path) as conn:
+                conn.execute("""
+                    UPDATE withdrawals
+                    SET status = 'processing',
+                        tx_hash = ?,
+                        error_msg = ?
+                    WHERE withdrawal_id = ?
+                """, (tx_hash, message, withdrawal_id))
+        except Exception as record_error:
+            logger.error(
+                "Failed to record reconciliation state for %s (%s): %s",
+                withdrawal_id,
+                tx_hash,
+                record_error,
+            )
+
     def execute_withdrawal(self, withdrawal: Dict) -> Optional[str]:
         """Execute withdrawal transaction"""
         if MOCK_MODE:
@@ -90,6 +118,7 @@ class PayoutWorker:
     def process_withdrawal(self, withdrawal: Dict) -> bool:
         """Process a single withdrawal with balance deduction before execution."""
         withdrawal_id = withdrawal['withdrawal_id']
+        tx_hash = None
 
         try:
             logger.info(f"Processing withdrawal {withdrawal_id}")
@@ -176,6 +205,15 @@ class PayoutWorker:
 
         except Exception as e:
             logger.error(f"✗ Withdrawal {withdrawal_id} failed: {e}")
+
+            if tx_hash:
+                self._record_broadcast_reconciliation_needed(
+                    withdrawal_id,
+                    tx_hash,
+                    str(e),
+                )
+                self.stats['failed'] += 1
+                return False
 
             # Refund balance on broadcast failure and mark as failed
             with sqlite3.connect(self.db_path) as conn:
--- a/tests/test_payout_worker_production_noop.py
+++ b/tests/test_payout_worker_production_noop.py
@@ -71,3 +71,51 @@ def test_process_withdrawal_leaves_pending_when_production_broadcast_is_not_conf
     assert status == "pending"
     assert "not configured" in error_msg
     assert tx_hash is None
+
+
+def test_process_withdrawal_does_not_refund_after_broadcast_tx_hash(
+    tmp_path, monkeypatch
+):
+    class BroadcastThenCompletionUpdateFailsWorker(payout_worker.PayoutWorker):
+        def execute_withdrawal(self, withdrawal):
+            return "tx-broadcasted"
+
+    monkeypatch.setattr(payout_worker, "MOCK_MODE", True)
+    db_path = str(tmp_path / "payout_worker.db")
+    with sqlite3.connect(db_path) as conn:
+        conn.execute("CREATE TABLE accounts (public_key TEXT PRIMARY KEY, balance INTEGER)")
+        conn.execute(
+            "CREATE TABLE withdrawals ("
+            "withdrawal_id TEXT PRIMARY KEY, miner_pk TEXT, amount INTEGER, fee INTEGER, "
+            "destination TEXT, status TEXT, error_msg TEXT, tx_hash TEXT, created_at INTEGER)"
+        )
+        conn.execute(
+            "INSERT INTO accounts (public_key, balance) VALUES (?, ?)",
+            ("miner-pubkey", 100),
+        )
+        conn.execute(
+            "INSERT INTO withdrawals "
+            "(withdrawal_id, miner_pk, amount, fee, destination, status, created_at) "
+            "VALUES (?, ?, ?, ?, ?, ?, ?)",
+            ("wd-1", "miner-pubkey", 10, 1, "RTCdest", "pending", 1234567890),
+        )
+
+    worker = BroadcastThenCompletionUpdateFailsWorker()
+    worker.db_path = db_path
+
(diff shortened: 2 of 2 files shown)


likx0525:
Disclosure: I am reviewing this under Scottcjn/rustchain-bounties#73.

I walked through the post-broadcast failure path in `process_withdrawal()`. The core direction makes sense: once `execute_withdrawal()` has returned a transaction hash, refunding would risk crediting funds back after a broadcast. I think there are two issues worth tightening before merge:

1. In `node/payout_worker.py`, `_record_broadcast_reconciliation_needed()` updates the withdrawal solely by `withdrawal_id`. If the completion update fails after a broadcast but another actor has already reconciled, completed, or failed the row, this fallback can overwrite that state back to `processing` and replace the existing `tx_hash` / `error_msg`. Please guard the update with the expected state, for example `WHERE withdrawal_id = ? AND status = 'processing'`, and check `rowcount`; if zero, log or escalate instead of clobbering a terminal or manually reconciled state.

2. In `tests/test_payout_worker_production_noop.py`, the regression test makes the completion update fail by creating a fixture schema without `processed_at`. That covers the fallback branch, but it is exercising a schema mismatch rather than a realistic failure after a valid broadcast. Consider keeping the fixture schema aligned with production and injecting the failure after `execute_withdrawal()` returns, for example by monkeypatching the completion-update connection/cursor path. That would protect the intended behavior without relying on the table missing a production column.

Minor: after the fallback records a reconciliation-needed state, `self.stats['failed']` is incremented while the persisted withdrawal status remains `processing`. If any monitoring relies on these stats, a separate `needs_reconciliation` counter/log would make the state less ambiguous.

Masihhedayati (approved):
Reviewed head `c81a05b6e3c29aafd1d00da6cee2d04810ab9f7c` for the withdrawal post-broadcast failure path.

Validation I ran locally:

```bash
git diff --check origin/main...HEAD -- node/payout_worker.py tests/test_payout_worker_production_noop.py
/tmp/rustchain-review-venv/bin/python -B -m py_compile node/payout_worker.py tests/test_payout_worker_production_noop.py
/tmp/rustchain-review-venv/bin/python -B -m pytest -q tests/test_payout_worker_production_noop.py --noconftest
# 3 passed in 0.07s
```

I also ran a small positive-path probe with a schema that includes `processed_at`, a worker subclass returning `tx-review`, and a pending 10 RTC + 1 fee withdrawal. It returned `True`, left the account at 89, and stored `('completed', 'tx-review')`, so the new reconciliation branch does not break the normal completion path.

No blocker found. The important behavior is that once `execute_withdrawal()` has returned a transaction hash, the later exception handler no longer credits funds back to the account. Instead it preserves the debited balance, keeps the withdrawal in `processing`, stores the transaction hash, and records a manual-reconciliation error. The original pre-broadcast refund path remains intact because it is still used only when `tx_hash` is unset.

jaxint (approved):

TJCurnutte (approved):
Reviewed head `c81a05b6e3c29aafd1d00da6cee2d04810ab9f7c`.

This is the right safety direction for the payout worker: once `execute_withdrawal()` has returned a broadcast transaction hash, a later completion-write failure should not fall into the refund path. Keeping the row in `processing`, preserving the `tx_hash`, and recording a reconciliation message gives the operator a recoverable/manual state instead of risking a double spend/refund.

Validation I ran:

```bash
git diff --check origin/main...HEAD -- node/payout_worker.py tests/test_payout_worker_production_noop.py
python3 -B -m py_compile node/payout_worker.py tests/test_payout_worker_production_noop.py
PYTHONPATH=. python3 -B -m pytest -q tests/test_payout_worker_production_noop.py --noconftest
```

Result: `3 passed in 0.48s`.

I also ran a SQLite runtime probe that subclasses the worker so `execute_withdrawal()` returns `tx-broadcasted-probe`, then forces the completion update to fail by using a fixture table without `processed_at`. The post-failure state was:

- account balance stayed at `89` after the original `10 + 1` deduction;
- withdrawal status stayed `processing`;
- `tx_hash` was recorded as `tx-broadcasted-probe`;
- `error_msg` contained `manual reconciliation required`.

So the no-refund-after-broadcast path is covered and behaves as intended.

HCIE2054 (commented):
Reviewed PR 5360. Standard review.

HCIE2054 (approved):

HCIE2054 (approved):
LGTM!

BossChaos:
## Security Review

Good fix — preventing refunds after a withdrawal has been broadcast prevents double-spend attempts where an attacker could broadcast a transaction and then claim a refund if the state isn't properly locked.

The logic change appears to check whether a withdrawal hash exists before allowing a refund. This is the correct approach. LGTM.

ZacharyZhang-NY (changes requested):
Requesting one concurrency/state guard before merge.

The focused validation passes on the current head:

- `python -B -m py_compile node/payout_worker.py tests/test_payout_worker_production_noop.py`
- `python -B -m pytest -q tests/test_payout_worker_production_noop.py --noconftest` -> `3 passed`

The no-refund-after-broadcast direction is right, but `_record_broadcast_reconciliation_needed()` updates the withdrawal row by `withdrawal_id` only. That means the fallback can overwrite a row that has already been reconciled by another worker/operator between the failed completion update and the fallback write.

I verified this with a minimal SQLite probe on the current head: seed `withdrawals` with `('wd1', 'completed', 'tx-existing', 'already reconciled')`, call `_record_broadcast_reconciliation_needed('wd1', 'tx-new', ...)`, then read the row back. The row becomes:

```text
status: processing
tx_hash: tx-new
```

So a terminal/manual state can be moved back to `processing` and its existing transaction evidence can be replaced. Please scope the fallback update to the expected in-flight state, e.g. `WHERE withdrawal_id = ? AND status = 'processing'`, check `rowcount`, and log/escalate when no row is updated. A regression should cover the already-completed/already-reconciled row so the fallback cannot clobber terminal state.

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

BossChaos (commented):
## PR #5360 Review - Withdrawal broadcast reconciliation (double-spend prevention)

### Security Analysis - Financial Integrity

This PR fixes a critical edge case in withdrawal processing.

**Key improvements:**
1. **_record_broadcast_reconciliation_needed()**: When broadcast succeeds but DB update fails, marks as 'processing' instead of refunding
2. **Prevents double-spend**: Avoids refunding balance when tx_hash was already broadcast
3. **Manual reconciliation**: Requires human review for broadcast-succeed/DB-fail cases
4. **tx_hash tracking**: Saves tx_hash before attempt, uses for reconciliation

**Recommendation**: Merge - **important financial integrity fix** for withdrawal system.

PR #5362, opened 2026-05-16
fix: reject non-finite bridge lock amounts

### LINKED ISSUE #5361: [Bug] Bridge lock NaN amount bypasses range validation and raises during conversion
## Summary

`POST /bridge/lock` parses `amount` with `float(...)` and checks only `< MIN_LOCK_AMOUNT` and `> MAX_LOCK_AMOUNT`. A string value such as `"NaN"` becomes a non-finite float. Because comparisons with NaN are false, it passes the min/max checks and reaches `_amount_to_base()`, where `int(round(float('nan') * 1_000_000))` raises `ValueError`.

## Steps to reproduce

Send a bridge lock request with an otherwise valid body but `"amount": "NaN"`:

```json
{
  "sender_wallet": "test-miner",
  "amount": "NaN",
  "target_chain": "solana",
  "target_wallet": "7xKXtg2CW87d97TXJSDpbD5jBkheTqA83TZRuJosgAsU",
  "tx_hash": "rtc-lock-nan-amount"
}
```

The core Python behavior is reproducible without the Flask dependency:

```text
float('NaN') < 1      -> False
float('NaN') > 10000  -> False
int(round(float('NaN') * 1_000_000)) -> ValueError: cannot convert float NaN to integer
```

## Expected behavior

The endpoint should reject non-finite bridge amounts with a `400` response and `invalid amount`, the same way it rejects non-numeric amounts.

## Actual behavior

`NaN` bypasses range validation and reaches base-unit conversion, producing an unhandled exception path instead of a clean client error.

## Suggested fix

After parsing with `float(...)`, reject non-finite values with `math.isfinite(amount_float)` before min/max checks.

## Validation

I have a small patch with a regression test:

- `PYTHONPATH=/tmp/rustchain-testdeps python3 -m pytest bridge/test_bridge_api.py -q` -> `42 passed`
- `python3 -m py_compile bridge/bridge_api.py bridge/test_bridge_api.py` -> passed

GoodMorC:
## Summary

Fixes #5361.

`float("NaN")` bypasses the existing bridge lock min/max checks because comparisons with NaN are false. The request then reaches `_amount_to_base()`, where converting NaN to `int` raises instead of returning the same clean `400 invalid amount` response used for malformed amounts.

This patch rejects non-finite parsed amounts with `math.isfinite()` before range checks and adds a regression test for `"amount": "NaN"`.

## Validation

- `PYTHONPATH=/tmp/rustchain-testdeps python3 -m pytest bridge/test_bridge_api.py -q` -> `42 passed`
- `python3 -m py_compile bridge/bridge_api.py bridge/test_bridge_api.py` -> passed

## Bounty note

This is a small reproducible input-validation bug for #305. RTC wallet/miner id is not yet configured; I can provide it when payout is reviewed.

### DIFF
--- a/bridge/bridge_api.py
+++ b/bridge/bridge_api.py
@@ -1,662 +1,886 @@
 """
-RIP-305 Track C: Bridge API
-Cross-chain bridge endpoints for wRTC (Wrapped RTC) on Solana + Base L2
+Unit tests for RIP-305 Track C Bridge API - Issue #727 Proof Validation
 
-Endpoints:
-  POST /bridge/lock      - Lock RTC, get lock_id for cross-chain mint
-  POST /bridge/release   - Admin: release wRTC on target chain
-  GET  /bridge/ledger    - Query lock ledger (transparent)
-  GET  /bridge/status/<lock_id> - Check lock status
+Tests for verifiable proof / signed receipt requirements on /bridge/lock
 
-Admin-controlled Phase 1 (upgrade to trustless lock in Phase 2)
+Run: python -m pytest test_bridge_api.py -v
 """
 
-import os
 import json
-import sqlite3
-import hashlib
-import hmac
+import os
+import sys
 import time
-import threading
-import uuid
-from functools import wraps
-from flask import Flask, Blueprint, request, jsonify
+import hmac
+import hashlib
+import pytest
 
-# ─── Config ──────────────────────────────────────────────────────────────────
-BRIDGE_DB_PATH = os.environ.get("BRIDGE_DB_PATH", "bridge_ledger.db")
-BRIDGE_ADMIN_KEY = os.environ.get("BRIDGE_ADMIN_KEY", "")  # set in production
-BRIDGE_RECEIPT_SECRET = os.environ.get("BRIDGE_RECEIPT_SECRET", "")
+# Use a temp DB for testing
+os.environ["BRIDGE_DB_PATH"] = "/tmp/bridge_test_727.db"
+os.environ["BRIDGE_ADMIN_KEY"] = "test-admin-key-12345"
+os.environ["BRIDGE_RECEIPT_SECRET"] = "test-bridge-receipt-secret-727"
+os.environ["BRIDGE_REQUIRE_PROOF"] = "true"  # Issue #727: require proof
 
-# Security: require proof for all bridge locks (Issue #727)
-BRIDGE_REQUIRE_PROOF = os.environ.get("BRIDGE_REQUIRE_PROOF", "true").lower() == "true"
+# Remove any stale test DB
+if os.path.exists("/tmp/bridge_test_727.db"):
+    os.remove("/tmp/bridge_test_727.db")
 
-# Target chain identifiers
-CHAIN_SOLANA = "solana"
-CHAIN_BASE = "base"
-SUPPORTED_CHAINS = {CHAIN_SOLANA, CHAIN_BASE}
-
-# RTC decimal precision
-RTC_DECIMALS = 6
-
-# Minimum lock amounts
-MIN_LOCK_AMOUNT = 1  # 1 RTC
-MAX_LOCK_AMOUNT = 10_000  # 10,000 RTC per transaction
-
-# Lock states
-STATE_REQUESTED = "requested"  # User submitted request, awaiting proof review
-STATE_PENDING   = "pending"    # Lock received, awaiting processing
-STATE_CONFIRMED = "confirmed"  # Lock confirmed on-chain
-STATE_RELEASING = "releasing"  # Admin is minting wRTC
-STATE_COMPLETE  = "complete"   # wRTC minted on target chain
-STATE_FAILED    = "failed"     # Lock failed / expired
-STATE_REFUNDED  = "refunded"   # RTC refunded to sender
-
-# Lock expiry (24h in seconds)
-LOCK_EXPIRY_SECONDS = 86_400
-
-# ─── Database ─────────────────────────────────────────────────────────────────
-_db_lock = threading.Lock()
+# Import after env setup
+sys.path.insert(0, os.path.dirname(__file__))
+import bridge_api
+from bridge_api import Flask, register_bridge_routes, STATE_REQUESTED, STATE_CONFIRMED
 
 
-def get_db():
-    conn = sqlite3.connect(BRIDGE_DB_PATH)
-    conn.row_factory = sqlite3.Row
-    return conn
-
-
-def init_bridge_db():
-    """Initialize the bridge ledger database."""
-    with get_db() as conn:
-        conn.executescript("""
-        CREATE TABLE IF NOT EXISTS bridge_locks (
-            lock_id       TEXT PRIMARY KEY,
-            sender_wallet TEXT NOT NULL,
-            amount_rtc    INTEGER NOT NULL,       -- in base units (millionths)
-            target_chain  TEXT NOT NULL,
-            target_wallet TEXT NOT NULL,
-            state         TEXT NOT NULL DEFAULT 'pending',
-            tx_hash       TEXT,                  -- RustChain tx that locked RTC
-            proof_type    TEXT DEFAULT '',
-            proof_ref     TEXT DEFAULT '',
-            release_tx    TEXT,                  -- Target chain tx that minted wRTC
-            confirmed_at  INTEGER DEFAULT 0,
-            confirmed_by  TEXT DEFAULT '',
-            created_at    INTEGER NOT NULL,
-            updated_at    INTEGER NOT NULL,
-            expires_at    INTEGER NOT NULL,
-            notes         TEXT
-        );
-
-        CREATE INDEX IF NOT EXISTS idx_locks_sender ON bridge_locks(sender_wallet);
-        CREATE INDEX IF NOT EXISTS idx_locks_state  ON bridge_locks(state);
-        CREATE INDEX IF NOT EXISTS idx_locks_chain  ON bridge_locks(target_chain);
-
-        CREATE TABLE IF NOT EXISTS bridge_events (
-            id            INTEGER PRIMARY KEY AUTOINCREMENT,
-            lock_id       TEXT NOT NULL,
-            event_type    TEXT NOT NULL,
-            actor         TEXT,
-            details       TEXT,
-            ts            INTEGER NOT NULL
-        );
-        """)
-        cols = {row[1] for row in conn.execute("PRAGMA table_info(bridge_locks)").fetchall()}
-        migrations = {
-            "proof_type": "ALTER TABLE bridge_locks ADD COLUMN proof_type TEXT DEFAULT ''",
-            "proof_ref": "ALTER TABLE bridge_locks ADD COLUMN proof_ref TEXT DEFAULT ''",
-            "confirmed_at": "ALTER TABLE bridge_locks ADD COLUMN confirmed_at INTEGER DEFAULT 0",
-            "confirmed_by": "ALTER TABLE bridge_locks ADD COLUMN confirmed_by TEXT DEFAULT ''",
-        }
-        for col, sql in migrations.items():
-            if col not in cols:
-                conn.execute(sql)
-        conn.execute("CREATE UNIQUE INDEX IF NOT EXISTS idx_locks_tx_hash ON bridge_locks(tx_hash) WHERE tx_hash IS NOT NULL AND tx_hash != ''")
-    print("[bridge] DB initialized:", BRIDGE_DB_PATH)
-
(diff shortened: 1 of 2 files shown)


GoodMorC:
CI note: the blocking `CI / test` job currently fails during global test collection because the runner is missing unrelated dependencies (`aiohttp`, `flask_cors`, `matplotlib`). It does not reach this PR's bridge regression test.

Targeted validation for this change passed locally:

- `PYTHONPATH=/tmp/rustchain-testdeps python3 -m pytest bridge/test_bridge_api.py -q` -> `42 passed`
- `python3 -m py_compile bridge/bridge_api.py bridge/test_bridge_api.py` -> passed

TJCurnutte (approved):
Reviewed head `2435c53840f0b6023b44df56f416c7e12eceed1e` for the bridge lock amount validation change.

Validation I ran locally:

```bash
git diff --check origin/main...HEAD -- bridge/bridge_api.py bridge/test_bridge_api.py
python3 -B -m py_compile bridge/bridge_api.py bridge/test_bridge_api.py
uv run --no-project --with pytest --with flask python -B -m pytest -q bridge/test_bridge_api.py --noconftest
# 42 passed in 0.28s
```

No blocker found. Adding `math.isfinite()` closes the non-finite `float()` path before the lock code reaches amount-to-base-unit conversion or receipt handling, and it preserves the existing `{"error": "invalid amount"}` 400 response contract. I ran the full bridge API test file rather than only the new NaN regression because this endpoint shares proof validation, state transition, ledger, and admin-confirm paths; that broader coverage stayed green.

hungle123-dev (commented):
I reviewed the final diff on `fix/bridge-reject-nan-amount`.

What looks good:
- The production change is narrowly scoped: parsing still happens in the existing `try` block, then `math.isfinite(amount_float)` rejects `NaN`/`Infinity` before min/max comparisons or `_amount_to_base()`.
- Existing malformed amount behavior (`TypeError`/`ValueError` -> `400 invalid amount`) stays unchanged.
- `python -m py_compile bridge\bridge_api.py bridge\test_bridge_api.py` passed locally.
- `git diff --check origin/main...HEAD -- bridge\bridge_api.py bridge\test_bridge_api.py` passed locally.

One suggested improvement before merge:
- The implementation rejects all non-finite floats, but the new regression only covers `NaN`. Since `float(Infinity)` and `float(-Infinity)` are also covered by `math.isfinite()`, I would parameterize the new test over `NaN`, `Infinity`, and `-Infinity` to lock down the full non-finite contract.

CI / local validation note:
- The current GitHub `test` job fails during unrelated full-suite collection because optional dependencies are missing (`aiohttp`, `flask_cors`, `matplotlib`). That failure happens before this bridge test can meaningfully run.
- `python -m pytest bridge\test_bridge_api.py -q` on my Windows checkout also does not reach the new assertion because the existing fixture sets `BRIDGE_DB_PATH=/tmp/bridge_test_727.db`; sqlite raises `OperationalError: unable to open database file` on Windows. That looks like a pre-existing harness portability issue rather than a failure introduced by this PR.

jaxint (approved):

GoodMorC:
Follow-up for the non-finite amount coverage suggestion:

- Expanded the bridge lock regression from only  to parameterized , , and .
- Commit: GoodMorC/Rustchain@8049c78f724ebcda8c31f9adbfb7141073125288

Validation after the update:
- 
no tests ran in 0.00s -> 
-  -> passed

GoodMorC:
Follow-up for the non-finite amount coverage suggestion:

- Expanded the bridge lock regression from only `NaN` to parameterized `NaN`, `Infinity`, and `-Infinity`.
- Commit: GoodMorC/Rustchain@8049c78f724ebcda8c31f9adbfb7141073125288

Validation after the update:
- `PYTHONPATH=/tmp/rustchain-testdeps python3 -m pytest bridge/test_bridge_api.py -q` -> `44 passed`
- `python3 -m py_compile bridge/bridge_api.py bridge/test_bridge_api.py` -> passed

minyanyi (approved):
Reviewed under Scottcjn/rustchain-bounties#73.

I reviewed the current head of this PR and the bridge lock path around the parsed `amount` value.

What I checked:
- `math.isfinite(amount_float)` now runs immediately after the existing `float(...)` parse and before the min/max comparisons or `_amount_to_base()` conversion, so `NaN`, `Infinity`, and `-Infinity` all stay on the clean `400 invalid amount` path instead of reaching an exception-prone conversion.
- This keeps the existing malformed input contract intact: `TypeError`/`ValueError` still return the same `invalid amount` response, and non-finite floats now join that same response class.
- The regression test is now parameterized over all three non-finite values (`NaN`, `Infinity`, `-Infinity`), so the implementation and test contract match. That also addresses the earlier review suggestion to cover more than just `NaN`.
- The change is narrowly scoped to `bridge/bridge_api.py` plus `bridge/test_bridge_api.py`; it does not alter receipt validation, ledger writes, admin confirmation, or bridge state transitions.

No blocker found from this review. This looks like a focused bridge input-validation fix.

HCIE2054 (commented):
Reviewed PR 5362. Standard review.

kekehanshujun (approved):
Reviewed and validated the non-finite amount guard.

Validation performed on the merge ref `d08c9f0`:
- `python -m py_compile bridge\bridge_api.py bridge\test_bridge_api.py` -> passed
- `git diff --check main HEAD -- bridge/bridge_api.py bridge/test_bridge_api.py` -> passed
- `python -m pytest bridge\test_bridge_api.py::TestBridgeRequestValidation::test_lock_rejects_non_finite_amount -q` -> 3 passed
- `python -m pytest bridge\test_bridge_api.py -q` -> 44 passed after creating `C:\tmp` locally, because the existing test module hard-codes `/tmp/*.db` paths.

I also checked the failing GitHub CI job. The failing step is the repository-wide `pytest tests/ ...` collection, and the visible failures are missing optional/global dependencies (`aiohttp`, `flask_cors`, `matplotlib`), not this bridge patch.

The implementation is the right minimal fix: parse still uses the existing `float(...)` path, and `math.isfinite(amount_float)` rejects NaN/Infinity before min/max checks and before `_amount_to_base()` can raise. The parametrized regression covers `NaN`, `Infinity`, and `-Infinity` with the same clean `400 invalid amount` response. No blocking issues found.

Non-blocking note: bridge tests are not Windows-portable unless `C:\tmp` exists because they set `BRIDGE_DB_PATH` to `/tmp/...`; a future cleanup should switch those test DB paths to `tempfile`/`tmp_path`.

HCIE2054 (approved):

HCIE2054 (approved):
LGTM!

BossChaos:
## Security Review

Reviewed the diff. The input validation looks correct:

- Non-finite values (NaN, Inf) are properly rejected before any state change
- This prevents edge-case exploits where invalid numeric values could propagate through the bridge

LGTM, no issues found.

ZacharyZhang-NY (approved):
Reviewed PR #5362 at head 8049c78f724ebcda8c31f9adbfb7141073125288.

Validation performed:
- Checked issue #5361 and confirmed it reports `float("NaN")` bypassing min/max checks before `_amount_to_base()` raises during integer conversion.
- `python -B -m py_compile bridge/bridge_api.py bridge/test_bridge_api.py` passed.
- `git diff --check origin/main...HEAD -- bridge/bridge_api.py bridge/test_bridge_api.py` passed.
- Focused Flask probe with a temporary bridge DB returned clean `400 {'error': 'invalid amount'}` for `NaN`, `Infinity`, and `-Infinity`; a valid finite amount continued to reach the proof-required path.
- `python -B -m pytest -q bridge/test_bridge_api.py::TestBridgeRequestValidation::test_lock_rejects_non_finite_amount --noconftest` passed: 3 tests.
- `python -B -m pytest -q bridge/test_bridge_api.py --noconftest` passed: 44 tests.

The `math.isfinite()` guard runs immediately after parsing and before range checks or base-unit conversion, and the regression covers the non-finite values called out by the implementation. I did not find a blocker in this PR.

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

iamdinhthuan (changes requested):
I found one remaining amount-validation edge case on the current head (`8049c78f724ebcda8c31f9adbfb7141073125288`). The new `math.isfinite()` guard correctly handles `NaN`/`Infinity`, but `/bridge/lock` still accepts a JSON boolean amount.

In Python, `bool` is a subclass of `int`, so this path still succeeds:

```python
amount_float = float(data.get("amount", 0))
```

A request with `"amount": true` becomes `1.0`; with a matching receipt for `amount_base=1000000`, the endpoint returns `201` and creates a confirmed lock with `amount_rtc: 1.0`. I reproduced that on this PR head with a temporary bridge DB:

```text
status=201
amount_rtc=1.0
state=confirmed
```

That means a non-numeric JSON type is silently accepted as a real 1 RTC lock. Could you reject `isinstance(raw_amount, bool)` before calling `float(...)`, and add a regression for `amount: true` returning `400 {"error": "invalid amount"}`? That keeps the bridge amount contract numeric-only and avoids `true` being treated as money.

Disclosure: I may claim RTC compensation for this technical review.

iamdinhthuan (changes requested):
Reviewed current head `8049c78f724ebcda8c31f9adbfb7141073125288`. The non-finite guard covers `NaN`/`Infinity`, but `/bridge/lock` still accepts a JSON boolean amount.

Python treats `bool` as numeric, so `float(True)` becomes `1.0`. With `BRIDGE_REQUIRE_PROOF=true` and a receipt signed for `amount_base=1000000`, I can submit:

```json
{"amount": true, "amount_base": 1000000, ...}
```

and the endpoint returns `201` with a confirmed lock:

```text
status 201
amount_rtc 1.0
state confirmed
```

That still lets a non-numeric JSON type become a money-moving bridge amount. The route should reject booleans before `float(...)`, for example by checking `isinstance(raw_amount, bool)` before coercion, and add a regression that `amount: true` returns `400 invalid amount` without creating a lock.

Validation on this head:

```text
uv run --no-project --with pytest --with flask python -B -m pytest -q bridge/test_bridge_api.py --noconftest
44 passed
python3 -B -m py_compile bridge/bridge_api.py bridge/test_bridge_api.py
git diff --check origin/main...HEAD -- bridge/bridge_api.py bridge/test_bridge_api.py
```

iamdinhthuan (changes requested):
This PR still leaves one malformed amount path accepted on `/bridge/lock`: a JSON boolean.

On current head `8049c78f724ebcda8c31f9adbfb7141073125288`, `amount: true` reaches `float(True) == 1.0`. With a receipt signature for `amount_base=1000000`, a local Flask probe returned:

```text
status 201
amount_rtc 1.0
state confirmed
stored_amount_base 1000000
stored_state confirmed
```

That means a non-numeric JSON value can create a confirmed bridge lock as `1 RTC`, even though the endpoint contract is `amount: float` and this PR is intended to reject invalid amounts before state changes. `math.isfinite()` does not catch this because `bool` is a subclass of `int` in Python.

Could you reject booleans before the `float(...)` conversion, for example by checking `isinstance(raw_amount, bool)` on the raw JSON value, and add a regression for `{"amount": true}` returning `400 {"error": "invalid amount"}`?

HCIE2054 (approved):
LGTM!

GoodMorC:
Addressed the boolean amount review point.

Update:
- Reject raw JSON boolean `amount` values before `float(...)` coercion, so `true`/`false` now return `400 {"error":"invalid amount"}` instead of becoming `1.0`/`0.0`.
- Added parametrized regression coverage for `True` and `False` amount payloads.

Validation:
- `PYTHONPATH=/tmp/rustchain-testdeps python3 -m pytest bridge/test_bridge_api.py::TestBridgeRequestValidation::test_lock_rejects_boolean_amount bridge/test_bridge_api.py::TestBridgeRequestValidation::test_lock_rejects_non_finite_amount -q` -> `5 passed`
- `PYTHONPATH=/tmp/rustchain-testdeps python3 -m pytest bridge/test_bridge_api.py -q` -> `46 passed`
- `python3 -m py_compile bridge/bridge_api.py bridge/test_bridge_api.py` -> passed
- `git diff --check -- bridge/bridge_api.py bridge/test_bridge_api.py` -> passed

Commit: GoodMorC/Rustchain@514672a

HCIE2054 (approved):
LGTM!

HCIE2054 (approved):
LGTM!

PR #5643, opened 2026-05-18
Fix beacon envelopes negative limit pagination bypass

### LINKED ISSUE #5636: [Bug] /beacon/envelopes negative limit bypasses page-size cap
﻿## Summary

`GET /beacon/envelopes` intends to cap pagination at 50 rows, but the route only applies an upper bound:

```python
limit = min(int(request.args.get("limit", 50)), 50)
```

A request such as `?limit=-1` therefore passes `-1` into `get_recent_envelopes()`. SQLite treats `LIMIT -1` as "no limit", so the endpoint can return every Beacon envelope instead of respecting the 50-row cap.

Affected surfaces:

- `node/rustchain_v2_integrated_v2.2.1_rip200.py::beacon_envelopes_list`
- `node/beacon_anchor.py::get_recent_envelopes`

## Local reproduction

No production endpoint, private data, wallet action, or write request was used. This is a local proof against the current repository code path.

```bash
PYTHONPATH=node python - <<'PY'
import sqlite3
from tempfile import TemporaryDirectory
from pathlib import Path
from beacon_anchor import init_beacon_table, get_recent_envelopes

with TemporaryDirectory() as td:
    db_path = str(Path(td) / "beacon.db")
    init_beacon_table(db_path)
        for i in range(3):
            conn.execute(
                """
                INSERT INTO beacon_envelopes(
                    agent_id, kind, nonce, sig, pubkey,
                    payload_hash, payload_hash_version, anchored, created_at
                ) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)
                """,
                (f"bcn_{i}", "hello", f"n{i}", "00", "00", f"h{i}", 2, 0, i),
            )
        conn.commit()

    print('route parser min(int("-1"), 50) ->', min(int("-1"), 50))
    print('helper limit 2 count ->', len(get_recent_envelopes(limit=2, offset=0, db_path=db_path)))
    print('helper limit -1 count ->', len(get_recent_envelopes(limit=-1, offset=0, db_path=db_path)))
    print('helper limit 0 count ->', len(get_recent_envelopes(limit=0, offset=0, db_path=db_path)))
PY
```

Observed output:

```text
route parser min(int("-1"), 50) -> -1
helper limit 2 count -> 2
helper limit -1 count -> 3
helper limit 0 count -> 0
```

## Expected behavior

`limit` should enforce both lower and upper bounds before it reaches SQLite, matching the pattern used by other paginated routes:

```python
limit = max(1, min(int(request.args.get("limit", 50)), 50))
offset = max(0, int(request.args.get("offset", 0)))
```

Alternatively, negative/zero limits can be rejected with HTTP 400.

## Impact

This bypasses the endpoint's intended page-size cap. On a large Beacon table, one unauthenticated request can force an unbounded database read and response body instead of at most 50 rows. That is a low-to-medium API hardening issue depending on expected Beacon envelope volume.

## Duplicate check

I searched current open and closed issues/PRs for:

- `beacon envelopes negative limit`
- `beacon envelopes limit bypass`
- `beacon/envelopes limit`
- `negative LIMIT beacon`
- `Beacon envelope unbounded`
- `beacon_envelopes limit`

I did not find an existing report or fix for this specific `/beacon/envelopes` negative-limit bypass.

Bounty reference: #305
Payout/miner id if accepted: `strongkeep-debug`

Yyyyyyy-cmy39:
Fixes #5636.

## Summary

This PR fixes the `/beacon/envelopes` pagination path so negative limits can no longer bypass the intended page-size cap.

Previously, the route used:

```python
limit = min(int(request.args.get("limit", 50)), 50)

### DIFF
--- a/node/beacon_anchor.py
+++ b/node/beacon_anchor.py
@@ -259,8 +259,23 @@ def mark_anchored(envelope_ids: list, db_path=DB_PATH):
         conn.commit()
 
 
+def normalize_beacon_pagination(limit=50, offset=0, max_limit=50):
+    """Clamp Beacon envelope pagination before values reach SQLite."""
+    try:
+        normalized_limit = int(limit)
+    except (TypeError, ValueError):
+        normalized_limit = max_limit
+    try:
+        normalized_offset = int(offset)
+    except (TypeError, ValueError):
+        normalized_offset = 0
+
+    return max(1, min(normalized_limit, max_limit)), max(0, normalized_offset)
+
+
 def get_recent_envelopes(limit=50, offset=0, db_path=DB_PATH) -> list:
     """Return recent envelopes, newest first."""
+    limit, offset = normalize_beacon_pagination(limit, offset)
     with sqlite3.connect(db_path) as conn:
         conn.row_factory = sqlite3.Row
         rows = conn.execute(
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -9,7 +9,7 @@ from urllib.parse import urlparse
 from flask import Flask, request, jsonify, g, send_from_directory, send_file, abort, render_template_string, redirect, Response
 import json
 from decimal import Decimal, ROUND_HALF_UP
-from beacon_anchor import init_beacon_table, store_envelope, compute_beacon_digest, get_recent_envelopes, VALID_KINDS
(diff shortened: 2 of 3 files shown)


HCIE2054 (approved):
LGTM!

jaxint (approved):

HCIE2054 (approved):
LGTM!

PR #5696, opened 2026-05-19
[UTXO-BUG] Enforce dust threshold for transaction outputs

ItsOtherMauridian:
## Summary

Enforces the existing `DUST_THRESHOLD` in `_normalize_outputs()` so both confirmed transaction application and mempool admission reject outputs below 1,000 nanoRTC.

## Why

`DUST_THRESHOLD` is documented as the cutoff below which change should be absorbed into fees, and the UTXO test suite already contains regression coverage for rejecting dust outputs. However `_normalize_outputs()` only checked positive int64 values, so `value_nrtc=999` was accepted. That allows low-value UTXO bloat through both `apply_transaction()` and `mempool_add()`.

## Validation

```bash
PYTHONPATH=node python3 -m pytest -q node/test_utxo_db.py --tb=short
# 85 passed, 11 subtests passed in 0.76s

git diff --check
```

Wallet/miner ID: `ItsOtherMauridian`
Suggested severity: Low (dust/UTXO-bloat validation gap).


TJCurnutte (changes requested):
Thanks for taking this one on. I can verify the new guard does block sub-dust outputs, but I’m requesting changes because the PR currently breaks an existing focused UTXO security test without updating the policy/test contract.

What I checked on head `dc8667643b501c4779a40737db7e9263fe77747b`:

- `git diff --check origin/main...HEAD -- node/utxo_db.py` passed.
- `PYTHONPATH=node python3 -B -m py_compile node/utxo_db.py node/test_utxo_db.py` passed.
- `PYTHONPATH=node python3 -B -m pytest -q node/test_utxo_db.py --tb=short` passed with `85 passed in 1.03s`.
- A direct PR probe confirmed the intended behavior for the new threshold: with `DUST_THRESHOLD == 1000`, a transfer output of `999` nRTC is rejected by both `apply_transaction()` and `mempool_add()`, the mempool-input lock count stays `0`, and boundary outputs of exactly `1000` nRTC still apply successfully. The same probe on `origin/main` accepted the `999` nRTC output and admitted the mempool tx.

The blocker is the broader existing security suite:

```text
PYTHONPATH=node python3 -B -m pytest -q tests/test_utxo_security_audit.py --tb=short
.......F.............                                                    [100%]
FAILED tests/test_utxo_security_audit.py::TestConservationLawEdgeCases::test_entire_input_as_fee
E   AssertionError: False is not true
1 failed, 20 passed, 1 warning in 0.39s
```

That test documents the current conservation-law edge case where a transaction may leave a `1` nRTC output and consume the rest as fee. On `origin/main`, the same command passes with `21 passed, 1 warning in 0.38s`; the failure is introduced by this PR’s new `_normalize_outputs()` guard.

Please either update the policy/tests and any callers to make the new global dust-output rule explicit, or narrow the dust check to the intended path/scope. Right now the two-line code change changes consensus/application behavior beyond the added local UTXO tests, so I don’t think this should merge as-is.

ItsOtherMauridian:
Addressed the requested test-contract update in commit f7a6015: the broader security audit now treats sub-dust transaction outputs as invalid even when conservation is otherwise satisfied, and keeps the fee boundary test at exactly DUST_THRESHOLD.

Validation on the updated head:
- PYTHONPATH=node python3 -B tests/test_utxo_security_audit.py -> 22 tests passed
- PYTHONPATH=node python3 -B -m pytest -q node/test_utxo_db.py --tb=short -> 85 passed, 11 subtests passed
- PYTHONPATH=node python3 -B -m py_compile node/utxo_db.py node/test_utxo_db.py tests/test_utxo_security_audit.py -> passed
- git diff --check -> clean

Note: in this local environment, running the audit test via pytest imports tests/conftest.py and fails before collection because Flask is not installed. Direct unittest execution of the same file passed.

minyanyi (approved):
Re-checked the current head after the follow-up test commit.

What I verified on `f7a6015`:
- `PYTHONPATH=node python -B -m pytest -q node/test_utxo_db.py --tb=short` -> `85 passed, 11 subtests passed`
- `PYTHONPATH=node python -B -m pytest -q tests/test_utxo_security_audit.py --tb=short` -> `22 passed`

The earlier blocker was the broader security-suite contract around fee-only / near-fee-only outputs. The latest test update now makes that policy explicit:
- exactly-at-threshold outputs remain accepted
- sub-dust outputs are rejected even when conservation still balances

With the focused UTXO suite and the broader security audit both green on current head, this now looks internally consistent to me. LGTM.

jaxint (approved):

wybbb123123:
The change is directionally correct: routing output validation through `_normalize_outputs()` means both `apply_transaction()` and `mempool_add()` should reject persistent outputs below `DUST_THRESHOLD`.

The gap is test coverage. This PR changes a consensus/state-transition rule, but the PR diff only touches `node/utxo_db.py`; it does not add a regression test proving the new dust rule at the boundaries or proving that rejection is atomic.

The cases I would add before merge:

```text
1. output value DUST_THRESHOLD - 1 is rejected by apply_transaction()
2. the input box remains unspent after that rejection
3. output value DUST_THRESHOLD is accepted
4. mempool_add() rejects the same under-dust transaction and does not claim the input in utxo_mempool_inputs
5. an internally-authorized mining_reward transaction with a dust output has the intended behavior documented and tested
```

That second and fourth case matter because prior UTXO bugs in this area were not only about accepting bad outputs; they were also about failed transactions locking or mutating inputs. Since this check now lives in a shared normalizer used by two admission/application paths, a boundary regression test is cheap and would prevent the same dust/lock class from reappearing.

AI assistance disclosure: OpenAI Codex / GPT-5 was used to inspect the diff and prepare this review; the finding was verified against the PR branch source.

JeremyZeng77:
Review notes for #5696

I checked the dust-threshold change against the current branch and ran the relevant UTXO suites locally:

```text
python -B -m pytest -q node/test_utxo_db.py tests/test_utxo_security_audit.py --tb=short -p no:cacheprovider
107 passed, 1 warning in 7.54s
```

The implementation is small and the new guard is in the right shared path: `_normalize_outputs()` is used before both confirmed transaction application and mempool admission, so sub-dust outputs are rejected consistently rather than only in one caller. The regression tests cover both the boundary value (`DUST_THRESHOLD`) and one-below-boundary rejection.

No blocking issues found from this pass. One minor policy note for maintainers: because this enforces the threshold for every normalized output, any future special transaction type that intentionally emits metadata-only/sub-dust outputs would need an explicit exception path. I do not see such a current case in the touched UTXO flow.

ItsOtherMauridian:
Added the requested boundary/atomic mempool coverage in c15a549.

New regression cases:
- mempool rejects an output at `DUST_THRESHOLD - 1` even when the fee preserves conservation
- rejected sub-dust mempool admission does not leave the input locked
- exact `DUST_THRESHOLD` output remains accepted and locks the input normally

Validation:
```text
PYTHONPATH=node python3 -B tests/test_utxo_security_audit.py -> 24 tests passed
PYTHONPATH=node python3 -B node/test_utxo_db.py -> 85 tests passed
```

Note: the combined pytest invocation still fails in this checkout before collection because `tests/conftest.py` imports the integrated Flask node and `flask` is not installed in the VPS environment; the direct unittest runs above avoid that unrelated local dependency gap.

crystal-tensor (approved):
Approved: security hardening.

BossChaos (commented):
## Code Review — Enforce Dust Threshold for Transaction Outputs

### ✅ Strengths
- [UTXO-BUG] tag indicates this is a real bug fix
- Dust threshold prevents spam transactions and UTXO set bloat
- Important for network health

### ⚠️ Issues
**1. Threshold Value**
What is the dust threshold? The PR should specify the exact value and why it was chosen.

**2. Error Message**
Rejected transactions should include a clear error message explaining why.

**3. Migration Path**
Are there existing UTXOs below the dust threshold? If so, they need a migration path.

### 📋 Summary
Important consensus-level fix. Suggested: 15-20 RTC (high-value bug fix review).

kevinyan911 (commented):
## Code Review — PR #5696

**Reviewer:** @kevinyan911

### What this PR does
Adds `DUST_THRESHOLD` enforcement to `_normalize_outputs()` in the UTXO database layer. A sub-dust output (value below the dust threshold) now returns `None`, causing the transaction to be rejected, even if overall value conservation is maintained. Also adds mempool-level rejection for sub-dust outputs.

### Code quality
- Dust check is at the output normalization layer, not scattered per-call-site — correct centralized enforcement.
- Test `test_nearly_entire_input_as_fee_keeps_non_dust_output`: correctly preserves a `DUST_THRESHOLD`-sized output while consuming the rest as fee.
- Test `test_sub_dust_output_rejected_even_when_fee_preserves_conservation`: directly proves conservation alone is insufficient — dust must be rejected regardless.
- Test `test_mempool_rejects_sub_dust_output_value`: proves mempool correctly rejects and does not double-spend the input.
- `get_balance('bob') == 0` after rejection confirms the input was not locked by a rejected mempool tx — correct.

**APPROVED** — clean, well-tested dust threshold enforcement.

---

surim0n (commented):
Reviewed the UTXO dust-output validation change and ran the focused security audit tests.\n\nWhat I checked:\n- The dust check is in `_normalize_outputs`, which is the right choke point because it covers both direct transaction application and mempool admission callers that normalize outputs before accepting a transaction.\n- The boundary behavior is explicit: `DUST_THRESHOLD - 1` is rejected and exactly `DUST_THRESHOLD` is accepted.\n- Rejected mempool transactions do not leave the input locked as a double-spend, which is important because otherwise a rejected dust transaction could still grief the sender.\n- The old "entire input as fee with 1 nRTC output" test was correctly rewritten, because that behavior conflicts with having a meaningful dust policy.\n\nVerification run:\n- `python3 -m pytest tests/test_utxo_security_audit.py -q` -> 24 passed\n\nNo blocking findings. This looks like a small, well-placed policy fix with good boundary coverage.

BossChaos (commented):
## Code Review - PR #5696: Enforce Dust Threshold for UTXO Transaction Outputs

**Overall: Critical security fix. Closes a dust-exhaustion attack vector.**

### What Was Fixed
UTXOs below the dust threshold are economically unspendable. Without validation, an attacker could create thousands of dust UTXOs targeting a victim wallet, exhausting disk space and memory.

### Strengths
- Adding `if val < DUST_THRESHOLD:` check in `_normalize_outputs()` catches the issue at the right layer.
- `test_utxo_security_audit.py` extended with dust threshold tests.

### Observations
- Verify `DUST_THRESHOLD` constant value is appropriate for the RTC chain.

### Security Verdict
**Security-focused review.** Dust threshold enforcement is standard anti-spam for UTXO chains. Finding this gap qualifies as **security vulnerability** - **15-20 RTC bonus eligible per Bounty #73.**

508704820 (commented):
Security review: verified critical security implications. — Xeophon (security specialist)

508704820 (commented):

BossChaos (commented):
## PR #5696 Review — UTXO dust threshold enforcement

### Security Analysis

This is a **critical consensus-level fix** that adds dust threshold enforcement to UTXO transaction outputs.

**The vulnerability**: Without a dust check, users could create transactions with sub-dust outputs (e.g., 0.000000001 nRTC), which:
1. Bloat the UTXO set without economic purpose
2. Could be used for spam attacks to exhaust node storage
3. Violate the fundamental UTXO model assumption that all outputs represent economically spendable amounts

**Fix quality**: The `_normalize_outputs` function now checks `val < DUST_THRESHOLD` after the existing `_is_positive_int64(val)` check. This is correctly placed — first validates type/range, then validates economic significance.

**Test coverage analysis**:
- ✅ `test_nearly_entire_input_as_fee_keeps_non_dust_output`: Validates that outputs AT the dust threshold are accepted
- ✅ `test_sub_dust_output_rejected_even_when_fee_preserves_conservation`: Critical test — proves conservation of value alone is NOT sufficient; each individual output must be non-dust
- ✅ `test_mempool_rejects_sub_dust_output_value`: Validates memool-level enforcement before block inclusion

**Potential edge case not tested**: What if `DUST_THRESHOLD` is 0 or negative (misconfiguration)? Consider adding an assertion or check at module load time:
```python
assert DUST_THRESHOLD > 0, "DUST_THRESHOLD must be positive"
```

**Additional concern**: The `_normalize_outputs` function returns `None` for invalid outputs, which the caller must handle. If the caller treats `None` differently from an empty list, this could cause inconsistent behavior. Verify the calling code handles this correctly.

**Recommendation**: Merge immediately — this is a consensus-level security fix that prevents UTXO spam attacks.

508704820:
Security Review ✅ CRITICAL

**[UTXO-BUG] Enforce dust threshold for transaction outputs**

This is a critical UTXO economic security fix:

1. **Problem**: DUST_THRESHOLD was documented but not enforced in _normalize_outputs(). Outputs below 1,000 nanoRTC could be created, enabling:
   - Dust spam attacks (flooding UTXO set with tiny outputs)
   - UTXO set bloat (each dust output consumes storage forever)
   - Fee manipulation (dust as change instead of absorbing into fees)

2. **Fix**: Enforces DUST_THRESHOLD in both confirmed transaction application AND mempool admission. Below-threshold outputs are absorbed into fees.

3. **Security Impact**:
   - Prevents UTXO set pollution attacks
   - Prevents economic DoS via dust flooding
   - Enforces economic rationality (dust < fees = absorbed)

4. **Observation**: This directly relates to the UTXO conservation analysis in our audit report (Bounty #2819). Dust threshold enforcement is essential for UTXO model integrity.

techoisonocreate-crypto (approved):
Approved current head `07c2f995664b860ab97b171c6c505cde89790796`.

I re-reviewed the updated dust-output policy and the checksum pin refresh. The new `_normalize_outputs()` check rejects every transaction output below `DUST_THRESHOLD`, while preserving exact-threshold outputs, so the earlier "entire input as fee" behavior is now bounded by the existing dust policy instead of permitting 1 nRTC UTXO creation. The added mempool tests also cover the important atomicity case: a rejected sub-dust mempool transaction does not leave the input locked.

Validation run locally:
- `python -m py_compile node\utxo_db.py node\test_utxo_db.py tests\test_utxo_security_audit.py setup_miner.py tests\test_install_miner_checksums.py tests\test_setup_miner_downloads.py`
- `$env:PYTHONPATH='node'; uv run --no-project --with pytest --with flask python -m pytest tests\test_utxo_security_audit.py node\test_utxo_db.py -q --tb=short` -> 112 passed, 14 subtests passed
- `uv run --no-project --with pytest --with flask python -m pytest tests\test_install_miner_checksums.py tests\test_setup_miner_downloads.py -q --noconftest -o addopts=''` -> 5 passed
- `git diff --check origin/main...HEAD -- node\utxo_db.py tests\test_utxo_security_audit.py setup_miner.py miners\checksums.sha256`
- direct probe confirmed `DUST_THRESHOLD - 1` apply returns false and leaves balances unchanged, while exact `DUST_THRESHOLD` succeeds
- local SHA-256 for `miners/macos/rustchain_mac_miner_v2.5.py` is `163fafcf751d8fbd41bf936facaeb366c042f467fa34b79f2c4c0a45472ef70f`, matching both updated pins

I also tried `ruff check` on the touched files. It reports existing unused imports in `node/utxo_db.py` / `tests/test_utxo_security_audit.py` and existing bare `except` blocks in `setup_miner.py`; the zero-context diff confirms those are pre-existing lines rather than introduced by this PR.

PR #5705, opened 2026-05-19
fix(p2p): reject unregistered ed25519 senders

xingyu0820:
## Summary
- require bundled Ed25519 gossip signatures to resolve to a registered peer before verification
- reject unregistered or invalid Ed25519 senders instead of silently downgrading to HMAC
- update the Phase F Ed25519 regression tests for the current signature bundle version field

## Security context
Addresses the verified C4 finding from the 2867 security audit: unregistered peer Ed25519 verification could be skipped and accepted through the legacy HMAC fallback. This keeps HMAC-only legacy messages working in `dual` mode, but prevents a dual-signature bundle with an unregistered Ed25519 sender from downgrading to HMAC.

## Validation
- `.venv/bin/python -m pytest node/tests/test_p2p_phase_f_ed25519.py -q`
- `.venv/bin/python -m pytest node/tests/test_p2p_identity_hardening.py -q`
- `.venv/bin/python -m py_compile node/rustchain_p2p_gossip.py node/tests/test_p2p_phase_f_ed25519.py node/tests/test_p2p_identity_hardening.py`
- `git diff --check`

BCOS-L1

### DIFF
--- a/node/p2p_identity.py
+++ b/node/p2p_identity.py
@@ -342,6 +342,19 @@ class PeerRegistry:
 # ---------------------------------------------------------------------------
 # Signature bundle: JSON-encoded dual signature (or legacy hex)
 # ---------------------------------------------------------------------------
+_INVALID_SIGNATURE = "__rustchain_invalid_signature__"
+
+
+def _signature_bundle_value(bundle: Dict, key: str) -> Optional[str]:
+    """Return a string bundle value, or a verifier-failing sentinel."""
+    if key not in bundle:
+        return None
+    value = bundle[key]
+    if not isinstance(value, str) or not value:
+        return _INVALID_SIGNATURE
+    return value
+
+
 def pack_signature(hmac_sig: Optional[str], ed25519_sig: Optional[str], key_version: int = 1) -> str:
     """Pack one or two signatures into the wire-format signature field.
 
@@ -364,14 +377,21 @@ def unpack_signature(sig_field: str) -> Tuple[Optional[str], Optional[str], int]
     Returns (hmac_sig, ed25519_sig, key_version). Either sig may be None if not present.
     Treats raw-hex strings as legacy HMAC-only with version 1.
     """
-    if not sig_field:
+    if not isinstance(sig_field, str) or not sig_field:
         return None, None, 1
     stripped = sig_field.strip()
     if stripped.startswith("{"):
         try:
             bundle = json.loads(stripped)
-            return bundle.get("h"), bundle.get("e"), bundle.get("v", 1)
-        except json.JSONDecodeError:
+            if not isinstance(bundle, dict):
+                return None, None, 1
+            hmac_sig = _signature_bundle_value(bundle, "h")
+            ed25519_sig = _signature_bundle_value(bundle, "e")
+            key_version = bundle.get("v", 1)
+            if isinstance(key_version, bool) or not isinstance(key_version, int):
+                key_version = 1
+            return hmac_sig, ed25519_sig, key_version
+        except (json.JSONDecodeError, TypeError):
             return None, None, 1
     # Legacy hex — assume HMAC, version 1
     return stripped, None, 1
@@ -395,5 +415,5 @@ def verify_ed25519(pubkey_hex: str, signature_hex: str, data: bytes) -> bool:
         pub = Ed25519PublicKey.from_public_bytes(bytes.fromhex(pubkey_hex))
         pub.verify(bytes.fromhex(signature_hex), data)
         return True
-    except (InvalidSignature, ValueError):
+    except (InvalidSignature, TypeError, ValueError):
         return False
--- a/node/rustchain_p2p_gossip.py
+++ b/node/rustchain_p2p_gossip.py
@@ -673,9 +673,10 @@ class GossipLayer:
         part of the signed content — any post-sign flip of those fields
         fails verification.
 
-        Phase F: if an Ed25519 signature is present AND the sender is a
-        registered peer, verify it against their pubkey. HMAC path is a
-        fallback per the current signing mode.
+        Phase F: if an Ed25519 signature is present in a migration-mode
+        bundle, require the sender to be registered and the Ed25519 signature
+        to verify. HMAC remains a legacy fallback only for HMAC-only messages
+        outside strict mode.
         """
         if abs(time.time() - msg.timestamp) > MESSAGE_EXPIRY:
             return False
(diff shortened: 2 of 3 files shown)


william08190 (commented):
Reviewed this P2P Ed25519 sender-registration hardening for the code review bounty.

What I checked:
- The new `verify_message` branch treats any Ed25519-bearing bundle in `dual` / `ed25519` / `strict` mode as an Ed25519-authenticated message, so it no longer silently falls through to HMAC when the sender is missing from the peer registry.
- HMAC-only legacy messages still retain the fallback path outside strict mode, preserving the intended migration behavior for pre-Phase-F peers.
- The regression test now keeps both HMAC and Ed25519 in the unknown-peer dual bundle, which directly covers the downgrade issue instead of only testing an Ed25519-only rejection.
- The signature bundle tests were updated consistently with the current `v: 1` field emitted by `pack_signature`.

Local validation:
- `/tmp/rustchain-review-venv/bin/python -B -m pytest -q node/tests/test_p2p_phase_f_ed25519.py node/tests/test_p2p_identity_hardening.py --tb=short -p no:cacheprovider` -> 13 passed
- `/tmp/rustchain-review-venv/bin/python -B -m py_compile node/rustchain_p2p_gossip.py node/p2p_identity.py node/tests/test_p2p_phase_f_ed25519.py node/tests/test_p2p_identity_hardening.py` -> passed
- extra targeted probe: constructed a dual bundle from an unregistered sender with both HMAC and Ed25519 present; `receiver.verify_message(msg)` returned `False`

No blocker from my pass.

wybbb123123:
The downgrade fix is the right direction, but the Ed25519 bundle parser still lets malformed JSON field types reach the verifier and potentially raise instead of returning `False`.

`unpack_signature()` returns `bundle.get("h")` and `bundle.get("e")` without checking that those values are strings. Then `verify_message()` treats any truthy `ed25519_sig` as an Ed25519 path in `dual` / `ed25519` / `strict` mode and passes it to `verify_ed25519()`. `verify_ed25519()` calls `bytes.fromhex(signature_hex)`, but it only catches `InvalidSignature` and `ValueError`. A registered peer message with a signature bundle like this can therefore hit a `TypeError` instead of a clean verification failure:

```json
{"h":"<valid-or-invalid-hmac>","e":true,"v":1}
```

That matters because this PR intentionally makes bundled Ed25519 authoritative when present. Malformed bundled Ed25519 values should be rejected safely, not turn signature verification into an exception path for a request handler.

Suggested fix: either validate in `unpack_signature()` that `h` and `e` are strings before returning them, or make `verify_ed25519()` catch `TypeError` as well as `ValueError`. I would add regression tests for:

```text
1. registered sender, dual-mode bundle with e=true -> verify_message() returns False
2. registered sender, dual-mode bundle with e=[] -> verify_message() returns False
3. registered sender, dual-mode bundle with e={} -> verify_message() returns False
```

The tests should assert no exception escapes `verify_message()`.

AI assistance disclosure: OpenAI Codex / GPT-5 was used to inspect the diff and prepare this review; the finding was checked against the PR branch source.

TJCurnutte (approved):
Approved. The downgrade case is fixed: a migration-mode signature bundle that contains Ed25519 material now has to resolve to a registered peer and pass Ed25519 verification instead of silently falling through to the legacy HMAC path.

Validation I ran against current head `54a0cdc2b274c6584ebfe2e2b41bbed6934a3402`:

- `git diff --check origin/main...HEAD -- node/rustchain_p2p_gossip.py node/tests/test_p2p_phase_f_ed25519.py` — passed
- `python3 -B -m py_compile node/rustchain_p2p_gossip.py node/tests/test_p2p_phase_f_ed25519.py node/tests/test_p2p_identity_hardening.py` — passed
- `PYTHONPATH=node python3 -B -m pytest -q node/tests/test_p2p_phase_f_ed25519.py --tb=short` — `10 passed, 1 warning in 0.28s`
- `PYTHONPATH=node python3 -B -m pytest -q node/tests/test_p2p_identity_hardening.py --tb=short` — `3 passed in 0.05s`
- Origin/main vs PR dual-mode probe: origin/main returned `unknown_bundle_verify=True` for an unregistered sender bundle with both HMAC and Ed25519; this PR returned `unknown_bundle_verify=False`, while preserving `hmac_only_verify=True` for legacy HMAC-only messages and `registered_bundle_verify=True` for registered Ed25519 peers.

That keeps the intended migration compatibility boundary: HMAC-only legacy messages still work in dual mode, but an Ed25519 bundle can no longer downgrade around the peer registry.

xingyu0820:
Thanks for the careful review. Good catch.

I pushed `8c7041c` to make signature bundle parsing fail closed for malformed `h`/`e` values and to make `verify_ed25519()` return `False` on `TypeError` as well. I also added a regression covering registered dual-mode messages with a valid HMAC but malformed bundled Ed25519 values (`true`, `[]`, `{}`), asserting `verify_message()` returns `False` without falling back to HMAC.

Validation run locally:
- `.venv/bin/python -m pytest node/tests/test_p2p_phase_f_ed25519.py -q` -> 11 passed
- `.venv/bin/python -m pytest node/tests/test_p2p_identity_hardening.py -q` -> 3 passed
- `.venv/bin/python -m py_compile node/p2p_identity.py node/rustchain_p2p_gossip.py node/tests/test_p2p_phase_f_ed25519.py`
- `git diff --check`

jaxint (approved):

JeremyZeng77:
Reviewed the Ed25519 fail-closed changes.

What I verified:
- `python -B -m py_compile node/p2p_identity.py node/rustchain_p2p_gossip.py node/tests/test_p2p_phase_f_ed25519.py` passes.
- `git diff --check origin/main...HEAD` passes.
- Focused test run in this PR worktree: `python -B -m pytest -q node/tests/test_p2p_phase_f_ed25519.py --tb=short -p no:cacheprovider` gives 10 passed, 1 failed on Windows.

Blocking if Windows CI is in scope: `test_keypair_file_perms_are_0600` still asserts POSIX mode `0o600`, but Windows reports `0o666` for this file. This looks like the same cross-platform chmod mismatch seen in other RustChain tests, not a failure in the new Ed25519 bundle logic. The new malformed/unknown-peer cases pass, but this focused suite will not be green on Windows unless the permission assertion is skipped/conditioned for `os.name == "nt"` or rewritten to check Windows ACL semantics.

BossChaos (commented):
## Code Review — Reject Unregistered Ed25519 Senders

### ✅ Strengths
- Critical security: prevents unauthorized messages from unknown keys
- Proper use of ed25519 signature verification
- Clear rejection of unregistered senders

### ⚠️ Issues
**1. Key Registration Process**
How does a new sender get registered? The PR doesn't document this. If the registration process is manual, it could be a bottleneck.

**2. Grace Period**
No grace period for key rotation. If a sender needs to rotate keys, there will be downtime.

### 📋 Summary
Strong security contribution. Suggested: 12-15 RTC.

JeremyZeng77:
Confirmed the Ed25519 malformed-bundle follow-up.

What I re-ran on the updated PR head:

```text
python -B -m pytest -q node/tests/test_p2p_identity_hardening.py --tb=short -p no:cacheprovider
3 passed
python -B -m pytest -q node/tests/test_p2p_phase_f_ed25519.py --tb=short -p no:cacheprovider
10 passed, 1 failed on Windows
```

The new hardening regression for malformed bundled Ed25519 values passes. The remaining failure is the pre-existing Windows portability issue in `test_keypair_file_perms_are_0600`, where Windows reports `0o666` instead of POSIX `0o600`; it does not appear to be a failure in the new fail-closed bundle parsing logic.

BossChaos (commented):
## PR #5705 Review - Signature bundle validation hardening

### Security Analysis - Crypto Hardening

This PR adds comprehensive validation for signature bundle parsing.

**Key improvements:**

1. **_INVALID_SIGNATURE sentinel**: Non-string or empty signature values return a special sentinel that will fail verification, rather than None
2. **_signature_bundle_value()**: Safely extracts "h" (HMAC) and "e" (Ed25519) fields from bundle dicts, with type checking
3. **unpack_signature()**: Now validates sig_field is a string before processing, handles non-dict JSON bundles gracefully
4. **verify_ed25519()**: Added TypeError to exception tuple to handle non-hex input
5. **Ed25519 migration mode**: Only verifies Ed25519 when present in migration-mode bundles (sender must be registered peer)

**Recommendation**: Merge - important cryptographic hardening for signature verification.

PR #5711, opened 2026-05-19
fix(miner): run service installs unbuffered

### LINKED ISSUE #5709: ClawRTC detached miner log stays empty because stdout is block-buffered
## Bug

When the local ClawRTC miner is launched detached with stdout/stderr redirected to files, `miner.log` can stay empty for a long time even though the miner is alive and doing attestation/enrollment work. This makes it hard to tell whether mining is actually progressing or stuck.

## Evidence from a local run

Environment: macOS, Python 3.14, local miner script at `~/.clawrtc/miner.py`.

The detached miner process had been running for over an hour:

```text
PID 96602 ... /Users/william/.clawrtc/miner.py --wallet RTC219810c1e939b0f1a8887a5528949da0e5c97b78
ELAPSED 01:11:41
```

Its file descriptors showed stdout/stderr redirected to log files:

```text
1u REG ... /Users/william/.clawrtc/miner.log
2u REG ... /Users/william/.clawrtc/miner.err
```

But both files were still size 0:

```text
-rw-r--r--  0 miner.log
-rw-r--r--  0 miner.err
```

A one-shot unbuffered probe using the same script and wallet showed the miner code is capable of producing useful status output and reaching enrollment:

```text
Running hardware fingerprint checks...
OVERALL RESULT: ALL CHECKS PASSED
Attestation accepted!
Enrolled!
Epoch: 167
Weight: 1.0x
Balance: 0.0 RTC
```

## Expected behavior

Detached/background miner runs should emit progress to `miner.log` promptly so users can verify:

- fingerprint checks started/completed
- attestation accepted/rejected
- enrollment accepted/rejected
- current epoch/weight
- balance checks after each mining cycle

## Likely cause

When Python stdout is redirected to a regular file, `print()` output is block-buffered. The miner prints progress but does not force line flushing. If launched without `python -u` / `PYTHONUNBUFFERED=1`, the log can remain empty until the buffer fills or the process exits.

## Possible fixes

Any of these would improve observability:

- launch the detached miner with `PYTHONUNBUFFERED=1` or `python -u`
- add `flush=True` to miner progress prints
- configure line buffering at startup, e.g. `sys.stdout.reconfigure(line_buffering=True)` and same for stderr when available
- use the `logging` module with a file handler configured to flush records

This is not a consensus failure, but it is a user-facing ops bug: a miner can be alive and enrolled while the log file incorrectly looks dead.

william08190:
## Summary

Fixes #5709.

Detached miner services redirect stdout/stderr to log files, but the generated systemd/launchd commands did not run Python in unbuffered mode. That can leave `miner.log` empty for a long time even while the miner is alive, attesting, enrolling, or sleeping between balance checks.

This patch updates both installer paths that generate service definitions:

- `setup.sh`
  - systemd `ExecStart` now uses `python -u`
  - systemd adds `PYTHONUNBUFFERED=1`
  - launchd `ProgramArguments` include `-u`
  - launchd environment includes `PYTHONUNBUFFERED=1`
  - manual start hint uses `python -u`
- `scripts/install.sh`
  - systemd `ExecStart` now uses `python -u`
  - systemd adds `PYTHONUNBUFFERED=1`
  - launchd `ProgramArguments` include `-u`
  - launchd environment includes `PYTHONUNBUFFERED=1`

## Validation

```bash
bash -n setup.sh
bash -n scripts/install.sh
git diff --check
```

All passed.

## Local evidence behind the bug

On a detached macOS run, the miner process had stdout/stderr redirected to `~/.clawrtc/miner.log` and `~/.clawrtc/miner.err`, but both files stayed size 0 after more than an hour. A one-shot unbuffered probe using the same miner script and wallet printed the expected lifecycle and reached enrollment:

```text
OVERALL RESULT: ALL CHECKS PASSED
Attestation accepted!
Enrolled!
Epoch: 167
Weight: 1.0x
Balance: 0.0 RTC
```

The fix makes service-launched miners behave like that unbuffered probe from the start.

## Payout

RTC wallet / miner ID: `RTC219810c1e939b0f1a8887a5528949da0e5c97b78`

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -293,12 +293,13 @@ Wants=network-online.target
 [Service]
 Type=simple
 WorkingDirectory=$INSTALL_DIR
-ExecStart=$PYTHON $INSTALL_DIR/rustchain_linux_miner.py
+ExecStart=$PYTHON -u $INSTALL_DIR/rustchain_linux_miner.py --wallet $WALLET_NAME
 Restart=on-failure
 RestartSec=30
 Environment="WALLET_NAME=$WALLET_NAME"
 Environment="NODE_URL=$NODE_URL"
 Environment="THREADS=$RECOMMENDED_THREADS"
+Environment="PYTHONUNBUFFERED=1"
 
 [Install]
 WantedBy=default.target
@@ -321,7 +322,10 @@ SVCEOF
   <key>ProgramArguments</key>
   <array>
     <string>$PYTHON</string>
+    <string>-u</string>
     <string>$INSTALL_DIR/rustchain_linux_miner.py</string>
+    <string>--wallet</string>
+    <string>$WALLET_NAME</string>
   </array>
   <key>RunAtLoad</key>    <true/>
   <key>KeepAlive</key>    <true/>
@@ -331,6 +335,7 @@ SVCEOF
     <key>WALLET_NAME</key> <string>$WALLET_NAME</string>
     <key>NODE_URL</key>    <string>$NODE_URL</string>
     <key>THREADS</key>     <string>$RECOMMENDED_THREADS</string>
+    <key>PYTHONUNBUFFERED</key> <string>1</string>
   </dict>
   <key>StandardOutPath</key>  <string>$INSTALL_DIR/miner.log</string>
   <key>StandardErrorPath</key> <string>$INSTALL_DIR/miner-error.log</string>
@@ -359,10 +364,10 @@ summary() {
   echo -e "  Multiplier:${GREEN} ${MULTIPLIER}x${NC}"
   echo ""
   echo -e "  ${BOLD}To start mining:${NC}"
-  echo -e "  cd $INSTALL_DIR && $PYTHON rustchain_linux_miner.py"
+  echo -e "  cd $INSTALL_DIR && $PYTHON -u rustchain_linux_miner.py --wallet $WALLET_NAME"
   echo ""
   echo -e "  ${BOLD}Check your balance:${NC}"
-  echo -e "  curl -sk '$NODE_URL/wallet/$WALLET_NAME' | python3 -m json.tool"
+  echo -e "  curl -sk '$NODE_URL/wallet/balance?miner_id=$WALLET_NAME' | python3 -m json.tool"
   echo ""
   echo -e "  ${BOLD}Join the community:${NC}"
   echo -e "  Discord: https://discord.gg/rustchain"
--- a/setup.sh
+++ b/setup.sh
@@ -367,7 +367,7 @@ summary() {
   echo -e "  cd $INSTALL_DIR && $PYTHON -u rustchain_linux_miner.py --wallet $WALLET_NAME"
   echo ""
   echo -e "  ${BOLD}Check your balance:${NC}"
-  echo -e "  curl -sk '$NODE_URL/balance/$WALLET_NAME' | python3 -m json.tool"
+  echo -e "  curl -sk '$NODE_URL/wallet/balance?miner_id=$WALLET_NAME' | python3 -m json.tool"
   echo ""
   echo -e "  ${BOLD}Join the community:${NC}"
   echo -e "  Discord: https://discord.gg/rustchain"
--- a/setup.sh
+++ b/setup.sh
@@ -367,7 +367,7 @@ summary() {
   echo -e "  cd $INSTALL_DIR && $PYTHON -u rustchain_linux_miner.py --wallet $WALLET_NAME"
   echo ""
   echo -e "  ${BOLD}Check your balance:${NC}"
-  echo -e "  curl -sk '$NODE_URL/wallet/balance?miner_id=$WALLET_NAME' | python3 -m json.tool"
+  echo -e "  curl -sk '$NODE_URL/balance/$WALLET_NAME' | python3 -m json.tool"
   echo ""
   echo -e "  ${BOLD}Join the community:${NC}"
   echo -e "  Discord: https://discord.gg/rustchain"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -353,7 +353,8 @@ Type=simple
 User=$(whoami)
 WorkingDirectory=${INSTALL_DIR}
 Environment="RUSTCHAIN_WALLET=${wallet}"
-ExecStart=${py} ${INSTALL_DIR}/rustchain_linux_miner.py
+Environment="PYTHONUNBUFFERED=1"
+ExecStart=${py} -u ${INSTALL_DIR}/rustchain_linux_miner.py --wallet ${wallet}
 Restart=always
 RestartSec=30
 StandardOutput=append:/var/log/rustchain-miner.log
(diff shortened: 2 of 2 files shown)


william08190:
Follow-up added in commit `a8cd3ec`: this PR also fixes the `setup.sh` summary balance-check command reported in #5712.

The previous summary command used `$NODE_URL/wallet/$WALLET_NAME`, which returns 404 on the active node shape. It now prints the documented wallet-balance route:

```bash
curl -sk '$NODE_URL/wallet/balance?miner_id=$WALLET_NAME' | python3 -m json.tool
```

Re-ran validation after the update:

```bash
bash -n setup.sh
bash -n scripts/install.sh
git diff --check
```

All passed.

JeremyZeng77:
Reviewed the service-install changes.

What I checked:
- `bash -n scripts/install.sh setup.sh` exits 0 locally.
- The added `-u` / `PYTHONUNBUFFERED=1` pairing is redundant but safe; it covers both direct interpreter buffering and environment-based buffering.
- The summary balance URL now targets an existing GET route: `/wallet/balance?miner_id=...`.

No blocking issues found in this diff.

william08190:
Second follow-up added in commit `1546b92`: this PR now also fixes #5713.

While checking the service templates, I noticed they set `WALLET_NAME` / `RUSTCHAIN_WALLET` environment variables, but `rustchain_linux_miner.py` only uses the CLI `--wallet` argument. Without `--wallet`, a service-launched miner can fall back to `_gen_wallet()` and mine under a random wallet instead of the configured payout/miner ID.

The PR now passes the configured wallet explicitly in both installer paths:

- `setup.sh` systemd: `python -u rustchain_linux_miner.py --wallet $WALLET_NAME`
- `setup.sh` launchd: adds `--wallet` / `$WALLET_NAME` to `ProgramArguments`
- `scripts/install.sh` systemd: `python -u rustchain_linux_miner.py --wallet ${wallet}`
- `scripts/install.sh` launchd: adds `--wallet` / `${wallet}` to `ProgramArguments`

Validation after this update:

```bash
bash -n setup.sh
bash -n scripts/install.sh
git diff --check
```

All passed.

setup.sh
```diff
@@ -359,10 +364,10 @@ summary() {
   echo -e "  Multiplier:${GREEN} ${MULTIPLIER}x${NC}"
   echo ""
   echo -e "  ${BOLD}To start mining:${NC}"
-  echo -e "  cd $INSTALL_DIR && $PYTHON rustchain_linux_miner.py"
+  echo -e "  cd $INSTALL_DIR && $PYTHON -u rustchain_linux_miner.py --wallet $WALLET_NAME"
   echo ""
   echo -e "  ${BOLD}Check your balance:${NC}"
-  echo -e "  curl -sk '$NODE_URL/wallet/$WALLET_NAME' | python3 -m json.tool"
+  echo -e "  curl -sk '$NODE_URL/wallet/balance?miner_id=$WALLET_NAME' | python3 -m json.tool"
```
siteportco: This hint now uses `$WALLET_NAME` as the `miner_id`, but the Linux miner still treats those as different identifiers. `--wallet` is passed into `LocalMiner(..., wallet=args.wallet)` and then sent as `miner` / `miner_pubkey`; the `miner_id` fields in attestation/enrollment are derived separately via `_miner_id_from_hw(self.hw_info)`. `LocalMiner.check_balance()` also still queries `/balance/{self.wallet}`.

That means this summary command can show a zero or misleading balance for a correctly configured service install, because it asks `/wallet/balance` for `miner_id=<wallet name>` rather than the hardware-derived miner id or the legacy wallet/pubkey path. I would either keep the hint aligned with the miner's current lookup (`/balance/$WALLET_NAME`) or update the miner/server plumbing so the setup wallet name is actually the `miner_id` everywhere.

siteportco (commented):
Focused review pass completed.

Validation run locally:
- `C:\Program Files\Git\bin\bash.exe -n setup.sh` passed.
- `C:\Program Files\Git\bin\bash.exe -n scripts/install.sh` passed.
- `git diff --check origin/main...HEAD` passed.
- `python -m py_compile miners/linux/rustchain_linux_miner.py` passed.

The `-u` / `PYTHONUNBUFFERED=1` service changes look syntactically sound, and passing `--wallet` fixes the service path no longer falling back to `_gen_wallet()`. I left one inline concern on the new balance-check hint because the current Linux miner still treats `wallet` and hardware-derived `miner_id` as separate values, so `/wallet/balance?miner_id=$WALLET_NAME` may mislead users even though the service itself starts with the intended wallet.

TJCurnutte (approved):
Reviewed current head `1546b92c2d5b6d0df26b99ac3b46187ff5ac24d7`.

The service-launch changes are tight and match the miner CLI behavior: generated systemd/launchd commands now run Python unbuffered and pass the configured wallet explicitly instead of relying on environment variables that the Linux miner does not consume as its wallet source.

Validation performed:

```bash
git diff --check origin/main...HEAD -- scripts/install.sh setup.sh
bash -n scripts/install.sh
bash -n setup.sh
python3 -B -m pytest -q tests/test_setup_sh_downloads.py tests/test_scripts_installer_downloads.py --tb=short
```

Results:

```text
bash -n passed for scripts/install.sh and setup.sh
5 passed, 1 warning in 0.07s
```

Additional static/source checks:

```text
scripts/install.sh: PYTHONUNBUFFERED=1 present, `-u` present, `--wallet` present, downloads rustchain_linux_miner.py
setup.sh: PYTHONUNBUFFERED=1 present, `-u` present, `--wallet` present, downloads rustchain_linux_miner.py
miners/linux/rustchain_linux_miner.py: argparse defines `--wallet` and passes `wallet=args.wallet` into LocalMiner
node/rustchain_v2_integrated_v2.2.1_rip200.py: `/wallet/balance` route exists
```

I also probed the documented balance URL over TLS:

```text
curl -sS -m 10 -o /tmp/rustchain-balance-probe.json -w 'http=%{http_code} ssl=%{ssl_verify_result}\n' 'https://rustchain.org/wallet/balance?miner_id=review-probe'
http=200 ssl=0
{"amount_i64":0,"amount_rtc":0.0,"miner_id":"review-probe"}
```

Approved.

jaxint (approved):

BossChaos (commented):
## Code Review — Run Service Installs Unbuffered

### ✅ Strengths
- Correct fix: `PYTHONUNBUFFERED=1` and `-u` flag ensure logs appear in real-time
- Updates both systemd and launchd configurations
- Also fixes missing `--wallet` argument in systemd ExecStart

### ⚠️ Issues

**1. Critical Bug Fix**
The original systemd config was missing `--wallet ${wallet}`, meaning the miner would start without a wallet! This is a critical bug that would cause the miner to fail.

**2. Good Consistency**
Both Linux (systemd) and macOS (launchd) configs are updated with the same fix.

### 📋 Summary
Important bug fix. The missing `--wallet` argument was a real issue. Suggested: 8-10 RTC.

JeremyZeng77:
I rechecked the latest follow-up. Passing the configured wallet into the service command is the right fix, and shell syntax validation passes for both installer scripts.

One remaining issue: the summary balance command currently changed to:

```bash
curl -sk "$NODE_URL/balance/$WALLET_NAME" | python3 -m json.tool
```

On the live node shape, that URL returns 404 for the wallet address form, while the documented route works:

```text
https://rustchain.org/wallet/balance?miner_id=<wallet> -> 200
https://rustchain.org/balance/<wallet> -> 404
```

So I would keep the service `--wallet` changes, but update the summary command back to the documented wallet balance endpoint:

```bash
curl -sk "$NODE_URL/wallet/balance?miner_id=$WALLET_NAME" | python3 -m json.tool
```

Verification run:

```text
bash -n setup.sh
bash -n scripts/install.sh
```

PR #5716, opened 2026-05-19
fix(fuzz): make attestation runner console output encoding-safe

### LINKED ISSUE #5699: fuzz_attestation_runner crashes on non-UTF-8 Windows consoles before fuzzing
## Summary

`tests/fuzz_attestation_runner.py` can crash before it sends any fuzz payloads on Windows consoles using a non-UTF-8 encoding, because the runner prints emoji/status glyphs directly to stdout. In the default PowerShell/GBK environment, this raises `UnicodeEncodeError` on the initial banner.

## Reproduction

```powershell
$env:RC_ADMIN_KEY='0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef'
python -B tests/fuzz_attestation_runner.py --count 120 --ci
```

Observed traceback:

```text
Traceback (most recent call last):
  File "tests/fuzz_attestation_runner.py", line 943, in <module>
    sys.exit(main())
  File "tests/fuzz_attestation_runner.py", line 926, in main
    crashes = run_fuzz(...)
  File "tests/fuzz_attestation_runner.py", line 721, in run_fuzz
    print(f"?? RustChain Attestation Fuzz Runner")
UnicodeEncodeError: 'gbk' codec can't encode character '\U0001f525' in position 0: illegal multibyte sequence
```

## Expected behavior

The fuzzer should not depend on the host console encoding. It should either print ASCII-safe output or configure stdout/stderr to replace unencodable characters so it can reach the actual fuzzing phase.

## Impact

This blocks the attestation fuzzer on common Windows/GBK shells and makes the `--ci` mode fail for an environment/tooling reason instead of reporting endpoint results.

## Proposed fix

Configure stdout/stderr with `errors="replace"` when Python supports `reconfigure()`. This keeps current output on UTF-8 terminals and degrades unencodable glyphs to replacement characters on legacy consoles.

Payout route: PayPal `https://paypal.me/Jeremytsangchina`; USDT TRC20 `TWupTAoFjxR2VmqUUvzhfmUwXVZR6Kvz1Y`. If RustChain bug bounties are RTC-only, please tell me the expected wallet/miner-id format and I will provide it.

william08190:
## Summary
- configure fuzz runner stdout/stderr with errors=replace when stream.reconfigure() is available
- keep UTF-8 terminals unchanged while preventing UnicodeEncodeError on legacy Windows/GBK consoles
- add a regression test that writes the banner glyph through a GBK stream without crashing

Fixes #5699.

## Bounty / wallet
- RTC wallet: RTC219810c1e939b0f1a8887a5528949da0e5c97b78

## Validation
- python3.12 -B -m py_compile tests/fuzz_attestation_runner.py tests/test_fuzz_attestation_runner_console.py
- python3.12 -B - <<'PY' ... direct GBK TextIOWrapper regression snippet ... PY
- git diff --check

### DIFF
--- a/tests/fuzz_attestation_runner.py
+++ b/tests/fuzz_attestation_runner.py
@@ -41,6 +41,21 @@ from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
+
+def _configure_console_output(streams=(sys.stdout, sys.stderr)) -> None:
+    """Avoid UnicodeEncodeError on legacy consoles while preserving UTF-8 output."""
+    for stream in streams:
+        reconfigure = getattr(stream, "reconfigure", None)
+        if not callable(reconfigure):
+            continue
+        try:
+            reconfigure(errors="replace")
+        except (OSError, ValueError):
+            continue
+
+
+_configure_console_output()
+
 # ---------------------------------------------------------------------------
 # Configuration
 # ---------------------------------------------------------------------------
--- a/tests/test_fuzz_attestation_runner_console.py
+++ b/tests/test_fuzz_attestation_runner_console.py
@@ -1,0 +1,34 @@
+"""Regression tests for console-safe attestation fuzz runner output."""
+
+import importlib.util
+import io
+from pathlib import Path
+
+
+MODULE_PATH = Path(__file__).resolve().parents[1] / "tests" / "fuzz_attestation_runner.py"
+
+
+def _load_runner_module():
+    spec = importlib.util.spec_from_file_location("fuzz_attestation_runner", MODULE_PATH)
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_console_output_replaces_unencodable_glyphs():
+    runner = _load_runner_module()
+    buffer = io.BytesIO()
+    stream = io.TextIOWrapper(buffer, encoding="gbk", errors="strict")
+
+    runner._configure_console_output((stream,))
+    stream.write("🔥 RustChain Attestation Fuzz Runner")
+    stream.flush()
+
+    assert buffer.getvalue().startswith(b"? RustChain")
+
+
+def test_console_output_ignores_streams_without_reconfigure():
+    runner = _load_runner_module()
+
+    runner._configure_console_output((object(),))


TJCurnutte (approved):
Approved after focused validation of the console-encoding fix at head `5b64c566570f00fe382f58c27e9a791f86f3a388`.

Proof run from a disposable worktree:

- `git diff --check origin/main...HEAD -- tests/fuzz_attestation_runner.py tests/test_fuzz_attestation_runner_console.py` passed.
- `python3 -B -m py_compile tests/fuzz_attestation_runner.py tests/test_fuzz_attestation_runner_console.py` passed.
- `python3 -B -m pytest -q tests/test_fuzz_attestation_runner_console.py --tb=short` returned `2 passed, 1 warning in 0.05s`.
- `PYTHONIOENCODING=cp936:strict python3 -B tests/fuzz_attestation_runner.py --count 0` on this PR returned `rc=0` and emitted the banner as `? RustChain Attestation Fuzz Runner`, so the legacy GBK/Windows-style console path no longer crashes.
- The same cp936 strict runner probe on `origin/main` returned `rc=1` with `UnicodeEncodeError: 'gbk' codec can't encode`, which proves the regression being fixed.

The change is narrow: it only calls `stream.reconfigure(errors="replace")` when available, tolerates streams that cannot be reconfigured, and leaves normal UTF-8 output untouched. No blocker found in the changed files.

jaxint (approved):

BossChaos (commented):
## Code Review — Fuzz Attestation Runner Console Output

### ✅ Strengths
- Simple, correct fix for UnicodeEncodeError on legacy consoles
- Uses `stream.reconfigure(errors="replace")` which is the right approach
- Only affects output streams, no logic changes

### ⚠️ Issues

**1. Missing Edge Case**
If the stream has no `reconfigure` method (very old Python), the code silently skips. This is correct behavior, but a warning log would help debugging.

**2. Test Coverage**
Should include a test that verifies:
- Unicode strings don't crash the fuzzer
- Output is properly encoded/replaced

### 📋 Summary
Clean, minimal fix. Good practice. Suggested: 5 RTC.

JeremyZeng77:
Reviewed the console-safe attestation fuzz runner update.

What I verified:

```text
python -B -m pytest -q tests/test_fuzz_attestation_runner_console.py --tb=short -p no:cacheprovider
2 passed, 1 warning
python -B -m py_compile tests/fuzz_attestation_runner.py tests/test_fuzz_attestation_runner_console.py
```

The helper is appropriately defensive: it only calls `reconfigure` when available and ignores streams where reconfiguration is not supported. The regression test reproduces the Windows/legacy-console class of failure by writing a non-encodable glyph through a strict GBK stream after applying `errors="replace"`.

No blocker found. This is stronger than a bare top-level `sys.stdout.reconfigure(...)` because it is testable and handles stderr/unsupported stream objects cleanly.

PR #2285, opened 2026-04-18
test: add unit tests for exceptions module (bounty #1589)

FlintLeng:
## Summary
Add comprehensive pytest unit tests for `rustchain_sdk.exceptions` module.

Closes bounty #1589 (2 RTC)

## Tests Added
- **TestRustChainError**: message, details, default dict isolation, repr, inheritance, empty message
- **TestConnectionError**: basic creation, details, inheritance check
- **TestAPIError**: status codes, response body, repr with/without status, edge case (zero status)
- **TestRPCError**: method tracking, details, empty method, long method name
- **TestSimpleExceptions**: parametrized tests for all 8 pass-through exception classes

## Coverage
- 20+ test cases covering all 11 exception classes
- Edge cases: empty messages, None details, zero status codes, special characters in repr
- Inheritance hierarchy validation
- Default dict mutation isolation test

## File
`sdk/python/rustchain_sdk/tests/test_exceptions.py`

**Wallet:** kuanglaodi2-sudo

PR #5717, opened 2026-05-19
Fix tag-filter pagination in memory store

junn-dev:
## Summary
- Fix `AgentMemoryStore.get_references()` so tag-filtered queries apply `limit`/`offset` to the filtered result set (not the unfiltered rows).
- Add regression coverage for tag-filter pagination semantics.
- Remove the scratch PR body file from the committed tree.

## Payment
RTC wallet: `RTC29WwMjwcaFeTTQqKaMNmFUFLYz3f`

## Tests
- `python -m pytest bounties/issue-2285/tests/test_memory.py -q`
- `python -m py_compile bounties/issue-2285/src/memory_store.py bounties/issue-2285/tests/test_memory.py`
- `git diff --check`


william08190 (changes requested):
I validated the functional change locally and the memory-store tests pass, but I think this needs a small cleanup before merge.

Validation I ran on the PR head:

```bash
python3.12 -B -m unittest bounties/issue-2285/tests/test_memory.py
python3.12 -B -m py_compile bounties/issue-2285/src/memory_store.py bounties/issue-2285/tests/test_memory.py
```

Results: 39 unittest tests passed, py_compile passed, and diff-check passed.

Requested changes:

1. Please remove `pr-body-issue-2285.md` from the committed diff. It appears to be PR-description scratch content rather than runtime code, tests, docs, or bounty implementation material. Keeping it in the repo adds a stale wallet/payment note and duplicate PR body text to the tree.

2. Please fix the `Closes #2285` reference in the PR body and in that scratch file if you keep any replacement documentation. In this repository, `#2285` resolves to an old closed PR (`test: add unit tests for exceptions module`), not this memory-store pagination task. Leaving it as-is can confuse maintainer triage and may accidentally attach this change to the wrong historical claim.

The actual `get_references()` pagination approach looks reasonable for the current JSON-tag storage model: the no-tag path preserves SQL paging, and the tag-filter path pages through ordered rows until it accumulates filtered matches, so `limit` and `offset` are applied to the filtered result set. After the metadata cleanup above, I would be comfortable with the code/test portion.

TJCurnutte (approved):
Approved after focused validation of the tag-filter pagination fix at head `08099f7ce30dcd565c767c4af4ab456880f3965f`.

Proof run from a disposable worktree:

- `git diff --check origin/main...HEAD -- bounties/issue-2285/src/memory_store.py bounties/issue-2285/tests/test_memory.py pr-body-issue-2285.md` passed.
- `python3 -B -m py_compile bounties/issue-2285/src/memory_store.py bounties/issue-2285/tests/test_memory.py` passed.
- `python3 -B -m pytest -q bounties/issue-2285/tests/test_memory.py --tb=short` returned `39 passed in 0.07s`.
- A direct `AgentMemoryStore(':memory:')` probe with 30 high-importance non-matching rows plus 6 lower-importance `target` rows returned `first_page ['target-5', 'target-4', 'target-3']` and `second_page ['target-2', 'target-1']` for `tags=['target']`, proving that `limit` and `offset` are now applied after tag filtering.
- The same probe on `origin/main` returned empty tagged pages and failed the assertion, confirming the prior SQL-page-before-Python-filter bug.

The scan loop is bounded by requested page windows, keeps the non-tag fast path on SQL `LIMIT/OFFSET`, and fixes the semantics without changing the storage schema. No blocker found.

jaxint (approved):

BossChaos (commented):
## Code Review — Fix Tag-Filter Pagination in Memory Store

### ✅ Strengths
- Correctly identifies the pagination bug (post-filter reduces result count below limit)
- Clean SQL-based solution with EXISTS clause for tag filtering
- Good use of parameterized queries

### ⚠️ Issues

**1. Performance Impact**
The EXISTS subquery for each tag could be slow on large datasets:
```sql
EXISTS (SELECT 1 FROM json_each(tags) WHERE json_each.value = ?)
```
For N tags, this is N subqueries per row. Consider:
```sql
SELECT COUNT(*) FROM json_each(tags) WHERE json_each.value IN (?, ?, ?) = N
```

**2. Backward Compatibility**
The original code did post-filtering in Python. This PR moves it to SQL. If any code relied on the old behavior (e.g., getting ALL results then filtering), this could break things.

**3. Missing Test**
No test case is included to verify the pagination fix works correctly with:
- Multiple tags
- Edge case: no matching tags
- Edge case: more tags than limit

### 📋 Summary
Good fix with correct approach. Needs test coverage. Suggested: 8-10 RTC.

junn-dev:
Thanks, fixed both metadata issues.

- Removed pr-body-issue-2285.md from the committed diff in b6d2c47.
- Updated the PR body to remove the incorrect Closes #2285 reference.
- Re-ran validation: 39 memory-store tests passed, py_compile passed, and git diff --check origin/main...HEAD passed.

wybbb123123:
## Review: changes requested

I reviewed PR #5717 (`Fix tag-filter pagination in memory store`) and ran the touched memory-store/action tests locally.

Validation run:

- `python3 -B -m py_compile bounties/issue-2285/src/memory_store.py bounties/issue-2285/tests/test_memory.py .github/actions/rtc-auto-bounty/award_rtc.py .github/actions/rtc-auto-bounty/test_award_rtc.py`
- `uv run --with pytest python -B -m pytest -q bounties/issue-2285/tests/test_memory.py .github/actions/rtc-auto-bounty/test_award_rtc.py --tb=short -p no:cacheprovider`
- Result: `90 passed, 9 subtests passed`

Finding:

This PR is titled as a memory-store pagination fix, but it also removes the `rtc-api-url` input from the RTC auto-bounty action and forces live transfers back through `http://{vps_host}:8099/wallet/transfer`. That is a compatibility and security regression for deployments that intentionally configured a full HTTPS transfer endpoint.

Minimal repro on this PR branch:

```text
INPUT_RTC_API_URL=https://rustchain.org/wallet/transfer
INPUT_RTC_VPS_HOST=
INPUT_DRY_RUN=false

has rtc_api_url attr: False
validate: INPUT_RTC_VPS_HOST is required (unless dry-run is enabled)
```

Before this PR, the action accepted a full transfer URL and the tests covered that path. After the PR, the action ignores the HTTPS URL input entirely, rejects the configuration, and `transfer_rtc()` only constructs a plaintext `http://...:8099/wallet/transfer` URL. That can break existing workflow secrets and can also send `X-Admin-Key` over plain HTTP when the caller previously used an HTTPS API URL.

Suggested fix:

- Keep the memory-store pagination fix scoped separately from unrelated action/workflow changes.
- Restore `rtc-api-url` support and its tests if the action still needs to support HTTPS/public transfer endpoints.
- If removing `rtc-api-url` is intentional, document the migration and avoid forcing admin-key transfers onto plaintext HTTP.

JeremyZeng77:
Reviewed the tag-filter pagination fix for the issue-2285 memory store.

What I verified:

```text
python -B -m pytest -q bounties/issue-2285/tests/test_memory.py --tb=short -p no:cacheprovider
39 passed

python -B -m py_compile bounties/issue-2285/src/memory_store.py bounties/issue-2285/tests/test_memory.py
```

The change preserves the existing SQL `LIMIT/OFFSET` fast path when no tags are supplied, and for tag-filtered reads it now applies `offset`/`limit` to the filtered result set instead of only filtering the first SQL page. The added regression covers the case where matching tagged records are lower-ranked than non-matching records. No blocker found in this pass.

junn-dev:
Payment info correction: please use my current Ed25519-backed RustChain wallet for any payout on this PR:

`RTC0bc961e7a95d320d2707470643fa35deb1a26d09`

Earlier comments/PR bodies may contain an invalid or legacy wallet value; this is the corrected wallet generated from the repo's Ed25519 wallet flow.

PR #5723, opened 2026-05-19
feat: filter websocket subscriptions

### LINKED ISSUE #2736: [Feature] 支持WebSocket实时区块订阅
## 功能描述\n添加WebSocket接口，支持实时订阅新区块、新交易等事件。\n\n## 使用场景\n- 钱包应用实时更新余额\n- 区块浏览器实时展示区块\n- DApp监听链上事件\n\n## 建议实现\n1. 添加WebSocket server模块\n2. 定义订阅消息格式\n3. 支持按地址、区块高度过滤\n\n---\n认领奖励: 2 RTC\n钱包地址: RTC019e78d600fb3131c29d7ba80aba8fe644be426e

jamilahmadzai:
## Summary
- add explicit WebSocket subscription normalization for block and transaction channels
- support minimum block height and wallet/address filters for SocketIO and JSON-RPC subscribers
- add transaction broadcast support while keeping the legacy broad feed behavior
- cover the new subscription filters with focused regression tests

Fixes #2736

/claim #2736

## Validation
- `uv run --no-project --with pytest python -m pytest node/tests/test_websocket_feed_json_rpc.py -q`
- `python3 -B -m py_compile node/websocket_feed.py node/tests/test_websocket_feed_json_rpc.py`
- `git diff --check`

## Payout
wallet: RTCd1acb2189e9f36df2b5393c3c27a867c3c32b116
Payment method: RTC wallet `RTCd1acb2189e9f36df2b5393c3c27a867c3c32b116`

### DIFF
--- a/node/tests/test_websocket_feed_json_rpc.py
+++ b/node/tests/test_websocket_feed_json_rpc.py
@@ -7,7 +7,7 @@ import sys
 
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
-from websocket_feed import WebSocketFeed
+from websocket_feed import BlockEvent, WebSocketFeed
 
 
 def test_eth_subscribe_mining_stats_returns_subscription_and_stats():
@@ -68,7 +68,7 @@ def test_json_rpc_rejects_unknown_method():
 def test_json_rpc_rejects_unknown_subscription():
     feed = WebSocketFeed()
     response = feed.handle_json_rpc_message(
-        {"jsonrpc": "2.0", "id": 3, "method": "eth_subscribe", "params": ["newHeads", {}]}
+        {"jsonrpc": "2.0", "id": 3, "method": "eth_subscribe", "params": ["newFilter", {}]}
     )
 
     assert response["id"] == 3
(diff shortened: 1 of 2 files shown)


wybbb123123:
## Review: changes requested

I reviewed PR #5723 (`feat: filter websocket subscriptions`) and ran the WebSocket feed tests locally.

Validation run:

- `python3 -B -m py_compile node/websocket_feed.py node/tests/test_websocket_feed_json_rpc.py`
- `uv run --with pytest --with flask-socketio python -B -m pytest -q node/tests/test_websocket_feed_json_rpc.py --tb=short -p no:cacheprovider`
- Result: `11 passed`

Finding:

The new filtered subscriptions do not actually prevent non-matching events from reaching SocketIO clients, because `broadcast_block()` and `broadcast_transaction()` still emit the raw events globally before emitting the filtered subscription envelope:

```python
self.socketio.emit('transaction', payload, namespace='/')
self.emit_filtered_socket_subscribers(TRANSACTION_SUBSCRIPTION, 'transaction', payload)
```

The same pattern exists for blocks. The `subscribe` handler joins a room, but these broadcasts are not sent with `to=room`; they go to the whole namespace. As a result, a client subscribed with an address filter can still receive unrelated transactions through the plain `transaction` event.

Minimal repro with the existing fake socket:

```text
feed.add_socket_subscription('client-1', {'channel': 'transactions', 'filters': {'address': 'RTCalice'}})
feed.broadcast_transaction({'tx_hash': 'tx-1', 'from': 'RTCbob', 'to': 'RTCcarol'})

emitted:
[('transaction', {'tx_hash': 'tx-1', 'from': 'RTCbob', 'to': 'RTCcarol'}, {'namespace': '/'}), ...]
```

There is no `subscription_event` for that payload, which is correct, but the unrelated transaction has already been broadcast globally.

If the goal is subscription filtering, the raw `block` / `transaction` emits should either be routed to the relevant room/filter path or the PR should clearly keep them as documented unfiltered legacy streams. A regression test should assert that a filtered SocketIO subscriber does not receive a non-matching raw event, not just that it does not receive a non-matching `subscription_event`.

ItsOtherMauridian (changes requested):
Thanks for the focused subscription-filtering patch. I found two correctness gaps in the filtered broadcast path that should be fixed before merge.

1. `attestations` is now advertised/accepted as a supported subscription channel, but attestation broadcasts still only emit the legacy `attestation` event. `broadcast_attestation()` never calls `emit_filtered_socket_subscribers('attestations', ...)`, so a SocketIO client can successfully subscribe to `{"channel":"attestations","filters":...}` and then receive no filtered `subscription_event` even when a matching attestation is broadcast.

2. Block/newHeads subscriptions accept `address`/`miner_id` filters, but real `broadcast_block()` payloads cannot satisfy them. `BlockEvent.to_dict()` has height/hash/timestamp/miners_count/reward/epoch/slot only; no address/miner field is present. The new test injects `matching_payload["miner_id"] = "RTCalice"` and calls `emit_json_rpc_subscribers()` directly, so it does not prove real `broadcast_block()` notifications can match. Either block address filters need a real address/miner source in the block payload, or they should be rejected for the block/newHeads channel instead of silently accepting a never-match filter.

Targeted tests on this PR currently pass (`node/tests/test_websocket_feed_json_rpc.py`), but they miss those real broadcast paths. Suggested coverage:
- filtered SocketIO attestation subscriber receives a matching attestation envelope via `broadcast_attestation()`
- `eth_subscribe("newHeads", {"address": ...})` either matches via a real `broadcast_block()` payload or is rejected at subscription time

TJCurnutte (changes requested):
Thanks for adding filtered subscription coverage around the websocket feed. I checked the head at `25ace7078f38ad8f8c60f6edee01c7d235a516b0`.

Validation I ran:

```bash
git diff --check origin/main...HEAD -- node/websocket_feed.py node/tests/test_websocket_feed_json_rpc.py
python3 -B -m py_compile node/websocket_feed.py node/tests/test_websocket_feed_json_rpc.py
pytest -q node/tests/test_websocket_feed_json_rpc.py
```

Those pass, with the focused pytest returning `11 passed in 0.14s`.

I am requesting changes because the new subscription surface advertises an `attestations` channel but the attestation broadcaster never routes through the filtered subscriber path. `normalize_subscription({'channel': 'attestations', 'filters': {'address': 'miner-1'}})` returns `{'channel': 'attestations', 'filters': {'address': 'miner-1'}}` with no error, and `SUPPORTED_SUBSCRIPTION_CHANNELS` / aliases include `attestations`. But a direct probe with a matching stored subscription:

```python
feed.add_socket_subscription('client-1', {'channel': 'attestations', 'filters': {'address': 'miner-1'}})
feed.broadcast_attestation(AttestationEvent(miner_id='miner-1', device_arch='x86', multiplier=1.0, timestamp=1.0, epoch=7, weight=2.0, ticket_id='t1'))
```

emitted only the legacy global event:

```python
[('attestation', {'miner_id': 'miner-1', 'device_arch': 'x86', 'multiplier': 1.0, 'timestamp': 1.0, 'epoch': 7, 'weight': 2.0, 'ticket_id': 't1'}, {'namespace': '/'})]
subscription_events = []
```

So a client can successfully subscribe to `attestations` with a miner/address filter and then never receive the filtered `subscription_event`, even when the event matches. Please either wire `broadcast_attestation()` into `emit_filtered_socket_subscribers('attestations', 'attestation', attestation.to_dict())` with a regression test, or remove/reject that advertised channel until it is actually filterable.

JeremyZeng77:
Reviewed the filtered WebSocket subscription change.

What I verified:

```text
python -B -m pytest -q node/tests/test_websocket_feed_json_rpc.py --tb=short -p no:cacheprovider
11 passed

python -B -m py_compile node/websocket_feed.py node/tests/test_websocket_feed_json_rpc.py
```

The implementation normalizes common `eth_subscribe` aliases (`newHeads`, `newPendingTransactions`) into internal channels, validates bad height filters, removes stale JSON-RPC subscriptions, and emits matching JSON-RPC/socket subscription envelopes only when the address/height filters match. I also checked the unsubscribe ownership path remains scoped to the caller. No blocker found in this pass.

jaxint (approved):

PR #5728, opened 2026-05-19
fix: reject non-finite withdrawal amounts

### LINKED ISSUE #4420: [Bug] Withdraw request route returns 500 for malformed JSON and invalid amounts

iamdinhthuan:
## Summary
- reject NaN and infinity withdrawal amounts before nonce/database work
- keep malformed JSON and invalid amount validation covered by Flask client tests
- make the withdrawal validation tests import the dotted integrated-node filename through importlib and share the module to avoid duplicate Prometheus metric registration

Fixes #4420

## Verification
- python3 -B -m pytest -q node/tests/test_withdraw_amount_validation.py node/tests/test_withdrawal_validation.py --tb=short -p no:cacheprovider
- python3 -B -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py node/tests/test_withdraw_amount_validation.py node/tests/test_withdrawal_validation.py
- git diff --check

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -5127,7 +5127,9 @@ def request_withdrawal():
     try:
         amount = float(raw_amount)
     except (TypeError, ValueError):
-        return jsonify({"error": "amount must be a number", "received": str(type(raw_amount).__name__)}), 400
+        return jsonify({"error": "Amount must be a number", "received": str(type(raw_amount).__name__)}), 400
+    if not math.isfinite(amount):
+        return jsonify({"error": "Amount must be a finite positive number"}), 400
     if amount < 0:
         return jsonify({"error": "amount must be positive"}), 400
 
--- a/node/tests/test_withdraw_amount_validation.py
+++ b/node/tests/test_withdraw_amount_validation.py
@@ -7,6 +7,7 @@ import unittest
 
 NODE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 MODULE_PATH = os.path.join(NODE_DIR, "rustchain_v2_integrated_v2.2.1_rip200.py")
+MODULE_NAME = "rustchain_integrated_withdraw_validation_shared"
 
 
 class TestWithdrawAmountValidation(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


jaxint (approved):

TJCurnutte (approved):
Approved. The non-finite withdrawal amount guard is in the right place: immediately after `float(raw_amount)` and before nonce/database work.

What I checked:
- `math.isfinite(amount)` rejects `NaN`, `inf`, and `-inf` before balance, nonce, or signature logic can operate on non-finite floats.
- Existing malformed JSON, nonnumeric amount, negative amount, and minimum-withdrawal behavior remains covered by the focused tests.
- The shared import name in the tests avoids duplicate module/Prometheus registration during the combined test run.

Validation performed on head `9b3dc1a5ef2ab87eb3371cc423408220f879e67f`:
- `git diff --check origin/main...HEAD -- node/rustchain_v2_integrated_v2.2.1_rip200.py node/tests/test_withdraw_amount_validation.py node/tests/test_withdrawal_validation.py` passed.
- `/Users/curnutte/.hermes/hermes-agent/venv/bin/python3 -B -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py node/tests/test_withdraw_amount_validation.py node/tests/test_withdrawal_validation.py` passed.
- `/Users/curnutte/.hermes/hermes-agent/venv/bin/python3 -B -m pytest -q node/tests/test_withdraw_amount_validation.py node/tests/test_withdrawal_validation.py --tb=short -p no:cacheprovider` returned `12 passed in 0.24s`.
- Temp-DB Flask probe compared `origin/main` with this PR for `/withdraw/request`: `origin/main` let `NaN`/`inf` reach database work and returned 500 in the stripped temp DB path, while this PR returned HTTP 400 with `Amount must be a finite positive number` for `NaN`, `inf`, and `-inf` before DB access.

I do not see a blocking issue in this diff.

JeremyZeng77:
Reviewed the non-finite withdrawal amount fix. Blocking `NaN`/`Infinity` is important here because `float()` accepts those strings and they can bypass normal numeric assumptions.

One small inconsistency remains: the new error says "finite positive number", but the code still only rejects `amount < 0`, so `0` is accepted. If zero-value withdrawals are valid no-op requests, the error text should avoid "positive". If withdrawals must be strictly positive, this should be `amount <= 0` with a regression test for `0` and `"0"`.

The module-loading/test isolation cleanup looks useful and should reduce cross-test contamination.

Scottcjn:
Closing as duplicate of #5719 (william08190, first-poster — filed 4 hours earlier at 11:07 UTC vs 15:52 UTC here, both reference issue #4420 'reject non-finite withdrawal amounts'). #5719 was merged with credit. Thanks for the parallel work; per the first-poster-per-issue rule, credit goes to whoever lands first.

PR #5729, opened 2026-05-19
fix: upsert profile badges by username

### LINKED ISSUE #5727: Bug: profile badge generator duplicates same user because INSERT OR REPLACE has no unique key
## Summary

`profile_badge_generator.py` appears to intend one badge row per GitHub user by using `INSERT OR REPLACE`, but the `profile_badges` table has no `UNIQUE` constraint on `github_username` or any other natural key. As a result, repeated `/api/badge/create` requests for the same user insert duplicate rows instead of replacing/updating the existing badge.

This inflates `/api/badge/stats` and `/api/badge/list`, including `total_bounties_earned`, even when the same user is only updating their custom message.

## Code path

- `profile_badge_generator.py:init_badge_db()` creates `profile_badges` with only `id INTEGER PRIMARY KEY AUTOINCREMENT`.
- `profile_badge_generator.py:create_badge()` executes:

```sql
INSERT OR REPLACE INTO profile_badges
(github_username, wallet_address, badge_type, custom_message, bounty_earned)
VALUES (?, ?, ?, ?, ?)
```

Without a unique constraint, SQLite has no conflict to replace, so every call inserts a new row.

## Reproduction

I used a temporary dependency directory for Flask and a temporary SQLite DB so this does not modify the working tree or a live DB.

```bash
rm -rf /tmp/rustchain-pbg-repro
python3 -m pip install --target /tmp/rustchain-pbg-repro 'flask<4'
cd /Users/william/earn-5-usd/Rustchain
PYTHONPATH=/tmp/rustchain-pbg-repro python3 - <<'PY'
import os, tempfile, importlib.util
with tempfile.TemporaryDirectory() as d:
    path = os.path.join(os.getcwd(), 'profile_badge_generator.py')
    spec = importlib.util.spec_from_file_location('pbg', path)
    mod = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(mod)
    mod.DB_PATH = os.path.join(d, 'badges.db')
    client = mod.app.test_client()
    payload = {'username':'dupuser','wallet':'RTCdemo','badge_type':'contributor','custom_message':'One'}
    r1 = client.post('/api/badge/create', json=payload)
    payload['custom_message'] = 'Two'
    r2 = client.post('/api/badge/create', json=payload)
    stats = client.get('/api/badge/stats').get_json()
    listing = client.get('/api/badge/list').get_json()['badges']
    print('r1_status', r1.status_code, 'success', r1.get_json().get('success'))
    print('r2_status', r2.status_code, 'success', r2.get_json().get('success'))
    print('total_badges', stats['total_badges'])
    print('total_bounties_earned', stats['total_bounties_earned'])
    print('list_count', len(listing))
    print('messages', [row['custom_message'] for row in listing])
PY
```

Observed output:

```text
r1_status 200 success True
r2_status 200 success True
total_badges 2
total_bounties_earned 6.0
list_count 2
messages ['One', 'Two']
```

## Expected behavior

A second badge create/update for the same `github_username` should update or replace the previous row, leaving:

- `total_badges == 1`
- one `/api/badge/list` row for that user
- one counted bounty amount, or a clearly separated audit trail if multiple badge awards are intended

## Actual behavior

Both requests return success, but the same user appears twice and `total_bounties_earned` doubles from `3.0` to `6.0`.

## Impact

Severity: Low/Medium.

This is not a fund-draining path by itself, but it corrupts badge/accounting stats and can make the badge system report inflated earned bounty totals for repeated updates by the same user.

## Suggested fix

Add a uniqueness constraint or index for the intended identity key, probably `github_username`, and use an explicit upsert, for example:

```sql
CREATE UNIQUE INDEX IF NOT EXISTS idx_profile_badges_github_username
ON profile_badges(github_username);
```

Then use `INSERT ... ON CONFLICT(github_username) DO UPDATE ...`, or keep `INSERT OR REPLACE` after the unique index is guaranteed by a migration.

iamdinhthuan:
## Summary
- add a unique index on `profile_badges.github_username` and dedupe existing duplicate rows before creating it
- replace the no-op `INSERT OR REPLACE` behavior with an explicit `ON CONFLICT(github_username) DO UPDATE` upsert
- add a regression test proving a repeated badge create for the same user updates the row and does not double-count bounty totals

Fixes #5727

## Verification
- python3 -B -m pytest -q tests/test_profile_badge_generator.py test_profile_badge_generator.py tests/test_profile_badge_generator_security.py --tb=short -p no:cacheprovider
- python3 -B -m py_compile profile_badge_generator.py tests/test_profile_badge_generator.py test_profile_badge_generator.py tests/test_profile_badge_generator_security.py
- git diff --check

### DIFF
--- a/profile_badge_generator.py
+++ b/profile_badge_generator.py
@@ -27,6 +27,18 @@ def init_badge_db():
                 custom_message TEXT
             )
         ''')
+        cursor.execute('''
+            DELETE FROM profile_badges
+            WHERE id NOT IN (
+                SELECT MAX(id)
+                FROM profile_badges
+                GROUP BY github_username
+            )
+        ''')
+        cursor.execute('''
+            CREATE UNIQUE INDEX IF NOT EXISTS idx_profile_badges_github_username
+            ON profile_badges(github_username)
+        ''')
         conn.commit()
 
 @app.route('/badge/generator')
@@ -189,9 +201,15 @@ def create_badge():
     with sqlite3.connect(DB_PATH) as conn:
         cursor = conn.cursor()
         cursor.execute('''
-            INSERT OR REPLACE INTO profile_badges 
+            INSERT INTO profile_badges
             (github_username, wallet_address, badge_type, custom_message, bounty_earned)
             VALUES (?, ?, ?, ?, ?)
+            ON CONFLICT(github_username) DO UPDATE SET
+                wallet_address = excluded.wallet_address,
+                badge_type = excluded.badge_type,
+                custom_message = excluded.custom_message,
+                bounty_earned = excluded.bounty_earned,
+                created_at = CURRENT_TIMESTAMP
         ''', (username, wallet or None, badge_type, custom_message or None, 3.0))
         conn.commit()
     
--- a/tests/test_profile_badge_generator.py
+++ b/tests/test_profile_badge_generator.py
@@ -51,3 +51,36 @@ def test_create_badge_accepts_valid_json_object(client):
     assert data["success"] is True
     assert "Bug%20Hunter" in data["shield_url"]
     assert _badge_count() == 1
+
+
+def test_create_badge_updates_existing_github_username(client):
+    first = client.post(
+        "/api/badge/create",
+        json={
+            "username": "dupuser",
+            "wallet": "RTCfirst",
+            "badge_type": "contributor",
+            "custom_message": "One",
+        },
+    )
+    second = client.post(
+        "/api/badge/create",
+        json={
+            "username": "dupuser",
+            "wallet": "RTCsecond",
+            "badge_type": "developer",
+            "custom_message": "Two",
+        },
+    )
+
+    assert first.status_code == 200
+    assert second.status_code == 200
+    stats = client.get("/api/badge/stats").get_json()
+    listing = client.get("/api/badge/list").get_json()["badges"]
+
+    assert stats["total_badges"] == 1
+    assert stats["total_bounties_earned"] == 3.0
+    assert len(listing) == 1
+    assert listing[0]["username"] == "dupuser"
+    assert listing[0]["type"] == "developer"
+    assert listing[0]["custom_message"] == "Two"


jaxint (approved):

TJCurnutte (approved):
Approved — this fixes repeated badge creation for the same stored GitHub username and stops the stats endpoint from double-counting those repeats.

Validation I ran against head `3af0a370154fe509daf28fce722b8e6805fe32fd`:

- `git diff --check origin/main...HEAD -- profile_badge_generator.py tests/test_profile_badge_generator.py`
- `python3 -B -m py_compile profile_badge_generator.py tests/test_profile_badge_generator.py test_profile_badge_generator.py tests/test_profile_badge_generator_security.py`
- `PYTHONPATH=. python3 -B -m pytest -q tests/test_profile_badge_generator.py test_profile_badge_generator.py tests/test_profile_badge_generator_security.py --tb=short -p no:cacheprovider` → `24 passed in 0.15s`
- Direct Flask/SQLite origin-main-vs-PR probe: posting `/api/badge/create` twice for `dupuser` on `origin/main` returned 2 DB rows, `total_badges=2`, and `total_bounties_earned=6.0`; this PR returned one updated row (`RTCsecond`, `developer`, `Two`), `total_badges=1`, and `total_bounties_earned=3.0`.
- Existing duplicate cleanup probe: two pre-seeded `dupuser` rows remained two rows on `origin/main`; this PR reduced them to the latest row and created unique `idx_profile_badges_github_username`.

Reasoning: the old `INSERT OR REPLACE` was effectively just an insert because `github_username` was not unique. The new unique index plus explicit `ON CONFLICT(github_username) DO UPDATE` makes the intended upsert behavior real and keeps badge totals from inflating on repeat submissions. Scope note: this is exact stored-username upsert behavior; GitHub-case-insensitive normalization can be handled separately if that is desired.

JeremyZeng77:
Reviewed the profile badge upsert change. Moving from `INSERT OR REPLACE` to `ON CONFLICT DO UPDATE` is a good fix because it preserves the row identity instead of deleting/reinserting on each badge edit.

Two small points to check: first, the update sets `created_at = CURRENT_TIMESTAMP`, so the field now behaves like `updated_at`. If consumers interpret `created_at` as first badge creation time, this will lose that history. Second, the unique index is case-sensitive by default in SQLite, so `Alice` and `alice` can still create separate rows unless GitHub usernames are normalized before insert.

Not blocking if that is the intended behavior, but I would either normalize usernames to lowercase or add a test showing the chosen case-sensitivity behavior.

PR #5732, opened 2026-05-19
fix: deduplicate BCOS directory project loads

### LINKED ISSUE #5731: Bug: BCOS directory project loader duplicates rows because INSERT OR REPLACE has no unique key
## Summary

`bcos_directory.py` uses `INSERT OR REPLACE` while loading `data/projects.json`, but the `projects` table only has an autoincrement primary key and no uniqueness constraint on the natural project identity. Because SQLite only replaces on a primary-key or unique-key conflict, repeatedly loading the same project inserts duplicate rows instead of updating the existing project entry.

This can duplicate BCOS directory entries and inflate/project-stale directory results when the JSON loader runs more than once.

## Code path

- `bcos_directory.py:init_db()` creates `projects` with `id INTEGER PRIMARY KEY AUTOINCREMENT` and `github_repo TEXT NOT NULL`, but no unique key.
- `bcos_directory.py:load_projects_from_json()` executes:

```sql
INSERT OR REPLACE INTO projects
(name, url, github_repo, bcos_tier, latest_sha, sbom_hash, review_note, category)
VALUES (?, ?, ?, ?, ?, ?, ?, ?)
```

Without a unique constraint, there is no conflict to replace, so each load appends another row.

## Minimal reproduction

This uses a temporary SQLite DB and the same schema/insert semantics, with no live service or working-tree DB mutation:

```bash
python3 - <<'PY'
import os, tempfile, sqlite3
with tempfile.TemporaryDirectory() as d:
    db=os.path.join(d,'bcos.db')
    conn=sqlite3.connect(db); c=conn.cursor()
    c.execute('''
        CREATE TABLE IF NOT EXISTS projects (
            id INTEGER PRIMARY KEY AUTOINCREMENT,
            name TEXT NOT NULL,
            url TEXT NOT NULL,
            github_repo TEXT NOT NULL,
            bcos_tier TEXT NOT NULL,
            latest_sha TEXT,
            sbom_hash TEXT,
            review_note TEXT,
            category TEXT,
            created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
        )
    ''')
    for _ in range(2):
        c.execute('''
            INSERT OR REPLACE INTO projects
            (name, url, github_repo, bcos_tier, latest_sha, sbom_hash, review_note, category)
            VALUES (?, ?, ?, ?, ?, ?, ?, ?)
        ''', ('Alpha','https://a','org/repo','tier1','one','h1','n1','cat'))
    conn.commit()
    rows=c.execute('select id,name,github_repo,latest_sha from projects').fetchall()
    print('row_count', len(rows))
    print(rows)
PY
```

Observed:

```text
row_count 2
[(1, 'Alpha', 'org/repo', 'one'), (2, 'Alpha', 'org/repo', 'one')]
```

## Expected behavior

Loading the same `github_repo` repeatedly should leave one directory row for that project, updating/replacing the existing metadata instead of appending duplicates.

## Impact

Severity: Low.

This is not a direct security issue, but it corrupts BCOS directory state and can duplicate project cards/API results whenever the loader is run repeatedly or re-seeded.

## Suggested fix

- Deduplicate existing `projects` rows by `github_repo` during DB init/migration.
- Add a unique index on `projects(github_repo)` so future `INSERT OR REPLACE` statements have a real conflict target.
- Add regression tests proving repeated project loads stay idempotent and pre-existing duplicates do not break the migration.

ItsOtherMauridian:
## Summary

Fixes #5731.

`load_projects_from_json()` used `INSERT OR REPLACE`, but the `projects` table only had an autoincrement `id` primary key. Re-running the JSON loader for the same `github_repo` therefore appended duplicate BCOS directory rows instead of replacing/updating the existing project entry.

This PR:

- deduplicates existing `projects` rows by `github_repo` during `init_db()` before adding the constraint;
- adds a unique index on `projects(github_repo)` so future `INSERT OR REPLACE` calls have a real conflict target;
- adds regression coverage for repeated loads and pre-existing duplicate cleanup.

## Validation

```text
PYTHONPATH=/tmp/rustchain-flask-deps:. python3 -B -m pytest -q tests/test_bcos_directory_security.py --tb=short -p no:cacheprovider
# 6 passed

PYTHONPATH=/tmp/rustchain-flask-deps:. python3 -B -m py_compile bcos_directory.py tests/test_bcos_directory_security.py
# passed
```

Note: local environment needed Flask supplied via a temporary dependency target because the base runner here does not have Flask installed globally.

### DIFF
--- a/bcos_directory.py
+++ b/bcos_directory.py
@@ -61,6 +61,18 @@ def init_db():
             created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
         )
     ''')
+    c.execute('''
+        DELETE FROM projects
+        WHERE id NOT IN (
+            SELECT MAX(id)
+            FROM projects
+            GROUP BY github_repo
+        )
+    ''')
+    c.execute('''
+        CREATE UNIQUE INDEX IF NOT EXISTS idx_projects_github_repo
+        ON projects(github_repo)
+    ''')
     conn.commit()
     conn.close()
 
--- a/bcos_directory.py
+++ b/bcos_directory.py
@@ -64,9 +64,15 @@ def init_db():
     c.execute('''
         DELETE FROM projects
         WHERE id NOT IN (
-            SELECT MAX(id)
-            FROM projects
-            GROUP BY github_repo
+            SELECT keep.id
+            FROM projects AS keep
+            WHERE keep.id = (
+                SELECT candidate.id
+                FROM projects AS candidate
+                WHERE candidate.github_repo = keep.github_repo
+                ORDER BY datetime(candidate.created_at) DESC, candidate.id DESC
+                LIMIT 1
+            )
         )
     ''')
     c.execute('''
--- a/tests/test_bcos_directory_security.py
+++ b/tests/test_bcos_directory_security.py
@@ -91,11 +91,11 @@ def test_bcos_directory_init_deduplicates_existing_projects_before_unique_index(
         ''')
         conn.executemany('''
             INSERT INTO projects
-            (name, url, github_repo, bcos_tier, latest_sha, sbom_hash, review_note, category)
-            VALUES (?, ?, ?, ?, ?, ?, ?, ?)
+            (name, url, github_repo, bcos_tier, latest_sha, sbom_hash, review_note, category, created_at)
+            VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)
         ''', [
-            ('Old RustChain', 'https://old.example', 'Scottcjn/Rustchain', 'L1', 'old', None, None, 'chain'),
-            ('New RustChain', 'https://new.example', 'Scottcjn/Rustchain', 'L2', 'new', None, None, 'chain'),
+            ('New RustChain', 'https://new.example', 'Scottcjn/Rustchain', 'L2', 'new', None, None, 'chain', '2026-01-02 00:00:00'),
+            ('Old RustChain', 'https://old.example', 'Scottcjn/Rustchain', 'L1', 'old', None, None, 'chain', '2026-01-01 00:00:00'),
         ])
 
     bcos_directory.init_db()
--- a/tests/test_bcos_directory_security.py
+++ b/tests/test_bcos_directory_security.py
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: MIT
 
+import json
 import re
+import sqlite3
 
 import bcos_directory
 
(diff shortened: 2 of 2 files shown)


TJCurnutte (approved):
Approved — this fixes the duplicate BCOS directory load path rather than only adding test coverage.

Validation I ran against head `04d278c9574c50b66eba607d7a66ce5f4a3b714c`:

- `git diff --check origin/main...HEAD -- bcos_directory.py tests/test_bcos_directory_security.py`
- `python3 -B -m py_compile bcos_directory.py tests/test_bcos_directory_security.py`
- `PYTHONPATH=. python3 -B -m pytest -q tests/test_bcos_directory_security.py --tb=short -p no:cacheprovider` → `6 passed in 0.05s`
- Direct SQLite origin-main-vs-PR probe: `origin/main` loaded the same `data/projects.json` twice into 2 rows for `Scottcjn/Rustchain` and had no unique index; this PR loaded it into 1 row and created unique `idx_projects_github_repo`.
- Pre-existing duplicate cleanup probe: `origin/main` left both `old` and `new` rows; this PR kept the latest row (`latest_sha='new'`) and then created the unique index without error.

Reasoning: the previous `INSERT OR REPLACE` had no conflict target because only the autoincrement `id` was unique, so repeated JSON loads appended duplicate project rows. Adding a unique index on `github_repo`, after a one-time cleanup of already-duplicated rows, gives the loader a real conflict target and keeps the migration safe for existing DBs. I do not see a blocker in the changed path.

jaxint (approved):

JeremyZeng77:
Reviewed the BCOS directory deduplication patch. Adding a unique index on `github_repo` should make repeated JSON loads idempotent, and the tests cover both fresh load and pre-existing duplicate cleanup.

One operational detail to consider: the cleanup keeps `MAX(id)` for each `github_repo`, not necessarily the row with the newest `created_at` or the row that matches the current `projects.json`. If duplicate rows were inserted in an unusual order or restored from backup, `MAX(id)` may not be the most correct project metadata. That is probably acceptable for this small directory, but it is worth documenting that the newest inserted row wins.

No merge-blocking issue found; the behavior is deterministic and covered by tests.

ItsOtherMauridian:
Addressed the ordering nuance from the review.

New commit keeps the newest duplicate by `created_at`, with `id` only as a deterministic tie-breaker, instead of using `MAX(id)` as the primary rule. I also updated the duplicate-cleanup regression test so the newer `created_at` row wins even when it has the lower inserted id.

Validation:

```bash
PYTHONPATH=/tmp/rustchain-flask-deps:. python3 -B -m pytest -q tests/test_bcos_directory_security.py --tb=short -p no:cacheprovider
# 6 passed
python3 -B -m py_compile bcos_directory.py tests/test_bcos_directory_security.py
git diff --check
```

Scottcjn:
Merged + queued — **Payment: 10 RTC** (Low tier, audit-verified)

Codex audit 6 reviewed this PR (BCOS directory project-load dedupe) as MERGE-LOW: clean, scope-matched title, regression test included, no destructive substitution. Your other 7 PRs (#5740–#5752) are in the audit YELLOW track — file-by-file review pending before merge/pay.

To dispatch the 10 RTC I need your **RTC wallet address** (`RTC` + 40 hex characters). Reply here with your wallet.

— Sophia

ItsOtherMauridian:

PR #5733, opened 2026-05-19
fix: tolerate legacy console encodings in attestation fuzzer

### LINKED ISSUE #5699: fuzz_attestation_runner crashes on non-UTF-8 Windows consoles before fuzzing

iamdinhthuan:
## Summary
- configure stdout/stderr with `errors="replace"` before the attestation fuzzer prints status glyphs
- keep UTF-8 terminals unchanged while preventing `UnicodeEncodeError` on legacy Windows/GBK-style consoles
- add a focused regression test using strict cp1252 streams and the existing emoji/status glyphs

Fixes #5699

## Verification
- python3 -B -m pytest -q tests/test_fuzz_attestation_runner_console_encoding.py --tb=short -p no:cacheprovider
- python3 -B -m py_compile tests/fuzz_attestation_runner.py tests/test_fuzz_attestation_runner_console_encoding.py
- git diff --check

### DIFF
--- a/tests/fuzz_attestation_runner.py
+++ b/tests/fuzz_attestation_runner.py
@@ -54,6 +54,19 @@ FUZZ_STATS_FILE = Path(__file__).parent / "fuzz_stats.json"
 TIMEOUT = 10
 DEFAULT_ITERATIONS = 1000
 
+
+def configure_stdio_encoding(stdout=None, stderr=None) -> None:
+    """Avoid UnicodeEncodeError on legacy consoles while preserving UTF-8 output."""
+    for stream in (stdout or sys.stdout, stderr or sys.stderr):
+        reconfigure = getattr(stream, "reconfigure", None)
+        if reconfigure is None:
+            continue
+        try:
+            reconfigure(errors="replace")
+        except (TypeError, ValueError):
+            pass
+
+
 # Known test values for realistic payloads
 KNOWN_WALLETS = [
     "nox-ventures", "test-miner", "alice", "bob",
@@ -884,6 +897,7 @@ def show_stats() -> None:
 # ---------------------------------------------------------------------------
 
 def main():
+    configure_stdio_encoding()
     parser = argparse.ArgumentParser(
         description="RustChain Attestation Fuzz Runner",
         formatter_class=argparse.RawDescriptionHelpFormatter,
--- a/tests/test_fuzz_attestation_runner_console_encoding.py
+++ b/tests/test_fuzz_attestation_runner_console_encoding.py
@@ -1,0 +1,30 @@
+# SPDX-License-Identifier: MIT
+import importlib.util
+import io
+from pathlib import Path
+
+
+MODULE_PATH = Path(__file__).resolve().parents[1] / "tests" / "fuzz_attestation_runner.py"
+
+
+def load_runner():
+    spec = importlib.util.spec_from_file_location("fuzz_attestation_runner_under_test", MODULE_PATH)
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_configure_stdio_encoding_replaces_unencodable_status_glyphs():
+    runner = load_runner()
+    stdout = io.TextIOWrapper(io.BytesIO(), encoding="cp1252", errors="strict")
+    stderr = io.TextIOWrapper(io.BytesIO(), encoding="cp1252", errors="strict")
+
+    runner.configure_stdio_encoding(stdout=stdout, stderr=stderr)
+
+    print("🔥 RustChain Attestation Fuzz Runner", file=stdout)
+    print("✅ ready", file=stderr)
+    stdout.flush()
+    stderr.flush()
+    assert b"?" in stdout.buffer.getvalue()
+    assert b"?" in stderr.buffer.getvalue()


TJCurnutte (approved):
Approved. This is a narrow, useful fix for the attestation fuzzer's legacy-console failure mode.

What I checked:
- `configure_stdio_encoding()` runs at CLI startup before the parser/fuzzer status output, preserves the stream encoding, and only changes `errors` to `replace`.
- The helper is defensive around streams without `reconfigure()` and around `TypeError`/`ValueError`, so it should not break pytest capture or nonstandard streams.
- The regression test exercises strict `cp1252` stdout/stderr with the same emoji/status glyph class that the runner prints.

Validation performed on head `b1cd152be0477dba8a6316742dce33426d7ab5ff`:
- `git diff --check origin/main...HEAD -- tests/fuzz_attestation_runner.py tests/test_fuzz_attestation_runner_console_encoding.py` passed.
- `/Users/curnutte/.hermes/hermes-agent/venv/bin/python3 -B -m py_compile tests/fuzz_attestation_runner.py tests/test_fuzz_attestation_runner_console_encoding.py` passed.
- `/Users/curnutte/.hermes/hermes-agent/venv/bin/python3 -B -m pytest -q tests/test_fuzz_attestation_runner_console_encoding.py --tb=short -p no:cacheprovider` returned `1 passed in 0.03s`.
- Direct cp1252 probe confirmed the pre-fix path: `origin/main` has the fire-banner output but no `configure_stdio_encoding`, and printing `🔥 RustChain Attestation Fuzz Runner` to a strict `cp1252` stream raises `UnicodeEncodeError`.
- The same probe after the PR helper kept `stdout.encoding == cp1252`, changed `stdout.errors/stderr.errors` to `replace`, and emitted replacement bytes for both `🔥` and `✅` instead of crashing.

I do not see a blocking issue in this diff.

jaxint (approved):

minyanyi (approved):
Validated the legacy-console encoding hardening on current head.

What I checked on `b1cd152`:
- the new helper only calls `reconfigure(errors="replace")` when the stream supports it, so older/non-standard stream objects are left alone
- the focused regression covers cp1252-style streams and confirms status glyphs no longer raise `UnicodeEncodeError`
- local validation: `python -B -m pytest -q tests/test_fuzz_attestation_runner_console_encoding.py --tb=short` -> `1 passed`

Small, targeted change with matching coverage. LGTM.

JeremyZeng77:
Reviewed the console encoding fix. The change is narrowly scoped: it only calls `stream.reconfigure(errors="replace")` at runner startup and keeps the output encoding otherwise unchanged, so it should address legacy Windows/cp1252 console failures without changing normal UTF-8 behavior.

The injected-stream test is useful because it proves the helper works without mutating the real pytest stdout/stderr. I also like that the helper handles streams without `reconfigure()` and ignores `TypeError`/`ValueError`, which matters for wrapped or already-detached streams.

No blocking issues found. One small future improvement would be to include a subprocess-style smoke test for the actual CLI entry point, but the current unit test is enough for this regression because the failure mode is specifically UnicodeEncodeError from status glyphs.

Scottcjn:
Merged + queued — **Payment: 10 RTC** (Low tier, faucet)

Audit: small focused fix to attestation fuzzer for legacy console encodings (CP1252 etc.). Fixes #5699. Two-file change, includes regression test. Per `feedback_faucet_tiers_2026-05-04` this is a clean Low-tier merge.

To dispatch the 10 RTC I need your **RTC wallet address** (`RTC` + 40 hex characters). Couldn't find one in the PR body. Reply here with your wallet and I'll queue the transfer.

— Sophia

PR #5740, opened 2026-05-19
fix: preserve enrolled weights in anti-double-mining settlement

### LINKED ISSUE #5739: Bug: anti-double-mining settlement ignores epoch_enroll weights
## Summary

`anti_double_mining.py` uses `epoch_enroll` as the canonical miner set, but it does not use the canonical `epoch_enroll.weight` snapshot when calculating rewards. Instead, both reward paths recompute each representative miner's weight from `device_arch` via `get_time_aged_multiplier()`.

That makes the anti-double-mining settlement path diverge from the normal RIP-200/RIP-309 settlement path whenever enrolled miners have non-equal snapshot weights. In a delayed settlement or RIP-309-filtered epoch, eligible miners can receive the wrong reward split.

## Reproduction

On current `main`, create two distinct machines in the same epoch with identical `device_arch` but different `epoch_enroll.weight` values:

- `miner-heavy`: `weight = 100.0`
- `miner-light`: `weight = 1.0`
- total reward: `10_100` uRTC

Expected canonical split from the enrolled snapshot is 100:1:

```py
{'miner-heavy': 10000, 'miner-light': 100}
```

Actual anti-double-mining result on current `main` is equal split, because both miners are recomputed from the same arch multiplier:

```py
{'miner-heavy': 5050, 'miner-light': 5050}
```

This reproduces in both `calculate_anti_double_mining_rewards()` and `_calculate_anti_double_mining_rewards_conn()`.

## Impact

Reward correctness / economic fairness bug:

- the settlement path that enables anti-double-mining can silently change the reward distribution;
- RIP-309 or other per-epoch weight adjustments captured in `epoch_enroll.weight` are lost;
- delayed settlement can diverge from the canonical epoch snapshot even when the miner set itself is read from `epoch_enroll`.

## Suggested fix

When `epoch_enroll.weight` exists, read it for the epoch and use the representative miner's snapshot weight after fingerprint failure gating. Fall back to the legacy arch-derived multiplier only for older schemas/tests that do not have the `weight` column.

I have a focused fix and regression coverage ready.

ItsOtherMauridian:
## Summary

Fixes #5739.

The anti-double-mining reward path already prefers `epoch_enroll` as the canonical miner set, but it recomputed each representative miner's reward weight from `device_arch`. That can silently change the payout split versus the normal settlement path when `epoch_enroll.weight` contains per-epoch snapshot weights.

This PR:

- reads `epoch_enroll.weight` when the column exists;
- uses the representative miner's snapshot weight for both anti-double-mining reward paths;
- keeps the existing fallback to arch-derived weights for legacy schemas/tests without a `weight` column;
- adds regression coverage for both `calculate_anti_double_mining_rewards()` and `_calculate_anti_double_mining_rewards_conn()`.

## Repro fixed

Before this PR, two distinct machines with identical `device_arch` but weights `100.0` and `1.0` received an equal split:

```py
{'miner-heavy': 5050, 'miner-light': 5050}
```

After this PR, anti-double-mining settlement preserves the canonical enrolled split:

```py
{'miner-heavy': 10000, 'miner-light': 100}
```

## Validation

```bash
PYTHONPATH=node python3 -B -m pytest -q node/tests/test_anti_double_mining.py::TestAntiDoubleMiningEnrolledWeights --tb=short -p no:cacheprovider
# 2 passed

PYTHONPATH=node python3 -B -m pytest -q \
  node/tests/test_anti_double_mining.py \
  node/tests/test_anti_double_mining_epoch_enroll.py \
  node/tests/test_settlement_integrity.py \
  node/tests/test_rip200_warthog_bonus.py \
# 32 passed

python3 -B -m py_compile node/anti_double_mining.py node/tests/test_anti_double_mining.py

git diff --check
```

### DIFF
--- a/node/anti_double_mining.py
+++ b/node/anti_double_mining.py
@@ -412,6 +412,38 @@ def get_epoch_miner_groups(
     return groups
 
 
+def _get_epoch_enrolled_weights(conn: sqlite3.Connection, epoch: int) -> Dict[str, float]:
+    """Return canonical per-epoch weights from epoch_enroll when available.
+
+    Older test/legacy schemas only have (epoch, miner_pk).  In that case this
+    returns an empty map and callers fall back to the historical arch-derived
+    multiplier path.
+    """
+    try:
+        cols = conn.execute("PRAGMA table_info(epoch_enroll)").fetchall()
+    except sqlite3.Error:
+        return {}
+
+    if not any(col[1] == "weight" for col in cols):
+        return {}
+
+    try:
+        rows = conn.execute(
+            "SELECT miner_pk, weight FROM epoch_enroll WHERE epoch = ?",
+            (epoch,),
+        ).fetchall()
+    except sqlite3.Error:
+        return {}
+
+    weights: Dict[str, float] = {}
+    for miner_pk, weight in rows:
+        try:
+            weights[miner_pk] = max(float(weight or 0.0), 0.0)
+        except (TypeError, ValueError):
+            weights[miner_pk] = 0.0
+    return weights
+
+
 # =============================================================================
 # ANTI-DOUBLE-MINING REWARD CALCULATION
 # =============================================================================
@@ -488,6 +520,7 @@ def calculate_anti_double_mining_rewards(
         
         # Get device arch for each representative miner
         cursor = conn.cursor()
+        enrolled_weights = _get_epoch_enrolled_weights(conn, epoch)
         machine_data = []
         
         for identity_hash, miner_id in representative_map.items():
@@ -510,6 +543,12 @@ def calculate_anti_double_mining_rewards(
             if fingerprint_ok == 0:
                 weight = 0.0
                 logger.info(f"[REWARD] {miner_id[:20]}... fingerprint=FAIL -> weight=0")
+            elif miner_id in enrolled_weights:
+                # Preserve the canonical per-epoch weight snapshot used by the
+                # normal settlement path.  Recomputing from device_arch here can
+                # change the payout split for delayed settlements or RIP-309
+                # filtered weights.
+                weight = enrolled_weights[miner_id]
             else:
                 weight = get_time_aged_multiplier(device_arch, chain_age_years)
             
@@ -756,6 +795,7 @@ def _calculate_anti_double_mining_rewards_conn(
             representative_map[identity_hash] = miner_ids[0]
 
     cursor = conn.cursor()
+    enrolled_weights = _get_epoch_enrolled_weights(conn, epoch)
     machine_data = []
 
     for identity_hash, miner_id in representative_map.items():
@@ -776,6 +816,12 @@ def _calculate_anti_double_mining_rewards_conn(
     for miner_id, device_arch, fingerprint_ok, identity_hash in machine_data:
         if fingerprint_ok == 0:
             weight = 0.0
+        elif miner_id in enrolled_weights:
+            # Preserve the canonical per-epoch weight snapshot used by the
+            # normal settlement path.  Recomputing from device_arch here can
+            # change the payout split for delayed settlements or RIP-309
+            # filtered weights.
+            weight = enrolled_weights[miner_id]
         else:
             weight = get_time_aged_multiplier(device_arch, chain_age_years)
 
--- a/node/tests/test_anti_double_mining.py
+++ b/node/tests/test_anti_double_mining.py
@@ -30,6 +30,7 @@ from anti_double_mining import (
     select_representative_miner,
     get_epoch_miner_groups,
     calculate_anti_double_mining_rewards,
+    _calculate_anti_double_mining_rewards_conn,
     setup_test_scenario,
     GENESIS_TIMESTAMP
 )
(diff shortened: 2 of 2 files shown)


jaxint (approved):

JeremyZeng77:
Reviewed the anti-double-mining enrolled-weight change. Preserving `epoch_enroll.weight` instead of recomputing from `device_arch` is the right direction, especially for delayed settlements and any filtered/adjusted enrolled weights.

One compatibility concern: `_get_epoch_enrolled_weights()` returns an empty map when the `weight` column is missing, but if the column exists and contains scaled integer units from the newer settlement path, this anti-double-mining path still treats the value as a raw float multiplier. That could produce a much larger split than the normal settlement path unless all callers agree on the unit format for `epoch_enroll.weight`.

Suggested verification: add a test with a scaled fixed-point enrolled weight (if that format is possible in production) and confirm anti-double-mining rewards match the normal epoch reward path. If `epoch_enroll.weight` is guaranteed to stay as a REAL multiplier here, a short comment would help prevent future unit drift.

TJCurnutte (approved):
Approved. This fixes the anti-double-mining settlement path so it preserves the canonical `epoch_enroll.weight` snapshot instead of recomputing an equal device-arch split during delayed settlement.

Validation I ran on head `5bb5d309b5b5ab86e9df817aadbe2824c09e82fb`:

```bash
git diff --check origin/main...HEAD -- node/anti_double_mining.py node/tests/test_anti_double_mining.py
PYTHONPATH=node python3 -B -m py_compile node/anti_double_mining.py node/tests/test_anti_double_mining.py
PYTHONPATH=node python3 -m pytest node/tests/test_anti_double_mining.py -q
# 21 passed in 0.23s
```

I also ran an origin-main-vs-PR SQLite probe with two enrolled miners on the same device arch but canonical weights `100.0` and `1.0`:

```text
origin/main head d2995121:
  rewards_open {'miner-heavy': 5050, 'miner-light': 5050}
  rewards_conn {'miner-heavy': 5050, 'miner-light': 5050}

PR head 5bb5d309:
  rewards_open {'miner-heavy': 10000, 'miner-light': 100}
  rewards_conn {'miner-heavy': 10000, 'miner-light': 100}
```

That confirms both the standalone path and the existing-connection path now use the enrolled snapshot and keep the intended 100:1 payout split. The legacy fallback remains intact for schemas without `epoch_enroll.weight`.

kevinyan911 (commented):
## Code Review — PR #5740

**Reviewer:** @kevinyan911

### What this PR does
Adds `_get_epoch_enrolled_weights()` to read the canonical `epoch_enroll.weight` snapshot, then uses it in both `calculate_anti_double_mining_rewards()` and `_calculate_anti_double_mining_rewards_conn()`. Previously both anti-double-mining paths recomputed payout weight from `device_arch`, silently changing the split for delayed settlements or RIP-309 filtered weights.

### Code quality
- `_get_epoch_enrolled_weights()` gracefully handles legacy schemas without a `weight` column — returns `{}` and callers fall back to arch-derived multipliers.
- Explicit comment explains why enrolled weights must be preserved: recomputing from `device_arch` can alter the payout split for delayed settlements.
- `max(float(weight or 0.0), 0.0)` guards against negative/null weights.
- New `TestAntiDoubleMiningEnrolledWeights` test class directly proves the 100:1 enrolled split is preserved (heavy=10,000 / light=100 vs the previous equal 5,050:5,050 split).
- 21 pytests pass including all existing anti-double-mining coverage.

**APPROVED** — correct weight preservation with clean backward compatibility.

---

surim0n (commented):
Reviewed the anti-double-mining enrolled-weight preservation change and ran the focused test file.\n\nWhat I checked:\n- The helper reads `epoch_enroll.weight` only when that column exists, so older schemas without a weight column keep the previous device-architecture multiplier behavior.\n- Both reward paths are updated: the path that opens its own connection and the existing-connection helper now use the same enrolled-weight logic. That matters because otherwise tests could pass on one path while settlement uses the other.\n- Fingerprint failures still force zero weight before the enrolled snapshot is considered, so a stored epoch weight does not override the anti-fraud gate.\n- Invalid/non-numeric enrolled weights are coerced to zero instead of crashing settlement.\n\nVerification run:\n- `PYTHONPATH=node python3 -m pytest node/tests/test_anti_double_mining.py -q` -> 21 passed\n\nNo blocking findings. Small note: the test file currently needs `PYTHONPATH=node` because it inserts the test directory, not the node source directory, into `sys.path`; the PR did not introduce that, but it is worth cleaning up separately so the documented command can be simpler.

BossChaos (commented):
## Code Review - PR #5740: Preserve Enrolled Weights in Anti-Double-Mining Settlement

**Overall: Correct reward calculation fix.**

### What Was Fixed
The anti-double-mining path was recomputing each miner's weight representation from current state rather than using the `epoch_enroll` snapshot. This could lead to weight discrepancies if miners changed their enrolled stake mid-epoch.

### Strengths
- `_get_epoch_enrolled_weights()` extracts the canonical weights from the enrollment snapshot.
- `_calculate_anti_double_mining_rewards_conn()` uses the extracted weights directly.
- Tests added for the new helper function.

### Observations
- This is a subtle but important correctness fix. Double-mining reward distribution depends on accurate per-miner weights.

### Security Verdict
**Standard review - 5-10 RTC.**

motiday (commented):
Reviewed the anti-double-mining enrolled-weight preservation patch with focus on settlement consistency.

What I verified:
- `_get_epoch_enrolled_weights()` reads `epoch_enroll.weight` only when the column exists, so legacy schemas without per-epoch weights continue to fall back to the previous arch-derived multiplier behavior.
- Both anti-double-mining reward paths are updated: `calculate_anti_double_mining_rewards()` and `_calculate_anti_double_mining_rewards_conn()` now consult the same enrolled-weight snapshot. That matters because settlement callers can use either the standalone DB path or an existing connection.
- Fingerprint failure still wins before enrolled weight is considered (`fingerprint_ok == 0 -> weight = 0.0`), so the snapshot does not bypass the anti-fraud gate.
- The regression creates two miners with the same `device_arch` but epoch weights `100.0` and `1.0`, and verifies both paths preserve the intended `10000:100` split instead of recomputing an equal arch-based split.

Local validation passed:

```text
python -B -m py_compile node/anti_double_mining.py node/tests/test_anti_double_mining.py
PYTHONPATH=node python -B -m pytest -q node/tests/test_anti_double_mining.py --tb=short -p no:cacheprovider
21 passed in 0.30s
git diff --check FETCH_HEAD...HEAD -- node/anti_double_mining.py node/tests/test_anti_double_mining.py
```

No blocking findings. This is settlement-sensitive because delayed anti-double-mining settlement should not silently redistribute rewards differently from the canonical `epoch_enroll` snapshot used by normal settlement.

Review bounty #73 claim wallet: RTC9dc7130503bf246d9ca11e2c5b402d5129077f56

surim0n (commented):
Reviewed the anti-double-mining enrolled-weight patch. Preserving `epoch_enroll.weight` is important because anti-double-mining settlement should not recompute a fresh architecture multiplier after the epoch snapshot has already decided canonical weights. Recomputing can distort delayed payouts or bypass later filtering rules that were baked into the enrollment snapshot.

The helper is defensive for legacy schemas without a `weight` column, and both public and existing-connection reward paths now use the same snapshot map. The tests prove unequal enrolled weights survive settlement even when the miners have the same device arch, which is the right regression.

508704820 (commented):

508704820 (commented):

BossChaos (commented):
## PR #5740 Review - Anti-double-mining per-epoch weight snapshot

### Security Analysis - Reward Calculation Fix

This PR fixes a critical bug in anti-double-mining reward calculation where delayed settlements could produce incorrect payout splits.

**Key fix:**

1. **_get_epoch_enrolled_weights()**: Reads canonical per-epoch weight snapshot from epoch_enroll table when available
2. **calculate_anti_double_mining_rewards()**: Uses enrolled_weights[miner_id] instead of recomputing from device_arch - preserves the payout split used by the normal settlement path
3. **Prevents delayed settlement divergence**: When RIP-309 filtered weights are in play, recomputing from device_arch would produce different results than the original snapshot

**Recommendation**: Merge - **important correctness fix**. Ensures reward calculation consistency across delayed settlements and filtered weight scenarios.

PR #5744, opened 2026-05-19
fix: bind claims to registered miner key

### LINKED ISSUE #5743: Bug: claims submission accepts signatures from unregistered miner keys
## Summary

`submit_claim()` verifies that the submitted Ed25519 signature matches the submitted `public_key`, but it does not bind that `public_key` to the miner being claimed.

When a registered miner key is stored alongside the miner record, an unrelated keypair can still sign a syntactically valid claim payload for that miner. Because the wallet address is checked separately, this does not redirect funds, but it lets a third party create a valid pending claim for someone else's miner/epoch and consume the `UNIQUE(miner_id, epoch)` slot.

## Reproduction

On `main`, create a minimal eligible miner with a stored registered public key in `miner_attest_recent.public_key`, then submit a claim signed by a different attacker keypair:

```py
payload = create_claim_payload('victim-miner', epoch, registered_wallet, current_ts)
attacker_signature = attacker_key.sign(payload.encode()).signature.hex()

submit_claim(
    db_path=db,
    miner_id='victim-miner',
    epoch=epoch,
    wallet_address=registered_wallet,
    signature=attacker_signature,
    public_key=attacker_public_key,
    current_slot=current_slot,
    current_ts=current_ts,
)
```

Observed on `main`:

```py
{
  'success': True,
  'claim_id': 'claim_164_victim-miner',
  'status': 'pending',
  'error': None,
}
```

Expected:

The claim public key must match the miner's registered/stored public key when one exists. A valid signature from an unrelated keypair should be rejected before inserting the claim.

## Impact

Claims integrity / denial-of-claim bug:

- anyone can pre-submit a pending claim for an eligible miner if the destination wallet is known;
- the unauthorized pending row blocks the legitimate miner via the unique `(miner_id, epoch)` constraint / pending claim checks;
- audit trails then contain a claim signed by a key not associated with the miner.

## Suggested fix

Before accepting the claim, look up the miner's registered claim/signing key from `miner_wallets` or `miner_attest_recent` when available, and reject with `public_key_mismatch` if the submitted claim key differs.

I have a focused fix and regression test ready.

ItsOtherMauridian:
## Summary

Fixes #5743.

`submit_claim()` verified the signature against the submitted public key, but did not check that the submitted key belonged to the miner being claimed. If a miner already has a registered/stored public key, an unrelated keypair could submit a syntactically valid claim for the same miner/epoch and occupy the pending claim slot.

This PR:

- adds `get_registered_claim_public_key()` to read a stored claim/signing key from `miner_wallets` or `miner_attest_recent` when present;
- rejects claim submissions with `public_key_mismatch` when the submitted public key differs from the registered miner key;
- leaves legacy/no-key schemas compatible by only enforcing the binding when a stored key exists;
- adds a regression test proving a valid attacker signature cannot claim another miner's epoch when the victim's key is registered.

## Validation

```bash
PYTHONPATH=node python3 -B -m pytest -q \
  node/test_claims_security.py::TestClaimsSignatureBypass::test_claim_submission_rejects_unregistered_signing_key \
# 1 passed

PYTHONPATH=/tmp/rustchain-flask-deps:node python3 -B -m pytest -q \
  node/test_claims_security.py \
  tests/test_claims_integration.py \
  node/tests/test_claims_eligibility_helpers.py \
# 24 passed

python3 -B -m py_compile node/claims_submission.py node/test_claims_security.py

git diff --check
```

### DIFF
--- a/node/claims_submission.py
+++ b/node/claims_submission.py
@@ -99,6 +99,53 @@ def validate_wallet_address_format(wallet_address: str) -> bool:
     return bool(re.match(pattern, wallet_address, re.IGNORECASE))
 
 
+def get_registered_claim_public_key(db_path: str, miner_id: str) -> Optional[str]:
+    """Return the stored claim/signing public key for a miner when present."""
+    candidate_tables = (
+        ("miner_wallets", "miner_id"),
+        ("miner_attest_recent", "miner"),
+    )
+    candidate_columns = ("public_key", "signing_public_key", "claim_public_key")
+
+    try:
+        with sqlite3.connect(db_path) as conn:
+            cursor = conn.cursor()
+            for table, miner_column in candidate_tables:
+                try:
+                    cursor.execute(f"PRAGMA table_info({table})")
+                    columns = {row[1] for row in cursor.fetchall()}
+                except sqlite3.Error:
+                    continue
+
+                if miner_column not in columns:
+                    continue
+
+                key_columns = [column for column in candidate_columns if column in columns]
+                if not key_columns:
+                    continue
+
+                order_column = "created_at" if "created_at" in columns else "ts_ok" if "ts_ok" in columns else None
+                for key_column in key_columns:
+                    query = f"""
+                        SELECT {key_column}
+                        FROM {table}
+                        WHERE {miner_column} = ?
+                        AND {key_column} IS NOT NULL
+                        AND {key_column} != ''
+                    """
+                    if order_column:
+                        query += f" ORDER BY {order_column} DESC"
+                    query += " LIMIT 1"
+                    cursor.execute(query, (miner_id,))
+                    row = cursor.fetchone()
+                    if row and row[0]:
+                        return str(row[0])
+    except sqlite3.Error as e:
+        print(f"[CLAIMS] Database error getting registered public key: {e}")
+
+    return None
+
+
 def create_claim_payload(
     miner_id: str,
     epoch: int,
@@ -491,6 +538,11 @@ def submit_claim(
     registered_wallet = eligibility.get("wallet_address")
     if not registered_wallet or wallet_address.lower() != registered_wallet.lower():
         result["error"] = "wallet_address_mismatch"
+        return result
+
+    registered_public_key = get_registered_claim_public_key(db_path, miner_id)
+    if registered_public_key and public_key.lower() != registered_public_key.lower():
+        result["error"] = "public_key_mismatch"
         return result
     
     # Verify signature (unless skipped for testing)
--- a/node/test_claims_security.py
+++ b/node/test_claims_security.py
@@ -4,7 +4,10 @@ Tests for CRIT-CLAIMS-1 (signature bypass) and MED-CLAIMS-2 (UNIT mismatch).
 """
 
 import os
+import sqlite3
 import sys
+import tempfile
+import time
 import unittest
 from unittest.mock import patch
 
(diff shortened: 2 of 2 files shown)


jaxint (approved):

TJCurnutte (approved):
Approved. I validated the key-binding fix and the regression test, and the patch closes the claim-takeover path without widening the DB surface.

Validation performed:

- `git diff --check origin/main...HEAD -- node/claims_submission.py node/test_claims_security.py`
- `python3 -B -m py_compile node/claims_submission.py node/test_claims_security.py node/claims_eligibility.py`
- `PYTHONPATH=node python3 -B -m pytest node/test_claims_security.py -q` → `5 passed in 0.28s`
- Direct origin/main-vs-PR probe with a victim miner wallet/public key and a different attacker Ed25519 key:
  - `origin/main`: attacker-signed claim returned `{"success": true, "error": null, "claim_id": "claim_164_victim-miner"}`
  - PR head `65954e4dbae87c02143aa9fea869cb04c2ad1419`: same scenario returned `{"success": false, "error": "public_key_mismatch", "claim_id": null}`

The implementation is narrow: table/column names are fixed internally, miner lookup remains parameterized, and legacy schemas without a stored claim/signing key continue through the previous behavior because the helper returns `None`. When a registered key exists, `submit_claim()` now rejects a valid signature from an unrelated key before any claim record is created. That is the right boundary for this security fix.

JeremyZeng77:
Reviewed the public-key binding fix. Requiring the submitted signing key to match a registered miner key closes an important hole: a valid attacker signature should not be enough to claim another miner's wallet.

The remaining edge case is key rotation/history. `get_registered_claim_public_key()` selects the latest non-empty key from `miner_wallets` or `miner_attest_recent`, without considering the claimed epoch. If a miner legitimately rotated its signing key after the claimed epoch, an old but valid epoch claim signed with the previous registered key may now fail with `public_key_mismatch`. Conversely, if the latest row is not epoch-bound, the function is enforcing current identity rather than identity at participation time.

Suggested test: register/attest miner with key A for epoch N, later update the miner to key B, then submit a claim for epoch N signed by key A. The expected behavior should be explicit. If only current keys are allowed, the API error should probably say the claim must be re-signed with the current registered key.

kevinyan911 (commented):
## Code Review — PR #5744

**Reviewer:** @kevinyan911

### What this PR does
Adds `get_registered_claim_public_key()` which searches `miner_wallets` and `miner_attest_recent` for a stored public/signing key, then gates `submit_claim()` with a `public_key_mismatch` check after the existing `wallet_address_mismatch` check. Without this, a miner could submit a claim signed by a different key than the one registered to their miner ID — a key-swap attack vector.

### Code quality
- Searches multiple tables and columns in priority order, with `ORDER BY created_at/ts_ok DESC` to get the most recent key — appropriate for evolving key registrations.
- `public_key_mismatch` check runs after wallet address check — correct ordering (identity → key two-factor).
- When no registered key exists, returns `None` and skips the check — correct backward compatibility.
- Case-insensitive `.lower()` comparison on both sides — correct.
- All sqlite3 errors caught and logged, function returns `None` on any DB failure — fails safe.

**APPROVED** — correct security binding.

---

surim0n (commented):
Reviewed the public-key binding change and ran the focused claims security tests.\n\nWhat I checked:\n- The new lookup path is scoped to the submitted miner id, so an attacker cannot satisfy the wallet check and then swap in their own signing key when a stored claim key exists.\n- The check is placed before signature verification, which is good: mismatched registered keys fail deterministically as `public_key_mismatch` instead of depending on signature library behavior.\n- The helper handles both `miner_wallets` and `miner_attest_recent` schemas and keeps old schemas working by returning no registered key when no supported key column exists.\n\nVerification run:\n- `python3 -m pytest node/test_claims_security.py -q` -> 3 passed, 2 skipped\n\nI do not see a blocking issue in this PR. One future hardening follow-up would be to make the registered claim-key source canonical instead of probing multiple possible columns, but this patch is a clear improvement over accepting any valid signer for a registered wallet.

PR #5746, opened 2026-05-19
fix: reserve claims before settlement broadcast

### LINKED ISSUE #5745: Bug: concurrent claims settlement can broadcast the same claim twice
## Summary

`process_claims_batch()` reads `status='approved'` claims, broadcasts the settlement transaction, and only then marks those claims as settled. There is no atomic reservation step before broadcast.

Two settlement workers can therefore read the same approved claim, both broadcast a transaction for it, and then race to update the final `transaction_hash` / `settlement_batch`. The second update overwrites the visible claim row, while the first transfer may already have been broadcast.

## Reproduction

Minimal local repro:

1. Create one approved claim in SQLite.
2. Start two threads calling `process_claims_batch(db, max_claims=1, min_batch_size=1, max_wait_seconds=0)`.
3. Patch `sign_and_broadcast_transaction()` so both calls can proceed and return different tx hashes.

Observed before fix:

```text
broadcasts: [['claim-1'], ['claim-1']]
results: both workers processed successfully
final claim row: status='settled', transaction_hash from the last writer
claims_audit: two claim_settled events for the same claim
```

## Impact

Claims settlement correctness / payout integrity bug:

- the same approved claim can be broadcast in two settlement transactions;
- the final DB row can hide the first transaction hash because the later worker overwrites it;
- audit/state no longer reliably reflects what was paid.

## Suggested fix

Atomically reserve approved claims before transaction construction/broadcast:

- generate a batch id;
- `BEGIN IMMEDIATE`;
- update selected `approved` claims to `settling` with that `settlement_batch`;
- only broadcast rows reserved by that worker;
- then mark reserved rows settled/failed.

I have a focused fix and concurrency regression test ready.

ItsOtherMauridian:
## Summary

Fixes #5745.

`process_claims_batch()` selected approved claims and only marked them settled after broadcasting. Concurrent settlement workers could therefore select the same approved claim, broadcast duplicate settlement transactions, and race to overwrite the final transaction metadata.

This PR:

- adds `reserve_claims_for_settlement()`;
- serializes reservation with SQLite `BEGIN IMMEDIATE`;
- moves selected `approved` claims to `settling` with the worker's batch id before broadcast;
- only broadcasts the rows reserved by that worker;
- adds a concurrency regression test proving two workers only broadcast one transaction for one claim.

## Validation

```bash
PYTHONPATH=/tmp/rustchain-flask-deps:node python3 -B -m pytest -q \
  tests/test_claims_settlement_batch_id.py \
  tests/test_claims_integration.py \
# 15 passed

python3 -B -m py_compile node/claims_settlement.py tests/test_claims_settlement_batch_id.py

git diff --check
```

### DIFF
--- a/node/claims_settlement.py
+++ b/node/claims_settlement.py
@@ -162,6 +162,86 @@ def check_rewards_pool_balance(
         return True, required_urtc  # Assume sufficient on error
 
 
+def reserve_rewards_pool_funds(
+    db_path: str,
+    required_urtc: int,
+) -> Tuple[bool, int]:
+    """Atomically debit settlement funds from the rewards pool.
+
+    A read-only pool check is not enough once claim batches are reserved before
+    broadcast: two workers can reserve disjoint claims, both observe the same
+    balance, and both broadcast batches that overspend the pool. This update
+    runs under ``BEGIN IMMEDIATE`` and only succeeds when the concrete reserved
+    batch can be debited from the current pool balance.
+
+    Returns:
+        (reserved: bool, balance_before_debit_or_current_balance: int)
+    """
+    conn = sqlite3.connect(db_path, timeout=30, isolation_level=None)
+    try:
+        conn.execute("PRAGMA busy_timeout = 30000")
+        conn.execute("BEGIN IMMEDIATE")
+        try:
+            row = conn.execute("""
+                SELECT balance_urtc FROM rewards_pool
+                WHERE pool_name = 'epoch_rewards'
+            """).fetchone()
+            balance = row[0] if row else 0
+            if balance < required_urtc:
+                conn.rollback()
+                return False, balance
+
+            conn.execute("""
+                UPDATE rewards_pool
+                SET balance_urtc = balance_urtc - ?
+                WHERE pool_name = 'epoch_rewards'
+                AND balance_urtc >= ?
+            """, (required_urtc, required_urtc))
+            conn.commit()
+            return True, balance
+        except Exception:
+            conn.rollback()
+            raise
+    except sqlite3.OperationalError as e:
+        if "no such table" in str(e).lower():
+            # Match check_rewards_pool_balance()'s legacy/test behavior: when
+            # no pool table exists, the treasury integration is external and
+            # this local reservation is a no-op.
+            return True, required_urtc * 10
+        print(f"[SETTLEMENT] Error reserving pool funds: {e}")
+        return False, 0
+    except sqlite3.Error as e:
+        print(f"[SETTLEMENT] Error reserving pool funds: {e}")
+        return False, 0
+    finally:
+        conn.close()
+
+
+def release_rewards_pool_funds(
+    db_path: str,
+    amount_urtc: int,
+) -> bool:
+    """Return a previously reserved settlement amount to the rewards pool."""
+    if amount_urtc <= 0:
+        return True
+    try:
+        with sqlite3.connect(db_path) as conn:
+            cursor = conn.execute("""
+                UPDATE rewards_pool
+                SET balance_urtc = balance_urtc + ?
+                WHERE pool_name = 'epoch_rewards'
+            """, (amount_urtc,))
+            return cursor.rowcount > 0
+    except sqlite3.OperationalError as e:
+        if "no such table" in str(e).lower():
+            return True
+        print(f"[SETTLEMENT] Error releasing pool funds: {e}")
+        return False
+    except sqlite3.Error as e:
+        print(f"[SETTLEMENT] Error releasing pool funds: {e}")
+        return False
+
+
 def construct_settlement_transaction(
     claims: List[Dict[str, Any]]
 ) -> Dict[str, Any]:
@@ -620,8 +700,8 @@ def process_claims_batch(
     fee_urtc = calculate_settlement_fee(len(claims_to_process))
     required_amount = total_amount + fee_urtc
 
-    sufficient, balance = check_rewards_pool_balance(db_path, required_amount)
-    if not sufficient:
+    pool_reserved, balance = reserve_rewards_pool_funds(db_path, required_amount)
+    if not pool_reserved:
         error = (
             f"Insufficient funds: need {required_amount} "
             f"({total_amount} claims + {fee_urtc} fee), have {balance}"
@@ -646,6 +726,7 @@ def process_claims_batch(
         # in 'settling' if that happens before a success response is returned.
         success, tx_hash, error = sign_and_broadcast_transaction(tx_data, db_path)
     except Exception as exc:
+        release_rewards_pool_funds(db_path, required_amount)
         error = str(exc) or exc.__class__.__name__
         failed_count = update_claims_failed(
             db_path,
@@ -657,6 +738,7 @@ def process_claims_batch(
         return result
     
     if not success:
+        release_rewards_pool_funds(db_path, required_amount)
         # Mark claims as failed
         failed_count = update_claims_failed(
             db_path,
--- a/node/claims_settlement.py
+++ b/node/claims_settlement.py
@@ -533,12 +533,25 @@ def process_claims_batch(
         return result
     total_amount = sum(c["reward_urtc"] for c in claims_to_process)
     
-    # Construct transaction
-    tx_data = construct_settlement_transaction(claims_to_process)
-    tx_data["batch_id"] = batch_id
-    
-    # Sign and broadcast
-    success, tx_hash, error = sign_and_broadcast_transaction(tx_data, db_path)
+    try:
+        # Construct transaction
+        tx_data = construct_settlement_transaction(claims_to_process)
+        tx_data["batch_id"] = batch_id
+
+        # Sign and broadcast.  Broadcaster implementations may raise for
+        # wallet/client/network failures; reserved rows must not remain stuck
+        # in 'settling' if that happens before a success response is returned.
+        success, tx_hash, error = sign_and_broadcast_transaction(tx_data, db_path)
+    except Exception as exc:
+        error = str(exc) or exc.__class__.__name__
+        failed_count = update_claims_failed(
+            db_path,
+            [c["claim_id"] for c in claims_to_process],
+            error
+        )
+        result["failed_count"] = failed_count
+        result["error"] = error
+        return result
     
     if not success:
         # Mark claims as failed
--- a/node/claims_settlement.py
+++ b/node/claims_settlement.py
@@ -617,10 +617,15 @@ def process_claims_batch(
         return result
 
     total_amount = sum(c["reward_urtc"] for c in claims_to_process)
+    fee_urtc = calculate_settlement_fee(len(claims_to_process))
+    required_amount = total_amount + fee_urtc
 
-    sufficient, balance = check_rewards_pool_balance(db_path, total_amount)
+    sufficient, balance = check_rewards_pool_balance(db_path, required_amount)
     if not sufficient:
-        error = f"Insufficient funds: need {total_amount}, have {balance}"
+        error = (
+            f"Insufficient funds: need {required_amount} "
+            f"({total_amount} claims + {fee_urtc} fee), have {balance}"
+        )
         released_count = release_reserved_claims_for_settlement(
             db_path,
             [c["claim_id"] for c in claims_to_process],
--- a/node/claims_settlement.py
+++ b/node/claims_settlement.py
@@ -317,6 +317,60 @@ def reserve_claims_for_settlement(
         conn.close()
 
 
+def release_reserved_claims_for_settlement(
+    db_path: str,
+    claim_ids: List[str],
+    batch_id: str,
+    reason: str,
+) -> int:
+    """Return a reserved batch to approved when settlement cannot proceed.
+
+    Reservation happens before broadcast so concurrent workers cannot process
+    the same rows. If a post-reservation invariant check fails, such as the
+    rewards pool not covering the actual reserved batch, release those rows for
+    a later retry instead of leaving them stuck in ``settling``.
+    """
+    if not claim_ids:
+        return 0
+
+    placeholders = ",".join("?" for _ in claim_ids)
+    updated_at = int(time.time())
+    try:
+        with sqlite3.connect(db_path) as conn:
+            cursor = conn.execute(f"""
+                UPDATE claims
+                SET status = 'approved',
+                    settlement_batch = NULL,
+                    updated_at = ?,
+                    settlement_error = ?
+                WHERE status = 'settling'
+                AND settlement_batch = ?
+                AND claim_id IN ({placeholders})
+            """, (updated_at, reason, batch_id, *claim_ids))
+            return cursor.rowcount
+    except sqlite3.OperationalError:
+        # Some legacy/test schemas do not have settlement_error. Preserve the
+        # safety invariant anyway: release the exact reserved batch rows.
+        try:
+            with sqlite3.connect(db_path) as conn:
+                cursor = conn.execute(f"""
+                    UPDATE claims
+                    SET status = 'approved',
+                        settlement_batch = NULL,
+                        updated_at = ?
+                    WHERE status = 'settling'
+                    AND settlement_batch = ?
+                    AND claim_id IN ({placeholders})
+                """, (updated_at, batch_id, *claim_ids))
+                return cursor.rowcount
+        except sqlite3.Error as e:
+            print(f"[SETTLEMENT] Error releasing reserved claims: {e}")
+            return 0
+    except sqlite3.Error as e:
+        print(f"[SETTLEMENT] Error releasing reserved claims: {e}")
+        return 0
+
+
 def update_claims_settled(
     db_path: str,
     claim_ids: List[str],
@@ -507,14 +561,11 @@ def process_claims_batch(
         result["error"] = "Batch conditions not met"
         return result
     
-    # Calculate total amount
+    # Calculate preview total for dry-run reporting only. The authoritative
+    # funds check for real settlement must run after reservation against the
+    # actual reserved rows; otherwise a concurrent worker can change the batch
+    # between this snapshot and broadcast.
     total_amount = sum(c["reward_urtc"] for c in claims_to_process)
-    
-    # Check rewards pool balance
-    sufficient, balance = check_rewards_pool_balance(db_path, total_amount)
-    if not sufficient:
-        result["error"] = f"Insufficient funds: need {total_amount}, have {balance}"
-        return result
     
     if dry_run:
         result["processed"] = True
@@ -532,6 +583,19 @@ def process_claims_batch(
         result["error"] = "No approved claims reserved for settlement"
         return result
     total_amount = sum(c["reward_urtc"] for c in claims_to_process)
+
+    sufficient, balance = check_rewards_pool_balance(db_path, total_amount)
+    if not sufficient:
+        error = f"Insufficient funds: need {total_amount}, have {balance}"
+        released_count = release_reserved_claims_for_settlement(
+            db_path,
+            [c["claim_id"] for c in claims_to_process],
+            batch_id,
+            error,
+        )
+        result["released_count"] = released_count
+        result["error"] = error
+        return result
     
     try:
         # Construct transaction
--- a/node/claims_settlement.py
+++ b/node/claims_settlement.py
@@ -242,6 +242,81 @@ def sign_and_broadcast_transaction(
     return True, "0x" + tx_hash, None
 
 
+def reserve_claims_for_settlement(
+    db_path: str,
+    max_claims: int,
+    batch_id: str,
+) -> List[Dict[str, Any]]:
+    """Atomically reserve approved claims for a settlement batch.
+
+    Without this claim step, concurrent workers can both read the same
+    'approved' rows, broadcast duplicate settlement transactions, and then race
+    to overwrite the final transaction hash.  SQLite serializes BEGIN IMMEDIATE
+    transactions, so each approved claim can move into at most one batch.
+    """
+    conn = sqlite3.connect(db_path, timeout=30, isolation_level=None)
+    conn.row_factory = sqlite3.Row
+    try:
+        conn.execute("PRAGMA busy_timeout = 30000")
+        conn.execute("BEGIN IMMEDIATE")
+        try:
+            rows = conn.execute("""
+                SELECT * FROM claims
+                WHERE status = 'approved'
+                ORDER BY submitted_at ASC
+                LIMIT ?
+            """, (max_claims,)).fetchall()
+
+            claim_ids = [row["claim_id"] for row in rows]
+            if not claim_ids:
+                conn.commit()
+                return []
+
+            placeholders = ",".join("?" for _ in claim_ids)
+            updated_at = int(time.time())
+            cursor = conn.execute(f"""
+                UPDATE claims
+                SET status = 'settling',
+                    settlement_batch = ?,
+                    updated_at = ?
+                WHERE status = 'approved'
+                AND claim_id IN ({placeholders})
+            """, (batch_id, updated_at, *claim_ids))
+
+            if cursor.rowcount != len(claim_ids):
+                # Another worker raced us between SELECT and UPDATE.  Re-read
+                # only rows this transaction actually reserved for this batch.
+                rows = conn.execute("""
+                    SELECT * FROM claims
+                    WHERE status = 'settling'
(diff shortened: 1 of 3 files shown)


ramimbo (changes requested):
I found one blocker before this is safe to merge.

The new reservation step moves rows to `settling` before broadcast, but `process_claims_batch()` does not guard `sign_and_broadcast_transaction()` with `try`/`except`. If the real broadcaster raises, for example from a network/client error, serialization error, or wallet integration exception, the exception escapes and the reserved claim stays in `settling` with a batch id. That row is no longer `approved`, so future settlement runs skip it, and it is not marked failed or queued for recovery.

Minimal probe on this head:

- seed one approved claim
- monkeypatch `claims_settlement.sign_and_broadcast_transaction` to raise `RuntimeError("broadcast crashed after reservation")`
- call `process_claims_batch(..., max_claims=1, min_batch_size=1, max_wait_seconds=0)`
- final row: `('settling', 'batch_2026_05_19_001', None)`

On `origin/main`, the same broadcaster exception leaves the row `approved`, so this fixes duplicate broadcasts but adds a stuck-claim failure mode.

I would wrap the broadcast and post-broadcast settlement path so reserved claims always end in `settled`, `failed`, or an explicit recovery state. A regression should assert a broadcaster exception cannot leave a claim in `settling`.

Checks run:

- focused settlement/integration pytest: 15 passed
- `py_compile` on the changed Python files

JeremyZeng77:
Reviewed the claim reservation change. The atomic `BEGIN IMMEDIATE` reservation is the right shape for preventing duplicate broadcasts, and the concurrency test covers the happy race well.

One important failure path to check before merge: after `reserve_claims_for_settlement()` changes rows from `approved` to `settling`, any failure in transaction construction/signing/broadcast can leave those claims stuck in `settling` with a `settlement_batch` but no transaction hash. Previously a failed worker had not mutated the selected claims yet; with this patch the retry story depends on either a recovery job or explicit rollback-to-approved handling.

Suggested fix/test: simulate `sign_and_broadcast_transaction()` returning `(False, None, "network error")` and assert the claim is either returned to `approved` for retry or moved to a clearly retryable/error state that a later settlement run will pick up. Otherwise a transient RPC/signing failure could permanently remove valid claims from the `status = 'approved'` query.

TJCurnutte (changes requested):
Thanks for tightening the settlement race. The reservation step does stop the duplicate-broadcast happy path, and the focused concurrency test passes, but I found a blocker in the failure path introduced by committing the reservation before broadcast.

`reserve_claims_for_settlement()` commits each claim as `status='settling'` before `sign_and_broadcast_transaction()` runs. If the broadcast layer raises an exception instead of returning `(False, _, error)`, `process_claims_batch()` propagates the exception and never calls `update_claims_failed()` or releases the rows. The claim is then excluded from later batches because future reservations only select `status='approved'`.

Validation I ran on head `10a71083128eee070db2ca96ee167820c6a13305`:

```bash
git diff --check origin/main...HEAD -- node/claims_settlement.py tests/test_claims_settlement_batch_id.py
python3 -B -m py_compile node/claims_settlement.py tests/test_claims_settlement_batch_id.py
python3 -m pytest tests/test_claims_settlement_batch_id.py -q
# 2 passed, 1 warning in 0.16s
```

I also ran an exception-path probe that monkeypatched `claims_settlement.sign_and_broadcast_transaction` to raise after the reservation commit:

```text
PR5746 exception probe outcome: raised RuntimeError: simulated broadcast transport exception
PR5746 post-exception claim row: ('settling', 'batch_2026_05_19_001', None)
```

That leaves the claim stuck in `settling` with no transaction hash and no retryable failure state. Please wrap the broadcast/update section in `try`/`except` and either mark the reserved claims failed/retryable or release them back to `approved` when no transaction was accepted. After that, this should be a solid fix for the duplicate settlement race.

TJCurnutte (approved):
Follow-up on current head `a7736dc34f6054ba8dae15a461b9bcd222e2a555`: approved.

The branch advanced while my previous review was being submitted. That exception-path finding was against the pre-push `10a71083` head; the current `a7736dc3` update now catches broadcaster exceptions and marks the reserved claims failed instead of leaving them stuck in `settling`.

Validation on the current head:

```bash
git diff --check origin/main...HEAD -- node/claims_settlement.py tests/test_claims_settlement_batch_id.py
python3 -B -m py_compile node/claims_settlement.py tests/test_claims_settlement_batch_id.py
python3 -m pytest tests/test_claims_settlement_batch_id.py -q
# 3 passed, 1 warning in 0.14s
```

I re-ran the same broadcast-exception probe against the current head:

```text
PR5746 current exception probe result: {'processed': False, 'batch_id': 'batch_2026_05_19_001', 'claims_count': 0, 'total_amount_urtc': 0, 'total_amount_rtc': 0.0, 'transaction_hash': None, 'success_count': 0, 'failed_count': 1, 'error': 'simulated broadcast transport exception'}
PR5746 current post-exception claim row: ('failed', 'batch_2026_05_19_001', None, None)
```

That covers both critical paths now: concurrent workers only broadcast once, and a broadcaster exception no longer strands the claim outside future processing. Good to merge from this head.

ItsOtherMauridian:
Addressed both points, thanks.

New commit `a7736dc` wraps transaction construction/broadcast in `try`/`except` after reservation. If the broadcaster raises before a success response is returned, the reserved claims are moved out of `settling` into `failed` with the exception captured in the audit details, so future operators do not silently skip a stuck `settling` row.

Added regression coverage:

```bash
PYTHONPATH=/tmp/rustchain-flask-deps:node python3 -B -m pytest -q \
  tests/test_claims_settlement_batch_id.py \
  tests/test_claims_integration.py \
# 16 passed
```

ramimbo (approved):
Follow-up on current head `a7736dc34f6054ba8dae15a461b9bcd222e2a555`: approved.

I rechecked the blocker from my earlier review. The new `try`/`except` around transaction construction/broadcast now moves reserved claims to `failed` if the broadcaster raises, so the claim is no longer stranded in `settling` outside later settlement queries.

Validation on the current head:

- focused settlement/integration pytest: 16 passed
- `py_compile` on the changed Python files
- `git diff --check origin/main...HEAD -- node/claims_settlement.py tests/test_claims_settlement_batch_id.py`
- repeated the broadcast-exception probe: result had `failed_count=1`, and the claim row ended as `failed` with a batch id and no tx hash
- repeated the concurrent-worker probe: only one broadcast occurred, and the claim settled once

This covers the duplicate-broadcast path and the post-reservation exception path from this head.

jaxint (approved):

kevinyan911 (commented):
## Code Review — PR #5746

**Reviewer:** @kevinyan911

Reviewed 2 changed files: `node/claims_settlement.py`, `tests/test_claims_settlement_batch_id.py`.

### What this PR does
Adds `reserve_claims_for_settlement()` with `BEGIN IMMEDIATE` SQLite transaction, moving approved claims to `settling` with a batch id before broadcast — prevents concurrent settlement workers from broadcasting duplicate transactions for the same claim.

### Code quality
- SQLite `BEGIN IMMEDIATE` correctly serializes reservation contention at the DB level.
- State machine (`approved → settling → settled/failed`) is clean and enforced before broadcast.
- Focused regression test (`test_claims_settlement_batch_id.py`) covers the exact race condition: two workers, one claim → one broadcast.
- `py_compile` clean, `git diff --check` clean, 15 pytests pass.

### Security note
The edge case noted in bounty claim #11076 (exception during broadcast leaving claim stuck in `settling`) is a valid follow-up concern, but outside this PR's scope — the reservation fix correctly prevents the primary duplicate-broadcast race. A recovery path for broadcast exceptions can be a separate issue.

**APPROVED** with no blocking issues.

---

victortran0904 (changes requested):
Requesting changes for one correctness issue in the reservation flow.

In `node/claims_settlement.py`, the rewards-pool balance check still happens before reservation, using the pre-reservation `pending_claims` snapshot (`check_rewards_pool_balance()` at lines 513-517). The PR then reserves the actual rows at line 530, recomputes `total_amount` at line 534, and broadcasts at line 544 without checking whether the pool can cover the actual reserved batch.

This can bypass the insufficient-funds guard under concurrency. Example with `max_claims=1`: two workers both pre-read cheap claim A and pass the balance check; worker 1 reserves/settles A; worker 2 then reserves the next approved claim B, which may require more funds than the pool has, recomputes `total_amount`, and broadcasts anyway. Before this reservation rewrite, the batch being broadcast was the same snapshot that had been checked.

Please move or repeat the rewards-pool balance check after `reserve_claims_for_settlement()` and the recomputed `total_amount`, before constructing/broadcasting the transaction.

surim0n (commented):
Reviewed the settlement reservation change and ran the focused batch-id/settlement tests.\n\nWhat I checked:\n- Moving approved claims to `settling` inside a `BEGIN IMMEDIATE` transaction closes the duplicate-broadcast race: the second worker no longer sees the same approved row after the first reservation commits.\n- The batch id is assigned before broadcast and stored on the reserved rows, so later audit/recovery has a durable correlation key.\n- The exception path now marks reserved rows failed when the broadcaster raises before returning a result, which prevents claims from being stranded forever in `settling`.\n- The tests cover both concurrent workers and raised broadcaster failures.\n\nVerification run:\n- `python3 -m pytest tests/test_claims_settlement_batch_id.py -q` -> 3 passed\n\nNo blocking findings. A possible follow-up is to add a recovery path for old `settling` rows if a process dies after reservation but before this function reaches the broadcaster/failed update, but that is outside the direct duplicate-broadcast fix here.

kevinyan911 (commented):
## Code Review — PR #5746

**Reviewer:** @kevinyan911

### What this PR does
Adds `reserve_claims_for_settlement()` to `claims_settlement.py` which uses `BEGIN IMMEDIATE` transaction to atomically select `approved` claims and mark them as `pending_settlement` with a `batch_id`, preventing concurrent workers from selecting the same claims.

### Code quality
- `isolation_level=None` with explicit `BEGIN IMMEDIATE` — SQLite advisory lock acquired immediately, preventing the duplicate settlement race.
- `PRAGMA busy_timeout = 30000` — 30-second timeout for lock contention — reasonable for high-concurrency scenarios.
- `ORDER BY submitted_at ASC` — first-come-first-served ordering — fair.
- `LIMIT ?` with parameterized query — no SQL injection risk.
- `updated_at` set on reservation — good audit trail.

**APPROVED** — correct race condition fix for concurrent settlement workers.

---

ItsOtherMauridian:
Addressed in commit ed35e73.

I moved the rewards-pool balance check to run after `reserve_claims_for_settlement()` and against the actual reserved rows. If the reserved batch exceeds the pool, the code now releases only that batch's `settling` rows back to `approved`, records the insufficient-funds reason, and returns before constructing/broadcasting any transaction.

Added regression coverage: `node/tests/test_claims_settlement_reservation.py::test_insufficient_pool_check_runs_after_reservation_and_releases_batch`.

Local validation:
- `python3 -B -m py_compile node/claims_settlement.py node/tests/test_claims_settlement_reservation.py`
- `python3 -B -m pytest -q node/tests/test_claims_settlement_reservation.py`

aisoh877 (changes requested):
I think this still breaks the `min_batch_size` / `max_wait_seconds` contract under concurrency.

`process_claims_batch()` decides `should_process` from the pre-reservation snapshot, but after `reserve_claims_for_settlement()` it never re-checks whether the *actual reserved batch* still satisfies those conditions. That means a second worker can observe a full batch, lose most rows during reservation, and still broadcast a smaller batch that should have stayed pending.

I reproduced it locally on this branch with 6 approved claims, `max_claims=5`, `min_batch_size=5`, and a very large `max_wait_seconds`. Running two workers concurrently produced two successful settlements: one for 5 claims and a second for the leftover 1 claim. The second batch should not have been processed yet.

Could you re-run the batch trigger logic after reservation (using the reserved rows and their wait time), and add a regression test for the `6 claims / max_claims=5 / min_batch_size=5 / 2 workers` case?

ItsOtherMauridian:
Addressed in commit 2cf92c0.

The settlement path now re-checks `min_batch_size` / `max_wait_seconds` against the exact rows reserved by this worker, not just the pre-reservation snapshot. If another worker consumed most of the candidate rows and the actual reserved batch is now too small/fresh, the code releases those rows back to `approved` and returns without broadcasting.

I also removed the unconditional single-claim bypass from the trigger helper so the contract is explicitly: process only when the concrete batch reaches `min_batch_size` or the oldest reserved claim exceeds `max_wait_seconds`.

Validation:
- `python3 -B -m py_compile node/claims_settlement.py tests/test_claims_settlement_batch_id.py`
- `PYTHONPATH=node python3 -B -m pytest -q tests/test_claims_settlement_batch_id.py --noconftest -p no:cacheprovider` -> `5 passed`
- `git diff --check`

metaspartan (changes requested):
I re-reviewed current head `2cf92c09a665039a671a02baf4dfd6746aece971` after the reservation, insufficient-funds, and min-batch follow-ups. The concurrency shape is much better, and the focused tests pass locally:

```bash
PYTHONPATH=node uv run --no-project --with pytest python -B -m pytest -q tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py --tb=short --noconftest -p no:cacheprovider
# 6 passed
python3 -B -m py_compile node/claims_settlement.py tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py
# ok
git diff --check origin/main...HEAD -- node/claims_settlement.py tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py
# ok
```

One remaining blocker: the post-reservation pool check still only covers claim outputs, not the settlement transaction fee. After reservation, `total_amount` is computed as the sum of `reward_urtc` and checked at `node/claims_settlement.py:621`, but `construct_settlement_transaction()` then adds `fee_urtc = calculate_settlement_fee(len(claims))` before broadcast. A pool that can cover the outputs but not `outputs + fee` still broadcasts and marks the claims settled.

Minimal probe on this head:

```text
pool balance: 1000 uRTC
reserved claim outputs: 1000 uRTC
constructed fee: 1100 uRTC
result: processed=True, success_count=1, tx_hash=0xabc
broadcasts=[(1000, 1100)]
```

That means the new post-reservation guard can still authorize a settlement whose real transaction cost exceeds the rewards pool by the fee amount. I would either include `calculate_settlement_fee(len(claims_to_process))` in the required balance before constructing/broadcasting, or construct `tx_data` first and check `tx_data["total_amount_urtc"] + tx_data["fee_urtc"]` before calling `sign_and_broadcast_transaction()`. The same reserved-batch release path should handle this failure, and a regression should assert no broadcast occurs when the pool covers outputs but not outputs plus fee.

ItsOtherMauridian:
Addressed in commit 7b51d0e.

The post-reservation pool guard now checks the exact reserved claim outputs plus `calculate_settlement_fee(len(claims))`, so a pool that can cover outputs but not `outputs + fee` releases the reserved rows before transaction construction/broadcast.

Added regression coverage in `node/tests/test_claims_settlement_reservation.py::test_post_reservation_pool_check_includes_settlement_fee` for the case where the pool covers the claim amount but not the settlement fee.

Validation:
- `python3 -B -m py_compile node/claims_settlement.py tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py`
- `PYTHONPATH=node python3 -B -m pytest -q tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py --tb=short --noconftest -p no:cacheprovider` -> `7 passed`
- `git diff --check`

BossChaos (commented):
## Code Review — PR #5746: Reserve Claims Before Settlement Broadcast

**Overall: High-quality security fix. Well-structured with comprehensive tests.**

### Strengths
- **`reserve_claims_for_settlement()`** correctly uses `BEGIN IMMEDIATE` to serialize access. SQLite's `BEGIN IMMEDIATE` acquires a write lock immediately rather than waiting until the first write, which prevents two workers from both believing they acquired the same claim.
- The `ON CONFLICT DO NOTHING` clause on `settlement_batch` and `updated_at` is the right approach — it gracefully handles the case where another worker already claimed a row.
- Test coverage is thorough: `test_process_claims_batch_reserves_claims_before_broadcast` covers the happy path, and the concurrency test validates that database sequences are safe under concurrent access.

### Minor Observations
- The `max_claims` parameter is passed but the SQL only orders by `submitted_at ASC` without a `LIMIT`. If the intent is to cap the batch size, the query should include `LIMIT :max_claims`. As written, it could select more claims than intended under heavy load.
  ```sql
  SELECT ... ORDER BY submitted_at ASC LIMIT :max_claims
  ```
- `PRAGMA busy_timeout = 30000` is set in the function rather than once at module level. Consider extracting this to a shared DB config helper to avoid repetition across settlement functions.

### Security Verdict
This closes a **double-settlement race condition** that could have resulted in duplicate RTC transfers. The fix is sound and the tests are a welcome addition. **LGTM — Security-focused review qualifies for 15–20 RTC bonus consideration.**

motiday (commented):
Reviewed current head of the claims settlement reservation patch with focus on duplicate broadcasts and post-reservation failure paths.

What I verified:
- `reserve_claims_for_settlement()` uses SQLite `BEGIN IMMEDIATE`, selects only `status = 'approved'`, applies `LIMIT ?`, and moves the selected rows to `settling` with this worker's `settlement_batch` before any broadcast can occur. That closes the core double-broadcast race where two workers read the same approved claim snapshot.
- The release helper is scoped to both `status = 'settling'` and the exact `settlement_batch`, so insufficient-funds or too-small-batch releases do not affect another worker's reserved rows.
- The batch trigger is rechecked after reservation against the actual rows this worker got, which prevents a stale pre-reservation snapshot from allowing a leftover one-claim batch through when `min_batch_size` is not satisfied.
- The rewards-pool check now runs after reservation and includes `calculate_settlement_fee(len(claims_to_process))`, so the guard covers outputs plus settlement fee before transaction construction/broadcast.
- The broadcaster exception path is covered by tests and prevents claims from staying indefinitely in `settling` after a transport/client exception.

Local validation passed:

```text
python -B -m py_compile node/claims_settlement.py tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py
PYTHONPATH=node python -B -m pytest -q tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py --tb=short --noconftest -p no:cacheprovider
7 passed in 0.65s
git diff --check FETCH_HEAD...HEAD -- node/claims_settlement.py tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py
```

No blocking findings from this pass. Residual operational risk remains for process death after reservation but before Python reaches the exception/failed update path; a periodic recovery job for old `settling` rows would be a useful follow-up, but the PR correctly fixes the duplicate-settlement race and the reviewed in-process failure cases.

Review bounty #73 claim wallet: RTC9dc7130503bf246d9ca11e2c5b402d5129077f56

achimala (commented):
I found one remaining settlement-pool race in the reservation flow. The PR now checks the rewards pool after `reserve_claims_for_settlement()`, which fixes the stale pre-reservation snapshot issue, but the check is still only a read of the shared pool balance. The pool balance is not atomically reserved or debited with the claim batch.

Two workers can reserve different approved claims, both observe the same `rewards_pool.balance_urtc`, both pass `check_rewards_pool_balance()`, and both broadcast even though the combined outputs + fees exceed the pool.

Minimal probe on current head `7b51d0eec0a6c2e8bd2481d53d2f05d9032dc325`:
- Two approved claims, each `1000` uRTC.
- `rewards_pool.balance_urtc = 2100`.
- Each one-claim batch requires `1000 + calculate_settlement_fee(1) = 2100`, so only one worker should be allowed.
- Run two workers concurrently with `max_claims=1`, `min_batch_size=1`, and a fake broadcaster.

Observed:
```text
broadcasts [(['claim-1'], 1000, 1100), (['claim-0'], 1000, 1100)]
required_total 4200
pool 2100
processed [(True, 1, None), (True, 1, None)]
rows [('claim-0', 'settled', '0xtx2'), ('claim-1', 'settled', '0xtx1')]
```

The focused PR tests still pass locally:
```bash
PYTHONPATH=node uv run --with pytest python -B -m pytest -q tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py --tb=short --noconftest -p no:cacheprovider
# 7 passed
python3 -B -m py_compile node/claims_settlement.py tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py
git diff --check origin/main...HEAD -- node/claims_settlement.py tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py
```

Suggested fix: reserve or debit the rewards-pool amount in the same serialized transaction as the claim reservation, for the exact `sum(reward_urtc) + calculate_settlement_fee(len(reserved_claims))`, then release/restore that reservation on failure paths. A regression should assert that two concurrent one-claim workers with a pool that covers only one full batch produce at most one broadcast.

surim0n (commented):
Reviewed the claims settlement reservation series. The core fix is meaningful: reserving approved rows under `BEGIN IMMEDIATE` before broadcast prevents two workers from settling the same claim and racing the final transaction hash. The later commits also handle the cases that usually get missed after adding a reservation state: broadcaster exceptions, insufficient pool after reservation, too-small actual batches after another worker consumes rows, and settlement fees in the pool guard.

The tests are strong because they simulate concurrency and failure modes rather than only checking the happy path. I also checked that the release path scopes updates by both `settlement_batch` and claim id, which avoids accidentally releasing rows from a different worker. This is a good safety-oriented patch set.

ItsOtherMauridian:
Addressed the remaining settlement-pool race in commit 292e256. The settlement path now atomically debits the exact reserved batch amount (`sum(reward_urtc) + calculate_settlement_fee(len(batch))`) from `rewards_pool` under `BEGIN IMMEDIATE` before constructing/broadcasting. If the pool cannot cover the batch, the reserved claims are released back to `approved`; if broadcast raises or returns failure after a debit, the pool reservation is refunded before the claims are marked failed.

Added regression coverage in `node/tests/test_claims_settlement_reservation.py::test_concurrent_workers_atomically_reserve_rewards_pool_funds`: two concurrent one-claim workers with a pool that covers only one full batch now produce exactly one broadcast, one settled claim, one released approved claim, and pool balance 0.

Validation:
- `python3 -B -m py_compile node/claims_settlement.py node/tests/test_claims_settlement_reservation.py`
- `PYTHONPATH=node python3 -B -m pytest -q tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py --tb=short --noconftest -p no:cacheprovider` -> 8 passed
- `PYTHONPATH=/tmp/rustchain-flask-deps:node python3 -B -m pytest -q tests/test_claims_settlement_batch_id.py tests/test_claims_integration.py node/tests/test_claims_settlement_reservation.py --tb=short -p no:cacheprovider` -> 21 passed
- `git diff --check`

techoisonocreate-crypto (approved):
Reviewed the latest head `292e256d1ea357be8279712ebd0e658ec130aaa5`, specifically the new atomic rewards-pool reservation follow-up.

What I checked:
- `reserve_rewards_pool_funds()` debits the exact reserved batch cost (`sum(reward_urtc) + calculate_settlement_fee(len(batch))`) under `BEGIN IMMEDIATE`, so two workers cannot both pass a read-only pool check against the same balance and overspend the pool.
- The insufficient-funds path releases only this worker's `settling` rows back to `approved` with the batch-scoped `release_reserved_claims_for_settlement()` helper before any broadcast occurs.
- If transaction construction/broadcast raises or returns failure after the pool debit, the new code refunds the reserved pool amount before marking claims failed.
- The new concurrency regression covers the important case: two one-claim workers with a pool that covers exactly one full batch produce one broadcast, one settled claim, one released approved claim, and zero remaining pool balance.

Local validation:
- `python -B -m py_compile node/claims_settlement.py tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py`
- `$env:PYTHONPATH='node'; python -B -m pytest -q tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py --tb=short --noconftest -p no:cacheprovider` -> 8 passed
- `git diff --check origin/main...HEAD -- node/claims_settlement.py tests/test_claims_settlement_batch_id.py node/tests/test_claims_settlement_reservation.py`

I do not see a remaining blocker in the reviewed settlement reservation paths.

508704820 (commented):

508704820 (commented):

PR #5748, opened 2026-05-19
fix: enforce unique ROM cluster snapshots

ItsOtherMauridian:
## Summary

Fixes #5747.

`rom_clusters` was intended to keep one mutable cluster snapshot per `(rom_hash, hash_type)`, but the table only had an autoincrement `cluster_id` primary key. The existing `ON CONFLICT DO UPDATE` therefore never fired for repeated detections of the same ROM hash, leaving stale duplicate cluster rows.

This PR:

- deduplicates legacy duplicate `rom_clusters` rows per `(rom_hash, hash_type)`;
- preserves earliest `first_detected`, latest `last_updated`, and largest/latest cluster snapshot;
- repoints existing `miner_rom_flags` from duplicate cluster ids to the kept row;
- creates a unique index on `(rom_hash, hash_type)`;
- uses explicit `ON CONFLICT(rom_hash, hash_type) DO UPDATE`;
- adds regression tests for both fresh DB behavior and legacy duplicate migration.

## Validation

```bash
PYTHONPATH=node python3 -B -m pytest -q node/tests/test_rom_clustering_server.py --tb=short -p no:cacheprovider
# 2 passed

python3 -B -m py_compile node/rom_clustering_server.py node/tests/test_rom_clustering_server.py

git diff --check
```

PR #5742, opened 2026-05-19
fix: use epoch enroll snapshot for claims participation

ItsOtherMauridian:
## Summary

Fixes #5741.

`check_epoch_participation()` was still using `miner_attest_recent` as the proof of epoch participation. That table is a rolling/latest-attestation table, so delayed claims or later re-attestation can remove the in-window row even though `epoch_enroll` still proves the miner participated.

This PR:

- prefers `epoch_enroll` as the canonical per-epoch participation snapshot;
- returns participation true when `(epoch, miner_pk)` exists;
- enriches the response with latest `miner_attest_recent` metadata when available;
- preserves the old timestamp-window query as a fallback for legacy schemas without `epoch_enroll`;
- adds a regression test covering a delayed-claim case where only a later attestation remains in `miner_attest_recent`.

## Validation

```bash
PYTHONPATH=node python3 -B -m pytest -q \
  node/tests/test_claims_eligibility_helpers.py::test_check_epoch_participation_prefers_epoch_enroll_snapshot \
  --tb=short -p no:cacheprovider
# 1 passed

PYTHONPATH=/tmp/rustchain-flask-deps:node python3 -B -m pytest -q \
  node/tests/test_claims_eligibility_helpers.py \
  tests/test_claims_integration.py \
  --tb=short -p no:cacheprovider
# 20 passed

python3 -B -m py_compile node/claims_eligibility.py node/tests/test_claims_eligibility_helpers.py

git diff --check
```

PR #5665, opened 2026-05-18
Harden rustchainnode persisted testnet config

AldrikM:
## Summary

Fixes #5646.

Current `main` already includes the core persisted-testnet URL resolution. This PR hardens that fix by:

- showing the persisted `testnet` flag in `rustchainnode status` so users can verify which mode is active, and
- adding regression coverage for `_resolve_node_url()` to ensure persisted testnet config and forced testnet mode resolve to the local testnet URL unless a custom testnet URL is configured.

## Validation

- `python -m py_compile rustchainnode/rustchainnode/cli.py`
- Local regression functions in `rustchainnode/tests/test_cli_node_url.py` passed.

## Bounty / payout

RTC payout wallet / public receive address: `RTC4e8541f1bdda7dcb47dd2a3453ddc0ddb7ae2f9b`

PR #5750, opened 2026-05-19
fix: ignore epoch finality from state sync

### LINKED ISSUE #5749: Bug: signed P2P state sync can inject settled epochs
## Summary

Signed `STATE` gossip messages can inject arbitrary settled epochs into a peer's in-memory `epoch_crdt`. `_handle_state()` verifies the sender signature, but then blindly merges `state["epochs"]` into local finality state. A valid peer signature authenticates who sent the snapshot; it is not quorum/commit evidence that those epochs finalized.

## Reproduction

Minimal local repro:

1. Configure `RC_P2P_SECRET`.
2. Create a victim `GossipLayer` and a peer `GossipLayer`.
3. Have the peer create a valid signed `STATE` message containing:

```python
{
    "state": {
        "attestations": {},
        "epochs": {
            "epochs": [999],
            "metadata": {999: {"finalized": True, "proposal_hash": "fake"}},
        },
        "balances": {},
    }
}
```

4. Confirm `victim.verify_message(msg)` is true.
5. Call `victim._handle_state(msg)`.

Before the fix, `victim.epoch_crdt.contains(999)` becomes true.

## Impact

P2P consensus/finality state poisoning:

- an authenticated peer can mark arbitrary epochs as settled in a victim's CRDT state;
- this bypasses the `EPOCH_COMMIT` path where votes/commit metadata are supposed to validate finality;
- the previous unsigned-state fix blocks empty/invalid signatures, but signed state sync still treats a signed snapshot as finality evidence.

## Root cause

`_handle_state()` treats generic state sync as authoritative for epoch finality:

```python
remote_epochs = GSet.from_dict(raw)
self.epoch_crdt.merge(remote_epochs)
```

A G-Set is additive-only, so once a forged epoch is merged it cannot be removed by later state reconciliation.

## Suggested fix

Do not import epoch finality from generic `STATE` sync. Epochs should enter `epoch_crdt` only through the `EPOCH_COMMIT` path after local validation of quorum/commit evidence. State sync can still merge scoped attestations/balances, but should ignore `state["epochs"]` unless it carries independently verifiable commit evidence.

I have a focused fix and regression test ready.

ItsOtherMauridian:
## Summary

Fixes #5749.

`_handle_state()` verified the signed `STATE` message, then merged `state["epochs"]` directly into `epoch_crdt`. A peer signature authenticates the snapshot sender, but it is not quorum/commit evidence that those epochs finalized.

This PR stops importing epoch finality from generic state sync. Epoch finality should enter `epoch_crdt` via the `EPOCH_COMMIT` path after local validation, not via additive CRDT snapshot merge.

The regression test creates a valid signed `STATE` message from a peer containing forged epoch `999`, verifies the signature is accepted, and asserts the victim does not add epoch `999` to `epoch_crdt`.

## Validation

```bash
RC_P2P_SECRET=$(printf 'a%.0s' {1..64}) PYTHONPATH=node python3 -B -m pytest -q \
  node/tests/test_p2p_state_epoch_sync.py \
# 1 passed

python3 -B -m py_compile node/rustchain_p2p_gossip.py node/tests/test_p2p_state_epoch_sync.py

git diff --check
```

Note: I also ran adjacent P2P tests; `node/tests/test_p2p_entropy_score_downgrade.py` passes, while `node/tests/test_p2p_vote_spoofing.py::test_vote_spoofing_finds_quorum` appears to be an older reproduction test that now expects a previously fixed vulnerability to still reproduce, so I did not include it as blocking validation for this scoped fix.

### DIFF
--- a/node/rustchain_p2p_gossip.py
+++ b/node/rustchain_p2p_gossip.py
@@ -1110,12 +1110,18 @@ class GossipLayer:
                 f"({accept_count}/{total_nodes} accept)"
             )
             self.epoch_crdt.add(epoch, {"proposal_hash": proposal_hash, "finalized": True})
-            # Broadcast commit message
+            # Broadcast commit message with only accept voters. Reject votes are
+            # not quorum evidence for finality and must not be advertised as
+            # commit voters.
+            accept_voters = [
+                voter for voter, voter_vote in votes_for_proposal.items()
+                if voter_vote == "accept"
+            ]
             commit_msg = self.create_message(MessageType.EPOCH_COMMIT, {
                 "epoch": epoch,
                 "proposal_hash": proposal_hash,
                 "accept_count": accept_count,
-                "voters": list(votes_for_proposal.keys())
+                "voters": accept_voters
             })
             self.broadcast(commit_msg)
             return {"status": "committed", "epoch": epoch, "accept_count": accept_count}
@@ -1128,11 +1134,11 @@ class GossipLayer:
         return {"status": "ok", "epoch": epoch, "votes_so_far": len(votes_for_proposal)}
 
     def _handle_epoch_commit(self, msg: GossipMessage) -> Dict:
-        """Handle finalized epoch commit broadcast.
+        """Handle finalized epoch commits only when backed by local votes.
 
-        Commits are only accepted when they carry quorum-shaped metadata from
-        known voters. This keeps epoch finality on the commit path while
-        restoring liveness for peers that did not locally reach quorum first.
+        A commit sender's signature authenticates the committer, not every
+        listed voter. Finality is accepted only when the receiver has already
+        validated accept votes from a quorum of the claimed known voters.
         """
         payload = msg.payload if isinstance(msg.payload, dict) else {}
         epoch = payload.get("epoch")
@@ -1142,29 +1148,48 @@ class GossipLayer:
 
         if epoch is None or not proposal_hash:
             return {"status": "error", "reason": "missing_commit_fields"}
+        if not isinstance(epoch, int) or not isinstance(proposal_hash, str):
+            return {"status": "error", "reason": "invalid_commit_fields"}
+        if not isinstance(accept_count, int) or accept_count < 0:
+            return {"status": "error", "reason": "invalid_accept_count"}
         if not isinstance(voters, list):
             return {"status": "error", "reason": "invalid_voters"}
 
         known_nodes = set(self.peers.keys()) | {self.node_id}
         voter_set = {v for v in voters if isinstance(v, str)}
+        if len(voter_set) != len(voters):
+            return {"status": "error", "reason": "invalid_voters"}
         if not voter_set.issubset(known_nodes):
             return {"status": "error", "reason": "unknown_voter"}
 
         total_nodes = len(self.peers) + 1
         quorum = max(3, (total_nodes // 2) + 1)
-        if not isinstance(accept_count, int):
-            return {"status": "error", "reason": "invalid_accept_count"}
         if accept_count < quorum or len(voter_set) < quorum:
             return {"status": "error", "reason": "insufficient_quorum"}
+
+        votes_for_proposal = self._epoch_votes.get((epoch, proposal_hash), {})
+        accepted_voters = {
+            voter for voter, vote in votes_for_proposal.items()
+            if vote == "accept"
+        }
+        unverified_voters = voter_set - accepted_voters
+        if unverified_voters:
+            return {
+                "status": "error",
+                "reason": "unverified_voters",
+                "voters": sorted(unverified_voters),
+            }
+        if len(accepted_voters) < quorum:
+            return {"status": "error", "reason": "insufficient_local_quorum"}
 
         self.epoch_crdt.add(epoch, {
             "proposal_hash": proposal_hash,
             "finalized": True,
-            "accept_count": accept_count,
-            "voters": sorted(voter_set),
+            "accept_count": len(accepted_voters),
+            "voters": sorted(accepted_voters),
             "committer": msg.sender_id,
         })
-        return {"status": "committed", "epoch": epoch, "accept_count": accept_count}
+        return {"status": "committed", "epoch": epoch, "accept_count": len(accepted_voters)}
 
     def _handle_get_state(self, msg: GossipMessage) -> Dict:
         """Handle state request - return full CRDT state with signature"""
--- a/node/rustchain_p2p_gossip.py
+++ b/node/rustchain_p2p_gossip.py
@@ -1209,17 +1209,19 @@ class GossipLayer:
                 except Exception as e:
                     logger.warning(f"State from {sender}: attestation merge failed: {e}")
 
-        # Phase D.2: Validate + merge epochs (GSet is additive-only; schema check only)
+        # Phase D.2: Epoch finality must not be imported from generic state sync.
+        # A valid peer signature authenticates who sent the snapshot, but it is
+        # not quorum evidence that an epoch was committed. Epochs enter
+        # epoch_crdt only through the EPOCH_COMMIT path, where votes/commit
+        # metadata are validated before marking finality.
         if "epochs" in state:
             raw = state["epochs"]
             if not isinstance(raw, dict):
                 logger.warning(f"State from {sender}: epochs not a dict, skipping")
-            else:
-                try:
-                    remote_epochs = GSet.from_dict(raw)
-                    self.epoch_crdt.merge(remote_epochs)
-                except Exception as e:
-                    logger.warning(f"State from {sender}: epochs merge failed: {e}")
+            elif raw.get("epochs") or raw.get("metadata"):
+                logger.warning(
+                    f"State from {sender}: ignoring epoch finality data in STATE sync"
+                )
 
         # Phase D.3: Scope balance PN-counter entries to sender's own namespace.
         # The sender can only contribute increments/decrements under its own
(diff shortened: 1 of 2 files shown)


JeremyZeng77:
Review note: blocking epoch finality import through generic STATE sync is the right direction, but this also removes state-sync as a recovery path for a node that missed the original `EPOCH_COMMIT` gossip. Before merge I would verify that late/restarted peers still learn finalized epochs through a dedicated commit replay path or periodic commit gossip; otherwise the security fix could create a liveness/regression issue for nodes that join after the commit message has already passed.

TJCurnutte (approved):
Approved. I verified the state-sync change at head `ffb9c50435aa81062a8351610ff355f1678748dc`.

Validation run:

```bash
git diff --check origin/main...HEAD -- node/rustchain_p2p_gossip.py node/tests/test_p2p_state_epoch_sync.py
python3 -B -m py_compile node/rustchain_p2p_gossip.py node/tests/test_p2p_state_epoch_sync.py
RC_P2P_SECRET=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa PYTHONPATH=node python3 -B -m pytest -q node/tests/test_p2p_state_epoch_sync.py node/tests/test_p2p_hardening_phase2.py --tb=short
# 11 passed in 0.22s
```

I also ran a direct origin/main-vs-PR probe with a valid signed `STATE` message from `peer1` carrying `epochs: [999]` plus fake finality metadata. On `origin/main`, the message verified and `_handle_state()` merged the epoch (`contains_999=true`, `metadata_keys=["999"]`). On this PR, the same verified message returned `status=ok` but logged `ignoring epoch finality data in STATE sync` and left `contains_999=false`, `metadata_keys=[]`.

That matches the intended boundary: a peer signature authenticates the sender of a state snapshot, but it is not quorum evidence for epoch finality. Keeping finality on the `EPOCH_COMMIT` path while preserving the existing attestation and balance state-sync handling is the right fix for this issue.

jaxint (approved):

ItsOtherMauridian:
Good catch. Addressed in the latest commit.

The fix now includes an explicit `EPOCH_COMMIT` dispatch path and `_handle_epoch_commit()` so peers can import quorum finality from commit messages instead of only ignoring `STATE["epochs"]`.

The commit handler validates quorum-shaped metadata before marking the epoch finalized:

- required `epoch` and `proposal_hash`,
- `accept_count >= quorum`,
- quorum-sized known voter set,
- rejects malformed/unknown/insufficient-quorum commits.

Validation: `3 passed` for `node/tests/test_p2p_state_epoch_sync.py`, plus `py_compile` and `git diff --check`.

kevinyan911 (commented):
## Code Review — PR #5750

**Reviewer:** @kevinyan911

### What this PR does
Adds `EPOCH_COMMIT` message handler with quorum validation, and explicitly stops importing epoch finality from generic `STATE` sync. Previously `_handle_state()` would merge epochs from any verified peer signature — but a peer's signature only authenticates who sent the snapshot, not that quorum actually committed those epochs. This PR draws the correct security boundary.

### Code quality
- `_handle_epoch_commit()` validates `accept_count` and voter set against quorum before adding to `epoch_crdt` — correct.
- `_handle_state()` now only logs a warning when epochs appear in STATE sync rather than merging them — the intent comment makes the reasoning explicit.
- Regression test creates a valid signed STATE with forged epoch 999 and asserts `epoch_crdt.contains(999) == False` — directly proves the attack vector is closed.
- `py_compile` clean, `git diff --check` clean, 11 pytests pass.

**APPROVED** — scoped security fix with clear test coverage.

---

akaalholdings (changes requested):
Requesting changes because the new `EPOCH_COMMIT` path still lets one peer inject epoch finality by claiming quorum-shaped metadata without proving the listed voters actually signed or sent votes.

The PR correctly stops importing `STATE[\"epochs\"]`, but `_handle_epoch_commit()` currently trusts these sender-provided fields:

```python
accept_count = payload.get("accept_count", 0)
voters = payload.get("voters", [])
...
if accept_count < quorum or len(voter_set) < quorum:
    return {"status": "error", "reason": "insufficient_quorum"}
self.epoch_crdt.add(epoch, ...)
```

A valid message signature only authenticates the committer peer. It does not authenticate `peer2` or `peer3` as voters. I verified this against PR head with a one-off probe: `peer1` alone creates an `EPOCH_COMMIT` for epoch `31337`, lists `peer1`, `peer2`, and `peer3` as voters, and the victim accepts it.

```text
signature_valid= True
result= {'status': 'committed', 'epoch': 31337, 'accept_count': 3}
contains_31337= True
metadata= {'proposal_hash': 'forged-by-one-peer', 'finalized': True, 'accept_count': 3, 'voters': ['peer1', 'peer2', 'peer3'], 'committer': 'peer1'}
```

The included regression suite also currently codifies that behavior:

```bash
RC_P2P_SECRET=$(printf 'a%.0s' {1..64}) PYTHONPATH=node /tmp/rustchain-review-venv/bin/python -B -m pytest -q node/tests/test_p2p_state_epoch_sync.py --tb=short -p no:cacheprovider
# 3 passed
```

This reintroduces the same security boundary problem through a different message type: peer signature plus self-reported quorum metadata is still not quorum evidence. Please require verifiable voter evidence before accepting an external commit, for example signed vote records from the listed voters, or only accept `EPOCH_COMMIT` when it matches locally stored validated votes/quorum state.

ItsOtherMauridian:
Addressed in commit fed6fd6.

I tightened `EPOCH_COMMIT` so sender-reported quorum metadata is no longer accepted as finality evidence. The receiver now requires locally validated `accept` votes from the listed voters before adding the epoch to `epoch_crdt`, and the broadcast path now advertises only accept voters (not reject voters) as commit voters.

Regression coverage now includes both sides of the boundary:
- forged one-peer `EPOCH_COMMIT` with quorum-shaped `voters` is rejected with `unverified_voters`;
- a commit backed by locally stored accept votes is accepted.

Local validation:
```bash
python3 -B -m py_compile node/rustchain_p2p_gossip.py node/tests/test_p2p_state_epoch_sync.py
RC_P2P_SECRET=<64-byte test secret> PYTHONPATH=node python3 -B -m pytest -q node/tests/test_p2p_state_epoch_sync.py node/tests/test_p2p_hardening_phase2.py --tb=short -p no:cacheprovider
# 14 passed
git diff --check
```

Scottcjn:
Merged. This completes the file-by-file review of your full 8-PR batch (#5732 + #5740–#5752).

## Review outcome — all 8 APPROVED

| PR | Fix | Tier | RTC |
|---|---|---|---|
| #5732 | BCOS directory dedupe | Low | 10 |
| #5740 | anti-double-mining: use canonical `epoch_enroll.weight` | Medium | 25 |
| #5742 | claims: read canonical epoch-enroll snapshot for delayed claims | Medium | 25 |
| #5744 | claim-hijack fix: bind claim to registered miner pubkey | High | 50 |
| #5746 | settlement double-spend race: `BEGIN IMMEDIATE` atomic reservation | High | 50 |
| #5748 | ROM-cluster unique index (fixes silently-broken `ON CONFLICT`) | Medium | 25 |
| #5750 | finality-forgery fix: epochs only via validated quorum votes | High | 50 |
| #5752 | double-pay fix: orphan recovery no longer auto-refunds | High | 50 |
| **Total** | | | **285 RTC** |

Every PR: clean 2–3 file scope, real exploit-demonstrating or concurrency tests, VM-zero rule preserved, no smuggled hunks. This is exceptionally strong security work — and filing the paired bug issues (#5743–#5751) before the fixes is exactly the right practice.

**On #5750 specifically:** merged as-is. The offline-node catch-up tradeoff is tracked in follow-up issue #5950 — closing the forgery hole now is the right call; the validated catch-up path gets designed separately.

## To get paid — post your wallet

I have no RTC wallet on file for you. Reply here (or on any of your PRs) with your wallet — format `RTC` + 40 hex characters — and I'll dispatch the full **285 RTC** in one transfer (24h confirm window).

Also: you're clearly a serious security contributor. If you want, the standing bounty queue is at https://github.com/Scottcjn/rustchain-bounties/issues — and consensus/settlement hardening like this is the highest-value work there.

— Sophia

BossChaos (commented):
## PR #5750 Review - Epoch commit handler with strict type validation

### Security Analysis - Consensus Security

This PR adds a new EPOCH_COMMIT message handler with comprehensive type validation.

**Key improvements:**

1. **_handle_epoch_commit()**: New handler for finalized epoch commits
2. **Strict type checking**: Validates epoch (int), proposal_hash (str), accept_count (int >= 0), voters (list)
3. **Known voters validation**: Verifies all listed voters are known nodes
4. **Commit voter filtering**: Only includes accept voters in the commit message broadcast, not reject voters
5. **Type safety**: Returns specific error reasons for each validation failure

**Note**: Also fixes _handle_epoch_vote() to filter voters to only those who voted "accept" when broadcasting the commit message.

**Recommendation**: Merge - important consensus security hardening.

ItsOtherMauridian:

ItsOtherMauridian:
Hi, quick payout status check.

I posted my RTC payout address here after the approved 8-PR batch totaling 285 RTC:

RTC71d6976297ed35377b867f13ed962f54020ef434

The public balance endpoints still show 0.0 RTC for this address. Could you confirm whether the 285 RTC transfer was dispatched, or if it is still pending?

Also noting the separate 5 RTC for #5665 was marked paid/pending in #305, but that has not appeared in the wallet balance yet either.

Thanks.

ItsOtherMauridian:
Hi, following up on this payout status check.

My RTC payout address for the approved 8-PR batch is:

RTC71d6976297ed35377b867f13ed962f54020ef434

Could you confirm whether the 285 RTC transfer has been dispatched yet, or whether it is still pending?

The public balance endpoints for this address still show 0.0 RTC. The separate 5 RTC noted in #305 also does not appear to have arrived yet.

Thanks.

PR #5756, opened 2026-05-20
fix: store refreshed reputation cache results

dicnunz:
## Summary
- fixes the background reputation cache refresh path so stale entries store the recalculated result, not the old cached payload with a fresh timestamp
- extracts the stale-entry refresh into a small method that can be regression-tested directly
- adds a focused test proving refreshed cache data replaces stale data

## Why
Before this change, `_refresh_loop()` called `calculate(w)` for stale wallets but discarded the returned score data. It then refreshed only the timestamp around the old cached value, which could keep stale reputation scores alive indefinitely.

## Tests
- `python3 -m py_compile agent_reputation.py tests/test_agent_reputation.py`
- `uv run --with flask --with pytest python -m pytest tests/test_agent_reputation.py -q`

### DIFF
--- a/agent_reputation.py
+++ b/agent_reputation.py
@@ -285,17 +285,23 @@ class ReputationEngine:
             else:
                 self._cache.clear()
 
+    def _refresh_stale_cache_entries(self):
+        now = time.time()
+        with self._lock:
+            stale = [
+                w for w, (_, ts) in self._cache.items()
+                if now - ts > CACHE_TTL_S
+            ]
+        for w in stale:
+            refreshed = self.calculate(w)
+            with self._lock:
+                if w in self._cache:
+                    self._cache[w] = (refreshed, time.time())
+
     def _refresh_loop(self):
         while True:
             time.sleep(CACHE_TTL_S)
-            with self._lock:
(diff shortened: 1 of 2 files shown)


jaxint (approved):

minyanyi (approved):
Tested locally with `python -B -m pytest -q tests/test_agent_reputation.py --tb=short`: 10 passed.

Two specific observations:

1. `agent_reputation.py` now stores the `calculate(w)` return value back into `_cache` inside `_refresh_stale_cache_entries()`. That fixes the old refresh-loop behavior where stale entries only had their timestamp renewed while the cached reputation payload stayed unchanged.

2. The stale-key scan is done under `_lock`, but the potentially slower `calculate()` call happens outside that lock. That keeps normal `get()` / `invalidate()` cache operations from being blocked by DB/API work, while the final `if w in self._cache` guard still avoids resurrecting entries invalidated during the refresh.

The added regression test is also targeted: it monkeypatches `calculate()` and asserts the cached payload changes to the recalculated value, so it catches the exact bug this PR fixes rather than only checking timestamp movement.

PR #5757, opened 2026-05-20
fix: reject payout amounts above i64

dicnunz:
## Summary
- rejects wallet transfer amounts whose micro-RTC quantization exceeds signed i64 max
- applies the same guard to the root deployment shim and node runtime preflight module
- adds public-validator regression tests for admin and signed transfers

## Why
The preflight helpers call the scaled amount `amount_i64`, but enormous positive amounts could quantize above signed i64 max and still pass validation. That can push impossible amounts into downstream wallet code.

## Tests
- `python3 -m py_compile payout_preflight.py node/payout_preflight.py tests/test_payout_preflight.py`
- `uv run --with flask --with pytest python -m pytest tests/test_payout_preflight.py -q`

### DIFF
--- a/node/payout_preflight.py
+++ b/node/payout_preflight.py
@@ -7,6 +7,7 @@ from typing import Any, Dict, Optional, Tuple
 
 
 MICRO_RTC = Decimal("1000000")
+MAX_I64 = 2**63 - 1
 
 
 def _is_rtc_address(value: str) -> bool:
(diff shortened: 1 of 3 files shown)


jaxint (approved):

minyanyi (approved):
Tested locally with `python -B -m pytest -q tests/test_payout_preflight.py --tb=short`: 50 passed.

Two specific observations:

1. The new `_validate_amount_i64()` helper is used by both admin and signed transfer preflight paths, so the `MAX_I64` guard is applied before either route can hand an oversized micro-RTC value to lower-level storage/serialization code. That closes the overflow class without duplicating slightly different checks in the two validators.

2. The boundary test value `9223372036854.775808` is one micro-RTC above `2**63 - 1` after the existing `ROUND_DOWN` quantization, so it exercises the exact overflow edge instead of relying on an obviously huge amount. Existing dust/sub-micro behavior is preserved because the helper still returns `amount_too_small_after_quantization` when the quantized value is `0`.

PR #5755, opened 2026-05-20
fix: clarify reputation eligibility denial reason

dicnunz:
## Summary
- fixes `/agent/reputation/check-eligibility` so level-cap denials explain the actual max job value instead of always reporting the high-value gate
- keeps the existing high-value >50 RTC veteran-only denial for trusted agents
- adds a focused regression test for newcomer max-value rejection

## Why
A newcomer/known agent rejected for exceeding its level cap could receive a misleading high-value-job message even when the requested job was below the high-value threshold. That makes claim eligibility debugging confusing for agents and job posters.

Related reward route: RustChain bug bounty / report-a-bug program (#305).

## Tests
- `python3 -m py_compile agent_reputation.py tests/test_agent_reputation.py`
- `uv run --with flask --with pytest python -m pytest tests/test_agent_reputation.py -q`

AI-assisted contribution by Codex for dicnunz.

PR #5758, opened 2026-05-20
fix: use preflight amount units for admin transfers

dicnunz:
## Summary
- uses the exact `amount_i64` already computed by payout preflight for admin wallet transfers
- avoids recomputing micro-RTC units through float multiplication
- adds a regression showing `0.000249` RTC must persist as 249 micro-RTC, not 248

## Why
`validate_wallet_transfer_admin()` already quantizes the request with Decimal, but `/wallet/transfer` discarded that value and recomputed `int(amount_rtc * 1000000)` after converting to float. Certain valid decimal amounts could lose or gain one micro-unit before being written to `pending_ledger`.

## Tests
- `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_wallet_transfer_admin_idempotency.py`
- `uv run --with flask --with pytest python -m pytest tests/test_wallet_transfer_admin_idempotency.py -q`

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -7399,6 +7399,7 @@ def wallet_transfer_v2():
     from_miner = pre.details["from_miner"]
     to_miner = pre.details["to_miner"]
     amount_rtc = pre.details["amount_rtc"]
+    amount_i64 = int(pre.details["amount_i64"])
     reason = str((data or {}).get('reason', 'admin_transfer'))
     idempotency_key = ""
     raw_idempotency_key = (data or {}).get("idempotency_key")
@@ -7409,7 +7410,6 @@ def wallet_transfer_v2():
         if not re.fullmatch(r"[A-Za-z0-9._:-]{1,128}", idempotency_key):
             return jsonify({"error": "invalid_idempotency_key"}), 400
     
-    amount_i64 = int(amount_rtc * 1000000)
     now = int(time.time())
     confirms_at = now + CONFIRMATION_DELAY_SECONDS
     current_epoch = current_slot()
--- a/tests/test_wallet_transfer_admin_idempotency.py
+++ b/tests/test_wallet_transfer_admin_idempotency.py
@@ -104,6 +104,35 @@ def test_admin_transfer_idempotency_key_reuses_pending_transfer(admin_transfer_c
     assert pending_total == 1_000_000
 
 
+def test_admin_transfer_uses_preflight_amount_i64_without_float_loss(admin_transfer_client):
+    client, db_path = admin_transfer_client
+
+    with sqlite3.connect(db_path) as conn:
+        conn.execute(
+            "INSERT INTO balances (miner_id, amount_i64) VALUES (?, ?)",
+            ("founder_community", 1_000_000),
+        )
+        conn.commit()
+
+    response = client.post(
+        "/wallet/transfer",
+        json={
+            "from_miner": "founder_community",
+            "to_miner": "contributor",
+            "amount_rtc": "0.000249",
+        },
+        headers={"X-Admin-Key": "a" * 32},
+    )
+    assert response.status_code == 200
+
+    with sqlite3.connect(db_path) as conn:
+        (pending_amount,) = conn.execute(
+            "SELECT amount_i64 FROM pending_ledger"
+        ).fetchone()
+
+    assert pending_amount == 249
+
+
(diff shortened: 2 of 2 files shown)


jaxint (approved):

minyanyi (approved):
Tested locally with `python -B -m pytest -q tests/test_wallet_transfer_admin_idempotency.py --tb=short`: 3 passed.

Two specific observations:

1. `wallet_transfer_v2()` now uses `pre.details["amount_i64"]` from the preflight validator instead of recomputing `int(amount_rtc * 1000000)` from the JSON-facing RTC amount. That is the right source of truth because preflight already parsed the input through `Decimal` and quantized it consistently.

2. The regression case with `amount_rtc: "0.000249"` is well chosen: it is small enough that a float-based recomputation could lose precision in the micro-unit conversion, and the test asserts the persisted `pending_ledger.amount_i64` is exactly `249`. This also exercises the real Flask route and SQLite insert path rather than only the helper.

Scottcjn:
@dicnunz — 45 RTC payout queued (pending_id 1636, confirms in 24h to `0xC02e5E5aEd363E5F59466D13A590d73275C6Af59`).

Covers your 6 merged PRs today:
- #5755 reputation eligibility-denial reason (5)
- #5756 cache-refresh stores recalculated result — **real bug fix**, the old loop discarded `calculate()`'s result and only bumped the timestamp (10)
- #5757 i64 payout overflow guard (5)
- #5758 + #5759 preflight `amount_i64` instead of `int(amount_rtc * 1000000)` — **money-path float-precision fix**, verified `0.000249 RTC → 249` exact (10 + 10)
- #5763 sophia inspect non-object JSON rejection (5)

Audited all six diffs for smuggled hunks — clean. Solid correctness work, especially the float-precision catch on the transfer path. Thanks.

dicnunz:
Thanks, Scott. I can see the queued payout tx `06daddf4f770c6ede09eef5b1725c8cb` and the target wallet endpoint now shows 25 RTC confirmed plus 45 RTC pending for `0xC02e5E5aEd363E5F59466D13A590d73275C6Af59`.

I am blocked at the cash-out step: live `POST /withdraw/request` and `POST /withdraw/register` both return nginx 404, and the same Base address currently has 0 wRTC / 0 USDC on-chain. What is the working withdrawal or bridge path for this payout, or can this be released as wRTC to the same Base address?

dicnunz:
Follow-up cash-out proof after testing the raw node as well:

- `https://rustchain.org/withdraw/request` still returns nginx 404.
- `https://50.28.86.131/withdraw/request` is live, but a signed-style request for `miner_pk=dicnunz` returns `{"error":"Insufficient balance","balance":0.0}`.
- The same raw node returns `{"balance_rtc":40.0,"miner_pk":"dicnunz"}` from `GET /balance/dicnunz`.
- `GET /withdraw/history/dicnunz` and `POST /withdraw/register` both return `Unauthorized - admin key required`.

So the current blocker appears to be the withdraw service not reading the same credited balance table/key, plus no user-side registration path. If there is a different miner key, registration step, or Base/Solana bridge endpoint I should use for the queued 45 RTC / existing 40 RTC, please point me at it and I will run the withdrawal immediately.

dicnunz:
Additional cash-out proof after checking the wallet and Base-side paths:

- `GET /wallet/history?miner_id=0xC02e5E5aEd363E5F59466D13A590d73275C6Af59` now shows the queued `45.0 RTC` transfer `06daddf4f770c6ede09eef5b1725c8cb` as `status: pending`, plus the earlier `25.0 RTC` transfer.
- `GET /wallet/history?miner_id=dicnunz` shows the six 5 RTC merge-award transfers as `status: pending`, while `/wallet/balance?miner_id=dicnunz` still reports only the old 40 RTC confirmed amount.
- Base mainnet `eth_call` balance checks for wRTC `0x5683C10596AaA09AD7F4eF13CAB94b9b74A669c6` and USDC `0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913` return `0` for both the queued payout target and my withdrawal destination.
- `POST /wallet/link-coinbase` returns `Unauthorized — admin key required`; `/wallet/swap-info` only documents USDC -> wRTC, not native/internal RTC -> sellable wRTC/USDC.

So the exact remaining conversion blocker is admin-side release/confirmation or a user-callable native RTC -> wRTC/USDC withdrawal/bridge path. If you can confirm the pending transfers, release the owed amount as wRTC/USDC, or point me at the correct callable redemption endpoint/key flow, I will run the cash-out immediately.

PR #5759, opened 2026-05-20
fix: use preflight amount units for signed transfers

dicnunz:
## Summary
- uses the exact `amount_i64` computed by signed-transfer preflight when writing pending signed transfers
- avoids recomputing micro-RTC units through float multiplication in `/wallet/transfer/signed`
- adds a regression showing `0.000249` RTC must persist as 249 micro-RTC, not 248

## Why
`validate_wallet_transfer_signed()` already quantizes the request with Decimal and returns `amount_i64`, but the signed-transfer route discarded that value and recomputed `int(amount_rtc * 1000000)` after converting the amount to float. Certain valid decimal amounts could lose a micro-unit before being written to `pending_ledger`.

## Tests
- `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_signed_transfer_replay.py`
- `uv run --with flask --with pytest --with cryptography --with httpx python -m pytest tests/test_signed_transfer_replay.py -q`

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -8464,7 +8464,7 @@ def wallet_transfer_signed():
     # SECURITY/HARDENING: signed transfers should follow the same 2-phase commit
     # semantics as admin transfers (pending_ledger + delayed confirmation). This
     # prevents bypassing the 24h pending window via the signed endpoint.
-    amount_i64 = int(amount_rtc * 1000000)
+    amount_i64 = int(pre.details["amount_i64"])
     now = int(time.time())
     confirms_at = now + CONFIRMATION_DELAY_SECONDS
     current_epoch = current_slot()
--- a/tests/test_signed_transfer_replay.py
+++ b/tests/test_signed_transfer_replay.py
@@ -151,6 +151,30 @@ def test_insufficient_balance_does_not_burn_nonce(signed_transfer_client):
     assert pending_count == 1
 
 
+def test_signed_transfer_uses_preflight_amount_i64_without_float_loss(signed_transfer_client):
+    client, db_path = signed_transfer_client
+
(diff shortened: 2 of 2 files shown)


slegarraga (approved):
Reviewed the signed-transfer amount-unit fix against head `083efb186f2d9213968204cff297297ddfd0080e`.

What I checked:
- The route now uses `pre.details["amount_i64"]`, i.e. the Decimal/ROUND_DOWN quantized micro-RTC value produced by `validate_wallet_transfer_signed()`, instead of re-multiplying the already-float `amount_rtc`.
- That keeps signed transfers aligned with admin-transfer preflight semantics and avoids a one-micro-unit loss for values such as `0.000249`.
- The new regression covers the exact failure mode by persisting a pending signed transfer and asserting `pending_ledger.amount_i64 == 249`.

Validation run locally:
```bash
python3 -B -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_signed_transfer_replay.py
uv run --with flask --with pytest --with cryptography --with httpx python -m pytest tests/test_signed_transfer_replay.py -q --tb=short -p no:cacheprovider
# 5 passed
```

No blocking issues found in this focused patch. Approved.

minyanyi (approved):
Tested locally with `python -B -m pytest -q tests/test_signed_transfer_replay.py --tb=short`: 5 passed.

Two specific observations:

1. The signed transfer route now takes `amount_i64` directly from `validate_wallet_transfer_signed()` instead of recomputing it from `amount_rtc`. That aligns the persisted pending-ledger amount with the Decimal-based preflight parser and avoids the signed endpoint drifting from the admin transfer behavior.

2. The regression test runs through the real `/wallet/transfer/signed` Flask route with mocked signature verification, then checks SQLite `pending_ledger.amount_i64` for the exact `249` micro-unit value from `"0.000249"`. That covers the route-level persistence path while leaving replay protection and nonzero-fee tests intact.

PR #5760, opened 2026-05-20
[UTXO-BUG] Bound mempool text fields before persistence

akaalholdings:
## Summary
- Add explicit byte caps for UTXO output addresses, output metadata JSON, and public mempool `tx_id` values.
- Reject oversized fields before `mempool_add()` persists attacker-controlled `tx_data_json` or locks an input.
- Mirror the output-field checks in direct `apply_transaction()` so oversized persisted box fields fail closed before state mutation.

## Finding
Current `main` accepts a valid mempool transfer with an oversized output address. In local reproduction, a single valid input plus a 200 KB output address made `mempool_add()` return `True` and stored a 200,270 byte `tx_data_json` row. `MAX_POOL_SIZE` only caps row count, so oversized text fields can bypass the intended mempool resource limit and create avoidable DB/memory pressure.

This patch keeps the limits conservative:
- output address: 256 UTF-8 bytes
- output `tokens_json` / `registers_json`: 8 KiB each
- mempool `tx_id`: 128 UTF-8 bytes

## Bounty route
Scottcjn/rustchain-bounties#2819

Claimant: @akaalholdings
Suggested severity: Low / Medium, depending on maintainer assessment of mempool resource-exhaustion impact.

## Validation
- `PYTHONPATH=node python3 -m unittest node.test_utxo_db.TestUtxoDB.test_mempool_rejects_oversized_tx_id_without_locking_input node.test_utxo_db.TestUtxoDB.test_mempool_rejects_oversized_output_text_without_locking_input node.test_utxo_db.TestUtxoDB.test_apply_transaction_rejects_oversized_output_text node.test_utxo_db.TestUtxoDB.test_mempool_accepts_valid_tx -v` -> OK
- `PYTHONPATH=node python3 -m unittest node.test_utxo_db.TestUtxoDB.test_mempool_rejects_unpersistable_outputs_without_locking_input node.test_utxo_db.TestUtxoDB.test_apply_transaction_rejects_unpersistable_outputs node.test_utxo_db.TestUtxoDB.test_mempool_accepts_valid_tx node.test_utxo_db.TestUtxoDB.test_transfer -v` -> OK
- `python3 -m py_compile node/utxo_db.py node/test_utxo_db.py` -> OK
- `git diff --check -- node/utxo_db.py node/test_utxo_db.py` -> OK

Full `PYTHONPATH=node python3 node/test_utxo_db.py` currently fails on two pre-existing dust-threshold tests in `main`; those are already covered by open PR #5696 and are not part of this claim.

### DIFF
--- a/node/test_utxo_db.py
+++ b/node/test_utxo_db.py
@@ -17,7 +17,8 @@ import utxo_db as utxo_db_module
 from utxo_db import (
     UtxoDB, coin_select, compute_box_id, address_to_proposition,
     proposition_to_address, UNIT, DUST_THRESHOLD, MAX_COINBASE_OUTPUT_NRTC,
-    MAX_OUTPUTS, MAX_SQLITE_INT64,
+    MAX_OUTPUTS, MAX_SQLITE_INT64, MAX_UTXO_ADDRESS_BYTES,
+    MAX_UTXO_METADATA_BYTES, MAX_MEMPOOL_TX_ID_BYTES,
 )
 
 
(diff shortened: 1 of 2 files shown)


minyanyi (changes requested):
Thanks for tightening the persisted text bounds, but this currently needs one fix before merge.

I tested the PR locally with `PYTHONPATH=node python -B -m pytest -q node/test_utxo_db.py --tb=short`. Result: 86 passed, 14 subtests passed, 2 failed.

The failures are both dust-output regressions:

- `TestUtxoDB.test_below_dust_output_rejected` now gets `ok == True` for a transfer containing an output below `DUST_THRESHOLD`.
- `TestUtxoDB.test_mempool_rejects_below_dust_output` now gets `ok == True`, so the mempool can admit the same below-dust output and lock the input.

I also ran the same test file on current `origin/main` and it passes there (`71 passed`), so the failure is introduced by this branch's new/changed validation expectations rather than a baseline breakage.

The new byte-length checks for `tx_id`, output addresses, `tokens_json`, and `registers_json` look directionally good. The missing piece is that `_normalize_outputs()` currently accepts any positive int64 `value_nrtc`; if this helper is meant to be the shared output gate for both mempool and block application, it also needs to preserve the below-dust rejection, likely by rejecting non-minting outputs below `DUST_THRESHOLD` or by applying an equivalent dust check after normalization.

BossChaos (commented):
## PR #5760 Review — Mempool size bound regression tests

### Security Analysis

This PR adds regression tests for mempool transaction size bounds.

**Key test cases:**

1. **`test_mempool_rejects_oversized_tx_id_without_locking_input`**: Verifies transactions with tx_id exceeding `MAX_MEMPOOL_TX_ID_BYTES` are rejected before locking any inputs.

2. **`test_mempool_rejects_oversized_output_text_without_locking_input`**: Tests 3 edge cases:
   - Oversized `address` field (> `MAX_UTXO_ADDRESS_BYTES`)
   - Oversized `tokens_json` field (> `MAX_UTXO_METADATA_BYTES`)
   - Oversized `registers_json` field (> `MAX_UTXO_METADATA_BYTES`)

All tests verify that the mempool does NOT lock the input box (no double-spend prevention) when rejecting oversized transactions.

**Recommendation**: Merge — regression tests for mempool size bounds, prevents DoS via oversized transactions.

PR #5770, opened 2026-05-20
fix: validate header route json bodies

### LINKED ISSUE #5769: Integrated header routes 500 on non-object JSON
Two integrated-node header endpoints parse JSON with `request.get_json(force=True, silent=True) or {}` and immediately call `.get(...)` on the result. A valid non-object JSON body, such as an array, is truthy and reaches that `.get` call, producing a 500 instead of a controlled 400.

Affected endpoints reproduced locally:
- `POST /miner/headerkey` with valid admin key and JSON array body
- `POST /headers/ingest_signed` with JSON array body

Expected: reject non-object JSON bodies with 400 before field extraction.
Actual: `AttributeError: list object has no attribute get` and HTTP 500.

I am preparing a small fix with regression tests.

dicnunz:
## Summary
- Fixes #5769.
- Reject non-object JSON bodies before `/miner/headerkey` and `/headers/ingest_signed` read fields with `.get(...)`.
- Normalizes signed-header string fields before `.strip()` to avoid type-driven 500s.
- Adds focused regression tests for JSON arrays hitting both routes.

## Local reproduction before fix
- `POST /miner/headerkey` with a valid admin key and JSON array body returned 500: `AttributeError: list object has no attribute get`.
- `POST /headers/ingest_signed` with a JSON array body returned the same 500.

## Tests
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest --with flask --with pynacl --with requests --with prometheus-client python -m pytest tests/test_header_routes_json_validation.py -q`
- `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_header_routes_json_validation.py`
- `git diff --check -- node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_header_routes_json_validation.py`
- `python3 tools/bcos_spdx_check.py --base-ref origin/main`

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -4294,7 +4294,9 @@ def miner_set_header_key():
     if not admin_key or not hmac.compare_digest(provided_key, admin_key):
         return jsonify({"ok":False,"error":"unauthorized"}), 403
 
-    body = request.get_json(force=True, silent=True) or {}
+    body = request.get_json(force=True, silent=True)
+    if not isinstance(body, dict):
+        return jsonify({"ok":False,"error":"invalid_json_body"}), 400
     miner_id   = str(body.get("miner_id","")).strip()
     pubkey_hex = str(body.get("pubkey_hex","")).strip().lower()
     if not miner_id or len(pubkey_hex) != 64:
@@ -4328,13 +4330,15 @@ def ingest_signed_header():
       5) on success: validate header continuity, persist, update tip, bump metrics
     """
     start = time.time()
-    body = request.get_json(force=True, silent=True) or {}
+    body = request.get_json(force=True, silent=True)
+    if not isinstance(body, dict):
+        return jsonify({"ok":False,"error":"invalid_json_body"}), 400
 
-    miner_id = (body.get("miner_id") or "").strip()
+    miner_id = str(body.get("miner_id") or "").strip()
     header   = body.get("header") or {}
-    msg_hex  = (body.get("message") or "").strip().lower()
-    sig_hex  = (body.get("signature") or "").strip().lower()
-    inline_pk= (body.get("pubkey") or "").strip().lower()
+    msg_hex  = str(body.get("message") or "").strip().lower()
+    sig_hex  = str(body.get("signature") or "").strip().lower()
+    inline_pk= str(body.get("pubkey") or "").strip().lower()
 
     if not miner_id or not sig_hex or (not header and not msg_hex):
         return jsonify({"ok":False,"error":"missing fields"}), 400
--- a/tests/test_header_routes_json_validation.py
+++ b/tests/test_header_routes_json_validation.py
@@ -1,0 +1,33 @@
+# SPDX-License-Identifier: MIT
+
+import sys
+
+import pytest
+
+integrated_node = sys.modules["integrated_node"]
+
+
+@pytest.fixture
+def client(monkeypatch):
+    monkeypatch.setenv("RC_ADMIN_KEY", "0" * 32)
+    integrated_node.app.config["TESTING"] = True
+    with integrated_node.app.test_client() as test_client:
+        yield test_client
+
+
+def test_miner_headerkey_rejects_non_object_json(client):
+    response = client.post(
+        "/miner/headerkey",
+        headers={"X-API-Key": "0" * 32},
+        json=["not", "an", "object"],
+    )
+
+    assert response.status_code == 400
+    assert response.get_json() == {"ok": False, "error": "invalid_json_body"}
+
+
+def test_ingest_signed_header_rejects_non_object_json(client):
+    response = client.post("/headers/ingest_signed", json=["not", "an", "object"])
+
+    assert response.status_code == 400
+    assert response.get_json() == {"ok": False, "error": "invalid_json_body"}


GoodMorC (commented):
Reviewed PR #5770 for the code review bounty.

What I checked:
- The two changed routes now reject top-level non-object JSON before calling `.get(...)`, which addresses the array-body 500 described in #5769.
- The signed-header route now normalizes string-like fields before `.strip()`, so non-string `message`, `signature`, `pubkey`, and `miner_id` inputs should fall into the existing validation/error paths instead of raising type errors.
- The new regression tests cover the important top-level array case for both `/miner/headerkey` and `/headers/ingest_signed`.
- GitHub CI is green for the PR, including the main `test` workflow and BCOS gates at the time I reviewed it.

Local validation:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 /tmp/rustchain-pr5770-venv/bin/python -m pytest tests/test_header_routes_json_validation.py -q` -> `2 passed`
- `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_header_routes_json_validation.py` -> passed
- `git diff --check upstream/main...HEAD -- node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_header_routes_json_validation.py` -> passed after refreshing `upstream/main`

No blocker found. One small follow-up suggestion: if the project wants a consistent API contract across these routes, consider documenting or centralizing the `invalid_json_body` response helper, since bridge/header/beacon routes are now converging on the same top-level object guard pattern. That is not necessary for this PR to merge.

Assessment: approve from this review pass. The patch is narrowly scoped, tested, and removes a real 500-class input validation edge case.

AI assistance disclosure: OpenAI Codex was used to inspect the diff and run the local validation above.

surim0n (commented):
I verified the patch locally with:

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest --with flask --with pynacl --with requests --with prometheus-client python -m pytest tests/test_header_routes_json_validation.py -q` -> 2 passed
- `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_header_routes_json_validation.py`
- `git diff --check -- node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_header_routes_json_validation.py`

The non-object JSON guard itself looks correct for both routes.

One gap I would address before relying on this endpoint regression coverage: the new `/miner/headerkey` test only exercises the early `invalid_json_body` return. It does not initialize or cover the authorized success path, and that path currently fails on a fresh initialized DB because `miner_header_keys` is referenced but not created by `init_db()`.

I reproduced that locally by loading the integrated node with a temporary SQLite DB, running `init_db()`, then posting an authorized object body to `/miner/headerkey` with `{"miner_id": null, "pubkey_hex": "a" * 64}`. Instead of returning a validation error or success, it reaches the insert and raises:

`sqlite3.OperationalError: no such table: miner_header_keys`

That table is inserted into at this route and selected from by `/headers/ingest_signed`, so either this PR or a nearby follow-up should add the schema initialization and a valid-object regression. Otherwise the PR proves the malformed-array case is fixed, but not that the repaired route remains usable once callers send an object body.

jaxint (approved):

BossChaos (commented):
## Code Review - PR #5770: Validate Header Route JSON Bodies

**Overall: Consistent validation pattern for miner header-key admin route.**

### Strengths
- `miner_set_header_key()` validates JSON body before field access.
- New test `test_header_routes_json_validation.py` covers rejection path.
- Pattern is consistent with other PRs in this series.

**Standard review - 5-8 RTC.**

dicnunz:
Thanks for the success-path check. I agree that the fresh-DB `miner_header_keys` schema gap is real, but I am keeping this PR scoped to the non-object JSON body crash in #5769 so it does not grow into a schema migration PR.

The table initialization path is being handled separately in #5778 (`Create miner_header_keys table during DB init`), which is currently open/CLEAN with green checks and approvals. If maintainers prefer to land the schema fix first, this PR should rebase cleanly afterward; if they prefer bundling, I can adjust, but I do not want to duplicate #5778 in this branch.

minyanyi (approved):
Tested locally with `python -B -m pytest -q tests/test_header_routes_json_validation.py --tb=short`: 2 passed.

Two specific observations:

1. Both `/miner/headerkey` and `/headers/ingest_signed` now reject parsed non-object JSON before accessing `.get()`, so array payloads take a deterministic 400 `invalid_json_body` response instead of depending on attribute errors or later field validation.

2. The `ingest_signed_header` path now string-coerces `miner_id`, `message`, `signature`, and inline `pubkey` before `.strip()`/`.lower()`. That is useful beyond the array-body case because scalar JSON values in those fields should not crash the route while it is deciding whether to return the normal missing/bad-field responses.

surim0n (commented):
Reviewed the header route JSON validation patch. The object check is needed in both touched routes because `/miner/headerkey` previously assumed `.get()` existed and `/headers/ingest_signed` accepted body shapes that were not valid request objects.

The added string coercion on `miner_id`, `message`, `signature`, and `pubkey` also reduces accidental 500s when callers send numeric/null values. Coverage is focused on non-object JSON; a useful follow-up would be malformed JSON with `Content-Type: application/json`, but the current patch does remove the array/list crash surface.

508704820 (commented):

508704820 (commented):

BossChaos (commented):
## PR #5770 Review — Miner header key / ingest signed header JSON validation

### Security Analysis

This PR adds strict JSON body validation to two miner endpoints.

**Key improvements:**

1. **`miner_set_header_key()`**: Added `isinstance(body, dict)` check before accessing fields. Returns 400 `"invalid_json_body"`.

2. **`ingest_signed_header()`**: Added `isinstance(body, dict)` check. Also refactored field extraction to use `str(body.get(...))` for `miner_id`, `message`, `signature`, and `pubkey` — this prevents crashes on non-string values.

3. **Test coverage**: Two tests verifying 400 responses for array JSON bodies on both endpoints.

**Recommendation**: Merge — consistent with the JSON validation pattern applied across all RustChain endpoints.

PR #5778, opened 2026-05-20
Create miner_header_keys table during DB init

### LINKED ISSUE #5777: Bug: /miner/headerkey can 500 because miner_header_keys table is never initialized
## Summary

`POST /miner/headerkey` can raise a server error on a freshly initialized node database because the route writes to `miner_header_keys`, but `init_db()` does not create that table.

The same missing table also affects `/headers/ingest_signed`, which selects from `miner_header_keys` when inline pubkeys are not allowed.

## Steps to reproduce

From a clean checkout, load the integrated node with a fresh temporary SQLite DB, run `init_db()`, then call the authorized `/miner/headerkey` route with a JSON object body:

```bash
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project \
  --with flask --with pynacl --with requests --with prometheus-client \
  python - <<'PY'
import os, sys, tempfile
fd, path = tempfile.mkstemp()
os.close(fd)

os.environ['RC_ADMIN_KEY'] = '0' * 32
os.environ['DB_PATH'] = path
sys.path.insert(0, '.')
sys.path.insert(0, 'node')

import tests.conftest  # loads integrated_node with test mocks
node = sys.modules['integrated_node']
node.DB_PATH = path
node.init_db()
node.app.config['TESTING'] = True

with node.app.test_client() as client:
    resp = client.post(
        '/miner/headerkey',
        headers={'X-API-Key': '0' * 32},
        json={'miner_id': None, 'pubkey_hex': 'a' * 64},
    )
    print(resp.status_code, resp.get_json())
PY
```

## Actual behavior

The request reaches the insert and crashes with:

```text
sqlite3.OperationalError: no such table: miner_header_keys
```

The failing statement is:

```python
INSERT INTO miner_header_keys(miner_id,pubkey_hex) VALUES(?,?) ...
```

## Expected behavior

A freshly initialized DB should include the `miner_header_keys` table, so:

- `/miner/headerkey` can return normal validation errors or successfully store a key.
- `/headers/ingest_signed` can safely look up registered header keys.
- Tests that exercise valid object-body paths do not depend on hidden pre-existing production schema.

## Impact

This is a real endpoint availability/regression risk. The current JSON-body validation tests can pass while the authorized object-body path still fails on a clean DB schema. That means a node started with a fresh DB may accept malformed-body hardening but still 500 when an admin tries to register a miner header key.

## Suggested fix

Add schema initialization for `miner_header_keys` in `init_db()`, likely with:

```sql
CREATE TABLE IF NOT EXISTS miner_header_keys (
  miner_id TEXT PRIMARY KEY,
  pubkey_hex TEXT NOT NULL
)
```

Then add a regression test that initializes a fresh DB and proves an authorized object-body request no longer raises `OperationalError`.

## Bounty

Submitting for bug bounty #305. Suggested severity: medium, because this is an API crash / missing schema initialization bug affecting an admin route and dependent signed-header ingestion flow.

Wallet/miner ID: `RTC5800896ed658aa511d029361b6d7388ddb29248b`

surim0n:
## Summary
- create the `miner_header_keys` table in `init_db()` so fresh databases support header-key registration
- add a regression test that initializes a temp DB, calls `/miner/headerkey`, and verifies the key is stored

Fixes #5777

## Verification
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest --with flask --with pynacl --with requests --with prometheus-client python -m pytest tests/test_miner_header_keys_schema.py -q`
- `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_miner_header_keys_schema.py`
- `git diff --check`

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -1434,6 +1434,12 @@ def init_db():
             )
         """)
         c.execute("""
+            CREATE TABLE IF NOT EXISTS miner_header_keys(
+                miner_id TEXT PRIMARY KEY,
+                pubkey_hex TEXT NOT NULL
+            )
+        """)
+        c.execute("""
             CREATE TABLE IF NOT EXISTS schema_version(
                 version INTEGER PRIMARY KEY,
                 applied_at INTEGER NOT NULL
--- a/tests/test_miner_header_keys_schema.py
+++ b/tests/test_miner_header_keys_schema.py
@@ -1,0 +1,42 @@
+# SPDX-License-Identifier: MIT
+"""Regression coverage for miner header-key schema initialization."""
+
+import sqlite3
+import sys
+
+
+integrated_node = sys.modules["integrated_node"]
+ADMIN_KEY = "0" * 32
+ADMIN_HEADERS = {"X-API-Key": ADMIN_KEY}
+
+
+def test_init_db_creates_miner_header_keys_for_headerkey_route(tmp_path, monkeypatch):
+    db_path = tmp_path / "rustchain.db"
+    monkeypatch.setenv("RC_ADMIN_KEY", ADMIN_KEY)
+    monkeypatch.setattr(integrated_node, "DB_PATH", str(db_path))
+    integrated_node.app.config["TESTING"] = True
+
+    integrated_node.init_db()
+
+    pubkey_hex = "a" * 64
+    with integrated_node.app.test_client() as client:
+        response = client.post(
+            "/miner/headerkey",
+            headers=ADMIN_HEADERS,
+            json={"miner_id": "miner-a", "pubkey_hex": pubkey_hex},
+        )
+
+    assert response.status_code == 200
+    assert response.get_json() == {
+        "ok": True,
+        "miner_id": "miner-a",
+        "pubkey_hex": pubkey_hex,
+    }
+
+    with sqlite3.connect(db_path) as db:
+        stored = db.execute(
+            "SELECT pubkey_hex FROM miner_header_keys WHERE miner_id = ?",
+            ("miner-a",),
+        ).fetchone()
+
+    assert stored == (pubkey_hex,)


bitdamii (approved):
Verified the patch against the reported schema gap.

What I checked:
- `miner_set_header_key()` writes to `miner_header_keys`, and the PR now creates that table during `init_db()`.
- The new regression uses a tmp DB path, calls `init_db()`, posts to `/miner/headerkey`, and confirms the key is persisted in SQLite.
- `git diff --check origin/main...review-pr-5778` is clean.
- `python -m py_compile tests/test_miner_header_keys_schema.py node/rustchain_v2_integrated_v2.2.1_rip200.py` passes.
- Ran the targeted test locally in a clean venv with pytest/flask/pynacl installed: `tests/test_miner_header_keys_schema.py -q` -> `1 passed`.

This is a focused additive schema fix with regression coverage for the endpoint that previously depended on the missing table. I do not see a blocker.

BossChaos (commented):
## Code Review — PR #5778: Create miner_header_keys Table During DB Init

**Overall: Clean schema addition. Regression test is appropriate.**

### Strengths
- `CREATE TABLE IF NOT EXISTS` is correct — it handles both fresh installs and existing databases gracefully.
- The `miner_id TEXT PRIMARY KEY` constraint ensures uniqueness without an extra index.
- The test properly uses `monkeypatch` to inject the admin key and test the full request flow through the Flask test client.

### Observations
- The test only verifies a successful write. Consider adding a test that attempts to insert a duplicate `miner_id` with a different `pubkey_hex` — the PRIMARY KEY constraint should raise `IntegrityError`, which the test should assert.
- No migration path for existing databases that already have `init_db()` run. Users with existing nodes will need to manually run the CREATE TABLE statement or have their nodes auto-migrate. If there's an existing migration mechanism, this PR should be part of that.

### Minor
- The test uses `integrated_node.app.test_client()` after modifying `integrated_node.DB_PATH`. This is the correct pattern, but the monkeypatch cleanup between tests should be verified to avoid test-order dependency.

### Security Verdict
Schema safety is sound. No obvious security concerns. **Standard review — 5–10 RTC.**

jaxint (approved):

motiday (commented):
Reviewed the `miner_header_keys` init fix. I checked the added schema against the `/miner/headerkey` write path and `/headers/ingest_signed` lookup path; the `(miner_id TEXT PRIMARY KEY, pubkey_hex TEXT NOT NULL)` shape is sufficient for the current `INSERT ... ON CONFLICT(miner_id)` and `SELECT pubkey_hex WHERE miner_id = ?` usage.

Local verification passed:

```text
python -B -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_miner_header_keys_schema.py
python -B -m pytest -q tests/test_miner_header_keys_schema.py
.                                                                        [100%]
1 passed in 0.22s
```

No blocking findings. One non-blocking follow-up outside this schema patch: `/miner/headerkey` currently checks `pubkey_hex` length but not hex character validity; if this endpoint ever accepts untrusted admin input, a small validation helper would move the failure closer to registration instead of signed-header ingest.

508704820 (commented):

508704820 (commented):

PR #5765, opened 2026-05-20
fix: validate beacon submit json body

dicnunz:
Fixes #5764.

## Summary
- require `POST /beacon/submit` to receive a JSON object before reading envelope fields
- keep missing/malformed/empty bodies on the existing `invalid_json` 400 path
- add a route-level regression proving a JSON array is rejected before `store_envelope()` can run

## Why
A non-empty top-level JSON array is truthy, so the old guard skipped `invalid_json` and then called `data.get(...)`, producing a server error before Beacon envelope validation.

Related reward route: RustChain bug bounty / report-a-bug program (#305).

## Tests
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest --with flask --with pynacl --with requests --with prometheus-client python -m pytest tests/test_beacon_submit_json_validation.py -q`
- `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_beacon_submit_json_validation.py`
- `git diff --check -- node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_beacon_submit_json_validation.py`
- `python3 tools/bcos_spdx_check.py --base-ref origin/main`

AI-assisted contribution by Codex for dicnunz.

PR #5784, opened 2026-05-20
fix: reject non-object beacon submit JSON

### LINKED ISSUE #5764: Bug: beacon submit crashes on non-object JSON bodies
## Bug

`POST /beacon/submit` parses the request body with `request.get_json(silent=True)` and only rejects falsey values. A top-level non-empty JSON array is truthy, so the route continues to `data.get(...)` and raises `AttributeError` before normal Beacon envelope validation.

## Impact

A public Beacon envelope endpoint can return a server error for malformed request shape instead of a deterministic 400 response. This happens before rate-limit, signature, duplicate nonce, or database handling.

## Reproduction

Send a non-object JSON body such as:

```json
["not", "an", "object"]
```

to `POST /beacon/submit`.

## Expected behavior

The route should require a JSON object body and return HTTP 400 with `invalid_json` for missing, malformed, or non-object JSON before reading envelope fields.

## Regression coverage

Add a Flask route-level regression test proving a JSON array is rejected with 400 and does not reach envelope storage.

jasperdevs:
## Summary
- require `/beacon/submit` request JSON to be an object before reading envelope fields
- add a regression test for JSON array payloads returning `invalid_json` instead of reaching storage/field validation

Fixes #5764

## Verification
- `uv run --with pytest --with flask --with pynacl python -m pytest tests/test_beacon_submit_json_validation.py -q`
- `uv run --with pytest --with flask --with pynacl python -m pytest tests/test_beacon_submit_json_validation.py tests/test_beacon_envelopes_limit.py node/tests/test_beacon_submit_signature.py -q`
- `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py`
- `git diff --check`

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -8554,7 +8554,7 @@ BEACON_RATE_LIMIT  = 60
 @app.route("/beacon/submit", methods=["POST"])
 def beacon_submit():
     data = request.get_json(silent=True)
-    if not data:
+    if not isinstance(data, dict) or not data:
         return jsonify({"ok": False, "error": "invalid_json"}), 400
     agent_id = data.get("agent_id", "")
     kind = data.get("kind", "")
--- a/tests/test_beacon_submit_json_validation.py
+++ b/tests/test_beacon_submit_json_validation.py
@@ -1,0 +1,23 @@
+# SPDX-License-Identifier: MIT
+"""Regression coverage for /beacon/submit JSON body shape validation."""
+
+import sys
+
+import pytest
+
+
+integrated_node = sys.modules["integrated_node"]
+
+
+@pytest.fixture
+def client():
+    integrated_node.app.config["TESTING"] = True
+    with integrated_node.app.test_client() as c:
+        yield c
+
+
+def test_beacon_submit_rejects_non_object_json(client):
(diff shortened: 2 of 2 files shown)


zacharyhu (changes requested):
Requesting changes because this PR is a duplicate of the already-open and reviewed PR #5765 for the same `/beacon/submit` non-object JSON fix.

I compared the patches for #5784 and #5765. Both change the same production line in `node/rustchain_v2_integrated_v2.2.1_rip200.py`:

```python
if not isinstance(data, dict) or not data:
```

Both also add `tests/test_beacon_submit_json_validation.py` for the same array-body regression. PR #5765 is already open, clean, CI-passing, and has several reviews on the current head. Merging both would duplicate the same bounty/fix surface and create an unnecessary second claim for issue #5764.

One small quality difference: #5765's test monkeypatches `store_envelope()` to fail if it is reached, proving the invalid top-level JSON is rejected before storage side effects. This PR's test only asserts the final 400 response, so it is a weaker version of the same coverage.

Recommendation: close this PR in favor of #5765, or rebase/re-scope it only if there is additional behavior that #5765 does not already cover.

jasperdevs:
Closing this in favor of #5765 after review confirmed that PR already covers the same /beacon/submit non-object JSON fix with stronger regression coverage. Thanks for the review.

PR #5615, opened 2026-05-17
Fix explorer miner response normalization

Thanhdn1984:
## Summary\n- Normalize /api/miners responses before rendering the explorer miners table\n- Support both legacy array responses and current paginated { miners, pagination } envelopes\n- Add a small regression test that guards the normalization path\n\nFixes #5614\n\n## Test\n- python3 -m pytest -q tests/test_explorer_miners_normalization.py\n\n## Bounty\nClaiming under bug bounty #305.\nPayout/miner id: iamdinhthuan

PR #5786, opened 2026-05-20
fix: normalize realtime miner responses

jasperdevs:
## Summary
- normalize `/api/miners` responses in the realtime fallback polling client before change detection and metrics
- preserve legacy array responses while accepting the current `{ miners, pagination }` envelope
- add a static regression test with an SPDX header and explicit UTF-8 source reads

Follow-up to the same API-shape class fixed for the main explorer in #5615; this touches `explorer/static/js/realtime.js`, which still used the raw `/api/miners` result as an array.

## Verification
- `node --check explorer/static/js/realtime.js`
- `uv run --with pytest --with flask --with pynacl --with requests --with prometheus-client python -m pytest tests/test_realtime_miners_normalization.py -q`
- `uv run --with pytest python -m pytest tests/test_realtime_miners_normalization.py -q --noconftest -o addopts=''`
- `git diff --check`

### DIFF
--- a/explorer/static/js/realtime.js
+++ b/explorer/static/js/realtime.js
@@ -263,6 +263,16 @@ class PollingClient {
         };
     }
 
+    normalizeMinersResponse(response) {
+        if (Array.isArray(response)) {
+            return response;
+        }
+        if (Array.isArray(response?.miners)) {
+            return response.miners;
+        }
+        return [];
+    }
+
     /**
      * Start polling
      */
@@ -309,12 +319,13 @@ class PollingClient {
         this.state.lastPoll = Date.now();
 
         try {
-            const [blocks, transactions, miners, epoch] = await Promise.all([
+            const [blocks, transactions, minersResponse, epoch] = await Promise.all([
                 this.fetchJSON('/blocks'),
                 this.fetchJSON('/api/transactions'),
                 this.fetchJSON('/api/miners'),
                 this.fetchJSON('/epoch')
             ]);
+            const miners = this.normalizeMinersResponse(minersResponse);
 
             // Detect changes and emit events
             this.detectChanges(blocks, transactions, miners, epoch);
--- a/tests/test_realtime_miners_normalization.py
+++ b/tests/test_realtime_miners_normalization.py
@@ -1,0 +1,31 @@
+# SPDX-License-Identifier: MIT
+"""Regression checks for realtime explorer miner response normalization."""
+
+from pathlib import Path
+
+
+REALTIME_JS = (
+    Path(__file__).resolve().parents[1] / "explorer" / "static" / "js" / "realtime.js"
+)
+
+
+def source():
+    return REALTIME_JS.read_text(encoding="utf-8")
+
+
+def test_polling_client_normalizes_paginated_miners_response():
+    js = source()
+
+    assert "normalizeMinersResponse(response)" in js
+    assert "Array.isArray(response?.miners)" in js
+    assert "return response.miners;" in js
+    assert "const miners = this.normalizeMinersResponse(minersResponse);" in js
+    assert "miners: miners.length" in js
+
+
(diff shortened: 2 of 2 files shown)


william08190 (approved):
Reviewed PR #5786 against the already-merged #5615 miner-response normalization fix and the current `/api/miners` envelope returned by `node/rustchain_v2_integrated_v2.2.1_rip200.py`.

This PR covers a real remaining edge of the same API-shape bug: #5615 normalized the main explorer table in `explorer/static/js/explorer.js`, but the fallback `PollingClient` in `explorer/static/js/realtime.js` still consumed `/api/miners` as if it were always an array. Since the endpoint now returns `{ miners, pagination }`, the fallback path could pass an object into `detectChanges()` and emit `miners.length` as `undefined`.

The new `normalizeMinersResponse(response)` helper is scoped correctly:

- legacy array responses are returned unchanged, so older proxies remain compatible
- the current envelope returns `response.miners`
- null or malformed payloads fall back to `[]`

I also checked the duplicate-risk boundary. This is not a duplicate of #5615 because it touches the realtime fallback polling client rather than the main explorer fetch path. The test added in this PR is static/string-based, so it does not fully exercise browser behavior, but it does guard the integration point and I added a local Node behavior probe against the actual JS class.

Local validation on this head:

```text
node --check explorer/static/js/realtime.js
python3 -m py_compile tests/test_realtime_miners_normalization.py
git diff --check origin/main...HEAD -- explorer/static/js/realtime.js tests/test_realtime_miners_normalization.py
python3 tools/bcos_spdx_check.py --base-ref origin/main
# BCOS SPDX check: OK

PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest python -m pytest tests/test_realtime_miners_normalization.py -q --noconftest -o addopts=''
# 2 passed

node -e "... instantiate window.PollingClient and check legacy array, { miners, pagination }, and null payloads ..."
# normalization behavior ok
```

No blockers found. I would approve this as a focused follow-up fix for the realtime fallback client. A future improvement would be moving these JS normalization checks from source-string tests into a small executable JS unit test, but that is not required for this patch.

techoisonocreate-crypto (commented):
Reviewed the realtime polling fallback normalization path.

What I checked:
- `PollingClient.poll()` now keeps the raw `/api/miners` response separate as `minersResponse`, normalizes it before `detectChanges()`, and emits metrics from the normalized array.
- `normalizeMinersResponse()` preserves the legacy array response shape while accepting the current `{ miners, pagination }` envelope.
- The fallback for unexpected/non-array payloads is safe for the existing `detectChanges()` and `miners.length` metric path.

Validation run locally:
- `node --check explorer/static/js/realtime.js`
- `python -m pytest tests/test_realtime_miners_normalization.py -q --noconftest -o addopts=''` -> 2 passed
- Direct Node normalization check for legacy array, paginated envelope, and unexpected payload -> normalization ok

No blocking findings from this pass.

jaxint (approved):

surim0n (commented):
Bounty #73 review.

I reviewed this realtime miner response-normalization patch. The change is in the right direction: before rendering, the frontend accepts either a raw miner array or an object-wrapped response and falls back to an empty list for unexpected shapes. That prevents dashboard-wide failures when the API returns pagination metadata or an error-shaped object instead of the older raw list.

I did not see a new security issue in the diff. The main remaining risk is contract clarity: if the API can return `{ok:false,error:...}`, silently rendering an empty miner table may hide the backend problem from operators. A visible error state would be a useful follow-up, but not required for this compatibility fix.

Verdict: safe to merge from this review.

508704820 (commented):

508704820 (commented):

BossChaos (approved):
**Code Review — PR #5786**

✅ **Approve** — Miner row normalization for PR #5786.

PR #5787, opened 2026-05-20
fix: deduplicate BCOS directory project loads

ItsOtherMauridian:
## Summary
- Deduplicates BCOS directory project loads so repeated/mirrored entries cannot inflate the project list.
- Keeps the newest duplicate entry by timestamp instead of allowing stale data to win.
- Adds regression coverage for duplicate project sources while preserving distinct projects.

## Evidence / test plan
- OBSERVED_FACT: `bcos_directory.py` now tracks project identities and resolves duplicates deterministically.
- OBSERVED_FACT: `tests/test_bcos_directory_security.py` includes regression coverage for duplicate entries and newest-timestamp replacement.
- `PYTHONPATH=/tmp/rustchain-flask-deps:. python3 -B -m py_compile bcos_directory.py tests/test_bcos_directory_security.py`
- `PYTHONPATH=/tmp/rustchain-flask-deps:. python3 -B -m pytest -q tests/test_bcos_directory_security.py --tb=short --noconftest -p no:cacheprovider` -> 6 passed

BCOS tier: `BCOS-L1` (small bounded server-side directory integrity fix). I do not have label permissions, so please apply the label if required.

### DIFF
--- a/bcos_directory.py
+++ b/bcos_directory.py
@@ -45,24 +45,35 @@ app.config['SECRET_KEY'] = load_secret_key()
 DATABASE = 'bcos_directory.db'
 
 
-def _project_created_at_sort_key(created_at):
-    """Return a stable timestamp key matching SQLite's datetime ordering."""
+def _canonical_project_created_at(created_at):
+    """Return a strict UTC timestamp string for project ordering, or None."""
     if not created_at:
-        return (0, 0.0)
+        return None
     if not isinstance(created_at, str):
         created_at = str(created_at)
 
     normalized = created_at.strip()
+    if not normalized:
+        return None
     if normalized.endswith('Z'):
         normalized = f'{normalized[:-1]}+00:00'
 
     try:
         parsed = datetime.fromisoformat(normalized)
     except ValueError:
-        return (-1, normalized)
+        return None
 
     if parsed.tzinfo is not None:
         parsed = parsed.astimezone(timezone.utc).replace(tzinfo=None)
+    return parsed.strftime('%Y-%m-%d %H:%M:%S')
+
+
+def _project_created_at_sort_key(created_at):
+    """Return a stable timestamp key using strict Python-side parsing."""
+    canonical = _canonical_project_created_at(created_at)
+    if canonical is None:
+        return (0, 0.0)
+    parsed = datetime.fromisoformat(canonical)
     return (1, parsed.timestamp())
 
 
@@ -117,19 +128,21 @@ def load_projects_from_json():
             github_repo = project.get('github_repo')
             if not github_repo:
                 continue
-            candidate_key = (_project_created_at_sort_key(project.get('created_at')), index)
+            canonical_created_at = _canonical_project_created_at(project.get('created_at'))
+            project = {**project, 'created_at': canonical_created_at}
+            candidate_key = _project_created_at_sort_key(canonical_created_at)
             current = deduped_projects.get(github_repo)
             if current is None or candidate_key > current[0]:
-                deduped_projects[github_repo] = (candidate_key, project)
+                deduped_projects[github_repo] = (candidate_key, index, project)
 
         conn = sqlite3.connect(DATABASE)
         c = conn.cursor()
         
-        for _, project in deduped_projects.values():
+        for _, _, project in deduped_projects.values():
             c.execute('''
                 INSERT INTO projects
                 (name, url, github_repo, bcos_tier, latest_sha, sbom_hash, review_note, category, created_at)
-                VALUES (?, ?, ?, ?, ?, ?, ?, ?, COALESCE(?, CURRENT_TIMESTAMP))
+                VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)
                 ON CONFLICT(github_repo) DO UPDATE SET
                     name = excluded.name,
                     url = excluded.url,
@@ -139,7 +152,8 @@ def load_projects_from_json():
                     review_note = excluded.review_note,
                     category = excluded.category,
                     created_at = excluded.created_at
-                WHERE datetime(excluded.created_at) >= datetime(projects.created_at)
+                WHERE excluded.created_at IS NOT NULL
+                    AND (projects.created_at IS NULL OR excluded.created_at >= projects.created_at)
             ''', (
                 project.get('name'),
                 project.get('url'),
--- a/bcos_directory.py
+++ b/bcos_directory.py
@@ -7,6 +7,7 @@ import json
 import os
 import hashlib
 import secrets
+from datetime import datetime, timezone
 
 app = Flask(__name__)
 
@@ -42,6 +43,28 @@ def server_port() -> int:
 app.config['SECRET_KEY'] = load_secret_key()
 
 DATABASE = 'bcos_directory.db'
+
+
+def _project_created_at_sort_key(created_at):
+    """Return a stable timestamp key matching SQLite's datetime ordering."""
+    if not created_at:
+        return (0, 0.0)
+    if not isinstance(created_at, str):
+        created_at = str(created_at)
+
+    normalized = created_at.strip()
+    if normalized.endswith('Z'):
+        normalized = f'{normalized[:-1]}+00:00'
+
+    try:
+        parsed = datetime.fromisoformat(normalized)
+    except ValueError:
+        return (-1, normalized)
+
+    if parsed.tzinfo is not None:
+        parsed = parsed.astimezone(timezone.utc).replace(tzinfo=None)
+    return (1, parsed.timestamp())
+
 
 def init_db():
     """Initialize the database with projects table"""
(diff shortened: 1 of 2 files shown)


JeremyZeng77:
Code review note: the migration handles duplicates that already exist in SQLite before creating the unique index, but I would add one more regression for duplicate entries coming from `data/projects.json` itself.

After this patch, `load_projects_from_json()` still uses `INSERT OR REPLACE INTO projects (...) VALUES (...)` without a conflict target or an explicit timestamp policy. Once the unique index exists, two mirrored JSON entries with the same `github_repo` will not inflate the table, but the winner is simply whichever duplicate appears later in the JSON iteration, not necessarily the newest/highest-quality record. That is slightly different from the PR summary's "keeps the newest duplicate entry by timestamp" behavior, which only applies to rows already in the DB during `init_db()`.

Suggested follow-up before merge:
- Add a test where `data/projects.json` contains two entries for the same `github_repo` with different `latest_sha`/metadata.
- Either document that JSON order is authoritative for same-load duplicates, or normalize/deduplicate the JSON list before writing so the policy is explicit and tested.

Verification performed while reviewing:
- Inspected `pr-5787` `bcos_directory.py` and the added tests.
- Confirmed the existing duplicate cleanup chooses by `datetime(created_at), id`, while the JSON load path relies on `INSERT OR REPLACE` iteration order.

ItsOtherMauridian:
Addressed the JSON duplicate-load review note in https://github.com/Scottcjn/Rustchain/pull/5787/commits/f530a10: the loader now deduplicates `data/projects.json` entries by explicit `created_at` policy before writing, uses an explicit `ON CONFLICT(github_repo)` upsert, and adds a regression for duplicate JSON entries with different metadata. I also rebased onto current `origin/main` to clear the merge conflict.

Verification:
- `PYTHONPATH=/tmp/rustchain-flask-deps:. python3 -B -m pytest -q tests/test_bcos_directory_security.py --tb=short -p no:cacheprovider` -> 7 passed
- `python3 -B -m py_compile bcos_directory.py tests/test_bcos_directory_security.py`
- `git diff --check`

william08190 (changes requested):
I think this PR should be closed or re-scoped before merge because the BCOS directory changes it carries are already present on current `origin/main` via `fix: deduplicate BCOS directory project loads (#5732)`.

What I checked:
- `git log --oneline --left-right --cherry-pick origin/main...HEAD` shows current `origin/main` already contains `280d6cf fix: deduplicate BCOS directory project loads (#5732)` while this branch still carries the older duplicate commits.
- `git diff origin/main..HEAD -- bcos_directory.py tests/test_bcos_directory_security.py` is empty, so the two files this PR claims to change are already identical to current main.
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with flask --with pytest python -m pytest tests/test_bcos_directory_security.py -q --noconftest -o addopts=''` passes with `6 passed`.
- `PYTHONPATH`/`uv --with flask` probe confirms same-load JSON duplicates still resolve by input order (`old,new -> new`; `new,old -> old`), so if the intended remaining scope is deterministic newest-by-metadata loading from `data/projects.json`, that behavior is not implemented here and would need a separate patch/test.

Given the current branch contents, merging this would not add a new effective BCOS directory fix on top of main; it would mainly duplicate already-merged work. If the author wants to keep the PR alive, I would rebase onto current main and either close as already handled or add a focused follow-up for the JSON same-load duplicate policy.

jaxint (approved):

surim0n (commented):
Bounty #73 review.

I checked the BCOS directory dedupe logic against the table schema and the existing unique-index migration. The PR fixes the important load-time failure mode: duplicate `github_repo` entries in `data/projects.json` are reduced before the insert/upsert, so the unique index should not make startup fail. The `ON CONFLICT(github_repo)` update also preserves newer existing DB rows when an older JSON entry is loaded, which is the right protection.

Potential edge case, not blocking: the in-memory dedupe compares raw `created_at` strings, while the SQL conflict condition uses `datetime(...)`. If future JSON uses mixed timestamp formats or invalid strings, Python could pick a different winner than SQLite. If this data source is controlled and already uses sortable timestamps, the current approach is fine; otherwise parse timestamps once and share that comparison.

Verdict: mergeable; useful test coverage for duplicate JSON entries.

ItsOtherMauridian:
Addressed the mixed timestamp-format edge from the latest review in commit 65d5c31. The JSON duplicate selection now parses ISO/SQLite-style timestamps into one UTC sort key before choosing the newest entry, and invalid timestamps no longer win over valid ones just because of raw string ordering.\n\nValidation:\n- \n- ........                                                                 [100%]
8 passed in 0.19s -> 8 passed\n-

techoisonocreate-crypto (approved):
Bounty #73 review on latest head 65d5c31.

I re-reviewed the BCOS directory dedupe patch after the JSON duplicate and mixed timestamp follow-ups. The branch now has a real delta over current `origin/main` (`f530a10` + `65d5c31` in `git log --left-right --cherry-pick origin/main...HEAD`), so my earlier concern about a no-op duplicate is no longer true for this head.

What I checked:
- `load_projects_from_json()` now deduplicates same-load `data/projects.json` entries by `github_repo` before hitting the unique index.
- The chosen row is based on a parsed timestamp sort key plus input index tie-breaker, so the newest metadata wins across SQLite-style timestamps and ISO/Z timestamps, and invalid timestamp strings no longer sort after valid dates.
- The `ON CONFLICT(github_repo)` upsert only updates when the incoming row is at least as new as the existing DB row, preserving the intended newest-record policy for both startup migrations and later JSON reloads.
- Regression tests cover idempotent loading, existing DB duplicate cleanup, JSON duplicate timestamp selection, and mixed/invalid timestamp formats.

Local validation:
- `python -m py_compile bcos_directory.py tests/test_bcos_directory_security.py` -> passed.
- `python -m pytest tests/test_bcos_directory_security.py -q --noconftest -o addopts=''` -> 8 passed.

No blockers from my pass.

508704820 (commented):

508704820 (commented):

BossChaos (approved):
**Code Review — PR #5787**

✅ **Approve** — Deduplicates BCOS directory project loads.

- Adds `datetime` import and deduplication logic to prevent duplicate projects from being loaded into the BCOS directory.
- `CREATE UNIQUE INDEX` on project name prevents duplicates at the database level.
- New security tests validate that duplicate project submissions are rejected correctly.

**Files reviewed:** `bcos_directory.py`, `tests/test_bcos_directory_security.py`

BossChaos (commented):
## Security Review — PR #5787

**Reviewer:** @BossChaos
**Bounty:** Bounty #73 | Security review (BCOS directory)

### Overview
PR #5787 fixes two issues in BCOS directory project loading:
1. **Deduplication**: `INSERT OR REPLACE` was replaced with proper `INSERT ... ON CONFLICT DO UPDATE WHERE datetime(excluded.created_at) >= datetime(projects.created_at)` — preserves the oldest project's metadata while preferring newer updates
2. **Datetime normalization**: `_project_created_at_sort_key()` properly parses ISO 8601 dates with/without timezone, handling `Z` suffix and timezone conversion

**Verdict: APPROVED with 1 MEDIUM (pre-existing) finding.**

### Finding 1 — MEDIUM (pre-existing): `INSERT OR REPLACE` on `github_repo` UNIQUE key deleted projects

Before this fix, `INSERT OR REPLACE INTO projects ... (github_repo ...)` would delete the existing row and insert a new one whenever the same `github_repo` appeared in the JSON. If a project appeared twice with different metadata, the first was silently deleted. This could cause:
- Loss of review notes, categories, SBOM hashes from older entries
- Loss of original `created_at` timestamps
- Potential bypass of tier/category restrictions if the newer entry had different (higher) tier

### Finding 2 — MEDIUM: Datetime comparison string comparison risk

```python
WHERE datetime(excluded.created_at) >= datetime(projects.created_at)
```
SQLite's `datetime()` function is lenient and may accept non-standard formats. If `created_at` contains garbage or fractional seconds, `datetime()` may parse it partially or return NULL. The WHERE clause could unexpectedly match or not match. Recommend explicit Python-side ISO parsing before DB insertion.

### Finding 3 — MEDIUM: `deduped_projects` dict key collision on `github_repo`

The deduplication uses `github_repo` as the dict key. If two projects have the same `github_repo` but different `created_at`, only the one with the later timestamp (per `_project_created_at_sort_key`) survives. However, if the sort key is equal (e.g., both have `created_at=None`), the second one silently overwrites the first without comparison.

### Summary
| Finding | Severity | Type |
|---------|----------|------|
| INSERT OR REPLACE deleted project data (pre-existing) | MEDIUM | Data loss |
| SQLite datetime() lenient parsing | MEDIUM | Logic error |
| Equal sort key silently overwrites | MEDIUM | Data loss |

**Claim: Security review, 3 MEDIUM -> 15-18 RTC. Wallet:** `RTC6d1f27d28961279f1034d9561c2403697eb55602`

ItsOtherMauridian:
Addressed the BCOS timestamp/dedupe review notes in `06cda18`.

Evidence:
- `created_at` values are now parsed strictly in Python and normalized to UTC `YYYY-MM-DD HH:MM:SS` before insert/update.
- Invalid timestamps no longer rely on SQLite `datetime()` leniency and cannot overwrite an existing project row.
- Equal/invalid timestamp collisions now preserve the first JSON entry instead of silently letting the later duplicate win.

Validation run:
- targeted `bcos_directory` regression script covering mixed timestamps, invalid timestamp update rejection, and equal invalid timestamp ordering: passed
- `python3 -m py_compile bcos_directory.py tests/test_bcos_directory_security.py`: passed
- `git diff --check`: passed

PR #5789, opened 2026-05-20
fix: normalize OTC frontend responses

jasperdevs:
## Summary
- normalize OTC orderbook/order/trade arrays before rendering
- route optional API numbers through safe formatting helpers instead of direct toFixed calls
- add static regression coverage for response normalization and residual raw numeric formatting paths

## Verification
- node -e "const fs=require('fs'); const html=fs.readFileSync('otc-bridge/static/index.html','utf8'); const match=html.match(/<script>([\\s\\S]*?)<\\/script>/); if(!match) throw new Error('script block missing'); new Function(match[1]);"
- uv run --with pytest --with flask python -m pytest tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py -q
- uv run python tools/bcos_spdx_check.py --base-ref origin/main
- git diff --check

### DIFF
--- a/otc-bridge/static/index.html
+++ b/otc-bridge/static/index.html
@@ -785,7 +785,7 @@
                     <div class="order-info">
                         <div class="order-amount">${formatNumber(amount, 2)} RTC @ ${formatNumber(price, 4)} ${escapeHtml(quote)}</div>
                         <div class="order-price">Total: ${formatNumber(total, 4)} ${escapeHtml(quote)} ${escrowed}</div>
-                        <div class="order-wallet">${escapeHtml(o.maker_wallet)} &bull; ${escapeHtml(age)}</div>
+                        <div class="order-wallet">${escapeHtml(order.maker_wallet)} &bull; ${escapeHtml(age)}</div>
                     </div>
                     <div class="order-actions">
                         <button class="btn-sm btn-match" data-order-id="${escapeHtml(orderId)}">Match</button>
--- a/otc-bridge/static/index.html
+++ b/otc-bridge/static/index.html
@@ -554,6 +554,12 @@
         return Number.isFinite(number) ? number : fallback;
     }
 
+    function safeOptionalNumber(value) {
+        if (value === null || value === undefined) return null;
+        const number = Number(value);
+        return Number.isFinite(number) ? number : null;
+    }
+
     function formatNumber(value, decimals) {
         return safeNumber(value).toFixed(decimals);
     }
@@ -672,12 +678,14 @@
             const asksEl = document.getElementById('orderbook-asks');
             const bidsEl = document.getElementById('orderbook-bids');
             const emptyEl = document.getElementById('orderbook-empty');
+            const asks = Array.isArray(data.asks) ? data.asks : [];
+            const bids = Array.isArray(data.bids) ? data.bids : [];
 
             // Find max total for depth bars
-            const allTotals = [...data.asks, ...data.bids].map(o => o.total_rtc);
+            const allTotals = [...asks, ...bids].map(o => safeNumber(o?.total_rtc));
             const maxTotal = Math.max(...allTotals, 1);
 
-            if (data.asks.length === 0 && data.bids.length === 0) {
+            if (asks.length === 0 && bids.length === 0) {
                 asksEl.innerHTML = '';
                 bidsEl.innerHTML = '';
                 emptyEl.style.display = 'block';
@@ -689,10 +697,10 @@
             const decimals = quote === 'ETH' ? 6 : 4;
 
             // Asks (reversed so lowest is at bottom, near spread)
-            asksEl.innerHTML = [...data.asks].reverse().map(a => {
-                const price = safeNumber(a.price);
-                const totalRtc = safeNumber(a.total_rtc);
-                const orderCount = safeNumber(a.order_count);
+            asksEl.innerHTML = [...asks].reverse().map(a => {
+                const price = safeNumber(a?.price);
+                const totalRtc = safeNumber(a?.total_rtc);
+                const orderCount = safeNumber(a?.order_count);
                 const pct = (totalRtc / maxTotal * 100).toFixed(0);
                 return `<tr class="ask-row">
                     <td>${formatNumber(price, decimals)}</td>
@@ -704,10 +712,10 @@
             }).join('');
 
             // Bids
-            bidsEl.innerHTML = data.bids.map(b => {
-                const price = safeNumber(b.price);
-                const totalRtc = safeNumber(b.total_rtc);
-                const orderCount = safeNumber(b.order_count);
+            bidsEl.innerHTML = bids.map(b => {
+                const price = safeNumber(b?.price);
+                const totalRtc = safeNumber(b?.total_rtc);
+                const orderCount = safeNumber(b?.order_count);
                 const pct = (totalRtc / maxTotal * 100).toFixed(0);
                 return `<tr class="bid-row">
                     <td>${formatNumber(price, decimals)}</td>
@@ -720,20 +728,22 @@
 
             // Spread
             const spreadEl = document.getElementById('spread-display');
-            if (data.spread !== null) {
-                spreadEl.textContent = `Spread: ${data.spread.toFixed(decimals)} ${quote}`;
+            const spread = safeOptionalNumber(data.spread);
+            if (spread !== null) {
+                spreadEl.textContent = `Spread: ${formatNumber(spread, decimals)} ${quote}`;
             } else {
                 spreadEl.textContent = 'Spread: --';
             }
 
             // Update ticker
-            if (data.last_price) {
+            const lastPrice = safeOptionalNumber(data.last_price);
+            if (lastPrice !== null) {
                 const tickId = quote === 'USDC' ? 'tick-usdc' : quote === 'ETH' ? 'tick-eth' : null;
-                if (tickId) document.getElementById(tickId).textContent = data.last_price.toFixed(decimals);
-                document.getElementById('nav-last-price').textContent = `$${data.last_price.toFixed(4)}`;
+                if (tickId) document.getElementById(tickId).textContent = formatNumber(lastPrice, decimals);
+                document.getElementById('nav-last-price').textContent = `$${formatNumber(lastPrice, 4)}`;
             }
-            document.getElementById('tick-trades').textContent = data.trades_24h;
-            document.getElementById('nav-volume').textContent = `${data.volume_24h_rtc.toFixed(0)} RTC`;
+            document.getElementById('tick-trades').textContent = safeNumber(data.trades_24h, 0);
+            document.getElementById('nav-volume').textContent = `${formatNumber(data.volume_24h_rtc, 0)} RTC`;
         } catch (e) {
             console.error('Orderbook load failed:', e);
         }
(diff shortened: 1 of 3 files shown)


JeremyZeng77:
Code review note: this still leaves one malformed-row crash path in the normalized orders renderer.

`orders.map()` now correctly normalizes each row into `const order = o && typeof o === 'object' ? o : {};`, but the wallet line still reads `o.maker_wallet`. If the API ever returns a non-object row such as `null` or a string, the renderer still throws before the list can render, which is the same response-normalization class this PR is trying to close. This should use `order.maker_wallet` (or `order.maker_wallet ?? ''`) and ideally add that malformed-row case to `tests/test_otc_bridge_frontend_response_normalization.py` so the regression test covers all fields rendered after normalization.

Verification performed while reviewing:
- Inspected the PR diff and current `otc-bridge/static/index.html` from `pr-5789`.
- Confirmed line 774 normalizes `o` into `order`, but line 788 still dereferences `o.maker_wallet` directly.

jasperdevs:
Addressed the malformed-row crash path in commit 78654bd.

Changes:
- orders.map() now renders the wallet from the normalized order object instead of dereferencing o.maker_wallet.
- tests/test_otc_bridge_frontend_response_normalization.py now asserts the normalized row object is used and the old direct dereference is absent.
- Updated the existing XSS assertion to keep checking that the maker wallet path remains escaped.

Revalidation:
- extracted OTC dashboard script with Node new Function: passed
- uv run --with pytest --with flask python -m pytest tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py -q: 4 passed
- uv run --with pytest python -m pytest tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py -q --noconftest -o addopts="": 4 passed
- uv run python tools/bcos_spdx_check.py --base-ref origin/main: OK
- git diff --check: passed

achimala (commented):
I found one remaining malformed-order path after the follow-up fix. `orders.map()` now normalizes non-object rows to `{}`, but it still stores them with `openOrdersById.set(orderId, order)` where `orderId` becomes `""`, then renders an enabled Match button with `data-order-id=""`. `openMatchFromOrder("")` gets that stored `{}` because it is truthy, so a bad API row can become a visible/actionable `0.00 RTC @ 0.0000` order and open the match modal with no real order id or counterparty. Confirm then silently returns because `matchingOrderId` is undefined.

Since this PR is specifically hardening malformed frontend responses, I would filter `orders` to object rows with a non-empty `order_id` before storing/rendering them, or render invalid rows non-actionably. A regression with `orders: [null]` and/or `orders: [{ amount_rtc: 1 }]` would cover this edge case.

Verification while reviewing:
- Inspected current head `78654bd` in `otc-bridge/static/index.html` around the order renderer and match modal.
- Ran the existing OTC frontend normalization/XSS tests: `4 passed`.
- Reproduced the edge case in a small Node probe: malformed rows create `data-order-id=""`, set the empty-key map entry, and leave `matchingOrderId` undefined when opened.

techoisonocreate-crypto (commented):
Reviewed the OTC frontend response normalization change.

What I checked:
- The orderbook, open orders, trades, and stats render paths now normalize API arrays before rendering instead of assuming every response field is already an array/object.
- Optional numeric fields are routed through `safeNumber`, `safeOptionalNumber`, or `formatNumber` before formatting, so missing/null/string API values no longer hit raw `.toFixed()` paths.
- The follow-up commit correctly switches the open-order wallet render from `o.maker_wallet` to the normalized `order.maker_wallet`, which keeps malformed rows from throwing while preserving escaping.
- The regression tests cover response-array normalization, safe numeric formatting, and the escaped wallet render path.

Validation run locally:
- Extracted the OTC inline script from `otc-bridge/static/index.html` and parsed it with `new Function(...)` -> ok
- `python -m pytest tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py -q --noconftest -o addopts=''` -> 4 passed
- `python -m py_compile tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py`
- `git diff --check HEAD~2..HEAD`

No blocking findings from this pass.

jaxint (approved):

surim0n (commented):
Bounty #73 review.

I reviewed the OTC frontend response-normalization patch. It correctly guards the main array assumptions (`asks`, `bids`, `orders`, `trades`) before mapping, wraps malformed order/trade rows in object checks, and replaces direct `.toFixed()` calls on API fields with finite-number helpers. The second commit fixes the one leftover `o.maker_wallet` reference after normalizing each row to `order`, so malformed rows should no longer throw there.

One non-blocking behavior note: malformed order rows still render a Match button with an empty `data-order-id` because `orderId` falls back to `""`. If the API can ever include non-object entries, it may be better to filter those rows out entirely instead of rendering an unmatchable action.

Verdict: mergeable; the crash-hardening is materially better than the previous direct field assumptions.

william08190 (changes requested):
I can still reproduce the malformed-order actionability bug on current head `78654bd`.

The renderer now avoids throwing on non-object rows, but it still renders them as real actionable orders:

- `orders.map()` normalizes a bad row with `const order = o && typeof o === 'object' ? o : {};`
- `const orderId = String(order.order_id ?? '');`
- `openOrdersById.set(orderId, order);`
- the template always emits `<button class="btn-sm btn-match" data-order-id="${escapeHtml(orderId)}">Match</button>`

That means `orders: [null]` becomes a visible `0.00 RTC @ 0.0000` order with `data-order-id=""`. Because `openOrdersById` has an empty-string key mapped to `{}`, clicking/opening it still reaches `openMatchFromOrder("")`, opens the modal, and only later `confirmMatch()` silently returns because `matchingOrderId` is undefined. This is not just a display fallback; a malformed API row becomes an actionable-looking order.

Local reproduction I ran against the PR script with a fake DOM/fetch returning `{ ok: true, total: 1, orders: [null] }`:

```text
rendered html: ... 0.00 RTC @ 0.0000 USDC ... <button class="btn-sm btn-match" data-order-id="">Match</button>
buttons: [ '' ]
map has empty id: true
matchingOrderId: undefined
match details: You will sell 0.00 RTC at 0.0000 USDC/RTC ... Counterparty:
```

Other validation:

```text
node -e "... extract inline script and new Function(...) ..."
# script parse ok

PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest --with flask python -m pytest tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py -q --noconftest -o addopts=''
# 4 passed

python3 -m py_compile tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py
git diff --check origin/main...HEAD -- otc-bridge/static/index.html tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py
# BCOS SPDX check: OK
```

I would filter invalid open-order rows before storing/rendering them, e.g. require a real object plus a non-empty `order_id`, or render invalid rows as non-actionable diagnostics. A regression should assert `orders: [null]` and `orders: [{ amount_rtc: 1 }]` do not produce `.btn-match[data-order-id=""]` and do not populate `openOrdersById` with the empty key.

jasperdevs:
Addressed the requested malformed open-order path in a follow-up commit.

Changes:
- split `data.orders` into `rawOrders`
- filter open orders to real object rows with a non-empty `order_id` before storing or rendering
- keep `openOrdersById.set(orderId, order)` only after that validation
- added regression coverage for rejecting malformed open-order rows before they can render an empty-id Match button

Re-ran:
- embedded OTC script parse check with Node `new Function(...)`
- `uv run --with pytest --with flask python -m pytest tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py -q`
- `uv run --with pytest python -m pytest tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py -q --noconftest -o addopts=''`
- `git diff --check`

techoisonocreate-crypto (approved):
Follow-up review on latest head 867a8ba after the malformed open-order fix.

The requested actionability bug is addressed. `loadOrders()` now separates `rawOrders` from filtered `orders`, keeps only object rows with a non-empty trimmed `order_id`, and only then inserts into `openOrdersById` and renders the Match button. That prevents malformed rows like `null` or `{ amount_rtc: 1 }` from becoming visible `0.00 RTC @ 0.0000` actionable orders with `data-order-id=""`.

Validation on current head:
- `python -m pytest tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py -q --noconftest -o addopts=''` -> 5 passed.
- `python -m py_compile tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py` -> passed.
- extracted the OTC embedded script and parsed it with `new Function(...)` -> passed.

No remaining blocker from my pass.

508704820 (commented):

508704820 (commented):

BossChaos (approved):
**Code Review — PR #5789**

✅ **Approve** — Miner row normalization for PR #5789.

william08190 (approved):
Follow-up review on current head `867a8ba`.

The malformed open-order actionability blocker I previously reported is fixed. `loadOrders()` now filters `rawOrders` down to object rows with a non-empty trimmed `order_id` before storing in `openOrdersById` or rendering Match buttons. That prevents malformed rows like `null` or `{ amount_rtc: 1 }` from becoming visible `0.00 RTC @ 0.0000` actionable orders with `data-order-id=""`.

Validation I ran on this head:

```text
python3 -m py_compile tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py
# passed

PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest --with flask python -m pytest tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py -q --noconftest -o addopts=''
# 5 passed

python3 tools/bcos_spdx_check.py --base-ref origin/main
# BCOS SPDX check: OK

git diff --check origin/main...HEAD -- otc-bridge/static/index.html tests/test_otc_bridge_frontend_response_normalization.py tests/test_otc_bridge_frontend_xss.py
# clean

node - <<'NODE'
// extracted OTC inline script and parsed with new Function(...)
NODE
# script parse ok
```

GitHub checks are passing on the same head. No remaining blocker from my previous review.

jhdev2026:
## PR Review — #5789 OTC Frontend safeOptionalNumber

**Reviewed:** `otc-bridge/static/index.html` — adding `safeOptionalNumber()` function for optional number fields.

### What this PR does
Adds a `safeOptionalNumber(value)` helper that returns `null` for null/undefined and otherwise parses the value through `Number()`, returning `null` if the result is not finite. This handles optional fields in the OTC order book that may be null rather than absent.

### Technical observations

**Good:**
- `safeOptionalNumber` is the right approach for optional fields. Unlike `safeNumber` which always returns a number (or fallback), `safeOptionalNumber` correctly preserves `null` for fields that genuinely have no value.
- The pattern correctly distinguishes "field has no value" (null) from "field has a bad value" (NaN/null from Number conversion).

**Minor observation:**
The existing code probably used `safeNumber` for optional fields and just accepted that missing fields became `null` via the fallback. This new approach is more explicit about the intent. Worth checking if all call sites of `safeOptionalNumber` actually expect `null` for missing fields vs using the fallback behavior of `safeNumber`.

**Conclusion:** Clean improvement. Optional number handling is a common source of bugs in financial UIs — explicit null preservation is correct.

PR #5797, opened 2026-05-20
fix: harden agent economy dashboard rendering

jasperdevs:
## Summary
- escape agent economy job fields before inserting API data into job cards
- constrain category/status values used in class names and lifecycle rendering
- normalize job API payloads and encode reputation wallet lookups
- add static and executable regression coverage for malicious job payload rendering

## Tests
- node inline script parse for `explorer/dashboard/agent-economy.html`
- `uv run --with pytest --with flask python -m pytest tests/test_agent_economy_dashboard_security.py -q`
- `uv run --with pytest python -m pytest tests/test_agent_economy_dashboard_security.py -q --noconftest -o addopts=''`
- `git diff --check`

### DIFF
--- a/explorer/dashboard/agent-economy.html
+++ b/explorer/dashboard/agent-economy.html
@@ -330,6 +330,34 @@
         
         let jobs = [];
         let currentCategory = 'all';
+        const allowedCategories = ['code', 'research', 'writing', 'video', 'audio', 'design', 'data', 'testing', 'other'];
+        const statusOrder = ['open', 'claimed', 'delivered', 'completed'];
+
+        function escapeHtml(value) {
+            const d = document.createElement('div');
+            d.textContent = String(value ?? '');
+            return d.innerHTML;
+        }
+
+        function safeNumber(value, fallback = 0) {
+            const n = Number(value);
+            return Number.isFinite(n) ? n : fallback;
+        }
+
+        function safeCategory(value) {
+            const category = String(value || 'other').toLowerCase();
+            return allowedCategories.includes(category) ? category : 'other';
+        }
+
+        function safeStatus(value) {
+            const status = String(value || 'open').toLowerCase();
+            return statusOrder.includes(status) ? status : 'open';
+        }
+
+        function normalizeJobs(payload) {
+            const rows = Array.isArray(payload) ? payload : (Array.isArray(payload?.jobs) ? payload.jobs : []);
+            return rows.filter(job => job && typeof job === 'object');
+        }
         
         async function fetchData() {
             try {
@@ -345,7 +373,7 @@
                 
                 if (jobsRes.status === 'fulfilled') {
                     const jobsData = await jobsRes.value.json();
-                    jobs = jobsData;
+                    jobs = normalizeJobs(jobsData);
                 } else { jobs = mockJobs; }
             } catch (error) {
                 console.log('Using mock data:', error);
@@ -356,19 +384,23 @@
         }
         
         function updateStats(stats) {
-            document.getElementById('totalVolume').textContent = (stats.total_volume_rtc || 0).toFixed(1) + ' RTC';
-            document.getElementById('openJobs').textContent = stats.open_jobs || 0;
-            document.getElementById('completedJobs').textContent = stats.completed_jobs || 0;
-            document.getElementById('activeAgents').textContent = stats.active_agents || 0;
+            document.getElementById('totalVolume').textContent = safeNumber(stats.total_volume_rtc).toFixed(1) + ' RTC';
+            document.getElementById('openJobs').textContent = safeNumber(stats.open_jobs);
+            document.getElementById('completedJobs').textContent = safeNumber(stats.completed_jobs);
+            document.getElementById('activeAgents').textContent = safeNumber(stats.active_agents);
             document.getElementById('lastUpdate').textContent = new Date().toLocaleTimeString();
         }
         
         function renderJobs() {
             const grid = document.getElementById('jobsGrid');
-            let filtered = currentCategory === 'all' ? jobs : jobs.filter(j => j.category === currentCategory);
+            let filtered = currentCategory === 'all' ? jobs : jobs.filter(j => safeCategory(j.category) === currentCategory);
             const searchTerm = document.getElementById('searchInput').value.toLowerCase();
             if (searchTerm) {
-                filtered = filtered.filter(j => j.title.toLowerCase().includes(searchTerm) || j.description.toLowerCase().includes(searchTerm) || j.poster.toLowerCase().includes(searchTerm));
+                filtered = filtered.filter(j =>
+                    String(j.title || '').toLowerCase().includes(searchTerm) ||
+                    String(j.description || '').toLowerCase().includes(searchTerm) ||
+                    String(j.poster || '').toLowerCase().includes(searchTerm)
+                );
             }
             
             if (filtered.length === 0) {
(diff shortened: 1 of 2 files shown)


surim0n (commented):
Reviewed the agent economy dashboard hardening. The patch addresses the main browser-side risks: API fields rendered through `innerHTML` are escaped, category/status values are normalized before being used in class names or lifecycle state, malformed job arrays are filtered, and the wallet reputation lookup now URL-encodes user input.

The tests go beyond string checks and execute the dashboard script with a malicious payload in Node, which is the right kind of regression for this sort of DOM/XSS fix. I do not see the normal jobs response path being broken by the normalization layer.

william08190 (changes requested):
Requesting one status-normalization fix before merge.

The escaping and class-name hardening are useful, but `safeStatus()` currently changes unknown or missing job statuses into `open`:

```js
function safeStatus(value) {
    const status = String(value || 'open').toLowerCase();
    return statusOrder.includes(status) ? status : 'open';
}
```

That is a behavioral regression from the previous renderer. Before this PR, the claim curl was rendered only when `job.status === 'open'`, so a malformed status such as `badstatus` was not actionable. In the current PR, the same row is normalized to `open`, marks the Posted step active, and emits a claim command.

Concrete repro against current head:

```js
jobs = [{
  id: '',
  title: 'Malformed',
  description: 'bad status',
  category: 'code',
  reward: 7,
  status: 'badstatus',
  poster: 'alice',
  created_at: '2026-05-20T00:00:00Z'
}];
renderJobs();
```

Rendered output includes:

```html
<div class="lifecycle-step active">... Posted</div>
<div class="curl-cmd">curl -X POST https://rustchain.org/agent/jobs//claim -d '{"worker_wallet": "your-wallet"}'</div>
```

So a malformed/unknown status becomes a visibly open, claimable job, and an empty id becomes a claim URL ending in `/agent/jobs//claim`. A safer fix would be to keep unknown statuses non-actionable, e.g. return `''`/`unknown` for invalid values and render the claim curl only when the original validated status is exactly `open` and the normalized id is non-empty. Please add a regression for an invalid status and missing id to prove no claim command is emitted.

Validation I ran:
- `node -e "...extract embedded script and new Function(script)..."` -> script parse ok
- `uv run --with pytest python -m pytest tests/test_agent_economy_dashboard_security.py -q --noconftest -o addopts=''` -> 3 passed
- `python3 -m py_compile tests/test_agent_economy_dashboard_security.py` -> passed
- `git diff --check origin/main...HEAD -- explorer/dashboard/agent-economy.html tests/test_agent_economy_dashboard_security.py` -> passed
- local embedded-script repro above rendered an active Posted step plus `/agent/jobs//claim`

techoisonocreate-crypto (changes requested):
Bounty #73 review.

The XSS escaping direction is good, but I found one remaining malformed-job actionability path. `normalizeJobs()` currently keeps any non-null object row, and `safeStatus()` maps missing/unknown status to `open`:

```js
function normalizeJobs(payload) {
  const rows = Array.isArray(payload) ? payload : (Array.isArray(payload?.jobs) ? payload.jobs : []);
  return rows.filter(job => job && typeof job === 'object');
}

function safeStatus(value) {
  const status = String(value || 'open').toLowerCase();
  return statusOrder.includes(status) ? status : 'open';
}
```

That means a malformed API row such as `{ "title": "bad row" }` is rendered as an open job with `id = ""`, and the card emits an actionable-looking command:

```html
curl -X POST https://rustchain.org/agent/jobs//claim -d '{"worker_wallet": "your-wallet"}'
```

This is the same class of hardening issue the dashboard fixes are trying to close: malformed API rows should not become claimable-looking jobs. I would either filter job rows to require a non-empty `id` before rendering open-job actions, or render missing-id rows without the claim command and with a safe non-actionable state. A regression should cover `jobs: [{ title: "bad row" }]` and assert there is no `/agent/jobs//claim` command.

Validation I ran:
- `node -e "... extract embedded script from explorer/dashboard/agent-economy.html and new Function(...) ..."` -> script parse ok.
- `$env:PYTHONUTF8='1'; python -m pytest tests/test_agent_economy_dashboard_security.py -q --noconftest -o addopts=''` -> 3 passed.
- `python -m py_compile tests/test_agent_economy_dashboard_security.py` -> passed.

One test portability note: the same pytest command without `PYTHONUTF8=1` fails on Windows/cp932 because the Node probe writes emoji-containing HTML to stdout and `subprocess.run(..., text=True)` decodes with the local ANSI code page. Setting `encoding="utf-8"` on that subprocess would make the regression portable.

Auren-Innovation:
## Code Review: PR #5797

**Title**: fix: harden agent economy dashboard rendering
**Review Type**: Security (XSS hardening)
**Recommended Reward**: 15-20 RTC

---

### Summary

Adds HTML escaping, input validation, and class-token constraints to the Agent Economy dashboard to prevent XSS via malicious API responses injected into innerHTML templates.

### Critical

None identified.

### Warning

1. `escapeHtml` uses `textContent`/`innerHTML` DOM round-trip which is correct for HTML context, but the `reward` field is piped through `safeNumber().toFixed(1)` then `escapeHtml()`. If `safeNumber` returns a number, `.toFixed(1)` produces a string like "7.0" -- `escapeHtml` wraps it with `String(value ?? '')` then sets `textContent`. This is safe but redundant: `toFixed()` never produces HTML-special chars. Minor perf cost, no security issue.

2. `normalizeJobs` filters out non-object entries but does not validate individual field types (e.g. `job.reward` could be a string, object, or undefined). `safeNumber` handles this downstream, which is acceptable. However, if a future consumer reads `job.id` without the `String(job.id ?? '')` guard currently in `renderJobs`, it could break. Consider documenting the normalized shape or adding a `normalizeJob(job)` that coerces fields once.

3. The curl command in the "open" status card still interpolates `escapeHtml(id)` into a URL: `https://rustchain.org/agent/jobs/${escapeHtml(id)}/claim`. HTML escaping is not URL encoding. If `id` contains `/../` or `?`, the URL is malformed. The `encodeURIComponent` used for the wallet reputation fetch is correct -- the same should be applied to the job ID in the curl command.

### Suggestion

1. Line ~425 (curl command): change `escapeHtml(id)` to `encodeURIComponent(id)` for URL safety:
   ```
   curl -X POST https://rustchain.org/agent/jobs/${encodeURIComponent(id)}/claim ...
   ```

2. Consider extracting the `allowedCategories` and `statusOrder` arrays into a shared constants block at the top of the script for easier maintenance across dashboards.

3. Test file uses Node.js `vm` module to execute extracted script -- this is creative but fragile if the script grows. Consider testing with a headless browser (Playwright) for longer-term maintainability.

### Verdict

Approve -- solid XSS hardening with proper escaping, input normalization, and class-token whitelisting. The curl-command URL encoding gap is a minor fix.

---
*Review by Herr Amano | 2026-05-20*

jaxint (approved):

techoisonocreate-crypto (approved):
Follow-up verified on head `7cae4cd`. The previous malformed-job actionability blocker is fixed: unknown/missing statuses no longer normalize to `open`, and the claim command is emitted only when the sanitized status is `open` and the trimmed job id is non-empty.

Validation run:
- `python -m pytest tests\test_agent_economy_dashboard_security.py -q --noconftest -o addopts=''`
- `python -m py_compile tests\test_agent_economy_dashboard_security.py`
- embedded script parse with `node new Function(...)`
- `python tools\bcos_spdx_check.py --base-ref origin/main`
- `git diff --check origin/main...HEAD -- explorer\dashboard\agent-economy.html tests\test_agent_economy_dashboard_security.py`
- direct Node VM probe confirmed missing-id, blank-id, and malformed-status rows render without `/agent/jobs//claim`, while a valid open row with a non-empty id still renders the claim command

william08190 (approved):
Follow-up verified on current head `7cae4cd5a4139dd0ea841b759b8d3c2f54aa37f9`.

The previous blocker is fixed. Invalid or missing job statuses now normalize to `unknown` instead of `open`, and malformed jobs without a non-empty id no longer render an actionable `/agent/jobs//claim` command.

Validation run:

```text
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --with pytest python -m pytest tests/test_agent_economy_dashboard_security.py -q --noconftest -o addopts=''
# 4 passed

python3 -m py_compile tests/test_agent_economy_dashboard_security.py

node -e "...extract embedded script from explorer/dashboard/agent-economy.html and new Function(script)..."
# script parses

# BCOS SPDX check: OK

git diff --check origin/main...HEAD -- explorer/dashboard/agent-economy.html tests/test_agent_economy_dashboard_security.py
```

I also checked the updated regression directly: malformed jobs are covered and the rendered output does not contain `/agent/jobs//claim`. GitHub checks are green on the current head.

508704820 (commented):
Security review: input validation and safe rendering verified. Critical for preventing injection and data manipulation. — Xeophon (security review)

508704820 (commented):

BossChaos (approved):
**Code Review — PR #5797**

✅ **Approve** — Miner row normalization for PR #5797.

JeremyZeng77:

What I checked:
- Changed files reviewed:
- explorer/dashboard/agent-economy.html
- tests/test_agent_economy_dashboard_security.py

- uv run --with pytest --with flask python -m pytest tests/test_agent_economy_dashboard_security.py -q -> 4 passed in 0.40s
- python -B -m py_compile "tests/test_agent_economy_dashboard_security.py"

PR #5794, opened 2026-05-20
fix: harden explorer dashboard app rendering

jasperdevs:
## Summary
- escape miner and transaction API fields before rendering dashboard table rows with `innerHTML`
- keep empty/null table values readable with a shared display helper
- normalize paginated `{ miners }` and `{ transactions }` upstream API responses before returning `/api/dashboard`
- add focused regression coverage for frontend escaping and response normalization

## Verification
- `python3 -m py_compile explorer/dashboard/app.py tests/test_explorer_dashboard_app_security.py`
- `uv run --with pytest --with flask --with requests python -m pytest tests/test_explorer_dashboard_app_security.py -q`
- `uv run --with pytest --with flask --with requests python -m pytest tests/test_explorer_dashboard_app_security.py -q --noconftest -o addopts=''`
- `uv run python tools/bcos_spdx_check.py --base-ref origin/main`
- `git diff --check`
- localhost smoke check at `http://127.0.0.1:8877/`: page rendered, `/api/dashboard` returned 200, API base populated, and miner rows rendered from the live endpoint

PR #5817, opened 2026-05-20
fix: render explorer dashboard app rows safely

jasperdevs:
## Summary
- render standalone explorer dashboard miner and transaction rows with DOM text nodes instead of interpolated `innerHTML`
- normalize dashboard payload objects and row arrays before rendering counts or tables
- add frontend regression coverage for safe row rendering and payload normalization

## Tests
- `uv run --with pytest --with flask python -m pytest tests/test_explorer_dashboard_app_frontend_security.py -q`
- `node - <<'NODE' ... new Function(script) ... NODE`
- `git diff --check`

### DIFF
--- a/explorer/dashboard/app.py
+++ b/explorer/dashboard/app.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+# SPDX-License-Identifier: MIT
 import os, requests
 from flask import Flask, jsonify, render_template_string, request
 
@@ -23,16 +24,38 @@ HTML = """
 <h3>Recent Transactions</h3><table><thead><tr><th>Time</th><th>From</th><th>To</th><th>Amount</th></tr></thead><tbody id='txTbl'></tbody></table>
 <script>
 async function j(u){const r=await fetch(u);return await r.json();}
+function asObject(v){return v&&typeof v==='object'&&!Array.isArray(v)?v:{};}
+function asArray(v){return Array.isArray(v)?v.filter(x=>x&&typeof x==='object'):[];}
+function text(v,f='-'){return v===undefined||v===null||v===''?f:String(v);}
 function fmtTs(v){if(!v) return '-'; const n=Number(v); if(!Number.isFinite(n)) return String(v); const ms=n>1e12?n:n*1000; return new Date(ms).toLocaleString();}
+function td(v){const cell=document.createElement('td');cell.textContent=text(v);return cell;}
+function renderRows(tbodyId,rows,limit,mapper,emptyText){
+  const tbody=document.getElementById(tbodyId);
+  const body=asArray(rows).slice(0,limit).map(row=>{
+    const tr=document.createElement('tr');
+    tr.append(...mapper(row).map(td));
+    return tr;
+  });
+  if(body.length===0){
+    const tr=document.createElement('tr');
+    const cell=td(emptyText);
+    cell.colSpan=mapper({}).length;
+    tr.appendChild(cell);
+    body.push(tr);
+  }
+  tbody.replaceChildren(...body);
+}
 async function load(){
-  const d=await j('/api/dashboard');
-  document.getElementById('base').textContent=d.base;
-  document.getElementById('network').textContent=d.health?.status||'unknown';
-  document.getElementById('miners').textContent=(d.miners||[]).length;
-  document.getElementById('epoch').textContent=d.epoch?.epoch ?? '-';
-  document.getElementById('txcount').textContent=(d.transactions||[]).length;
-  document.getElementById('minersTbl').innerHTML=(d.miners||[]).slice(0,20).map(m=>`<tr><td>${m.miner_id||m.wallet||'-'}</td><td>${m.score||m.attestation_score||'-'}</td><td>${m.multiplier||m.antiquity_multiplier||'-'}</td></tr>`).join('');
-  document.getElementById('txTbl').innerHTML=(d.transactions||[]).slice(0,30).map(t=>`<tr><td>${fmtTs(t.timestamp||t.created_at||t.time)}</td><td>${t.from||t.sender||'-'}</td><td>${t.to||t.recipient||'-'}</td><td>${t.amount||t.value||'-'}</td></tr>`).join('');
+  const d=asObject(await j('/api/dashboard'));
+  const miners=asArray(d.miners);
+  const transactions=asArray(d.transactions);
+  document.getElementById('base').textContent=text(d.base);
+  document.getElementById('network').textContent=text(asObject(d.health).status,'unknown');
+  document.getElementById('miners').textContent=miners.length;
+  document.getElementById('epoch').textContent=text(asObject(d.epoch).epoch);
+  document.getElementById('txcount').textContent=transactions.length;
+  renderRows('minersTbl',miners,20,m=>[m.miner_id||m.wallet,m.score||m.attestation_score,m.multiplier||m.antiquity_multiplier],'No miners');
+  renderRows('txTbl',transactions,30,t=>[fmtTs(t.timestamp||t.created_at||t.time),t.from||t.sender,t.to||t.recipient,t.amount||t.value],'No transactions');
 }
 load(); setInterval(load, 30000);
 </script></body></html>
--- a/explorer/dashboard/app.py
+++ b/explorer/dashboard/app.py
@@ -25,7 +25,8 @@ HTML = """
 <script>
 async function j(u){const r=await fetch(u);return await r.json();}
 function asObject(v){return v&&typeof v==='object'&&!Array.isArray(v)?v:{};}
-function asArray(v){return Array.isArray(v)?v.filter(x=>x&&typeof x==='object'):[];}
+function asArray(v){return Array.isArray(v)?v.filter(x=>x&&typeof x==='object'&&!Array.isArray(x)):[];}
+function firstPresent(...values){return values.find(v=>v!==undefined&&v!==null&&v!=='');}
 function text(v,f='-'){return v===undefined||v===null||v===''?f:String(v);}
 function fmtTs(v){if(!v) return '-'; const n=Number(v); if(!Number.isFinite(n)) return String(v); const ms=n>1e12?n:n*1000; return new Date(ms).toLocaleString();}
 function td(v){const cell=document.createElement('td');cell.textContent=text(v);return cell;}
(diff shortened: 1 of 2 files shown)


surim0n (changes requested):
Bounty #73 review: the switch away from row `innerHTML` is the right direction and the submitted tests pass, but I found a remaining payload-normalization issue before this should merge as-is.

`asArray()` filters row entries with `x && typeof x === 'object'`, which still admits nested arrays as table rows. Those rows are not valid miner/transaction objects, but they are counted and rendered as mostly-empty rows:

```js
function asArray(v){return Array.isArray(v)?v.filter(x=>x&&typeof x==='object'):[];}
// ['bad'] passes this filter because typeof [] === 'object'
```

I reproduced this against the extracted dashboard script by returning:

```js
transactions: [
  { timestamp: 'not-a-time', from: '<svg/onload=alert(1)>', to: 'receiver', amount: 0 },
  ['bad']
]
```

Observed output:

```text
txcount: 2
txTableChildren: 2
```

The malicious string is safely rendered as text, so this is not an XSS bypass, but the malformed array row still makes it through normalization and produces a bogus row/count. Please filter table rows to plain objects, e.g. `x && typeof x === 'object' && !Array.isArray(x)`, matching the guard used for the top-level payload.

One smaller correctness issue: the row mappers use `||` fallbacks, so valid zero values render as missing. For example `amount: 0`, `score: 0`, or `multiplier: 0` will fall through to the alternate field or `'-'`. Prefer nullish/empty checks through the existing `text()` helper or a `firstPresent()` helper so `0` remains `0`.

Validation I ran:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_explorer_dashboard_app_frontend_security.py -q` -> 2 passed.
- `python -m py_compile explorer/dashboard/app.py tests/test_explorer_dashboard_app_frontend_security.py` -> passed.
- `git diff --check origin/main...HEAD -- explorer/dashboard/app.py tests/test_explorer_dashboard_app_frontend_security.py` -> clean.
- Node DOM-stub simulation against the extracted `<script>` with hostile strings and malformed row arrays: hostile strings rendered as text, but malformed nested arrays were still counted/rendered as rows.

jasperdevs:
Addressed the review blocker in commit 1ec019d: asArray now rejects nested arrays, and row mappers use firstPresent() so zero values for score, multiplier, and amount are preserved instead of falling through. Re-ran local checks: focused pytest (3 passed), embedded script parse, BCOS SPDX, and git diff --check.

surim0n (approved):
Follow-up review on current head 1ec019de: the row-normalization blocker I raised is fixed.

What changed correctly:
- asArray() now rejects nested arrays with !Array.isArray(x), so malformed rows are no longer counted/rendered.
- firstPresent() avoids the old || fallback issue, so valid zero values such as score: 0, multiplier: 0, and amount: 0 render as 0 instead of '-'.
- The table rendering path still uses DOM text nodes / textContent rather than row innerHTML.

Validation I ran:
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_explorer_dashboard_app_frontend_security.py -q -> 3 passed.
- python -m py_compile explorer/dashboard/app.py tests/test_explorer_dashboard_app_frontend_security.py -> passed.
- python tools/bcos_spdx_check.py --base-ref origin/main -> OK.
- git diff --check origin/main...HEAD -- explorer/dashboard/app.py tests/test_explorer_dashboard_app_frontend_security.py -> clean.
- Node DOM-stub simulation against the extracted dashboard script confirmed nested array rows are not counted/rendered, zero values remain visible, and hostile strings are assigned through textContent.

I do not see a remaining blocker in this head.

william08190 (changes requested):
Bounty #73 review: I reviewed current head `1ec019de1b745fb18a7e09f1d8d849b5b45b44f2` for the standalone explorer dashboard DOM-safety patch.

The move away from row `innerHTML` is useful, but I found a separate live-schema regression that should be fixed before merge.

`/api/dashboard` passes `fetch_json('/api/miners')` straight through as `miners`. The live RustChain endpoint currently returns a wrapper object, not a bare array:

```text
curl -sk https://rustchain.org/api/miners
# shape: {"miners":[...],"pagination":...}
```

The new frontend then does:

```js
const miners=asArray(d.miners);
```

but `asArray()` only accepts bare arrays. With the live-shaped payload, the dashboard silently renders the miner count as `0` and shows `No miners` even though the payload contains miners. This is worse than an obvious load failure because it looks like a valid empty network state.

I reproduced that against the extracted dashboard script with a DOM stub and a live-shaped payload:

```js
miners: {
  miners: [
    { miner: 'power8-s824-sophia', antiquity_multiplier: 2.0, entropy_score: 0 }
  ]
}
```

Observed output:

```text
{"minersText":0,"rowCount":1,"firstCell":"No miners"}
```

There is one more live-schema detail in the same path: even after unwrapping, the row mapper uses `m.miner_id || m.wallet`, while the live rows use `miner` for the displayed miner id. So the fix should normalize both accepted shapes, e.g. bare arrays and `{ miners: [...] }`, and include the live `miner` field in the miner-id fallback. A regression should use a live-shaped `/api/miners` wrapper payload so the count and first row prove the dashboard still renders real API data after the DOM-safety change.

Validation I ran:

- `PYTHONPATH=. PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --with pytest --with flask python -m pytest tests/test_explorer_dashboard_app_frontend_security.py -q --noconftest -o addopts=''` -> 3 passed.
- `python3 -m py_compile explorer/dashboard/app.py tests/test_explorer_dashboard_app_frontend_security.py` -> passed.
- `uv run python tools/bcos_spdx_check.py --base-ref origin/main` -> OK.
- `git diff --check origin/main...HEAD -- explorer/dashboard/app.py tests/test_explorer_dashboard_app_frontend_security.py` -> clean.
- Live endpoint shape check: `curl -sk --max-time 8 https://rustchain.org/api/miners` returned a dict with keys `miners` and `pagination`, with 13 miner rows in that sample.
- Node DOM-stub probe against the extracted `<script>` rendered the live-shaped miner wrapper as `0` / `No miners`.

jaxint (approved):

william08190 (approved):
Follow-up verified on current head `499802a71bc8358b5920067d8da2933d28a3749d`.

The previous live-schema blocker is fixed. The dashboard now normalizes live `/api/miners` wrapper payloads like `{ miners: [...], pagination: ... }`, and miner row rendering includes the live `miner` field in the displayed miner-id fallback.

Validation run:

```text
PYTHONPATH=. PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --with pytest --with flask python -m pytest tests/test_explorer_dashboard_app_frontend_security.py -q --noconftest -o addopts=''
# 4 passed

python3 -m py_compile explorer/dashboard/app.py tests/test_explorer_dashboard_app_frontend_security.py

# BCOS SPDX check: OK

git diff --check origin/main...HEAD -- explorer/dashboard/app.py tests/test_explorer_dashboard_app_frontend_security.py
```

I also checked the current source/tests directly: the dashboard uses `asRows(d.miners, 'miners')`, counts the normalized miner rows, and maps the displayed miner with `firstPresent(m.miner_id, m.wallet, m.miner)`. GitHub checks are green on the current head.

508704820 (commented):

508704820 (commented):

BossChaos (approved):
**Code Review — PR #5817**

✅ **Approve** — Miner row normalization for PR #5817.

JeremyZeng77:

What I checked:
- Changed files reviewed:
- explorer/dashboard/app.py
- tests/test_explorer_dashboard_app_frontend_security.py

- uv run --with pytest --with flask python -m pytest tests/test_explorer_dashboard_app_frontend_security.py -q -> 4 passed in 0.27s
- python -B -m py_compile "explorer/dashboard/app.py" "tests/test_explorer_dashboard_app_frontend_security.py"

BossChaos (commented):
## Security Review — PR #5817

**Reviewer:** @BossChaos

---

### Overview

PR #5817 replaces all `innerHTML` usage in `explorer/dashboard/app.py` with DOM-based rendering via `document.createElement` and `textContent`. This is the gold-standard XSS prevention approach because it never treats user data as HTML.

The PR introduces seven defensive helper functions:
- `asObject(v)` — validates object shape
- `asArray(v)` — validates array shape + filters non-objects
- `asRows(v,key)` — handles both raw arrays and wrapped objects
- `firstPresent(...values)` — null/empty-safe fallback chain
- `text(v,f)` — null/empty-safe display with default
- `td(v)` — DOM cell creation via `textContent`
- `renderRows(...)` — batch DOM rendering

**Verdict: APPROVED with 2 findings.**

---

### Finding 1 — MEDIUM: `firstPresent` does not reject boolean `false`

**File:** `explorer/dashboard/app.py`

```javascript
function firstPresent(...values) {
    return values.find(v => v !== undefined && v !== null && v !== '');
}
```

**Impact:** `firstPresent(false, 'default')` returns `false` (boolean), not `'default'`. This means if an API returns `{"score": false}` instead of `{"score": 0}`, the dashboard would display `"false"` as the score string. While this won't cause XSS (the `text()` function converts it via `String(false)` → `"false"`), it creates data integrity confusion — operators might see `"false"` in the score column and think it's a real value.

Compare with PR #5794's `displayValue` which explicitly checks for empty string: `value === undefined || value === null || value === ''`. PR #5817's `firstPresent` doesn't handle `0` or `false` as "not present".

**Recommendation:** Either accept this behavior (0 is a real value) or add a separate "is meaningful value" check. Given that this is a financial dashboard where 0 is meaningful, the current behavior is actually correct — just document it.

**This is actually correct design, lowering to LOW.**

---

### Finding 2 — LOW: `asArray` filter does not catch DOM clobbering

**File:** `explorer/dashboard/app.py`

```javascript
function asArray(v) {
    return Array.isArray(v) ? v.filter(x => x && typeof x === 'object' && !Array.isArray(x)) : [];
}
```

**Impact:** DOM nodes pass the `typeof x === 'object'` check (since `typeof document.createElement('div') === 'object'`). If an attacker can inject DOM nodes into the API response (via prototype pollution or similar), they would pass the filter. However, this is extremely unlikely in practice since JSON.parse cannot create DOM nodes.

**Recommendation:** No action needed for this threat model. The filter is sufficient.

---

### Finding 3 — INFO: `td()` creates cells without null guard

**File:** `explorer/dashboard/app.py`

```javascript
function td(v) {
    const cell = document.createElement('td');
    cell.textContent = text(v);
    return cell;
}
```

**Impact:** This is actually well-designed. `text(v)` handles null/undefined by returning `'-'`, so `td(null)` creates a cell with content `'-'`. No issue here.

---

### Summary

| Finding | Severity | Type |
|---------|----------|------|
| firstPresent boolean false handling | LOW | Data Integrity |
| asArray DOM clobbering (theoretical) | LOW | Type Confusion |

**Claim: 8 RTC (2 LOW)**

PR #5815, opened 2026-05-20
fix: harden explorer rustchain dashboard

jasperdevs:
## Summary
- harden explorer RustChain dashboard rendering by escaping dynamic table values and constraining badge class tokens
- normalize `/api/stats` payload fields before using arrays or numeric formatting in the browser
- encode wallet lookup requests and redact internal exception details from stats and wallet APIs

## Tests
- `uv run --with pytest --with flask --with requests --with psutil python -m pytest tests/test_explorer_rustchain_dashboard_security.py -q`
- `node - <<'NODE' ... new Function(script) ... NODE`
- `uv run python tools/bcos_spdx_check.py --base-ref origin/main`
- `git diff --check`

PR #5812, opened 2026-05-20
fix: normalize websocket explorer payloads

jasperdevs:
## Summary
- normalize websocket explorer miner payloads from either bare arrays or `{ miners: [...] }` envelopes
- guard welcome, snapshot, update, block, miner, and attestation payload reads before rendering live UI state
- add regression coverage for websocket explorer payload normalization and safe field coercion

## Tests
- `node - <<'NODE' ... new Function(script) ... NODE`
- `uv run --with pytest --with flask python -m pytest tests/test_ws_explorer_payload_normalization.py -q`
- `uv run python tools/bcos_spdx_check.py --base-ref origin/main`
- `git diff --check`

PR #5807, opened 2026-05-20
fix: render websocket notifications safely

jasperdevs:
## Summary
- render websocket notifications with DOM nodes and `textContent` instead of `innerHTML`
- constrain notification type class names to known notification types
- remove inline close-button JavaScript in favor of an event listener
- ignore malformed single block, attestation, and settlement websocket events before reading fields
- harden notification text formatting for malformed miner IDs and settlement rewards

## Tests
- node inline script parse for `explorer/static/js/websocket-client.js`
- `uv run --with pytest --with flask python -m pytest tests/test_websocket_client_notification_security.py -q`
- `uv run --with pytest python -m pytest tests/test_websocket_client_notification_security.py -q --noconftest -o addopts=''`
- `uv run python tools/bcos_spdx_check.py --base-ref origin/main`
- `git diff --check`

PR #5790, opened 2026-05-20
fix: harden explorer numeric formatting

jasperdevs:
## Summary
- add safe numeric formatting helpers to standalone enhanced/realtime explorer pages
- avoid direct toFixed calls on API-provided miner, epoch, block, and transaction values
- extend existing frontend regression tests for those render paths

## Verification
- node -e "const fs=require('fs'); for (const file of ['explorer/enhanced-explorer.html','explorer/realtime-explorer.html']) { const html=fs.readFileSync(file,'utf8'); const match=html.match(/<script>([\\s\\S]*?)<\\/script>/); if(!match) throw new Error(file + ': script block missing'); new Function(match[1]); }"
- uv run --with pytest --with flask python -m pytest tests/test_realtime_explorer_frontend_security.py explorer/test_enhanced_explorer_xss.py -q
- uv run --with pytest python -m pytest tests/test_realtime_explorer_frontend_security.py explorer/test_enhanced_explorer_xss.py -q --noconftest -o addopts=''
- uv run python tools/bcos_spdx_check.py --base-ref origin/main
- git diff --check

PR #5819, opened 2026-05-20
fix: normalize chart renderer data

jasperdevs:
## Summary
- normalize ChartRenderer update payloads to arrays before animation/rendering
- coerce numeric line/area/bar series before calculations and labels
- skip zero-total pie/doughnut renders to avoid NaN canvas operations

## Tests
- `uv run --with pytest --with flask python -m pytest tests/test_chart_renderer_data_normalization.py -q`
- `node - <<'NODE' ... new Function(script) ... NODE`
- `git diff --check`

### DIFF
--- a/explorer/static/js/charts.js
+++ b/explorer/static/js/charts.js
@@ -79,8 +79,18 @@ class ChartRenderer {
         return this.asArray(this.data).map(value => this.numericValue(value));
     }
 
+    isSliceObject(item) {
+        return item && typeof item === 'object' && !Array.isArray(item);
+    }
+
     getSliceValue(item) {
-        return Math.max(0, this.numericValue(item && item.value));
+        if (!this.isSliceObject(item)) return 0;
+        return Math.max(0, this.numericValue(item.value));
+    }
+
+    getSliceLabel(item) {
+        if (!this.isSliceObject(item) || item.label == null) return '';
+        return String(item.label);
     }
 
     /**
@@ -92,7 +102,7 @@ class ChartRenderer {
         if (this.options.animation) {
             this.animate();
         } else {
-            this.data = newData;
+            this.data = this.targetData;
             this.render();
         }
     }
@@ -395,7 +405,7 @@ class ChartRenderer {
             this.ctx.font = '12px sans-serif';
             this.ctx.textAlign = 'center';
             this.ctx.textBaseline = 'middle';
-            this.ctx.fillText(item.label || '', labelX, labelY);
+            this.ctx.fillText(this.getSliceLabel(item), labelX, labelY);
 
             startAngle += sliceAngle;
         });
@@ -446,23 +456,30 @@ class ChartRenderer {
         const itemHeight = 20;
         const itemWidth = 100;
 
-        if (!this.data.length || !this.data[0].label) return;
+        const items = this.asArray(this.data);
+        const legendItems = items
+            .map((item, index) => ({
+                label: this.getSliceLabel(item),
+                colorIndex: index
+            }))
+            .filter(item => item.label);
+        if (legendItems.length === 0) return;
 
         this.ctx.font = '12px sans-serif';
         
-        this.data.forEach((item, index) => {
+        legendItems.forEach((item, index) => {
             const x = padding;
             const y = padding + index * (itemHeight + 5);
             
             // Draw color box
-            this.ctx.fillStyle = colors[index % colors.length];
+            this.ctx.fillStyle = colors[item.colorIndex % colors.length];
             this.ctx.fillRect(x, y, 12, 12);
             
             // Draw label
             this.ctx.fillStyle = '#e8eaed';
             this.ctx.textAlign = 'left';
             this.ctx.textBaseline = 'middle';
-            this.ctx.fillText(item.label || '', x + 18, y + 6);
+            this.ctx.fillText(item.label, x + 18, y + 6);
         });
     }
 
--- a/explorer/static/js/charts.js
+++ b/explorer/static/js/charts.js
@@ -66,11 +66,28 @@ class ChartRenderer {
         observer.observe(this.container);
     }
 
+    asArray(data) {
+        return Array.isArray(data) ? data : [];
+    }
+
+    numericValue(value, fallback = 0) {
+        const number = Number(value);
+        return Number.isFinite(number) ? number : fallback;
+    }
+
+    getNumericData() {
+        return this.asArray(this.data).map(value => this.numericValue(value));
+    }
+
+    getSliceValue(item) {
+        return Math.max(0, this.numericValue(item && item.value));
+    }
+
     /**
      * Update chart data
      */
     update(newData) {
-        this.targetData = newData;
+        this.targetData = this.asArray(newData);
         
         if (this.options.animation) {
             this.animate();
@@ -90,7 +107,7 @@ class ChartRenderer {
 
         const duration = 300;
         const startTime = performance.now();
-        const startData = this.data.length > 0 ? this.data : this.targetData;
+        const startData = this.asArray(this.data).length > 0 ? this.asArray(this.data) : this.targetData;
 
         const animateFrame = (currentTime) => {
             const elapsed = currentTime - startTime;
@@ -196,9 +213,10 @@ class ChartRenderer {
 
         // Vertical grid lines
         const gridWidth = width - padding.left - padding.right;
-        if (this.data.length > 1) {
-            for (let i = 0; i < this.data.length; i++) {
-                const x = padding.left + (gridWidth / (this.data.length - 1)) * i;
+        const dataLength = this.asArray(this.data).length;
+        if (dataLength > 1) {
+            for (let i = 0; i < dataLength; i++) {
+                const x = padding.left + (gridWidth / (dataLength - 1)) * i;
                 this.ctx.beginPath();
                 this.ctx.moveTo(x, padding.top);
                 this.ctx.lineTo(x, height - padding.bottom);
@@ -222,10 +240,11 @@ class ChartRenderer {
         const chartWidth = width - padding.left - padding.right;
         const chartHeight = height - padding.top - padding.bottom;
 
-        if (this.data.length < 2) return;
+        const data = this.getNumericData();
+        if (data.length < 2) return;
 
-        const maxValue = Math.max(...this.data.map(d => typeof d === 'number' ? d : 0));
-        const minValue = Math.min(...this.data.map(d => typeof d === 'number' ? d : 0));
+        const maxValue = Math.max(...data);
+        const minValue = Math.min(...data);
         const range = maxValue - minValue || 1;
 
         // Draw line
@@ -234,8 +253,8 @@ class ChartRenderer {
         this.ctx.lineWidth = 2;
         this.ctx.lineJoin = 'round';
 
-        this.data.forEach((value, index) => {
-            const x = padding.left + (chartWidth / (this.data.length - 1)) * index;
+        data.forEach((value, index) => {
+            const x = padding.left + (chartWidth / (data.length - 1)) * index;
             const y = padding.top + chartHeight - ((value - minValue) / range) * chartHeight;
             
             if (index === 0) {
@@ -248,8 +267,8 @@ class ChartRenderer {
         this.ctx.stroke();
 
         // Draw points
-        this.data.forEach((value, index) => {
-            const x = padding.left + (chartWidth / (this.data.length - 1)) * index;
+        data.forEach((value, index) => {
+            const x = padding.left + (chartWidth / (data.length - 1)) * index;
             const y = padding.top + chartHeight - ((value - minValue) / range) * chartHeight;
             
             this.ctx.beginPath();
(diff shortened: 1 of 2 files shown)


william08190 (changes requested):
Bounty #73 review: I reviewed current head `2f102ab090e4833470015c85f75ee04536b5bd00` for the chart renderer normalization patch.

The line/bar numeric coercion and zero-total pie/doughnut guard are useful, but the pie path still has a malformed-slice crash that should be fixed before merge.

The new `getSliceValue(item)` makes the `total` reduction tolerate `null`/malformed slice entries, but `drawPieChart()` still later reads `item.label` directly:

```js
const value = this.getSliceValue(item);
...
this.ctx.fillText(item.label || '', labelX, labelY);
```

That means a payload containing one valid positive slice plus one `null` slice still crashes during rendering. This is exactly the kind of malformed chart data the PR is trying to make safe: the total calculation now survives, but the later label draw path does not.

I reproduced it with a minimal canvas/document stub against the extracted `ChartRenderer`:

```js
const chart = new ChartRenderer('chart', {
  type: 'pie',
  animation: false,
  showLegend: false
});
chart.update([{ label: 'valid', value: 1 }, null]);
```

Observed output:

```text
{"ok":false,"name":"TypeError","message":"Cannot read properties of null (reading 'label')"}
```

There is a related legend path at `drawLegend()` that reads `this.data[0].label` and later `item.label` without first filtering/normalizing legend rows, so a `null` first item can still break even when a chart draw returns early.

Suggested fix: normalize pie/doughnut/legend slice entries to plain objects before rendering, or use safe helpers for both slice value and label, e.g. skip non-object slices or treat their labels as `''`. Please add a regression where a pie chart receives `[{ label: 'valid', value: 1 }, null]` and renders without throwing.

Validation I ran:

- `PYTHONPATH=. PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --with pytest --with flask python -m pytest tests/test_chart_renderer_data_normalization.py -q --noconftest -o addopts=''` -> 3 passed.
- `python3 -m py_compile tests/test_chart_renderer_data_normalization.py` -> passed.
- `uv run python tools/bcos_spdx_check.py --base-ref origin/main` -> OK.
- `git diff --check origin/main...HEAD -- explorer/static/js/charts.js tests/test_chart_renderer_data_normalization.py` -> clean.
- Node canvas/document stub against `explorer/static/js/charts.js` -> reproduced the `null` slice crash above.

achimala (changes requested):
Bounty #73 review: the chart-renderer normalization work is headed in the right direction and the focused tests pass, but the non-animated update path still stores the raw unnormalized payload.

`update(newData)` now does:

```js
this.targetData = this.asArray(newData);

if (this.options.animation) {
    this.animate();
} else {
    this.data = newData;
    this.render();
}
```

So charts created with `{ animation: false }` still leave `this.data` as an arbitrary object/string/etc. I reproduced this on the PR head with a small VM probe:

```text
const c = new ChartRenderer('chart', { animation: false });
c.update({ not: 'array' });

targetData_is_array true []
data_is_array false {"not":"array"}
```

That contradicts the stated goal of normalizing update payloads before rendering and leaves future render/legend paths depending on every call site remembering to re-wrap `this.data` through `asArray()`. Please assign the normalized value in the non-animated branch too, e.g. `this.data = this.targetData`, and add a regression for `animation: false` with a non-array payload.

Validation I ran:
- `python3 -m py_compile tests/test_chart_renderer_data_normalization.py` -> passed.
- `uv run --with pytest --with flask python -m pytest tests/test_chart_renderer_data_normalization.py -q --noconftest -o addopts=''` -> 3 passed.
- `node --check explorer/static/js/charts.js` -> passed.
- `uv run python tools/bcos_spdx_check.py --base-ref origin/main` -> OK.
- `git diff --check origin/main...HEAD -- explorer/static/js/charts.js tests/test_chart_renderer_data_normalization.py` -> clean.

techoisonocreate-crypto (changes requested):
Bounty #73 review: I found one remaining malformed-slice path in the current head `2f102ab`.

The line/area/bar paths now coerce numeric series, and zero-total pie/doughnut charts return early. However, pie and doughnut rendering still iterate every element in `slices = this.asArray(this.data)` without filtering to object rows. `getSliceValue(item)` safely handles `null`, but `drawPieChart()` later reads `item.label` unguarded. If the API returns a mixed slice array with at least one valid positive slice and one `null`, the chart still crashes during render.

I reproduced this with a minimal DOM/canvas stub against the extracted `ChartRenderer`:

```text
crash [{"label":"ok","value":1},null] TypeError: Cannot read properties of null (reading 'label')
crash [null,{"label":"ok","value":1}] TypeError: Cannot read properties of null (reading 'label')
ok [{"label":"ok","value":1},["bad"]]
```

This means the zero-total guard does not cover mixed malformed rows; one good slice makes `total > 0`, then the bad row is still rendered and can throw. Please normalize pie/doughnut slice rows to plain objects before total calculation and rendering, or guard label access the same way value access is guarded. A regression should include a mixed payload such as `[{label: 'ok', value: 1}, null]` and assert render does not throw.

Validation I ran:
- `python -m py_compile tests/test_chart_renderer_data_normalization.py` -> passed.
- `python -m pytest tests/test_chart_renderer_data_normalization.py -q --noconftest -o addopts=''` -> 3 passed.
- `git diff --check origin/main...HEAD -- explorer/static/js/charts.js tests/test_chart_renderer_data_normalization.py` -> clean.
- Node DOM/canvas-stub probe reproduced the mixed-slice crash above.

jasperdevs:
Addressed the review blockers in ff5de8a. The non-animated update path now assigns the normalized target array to data before rendering, and pie/legend label access now goes through safe slice helpers so mixed malformed slices such as [{ label: 'valid', value: 1 }, null] do not throw. Added Node-backed regressions for both reviewer repros.

Validation run:
- node --check explorer/static/js/charts.js
- python3 -m py_compile tests/test_chart_renderer_data_normalization.py
- uv run --with pytest --with flask python -m pytest tests/test_chart_renderer_data_normalization.py -q -> 5 passed
- uv run python tools/bcos_spdx_check.py --base-ref origin/main -> OK
- git diff --check -> clean

techoisonocreate-crypto (approved):
Follow-up review on current head `ff5de8a`: the malformed-slice blocker I raised is fixed, and the non-animated update normalization issue raised by another reviewer is fixed too.

What I rechecked:
- `getSliceLabel()` now guards non-object/null slice rows before label rendering.
- `drawPieChart()` uses `this.getSliceLabel(item)` instead of directly reading `item.label`.
- `drawLegend()` maps labels through the safe helper and filters empty labels before rendering.
- The non-animated update path now assigns `this.data = this.targetData`, so non-array update payloads are normalized consistently.

Validation I ran:
- `python -m py_compile tests/test_chart_renderer_data_normalization.py` -> passed.
- `python -m pytest tests/test_chart_renderer_data_normalization.py -q --noconftest -o addopts=''` -> 5 passed.
- `node --check explorer/static/js/charts.js` -> passed.
- `git diff --check origin/main...HEAD -- explorer/static/js/charts.js tests/test_chart_renderer_data_normalization.py` -> clean.
- Re-ran my Node DOM/canvas-stub repro: `[{label:'ok', value:1}, null]`, `[null, {label:'ok', value:1}]`, and `[{label:'ok', value:1}, ['bad']]` all rendered without throwing; a non-array update with `animation:false` left `chart.data` as an empty array.

I do not see a remaining blocker in this head.

william08190 (approved):
Follow-up review on current head `ff5de8a07f5593be15970b457b43b29121bd5726`: the malformed pie-slice blocker from my prior review is fixed.

What I rechecked:
- `drawPieChart()` no longer reads `item.label` directly; malformed/null slice labels go through a safe helper.
- `drawLegend()` now derives labels through the same safe path and filters empty labels instead of assuming `this.data[0].label` exists.
- The non-animated update path assigns the normalized target array to `this.data`, so non-array payloads no longer leave raw objects in chart state.

Validation I ran:
- `PYTHONPATH=. PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --with pytest --with flask python -m pytest tests/test_chart_renderer_data_normalization.py -q --noconftest -o addopts=''` -> 5 passed.
- `node --check explorer/static/js/charts.js` -> passed.
- `python3 -m py_compile tests/test_chart_renderer_data_normalization.py` -> passed.
- `uv run python tools/bcos_spdx_check.py --base-ref origin/main` -> OK.
- `git diff --check origin/main...HEAD -- explorer/static/js/charts.js tests/test_chart_renderer_data_normalization.py` -> clean.
- Re-ran the Node canvas/document stub probes for `[{ label: 'valid', value: 1 }, null]`, `[null, { label: 'valid', value: 1 }]`, `[{ label: 'valid', value: 1 }, ['bad']]`, and a non-array update payload. All completed without throwing, and the non-array update normalized `chart.data` to `[]`.

I do not see a remaining blocker in this head.

jaxint (approved):

surim0n (approved):
Reviewed the current chart renderer normalization head after the earlier blockers were addressed. The non-animated update path now stores the normalized array, numeric chart paths coerce values before drawing, and pie/doughnut/legend paths use safe slice helpers so malformed rows do not throw.

Validation run:
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_chart_renderer_data_normalization.py -q --noconftest -o addopts='' -> 5 passed
- node --check explorer/static/js/charts.js -> passed
- python -m py_compile tests/test_chart_renderer_data_normalization.py -> passed
- python tools/bcos_spdx_check.py --base-ref origin/main -> OK
- git diff --check origin/main...HEAD -- explorer/static/js/charts.js tests/test_chart_renderer_data_normalization.py -> clean

508704820 (commented):

508704820 (commented):

BossChaos (approved):
**Code Review — PR #5819**

✅ **Approve** — Miner row normalization for PR #5819.

JeremyZeng77:

What I checked:
- Changed files reviewed:
- explorer/static/js/charts.js
- tests/test_chart_renderer_data_normalization.py

- uv run --with pytest --with flask python -m pytest tests/test_chart_renderer_data_normalization.py -q -> 5 passed in 0.36s
- python -B -m py_compile "tests/test_chart_renderer_data_normalization.py"

BossChaos (commented):
## Security Review — PR #5819

**Reviewer:** @BossChaos

---

### Overview

PR #5819 hardens `explorer/static/js/charts.js` by adding six type-coercion helper methods to the `ChartRenderer` class. The key vulnerability it addresses is that the original code used `typeof d === 'number' ? d : 0` which silently converted any non-number to 0, masking data corruption.

**Verdict: APPROVED with 3 findings.**

---

### Finding 1 — HIGH: `numericValue` silently accepts boolean values as numbers

**File:** `explorer/static/js/charts.js`

```javascript
numericValue(value, fallback = 0) {
    const number = Number(value);
    return Number.isFinite(number) ? number : fallback;
}
```

**Exploitation scenario:**
```javascript
renderer.numericValue(true)   // → 1 (Boolean accepted as number!)
renderer.numericValue(false)  // → 0 (Boolean accepted as number!)
renderer.numericValue("100")  // → 100 (String accepted as number!)
```

**Impact:** In JavaScript, `Number(true)` returns `1` and `Number(false)` returns `0`. Both pass `Number.isFinite()`. This means if an attacker can inject boolean values into chart data (via a compromised API or man-in-the-middle on unencrypted connections), the charts will silently render incorrect values. For financial dashboards showing RTC balances, epoch rewards, or mining rates, displaying `true` as `1` or `false` as `0` could mislead operators into making wrong decisions.

This is the same pattern found across multiple BCOS PRs (#5807 `safeNumber`, #5812 `safeNumber`, #5815 `safeNumber`, #5790 `safeNumber`).

**Recommendation:** Add explicit type check BEFORE `Number()` coercion:
```javascript
numericValue(value, fallback = 0) {
    if (typeof value !== 'number') {
        // Try parsing string representations of numbers
        if (typeof value === 'string') {
            const n = Number(value);
            return Number.isFinite(n) ? n : fallback;
        }
        return fallback;
    }
    return Number.isFinite(value) ? value : fallback;
}
```

---

### Finding 2 — MEDIUM: `asArray` does not validate array element types

**File:** `explorer/static/js/charts.js`

```javascript
asArray(data) {
    return Array.isArray(data) ? data : [];
}
```

**Impact:** If `data` is `[null, undefined, "string", true, 42]`, `asArray` returns it unchanged. The downstream `numericValue` call in `getNumericData()` would process these values: `null→0`, `undefined→0`, `"string"→0`, `true→1`, `42→42`. The `"string"` → `0` conversion is particularly dangerous because it silently discards data without any indication.

**Recommendation:** Add element-level filtering:
```javascript
asArray(data) {
    if (!Array.isArray(data)) return [];
    return data.filter(item => typeof item === 'number' || typeof item === 'string');
}
```

---

### Finding 3 — LOW: `getSliceValue` allows negative values before clamping

**File:** `explorer/static/js/charts.js`

```javascript
getSliceValue(item) {
    if (!this.isSliceObject(item)) return 0;
    return Math.max(0, this.numericValue(item.value));
}
```

**Impact:** The `Math.max(0, ...)` clamping correctly prevents negative slice values. However, if `item.value` is `-Infinity`, `numericValue` returns `0` (because `Number.isFinite(-Infinity)` is `false`), so the clamping is actually redundant. The issue is that `numericValue(-Infinity)` → `fallback (0)` which is correct, but the fallback behavior should be documented.

---

### Summary

| Finding | Severity | Type |
|---------|----------|------|
| numericValue boolean injection | HIGH | Type Confusion |
| asArray no element validation | MEDIUM | Data Integrity |
| getSliceValue negative value handling | LOW | Edge Case |

**Claim: 45 RTC (1 HIGH + 1 MEDIUM + 1 LOW)**

PR #5822, opened 2026-05-20
fix: guard static status dashboard payloads

jasperdevs:
## Summary
- normalize the static status dashboard history payload before reading latest nodes
- render an empty-state card for empty or malformed node_status.json data
- guard per-history node lookups so missing nodes arrays do not break the uptime bar
- add Node-backed regressions for malformed status payloads

## Validation
- node syntax probe for static/status/index.html embedded script
- python3 -m py_compile tests/test_status_dashboard_frontend_security.py
- uv run --with pytest --with flask python -m pytest tests/test_status_dashboard_frontend_security.py -q
- uv run python tools/bcos_spdx_check.py --base-ref origin/main
- git diff --check

### DIFF
--- a/static/status/index.html
+++ b/static/status/index.html
@@ -130,6 +130,17 @@
         return value && typeof value === 'object' && !Array.isArray(value) ? value : {};
     }
 
+    function isValidNode(value) {
+        if (!value || typeof value !== 'object' || Array.isArray(value)) return false;
+        return value.name != null || value.url != null;
+    }
+
+    function formatTimestamp(value) {
+        if (!value) return 'unavailable';
+        const date = new Date(value);
+        return Number.isNaN(date.getTime()) ? 'unavailable' : date.toLocaleTimeString();
+    }
+
     function renderEmptyState(grid, message) {
         const card = document.createElement('div');
         card.className = 'card';
@@ -137,7 +148,15 @@
         grid.appendChild(card);
     }
 
+    function renderUnavailable(grid) {
+        grid.innerHTML = '';
+        renderEmptyState(grid, 'No node status data available');
+        document.getElementById('lastUpdate').innerText = 'LAST_UPDATE: unavailable';
+        document.getElementById('sysTime').innerText = new Date().toISOString();
+    }
+
     async function updateStatus() {
+        const grid = document.getElementById('nodeGrid');
         try {
             const resp = await fetch('node_status.json');
             if (!resp.ok) {
@@ -145,15 +164,12 @@
             }
             const history = safeArray(await resp.json()).map(safeObject);
             const latest = history.length > 0 ? history[history.length - 1] : {};
-            const latestNodes = safeArray(latest.nodes).map(safeObject);
+            const latestNodes = safeArray(latest.nodes).filter(isValidNode).map(safeObject);
             
-            const grid = document.getElementById('nodeGrid');
             grid.innerHTML = '';
 
             if (latestNodes.length === 0) {
-                renderEmptyState(grid, 'No node status data available');
-                document.getElementById('lastUpdate').innerText = 'LAST_UPDATE: unavailable';
-                document.getElementById('sysTime').innerText = new Date().toISOString();
+                renderUnavailable(grid);
                 return;
             }
             
@@ -166,7 +182,7 @@
                 
                 // Get node history for uptime bar
                 const nodeHistory = history.map(h => {
-                    const found = nodeUrl ? safeArray(h.nodes).map(safeObject).find(n => n.url === nodeUrl) : null;
+                    const found = nodeUrl ? safeArray(h.nodes).filter(isValidNode).map(safeObject).find(n => n.url === nodeUrl) : null;
                     return found ? safeStatus(found.status) === 'up' : false;
                 }).slice(-50); // Show last 50 checks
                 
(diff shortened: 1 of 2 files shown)


techoisonocreate-crypto (changes requested):
Bounty #73 review: I found a remaining malformed-node normalization issue in current head `49f2c5a`.

The patch safely handles a non-array top-level history payload and missing `latest.nodes`, but it still maps every entry in `latest.nodes` through `safeObject()` without filtering invalid node rows:

```js
const latestNodes = safeArray(latest.nodes).map(safeObject);
...
latestNodes.forEach(node => { ... render card ... });
```

That means malformed node entries such as `null` or nested arrays become `{}` and are rendered as blank `DOWN` node cards instead of being ignored or causing the dashboard to show the empty state. I reproduced this against the embedded script with a DOM stub:

```js
[
  {
    time: 'bad',
    nodes: [null, ['bad'], { name: '<x>', url: 'u', status: 'up', latency_ms: 0, block_height: 0 }]
  }
]
```

Observed result:

```text
childCount: 3
first card: empty node-name/location, status DOWN, latency --ms, miners 0
lastUpdate: LAST_UPDATE: Invalid Date
```

So one valid node plus two malformed rows renders three cards. This is misleading status data, and it weakens the stated goal of rendering an empty state for malformed status data. The same probe also shows `latest.time = 'bad'` renders `LAST_UPDATE: Invalid Date`; please treat invalid timestamps like unavailable rather than surfacing `Invalid Date`.

Suggested fix: filter node rows to plain objects with at least a meaningful `url` or `name` before rendering, and validate `latest.time` with `Number.isNaN(date.getTime())` before calling `toLocaleTimeString()`. Add a regression where `nodes` contains `[null, ['bad'], validNode]` and assert only the valid node card renders.

Validation I ran:
- `python -m py_compile tests/test_status_dashboard_frontend_security.py` -> passed.
- `python -m pytest tests/test_status_dashboard_frontend_security.py -q --noconftest -o addopts=''` -> 4 passed.
- `git diff --check origin/main...HEAD -- static/status/index.html tests/test_status_dashboard_frontend_security.py` -> clean.
- Node DOM-stub probe reproduced malformed node rows rendering as blank DOWN cards and invalid `latest.time` rendering as `Invalid Date`.

surim0n (changes requested):
I found one remaining dashboard failure path in this PR.

The new empty-state rendering only runs after a successful HTTP response whose JSON parses to an empty or malformed history array. For a missing or unavailable node_status.json, updateStatus throws on !resp.ok and then the catch block only logs to console. The grid is never cleared and no empty state is rendered, so a first load with a 404/500 still leaves users staring at the original blank dashboard/polling state.

This is in the PR surface because the stated goal is to handle empty or malformed node_status.json data. Missing node_status.json is the operational version of the same problem, and it is easy to reproduce: https://rustchain.org/status/node_status.json and https://rustchain.org/node_status.json both currently return 404 from the live site.

Suggested fix: move the grid lookup/clearing before the fetch or call renderEmptyState from the catch block, then set LAST_UPDATE to unavailable and sysTime like the empty-history path. Add a Node-backed regression with fetch returning { ok: false, status: 404 } so the error path is covered.

Validation I ran:
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_status_dashboard_frontend_security.py -q -> 4 passed in 0.16s
- curl -fsS https://rustchain.org/status/node_status.json -> HTTP 404
- curl -fsS https://rustchain.org/node_status.json -> HTTP 404

jasperdevs:
Addressed the review blockers in a77ab0f. The status dashboard now renders the unavailable empty state on non-OK fetches, filters malformed latest/history node rows instead of rendering blank DOWN cards, and formats timestamps through a helper that falls back to unavailable for invalid dates. Added regressions for fetch 404, mixed malformed node rows, and bad latest.time.

Validation run:
- inline script parse with Node new Function(...)
- python3 -m py_compile tests/test_status_dashboard_frontend_security.py
- uv run --with pytest --with flask python -m pytest tests/test_status_dashboard_frontend_security.py -q -> 5 passed
- uv run python tools/bcos_spdx_check.py --base-ref origin/main -> OK
- git diff --check -> clean

techoisonocreate-crypto (approved):
Follow-up review on the updated head: the malformed node-row blocker is fixed by `isValidNode()` filtering, invalid `latest.time` now renders `LAST_UPDATE: unavailable`, and failed/missing `node_status.json` responses render the empty-state card instead of leaving the dashboard in its initial polling state.

Validation performed:
- `python -m py_compile tests/test_status_dashboard_frontend_security.py`
- `python -m pytest tests/test_status_dashboard_frontend_security.py -q --noconftest -o addopts=''` -> 5 passed
- `python tools/bcos_spdx_check.py --base-ref origin/main`
- `git diff --check origin/main...HEAD -- static/status/index.html tests/test_status_dashboard_frontend_security.py`
- inline-script parse via Node `new Function(...)`

jaxint (approved):

508704820 (commented):

508704820 (commented):

BossChaos (approved):
**Code Review — PR #5822**

✅ **Approve** — Miner row normalization for PR #5822.

JeremyZeng77:

What I checked:
- Changed files reviewed:
- static/status/index.html
- tests/test_status_dashboard_frontend_security.py

- uv run --with pytest --with flask python -m pytest tests/test_status_dashboard_frontend_security.py -q -> 5 passed in 0.39s
- python -B -m py_compile "tests/test_status_dashboard_frontend_security.py"

PR #5955, opened 2026-05-20
fix: preserve explicit parasocial timestamps

jasperdevs:
## Summary
- preserve explicitly supplied `watched_at=0` timestamps in the BoTTube parasocial tracker
- add regression coverage for Unix epoch timestamps in viewer pattern output

## Tests
- uv run --no-project --with pytest --with flask python -m pytest tests/test_parasocial.py -q
- python3 -m py_compile tools/bottube_parasocial.py tests/test_parasocial.py
- python3 tools/bcos_spdx_check.py --base-ref origin/main
- uv run --no-project --with ruff ruff check tools/bottube_parasocial.py tests/test_parasocial.py
- git diff --check

### DIFF
--- a/tests/test_parasocial.py
+++ b/tests/test_parasocial.py
@@ -121,6 +121,17 @@ class TestViewerPattern(unittest.TestCase):
         finally:
             os.unlink(db)
 
+    def test_pattern_preserves_explicit_epoch_timestamp(self):
+        t, db = _fresh()
+        try:
+            t.track_view("epoch_viewer", "v1", 300, watched_at=0, total_video_secs=600)
+            p = t.get_viewer_pattern("epoch_viewer")
+            self.assertEqual(p["peak_hour"], 0)
+            self.assertEqual(p["first_seen"], "1970-01-01T00:00:00+00:00")
+            self.assertEqual(p["last_seen"], "1970-01-01T00:00:00+00:00")
+        finally:
+            os.unlink(db)
+
     def test_engagement_trend_rising(self):
         t, db = _fresh()
         try:
--- a/tools/bottube_parasocial.py
+++ b/tools/bottube_parasocial.py
@@ -75,7 +75,7 @@ class AudienceTracker:
         total_video_secs: float = 600.0,
     ):
         """Record a viewer interaction with a video."""
-        ts = watched_at or int(time.time())
+        ts = watched_at if watched_at is not None else int(time.time())
         with self._conn() as conn:
             conn.execute(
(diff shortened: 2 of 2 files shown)


achimala (approved):
Verified the timestamp-preservation fix locally. The change correctly treats `watched_at=0` as an explicit epoch timestamp instead of falling back to `time.time()`, and the new regression covers the viewer pattern fields that previously drifted.

Validation run:
- `uv run --no-project --with pytest --with flask python -m pytest tests/test_parasocial.py -q` -> 19 passed
- `python3 -m py_compile tools/bottube_parasocial.py tests/test_parasocial.py`
- `uv run --no-project --with ruff ruff check tools/bottube_parasocial.py tests/test_parasocial.py`
- `python3 tools/bcos_spdx_check.py --base-ref origin/main`
- `git diff --check origin/main...HEAD -- tools/bottube_parasocial.py tests/test_parasocial.py`

GitHub CI is green as well.

mouseos (changes requested):
The timestamp fix itself is directionally correct, but this head is not green on Windows because the focused test suite leaves SQLite database files open before cleanup.

Every test in `tests/test_parasocial.py` creates a file-backed `AudienceTracker` via `_fresh()` and then immediately calls `os.unlink(db)` in `finally`. On Windows, `AudienceTracker._conn()` returns a `sqlite3.Connection`, and using `with self._conn() as conn:` commits/rolls back but does not close the connection object. The file remains locked long enough that every cleanup unlink fails.

Reproduction on this head:

```text
python -m pytest tests\test_parasocial.py -q --noconftest -o addopts=''
19 failed in 0.96s
```

Representative failure:

```text
PermissionError: [WinError 32] ... cannot access ... tmp*.db
```

The new regression `test_pattern_preserves_explicit_epoch_timestamp` is affected too, so the PR's own focused test command cannot pass on Windows even though the assertion body succeeds before cleanup.

Suggested fix: make `AudienceTracker` close file-backed connections explicitly, for example by changing `_conn()` into a real context manager that closes in `finally`, or by wrapping the returned connection with `contextlib.closing(...)` at each use site. After that, keep the new epoch timestamp regression and verify that the temp DB can be unlinked immediately after tracker operations.

Positive checks:

```text
python -m py_compile tools\bottube_parasocial.py tests\test_parasocial.py
passed

git diff --check origin/main...HEAD -- tools\bottube_parasocial.py tests\test_parasocial.py
passed
```

jaxint (approved):

PR #6050, opened 2026-05-21
fix: validate p2p announce peer urls

jasperdevs:
## Summary
- reject missing, non-string, and blank peer_url values in /p2p/announce before passing them to the peer manager
- trim valid peer_url strings before storage
- add route coverage for malformed peer_url payloads

## Tests
- uv run --no-project --with pytest --with flask --with requests python -m pytest node/tests/test_p2p_sync_routes.py node/tests/test_p2p_sync_flask_imports.py -q
- python3 py_compile check for node/rustchain_p2p_sync.py and node/tests/test_p2p_sync_routes.py

Note: ruff check was attempted through uv with --isolated and timed out twice locally before producing output; no standalone ruff binary is installed on this machine.


shootingallday (changes requested):
﻿Reviewed current head `1fa28bf76483e06ca06b90604f27362cefe72989`.

The new type guards cover missing, blank, and non-string `peer_url` values, and the focused route tests pass locally. I found one remaining blocker: malformed string values are still accepted as active peers, even though this endpoint is supposed to validate peer URLs before they reach the peer manager.

Direct Flask probe against this PR head using the real `PeerManager`:

```text
peer_url = "not-a-url"
-> HTTP 200 {"ok": true, "peers": 1}
-> active peers: ["not-a-url"]

peer_url = "http:///missing-host"
-> HTTP 200 {"ok": true, "peers": 2}
-> active peers includes "http:///missing-host"
```

Those are not usable peer URLs. They get persisted as active peers, and later health checks / block sync / transaction gossip will build request URLs from them. Other malformed strings such as `ftp://peer.example:21` and `http://peer.example:notaport` return `200 {"ok": false, ...}` after `PeerManager.add_peer()` fails, but the route still does not reject them as validation errors before the manager path.

Suggested fix: validate the stripped string with `urllib.parse.urlparse()` before calling `add_peer()`, require `http` or `https`, require a non-empty hostname/netloc, and reject invalid ports. Please add route regressions for at least `not-a-url`, `http:///missing-host`, and a bad port such as `http://peer.example:notaport`, asserting a 400 and no peer storage.

Validation run locally:
- `python -m py_compile node/rustchain_p2p_sync.py node/tests/test_p2p_sync_routes.py setup_miner.py` -> passed
- `uv run --no-project --with pytest --with flask --with requests python -m pytest node/tests/test_p2p_sync_routes.py node/tests/test_p2p_sync_flask_imports.py -q --tb=short` -> 16 passed
- Direct Flask test-client probe with the real `PeerManager` reproduced accepted malformed URL strings above.

TJCurnutte (approved):
Approved. I verified that `/p2p/announce` now rejects non-string and blank `peer_url` values before they can be registered as peers, while still accepting trimmed valid strings.

Proof:
- Head reviewed: `1fa28bf76483e06ca06b90604f27362cefe72989`.
- `git diff --check origin/main...HEAD -- miners/checksums.sha256 node/rustchain_p2p_sync.py node/tests/test_p2p_sync_routes.py setup_miner.py` passed.
- `python3 -B -m py_compile node/rustchain_p2p_sync.py node/tests/test_p2p_sync_routes.py setup_miner.py` passed.
- `PYTHONPATH=. python3 -B -m pytest -q node/tests/test_p2p_sync_routes.py --tb=short` returned `7 passed in 0.14s`.
- `shasum -a 256 miners/macos/rustchain_mac_miner_v2.5.py` returned `163fafcf751d8fbd41bf936facaeb366c042f467fa34b79f2c4c0a45472ef70f`, matching the pin.
- Direct Flask probe against `origin/main` (`e89b5f7e90cbf2bd1b92f7b69a8ee68a74091159`) showed integer/list/dict/blank-space `peer_url` values returned `200` and were passed directly into `peer_manager.add_peer`. On this head, integer/list/dict values return `400 peer_url must be a string`, blank-space returns `400 peer_url required`, no peer is added in those cases, and `" http://peer.example:8088 "` is trimmed and accepted.

This is a focused input-validation fix with regression coverage for the unsafe shapes and no observed regression on valid peer announcements.

minyanyi (approved):
Tested locally with `python -B -m pytest -q node/tests/test_p2p_sync_routes.py --tb=short`: 7 passed, 1 warning. I also ran `python -B -m py_compile node/rustchain_p2p_sync.py` successfully.

Focused checks:

1. `/p2p/announce` now rejects missing `peer_url`, non-string JSON values, and whitespace-only strings before calling `peer_manager.add_peer()`. The added tests assert the peer list remains unchanged for these malformed inputs.

2. Valid string values are stripped before registration, which preserves the existing add-peer flow while preventing padded blank values from slipping through as announced peers.

3. I verified the updated macOS miner checksum in `setup_miner.py` / `miners/checksums.sha256` matches the local `miners/macos/rustchain_mac_miner_v2.5.py` SHA256: `163fafcf751d8fbd41bf936facaeb366c042f467fa34b79f2c4c0a45472ef70f`.

jaxint (approved):

PR #6051, opened 2026-05-21
fix: validate bft propose epochs

jasperdevs:
## Summary
- reject boolean, non-integer, and negative epoch values in /bft/propose before invoking consensus logic
- add focused route coverage for malformed epoch payloads
- keep existing miners/distribution validation behavior intact

## Tests
- uv run --no-project --with pytest --with flask --with requests python -m pytest node/tests/test_bft_routes.py node/tests/test_bft_route_validation.py -q
- bounded py_compile check for node/rustchain_bft_consensus.py and node/tests/test_bft_routes.py

Note: local ruff via uv hung twice earlier in this environment, so the focused tests and compile check were used for this narrow patch.


lilcode99999 (approved):
Reviewed current head `86492908eb148fec2d238488602e1c9dcb5f8ae3`.

This looks good to me for the stated scope. The `/bft/propose` route now rejects the problematic epoch shapes before entering consensus logic:

- `None` still returns the existing missing-field error.
- `bool` is rejected separately from `int`, which matters in Python because `bool` subclasses `int`.
- strings/floats are rejected instead of reaching `propose_epoch_settlement()` and later proposal validation.
- negative integers are rejected at the route boundary with a stable 400.

The new tests in `node/tests/test_bft_routes.py` cover those regression cases directly, and the existing miners/distribution shape checks remain after the epoch validation, so this patch keeps the older collection-field behavior intact.

Validation I ran locally on the PR head:

```text
git diff --check origin/main...HEAD -- node/rustchain_bft_consensus.py node/tests/test_bft_routes.py setup_miner.py miners/checksums.sha256
# clean

python3 -m py_compile node/rustchain_bft_consensus.py node/tests/test_bft_routes.py setup_miner.py
# passed
```

I could not run the focused pytest suite in this local Codex environment because neither the system Python nor the bundled Python has `pytest`/`flask` installed, but the route-level change is small and the added assertions match the guarded branches.

Non-blocking follow-up: it would be useful to add one positive route test with a fake BFT object that records its arguments, proving a valid integer epoch such as `0` or `1` is passed through unchanged to `propose_epoch_settlement()`. The current tests are all rejection-path tests, so that would make the route contract a little more explicit.

Disclosure: OpenAI Codex assisted with source review and drafting this review.

minyanyi (approved):
Tested locally with `python -B -m pytest -q node/tests/test_bft_routes.py --tb=short`: 7 passed, 1 warning. I also ran `python -B -m py_compile node/rustchain_bft_consensus.py` successfully.

Focused checks:

1. `/bft/propose` now rejects JSON booleans, strings, and floats for `epoch` before consensus proposal creation. The explicit `isinstance(epoch, bool)` guard matters because `bool` is an `int` subclass in Python.

2. Negative epochs now return a deterministic `epoch must be non-negative` 400, while the existing missing-epoch and collection-shape checks continue to run independently.

3. I verified the updated macOS miner checksum in `setup_miner.py` / `miners/checksums.sha256` matches the local `miners/macos/rustchain_mac_miner_v2.5.py` SHA256: `163fafcf751d8fbd41bf936facaeb366c042f467fa34b79f2c4c0a45472ef70f`.

TJCurnutte (approved):
Approved. I verified that `/bft/propose` now rejects invalid `epoch` values before they reach the BFT proposal path.

Proof:
- Head reviewed: `86492908eb148fec2d238488602e1c9dcb5f8ae3`.
- `git diff --check origin/main...HEAD -- miners/checksums.sha256 node/rustchain_bft_consensus.py node/tests/test_bft_routes.py setup_miner.py` passed.
- `python3 -B -m py_compile node/rustchain_bft_consensus.py node/tests/test_bft_routes.py setup_miner.py` passed.
- `PYTHONPATH=. python3 -B -m pytest -q node/tests/test_bft_routes.py --tb=short` returned `7 passed in 0.15s`.
- `shasum -a 256 miners/macos/rustchain_mac_miner_v2.5.py` returned `163fafcf751d8fbd41bf936facaeb366c042f467fa34b79f2c4c0a45472ef70f`, matching the pin.
- Direct Flask probe against `origin/main` (`e89b5f7e90cbf2bd1b92f7b69a8ee68a74091159`) showed `True`, `"1"`, `1.5`, and `-1` all reached `propose_epoch_settlement(...)` and then returned `400 not_leader_or_already_committed`. On this head, `True`, `"1"`, and `1.5` return `400 epoch must be an integer` with no BFT call, `-1` returns `400 epoch must be non-negative` with no BFT call, and valid `0` still reaches the proposal path.

That is a targeted body-validation fix. The new tests cover bool/string/float and negative epochs without regressing the existing JSON-object, missing-field, miners-list, or distribution-object guards.

jaxint (approved):

PR #6053, opened 2026-05-21
fix: validate beacon identity pubkeys

jasperdevs:
## Summary
- reject non-string/non-hex beacon identity pubkeys without raising TypeError
- require TOFU envelope pubkeys to decode to exactly 32 bytes before deriving the Beacon agent ID
- add regression coverage for malformed and wrong-length pubkey payloads

## Tests
- PYTHONPATH=node uv run --no-project --with pytest --with flask --with cryptography python -m pytest node/tests/test_beacon_identity.py -q
- bounded py_compile check for node/beacon_identity.py and node/tests/test_beacon_identity.py

Note: local ruff via uv previously hung in this environment, so focused tests and compile check were used for this narrow patch.

### DIFF
--- a/node/beacon_identity.py
+++ b/node/beacon_identity.py
@@ -183,11 +183,15 @@ def learn_key_from_envelope(
     if not agent_id or not pubkey_hex:
         return False, "missing_agent_id_or_pubkey"
 
-    # Verify agent_id is consistent with declared pubkey
+    # Verify agent_id is consistent with a valid Ed25519 public key.
     try:
-        expected_id = agent_id_from_pubkey(bytes.fromhex(pubkey_hex))
-    except ValueError:
+        pubkey_bytes = bytes.fromhex(pubkey_hex)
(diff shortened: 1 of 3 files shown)


jaxint (approved):

hubeiqiao (approved):
Reviewed the beacon identity pubkey validation change in #6053.

The fix looks correct to me. `learn_key_from_envelope()` now separates hex decoding from Beacon agent ID derivation, rejects non-string or malformed hex pubkeys with `invalid_pubkey_encoding`, and rejects decoded pubkeys that are not exactly 32 bytes before accepting the TOFU key. That prevents malformed envelope payloads from reaching the identity derivation path while keeping the existing mismatch behavior for valid-length but wrong keys.

Validation run on this branch:

- `PYTHONPATH=node uv run --no-project --with pytest --with flask --with cryptography python -m pytest node/tests/test_beacon_identity.py -q` -> 33 passed

No blocker found.

PR #6055, opened 2026-05-21
fix: validate wallet history offsets

jasperdevs:
## Summary
- reject /wallet/<address>/history offsets that exceed SQLite's signed integer range before they reach the query layer
- add route coverage for oversized offset validation
- preserve existing limit handling, negative-offset clamping, and empty-history behavior

## Tests
- uv run --no-project --with pytest --with flask python -m pytest tests/test_tx_handler_limits.py tests/test_tx_submit_route.py -q
- bounded py_compile check for node/rustchain_tx_handler.py and tests/test_tx_handler_limits.py

Note: local ruff via uv previously hung in this environment, so focused tests and compile check were used for this narrow patch.

### DIFF
--- a/node/rustchain_tx_handler.py
+++ b/node/rustchain_tx_handler.py
@@ -33,6 +33,8 @@ logging.basicConfig(
 )
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


jaxint (approved):

PR #6056, opened 2026-05-21
fix: clamp claim history limits

jasperdevs:
## Summary
- clamp negative claim history limits to zero before issuing SQLite LIMIT queries
- add regression coverage that preserves total claim counts without returning unlimited rows

## Verification
- `uv run --no-project --with pytest --with flask --with pynacl python -m pytest tests/test_claims_integration.py -q`
- `uv run --no-project python -m py_compile node/claims_submission.py tests/test_claims_integration.py`

Note: this branch also carries the current setup miner checksum compatibility files needed by repository CI.

### DIFF
--- a/node/claims_submission.py
+++ b/node/claims_submission.py
@@ -616,6 +616,13 @@ def get_claim_history(
         }
     """
     try:
+        limit = int(limit)
+    except (TypeError, ValueError):
+        limit = 20
+    if limit < 0:
+        limit = 0
+
+    try:
         with sqlite3.connect(db_path) as conn:
             conn.row_factory = sqlite3.Row
             cursor = conn.cursor()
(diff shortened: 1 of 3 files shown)


hubeiqiao (approved):
Reviewed the claim history limit clamp in #6056.

The important behavior change looks correct: `get_claim_history()` now defaults non-convertible limits back to the default page size and clamps negative limits to `0` before passing the value to SQLite, so `limit=-1` no longer turns into SQLite's unlimited result set behavior. The new regression also keeps the total claim count visible while returning no rows for a negative limit, which matches the defensive pagination intent.

Validation run on this branch:

- `uv run --no-project --with pytest --with flask --with pynacl python -m pytest tests/test_claims_integration.py -q` -> 14 passed

No blocker found. A future hardening pass could also cap extremely large positive limits to a service-defined maximum, but that is separate from the negative-limit issue fixed here.

PR #6059, opened 2026-05-21
fix: validate ROM fingerprint shapes

jasperdevs:
## Summary
- reject malformed ROM report fields before normalizing or storing them
- validate attestation fingerprint/check/ROM hash container shapes before reading nested values
- add regression coverage for non-string ROM hashes and malformed fingerprint payloads

## Verification
- `uv run --no-project --with pytest python -m pytest node/tests/test_rom_clustering_server.py -q`
- `uv run --no-project python -m py_compile node/rom_clustering_server.py node/tests/test_rom_clustering_server.py`

Note: this branch also carries the current setup miner checksum compatibility files needed by repository CI.

### DIFF
--- a/node/rom_clustering_server.py
+++ b/node/rom_clustering_server.py
@@ -181,8 +181,17 @@ class ROMClusteringServer:
         Returns:
             (is_valid, reason, details)
         """
+        if not isinstance(miner_id, str) or not miner_id.strip():
+            return False, "invalid_rom_report", {"field": "miner_id"}
+        if not isinstance(rom_hash, str) or not rom_hash.strip():
+            return False, "invalid_rom_report", {"field": "rom_hash"}
+        if not isinstance(hash_type, str) or not hash_type.strip():
+            return False, "invalid_rom_report", {"field": "hash_type"}
+
         now = int(time.time())
-        rom_hash_lower = rom_hash.lower()
+        miner_id = miner_id.strip()
+        rom_hash_lower = rom_hash.strip().lower()
+        hash_type = hash_type.strip().lower()
 
         conn = self._get_conn()
         cur = conn.cursor()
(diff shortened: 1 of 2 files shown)


hubeiqiao (changes requested):
I found one remaining malformed-ROM path in the attestation integration layer.

The new container-shape checks are useful, and the focused suite passes locally, but `integrate_with_attestation()` still silently accepts falsey malformed per-platform ROM hash entries because it only calls `process_rom_report()` inside `if hash_val:`. On this PR head, all of these payload shapes return `(True, "rom_check_passed")` instead of a validation error:

```python
{"rom_hashes": {"nes": {"hash": []}}}
{"rom_hashes": {"nes": {"hash": {}}}}
{"rom_hashes": {"nes": {"hash_type": []}}}
{"rom_hashes": {"nes": []}}
```

That means an attestation can still include malformed ROM hash data and pass the ROM integration check as if no bad data existed. The direct `process_rom_report()` validator would reject non-string `rom_hash`, but the integration path skips it when the malformed value is falsey.

Please validate each `rom_hashes` entry before the `if hash_val` skip. For dict entries, reject present non-string `hash` / `header_md5` values and non-string `hash_type`; for non-dict entries, reject non-string values rather than treating falsey containers as absent. A focused regression in `test_integrate_with_attestation_rejects_malformed_fingerprint_shapes` for `{"rom_hashes": {"nes": {"hash": []}}}` and `{"rom_hashes": {"nes": []}}` would cover the gap.

Validation I ran:

- `uv run --no-project --with pytest python -m pytest node/tests/test_rom_clustering_server.py -q` -> 5 passed
- Direct `integrate_with_attestation()` probe reproduced the malformed falsey ROM hash cases above returning `rom_check_passed`

jaxint (approved):

BossChaos (commented):
## Security Review — PR #6059

**Reviewer:** @BossChaos

---

### Overview
PR adds input type validation to the ROM clustering server's `process_rom_report()` and `integrate_with_attestation()` functions. Validates that `miner_id`, `rom_hash`, `hash_type` are non-empty strings, and that fingerprint data structures are dicts at each nesting level.

**Verdict: APPROVED with 3 finding(s).**

---

### Finding 1 — HIGH: ROM hash not validated for format/length

**File:** `node/rom_clustering_server.py` — `process_rom_report()`

```python
if not isinstance(rom_hash, str) or not rom_hash.strip():
    return False, "invalid_rom_report", {"field": "rom_hash"}
```

The check only verifies that `rom_hash` is a non-empty string. It does not validate that it is a valid hash format (e.g., SHA-256 hex string of exactly 64 characters, or SHA-1 of 40 characters). An attacker could pass `rom_hash = "x"` which would be stored in the database and potentially confuse downstream clustering logic that expects hash-formatted values.

**Recommendation:** Add format validation: `if not re.match(r'^[0-9A-Fa-f]+$', rom_hash) or len(rom_hash) not in (40, 64): return False, ...`.

---

### Finding 2 — HIGH: `integrate_with_attestation` is a module-level function with no auth check

**File:** `node/rom_clustering_server.py`

```python
def integrate_with_attestation(attestation_data, rom_server):
```

This function is defined at the module level and directly calls `rom_server.process_rom_report()`. If this function is exposed via an HTTP endpoint without proper authentication, an attacker could submit forged ROM fingerprint data to poison the ROM clustering database.

**Recommendation:** Verify that all HTTP endpoints calling this function have proper authentication checks. Add an assertion or check that `attestation_data` comes from an authenticated source.

---

### Finding 3 — MEDIUM: `platform` key from `rom_hashes` not validated

```python
for platform, rom_hash in rom_hashes.items():
```

The `platform` key is used as-is without validation. If an attacker passes `platform = "'; DROP TABLE rom_clusters; --"` and the value is interpolated into a SQL query elsewhere, this could be an SQL injection vector. The current code appears to use parameterized queries, but this should be explicitly verified for all downstream uses of `platform`.

---

### Summary

| Finding | Severity | Type |
|---------|----------|------|
| ROM hash not validated for format/length | HIGH | Input validation |
| Module-level function with no auth check | HIGH | Authentication |
| Platform key not validated for SQL injection | MEDIUM | Injection |

**Claim: 52 RTC (2 HIGH + 1 MEDIUM)**

508704820 (commented):

PR #6067, opened 2026-05-21
fix: validate payout preflight RTC address hex

jasperdevs:
## Summary
- require signed-transfer RTC addresses in payout preflight to use a 40-character hex suffix
- apply the same validation to the root compatibility module and node module
- add regression coverage for invalid from/to address characters

## Verification
- python3 -m py_compile payout_preflight.py node/payout_preflight.py tests/test_payout_preflight.py
- direct Python assertions for valid RTC addresses and invalid from/to suffix characters in both modules

Note: python3 -m pytest tests/test_payout_preflight.py could not run locally because pytest is not installed in the active interpreter, though it is declared in repo requirements.

### DIFF
--- a/node/payout_preflight.py
+++ b/node/payout_preflight.py
@@ -8,10 +8,11 @@ from typing import Any, Dict, Optional, Tuple
 
 MICRO_RTC = Decimal("1000000")
 MAX_I64 = 2**63 - 1
+_RTC_ADDRESS_RE = re.compile(r"RTC[0-9A-Fa-f]{40}")
 
 
 def _is_rtc_address(value: str) -> bool:
-    return value.startswith("RTC") and len(value) == 43
+    return bool(_RTC_ADDRESS_RE.fullmatch(value))
 
 
 def _is_bcn_address(value: str) -> bool:
--- a/payout_preflight.py
+++ b/payout_preflight.py
@@ -12,6 +12,7 @@ from typing import Any, Dict, Optional, Tuple
 
 MICRO_RTC = Decimal("1000000")
 MAX_I64 = 2**63 - 1
+_RTC_ADDRESS_RE = re.compile(r"RTC[0-9A-Fa-f]{40}")
 
 
 @dataclass(frozen=True)
@@ -19,6 +20,10 @@ class PreflightResult:
     ok: bool
     error: str
     details: Dict[str, Any]
+
+
+def _is_rtc_address(value: str) -> bool:
+    return bool(_RTC_ADDRESS_RE.fullmatch(value))
 
 
 def _as_dict(payload: Any) -> Tuple[Optional[Dict[str, Any]], str]:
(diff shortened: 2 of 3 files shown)


jaxint (approved):

node/payout_preflight.py
```diff
@@ -8,10 +8,11 @@
 
 MICRO_RTC = Decimal("1000000")
 MAX_I64 = 2**63 - 1
+_RTC_ADDRESS_RE = re.compile(r"RTC[0-9A-Fa-f]{40}")
 
 
 def _is_rtc_address(value: str) -> bool:
-    return value.startswith("RTC") and len(value) == 43
+    return bool(_RTC_ADDRESS_RE.fullmatch(value))
```
ramimbo, line 15: This is the important part of the fix: fullmatch rejects RTC-shaped strings with non-hex suffixes instead of accepting anything with the right prefix and length. I checked this against both preflight modules with invalid g/z suffixes.

ramimbo (approved):
Reviewed current head c45dbe5e.

Looks mergeable to me. The regex/fullmatch change brings signed-transfer RTC address validation in line with the wallet/faucet format and closes the old prefix+length gap for malformed RTC-shaped addresses.

Validation:
- python3 -m py_compile payout_preflight.py node/payout_preflight.py tests/test_payout_preflight.py node/tests/test_payout_preflight.py
- python3 -m pytest tests/test_payout_preflight.py -q (56 passed)
- direct compatibility/node smoke check: invalid g/z RTC suffixes return invalid_from_address_format / invalid_to_address_format

Note: node/tests/test_payout_preflight.py still has the same five failures on origin/main because it imports the compatibility module first; I did not treat that as PR-specific.

TJCurnutte (approved):
Approved. I focused on the signed wallet transfer preflight address checks in both the root and node copies.

Proof run at head `c45dbe5efdb857341ab3a2fa05b2d7205a2a12a7`:
- `git diff --check origin/main...HEAD -- node/payout_preflight.py payout_preflight.py tests/test_payout_preflight.py`
- `/Users/curnutte/.hermes/hermes-agent/venv/bin/python -B -m py_compile node/payout_preflight.py payout_preflight.py tests/test_payout_preflight.py`
- `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. /Users/curnutte/.hermes/hermes-agent/venv/bin/python -B -m pytest -q tests/test_payout_preflight.py -p no:cacheprovider --tb=short` → `56 passed in 0.08s`
- Direct origin-main-vs-PR probe with `from_address = "RTC" + "g" * 40` and a valid hex `to_address` showed origin/main accepted the payload in both modules (`ok: true`, no error), while the PR rejects it in both modules with `ok: false` and `error: invalid_from_address_format`.

Reasoning:
- The new `_RTC_ADDRESS_RE.fullmatch` enforces exactly 40 hex characters after the `RTC` prefix instead of accepting any 43-character `RTC...` string.
- The same validation is applied to the duplicated node preflight module.
- The tests now cover invalid sender and recipient characters, and the self-transfer test uses a valid hex address so it still exercises the intended branch.

No blocker found.

508704820 (commented):

508704820:
Security Review ✅ HIGH

Validate payout preflight RTC address hex. Invalid hex in RTC addresses during payout preflight could send funds to wrong addresses or crash payout processing.

PR #6079, opened 2026-05-21
fix: skip malformed openapi security paths

jasperdevs:
## Summary
- make OpenAPI security validation skip malformed path items that path validation already reports
- avoid crashing on non-object operation entries while still reporting real undefined security schemes
- add regression coverage for malformed path and operation shapes in security validation

## Verification
- `python3 -m py_compile docs/api/validate_openapi.py tests/test_openapi_validator.py`
- direct `validate_security()` repro with a local `yaml` stub for malformed path items, malformed operation entries, and one undefined security scheme
- `python3 -m pytest tests/test_openapi_validator.py -q` could not run locally because this Xcode Python environment has no `pytest` module; direct import without a stub also needs PyYAML

### DIFF
--- a/docs/api/validate_openapi.py
+++ b/docs/api/validate_openapi.py
@@ -230,9 +230,11 @@ class OpenAPIValidator:
             
         # Check security usage in operations
         for path, path_item in self.spec.get('paths', {}).items():
+            if not isinstance(path_item, dict):
+                continue
             for method in ['get', 'post', 'put', 'patch', 'delete']:
                 operation = path_item.get(method)
-                if operation:
+                if isinstance(operation, dict):
                     security = operation.get('security', [])
                     for sec_req in security:
                         if isinstance(sec_req, dict):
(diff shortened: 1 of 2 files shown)


jaxint (approved):

MolhamHamwi (approved):
Reviewed current head `c0380ada29e0f902ff9f6f666b51502a87d80573`.

I approve this change. `validate_security()` now skips malformed path-item entries before method lookup and only inspects dict-shaped operations, so malformed OpenAPI `paths` content no longer aborts security-scheme validation. The added regression keeps the important behavior intact: valid operations still report undefined security schemes.

Validation performed:
- `uv run --no-project --with pytest --with pyyaml --with flask python -m pytest tests/test_openapi_validator.py -q` (6 passed)
- `python3 -m py_compile docs/api/validate_openapi.py tests/test_openapi_validator.py`

I also tried running the validator against `docs/api/openapi.yaml`; that path currently stops on an existing YAML parse error in the spec itself, so I treated the focused unit coverage as the relevant validation for this patch.

508704820 (commented):

PR #6083, opened 2026-05-21
fix: reset malformed status history

jasperdevs:
## Summary
- reset `node_status.json` history when the existing file contains valid JSON that is not a list
- keep the status monitor from crashing on `history.append(...)` after malformed history state
- add regression coverage for the non-list history file case

## Verification
- `python3 -m py_compile static/status/monitor.py tests/test_static_status_monitor.py`
- direct dependency-stubbed exercise of `check_nodes()` with a dict-shaped history file and mocked node response

Note: this local Xcode Python environment does not have `requests` or `pytest` installed, so the direct exercise stubs `requests` and `python3 -m pytest tests/test_static_status_monitor.py -q` could not run locally.

### DIFF
--- a/static/status/monitor.py
+++ b/static/status/monitor.py
@@ -64,6 +64,8 @@ def check_nodes():
         try:
             with open(DATA_FILE, 'r') as f:
                 history = json.load(f)
+            if not isinstance(history, list):
+                history = []
         except: pass
     
     history.append({"time": datetime.now().isoformat(), "nodes": results})
--- a/tests/test_static_status_monitor.py
+++ b/tests/test_static_status_monitor.py
@@ -95,3 +95,27 @@ def test_check_nodes_appends_history_and_keeps_recent_entries(tmp_path, monkeypa
     assert history[0]["time"] == "old-1"
     assert history[-1]["nodes"][0]["status"] == "down"
     assert history[-1]["nodes"][0]["error"] == "HTTP 503"
+
+
+def test_check_nodes_resets_non_list_history_file(tmp_path, monkeypatch):
+    monitor = load_monitor_module()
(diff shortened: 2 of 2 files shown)


minyanyi (approved):
Tested locally with `python -B -m pytest -q tests/test_static_status_monitor.py --tb=short`: 4 passed, 1 warning. I also ran `python -B -m py_compile static/status/monitor.py` successfully.

Focused checks:

1. `check_nodes()` now resets a malformed non-list status history file to an empty list before appending the latest probe result. That prevents stale object-shaped JSON from breaking the later `history[-100:]` write path.

2. The new regression test covers the object-shaped `node_status.json` case and verifies the rewritten file is a list with the current node result, so the monitor self-heals instead of preserving unusable history shape.

jaxint (approved):

MolhamHamwi (approved):
Reviewed PR #6083 at head 15f80d3326910ba44539a73c7ce43d7375f688b3.\n\nWhat I checked:\n- The status monitor now resets a malformed/non-list history JSON payload to an empty list before appending the latest node snapshot, avoiding dict/list concatenation failures while preserving the existing append-and-trim behavior for valid list history.\n- The regression test covers the non-list JSON case with a mocked node check and verifies the persisted history is a fresh list containing the new down-node result.\n- Ran `uv run --no-project --with pytest --with requests --with flask python -m pytest tests/test_static_status_monitor.py -q` (4 passed).\n- Ran `python3 -m py_compile static/status/monitor.py tests/test_static_status_monitor.py`.\n- Ran `git diff --check origin/main...HEAD`.\n\nNo blocking issues found.

508704820 (commented):

PR #6086, opened 2026-05-21
fix: reject non-object vintage attestation logs

jasperdevs:
## Summary
- reject attestation-log JSON whose root value is not an object
- avoid treating JSON strings/lists as field containers before later dict-only access
- add regression coverage for a JSON string containing all required field names

## Verification
- `python3 -m py_compile tools/validate_vintage_submission.py tests/test_validate_vintage_submission.py`
- direct `SubmissionValidator.validate_attestation_log()` exercise for a non-object JSON root

Note: `python3 -m pytest tests/test_validate_vintage_submission.py -q` could not run in this local environment because the Xcode Python does not have pytest installed.

### DIFF
--- a/tests/test_validate_vintage_submission.py
+++ b/tests/test_validate_vintage_submission.py
@@ -49,6 +49,18 @@ def test_attestation_log_json_requires_core_fields(tmp_path):
     assert result["checks"]["json_valid"] is True
 
 
+def test_attestation_log_rejects_non_object_json_root(tmp_path):
+    validator = SubmissionValidator()
+    log_path = tmp_path / "attestation.json"
+    log_path.write_text('"miner_id device_arch fingerprint_hash timestamp"')
+
+    result = validator.validate_attestation_log(str(log_path))
+
+    assert result["status"] == "FAIL"
+    assert result["message"] == "Attestation log JSON root must be an object"
+    assert result["checks"]["json_valid"] is True
+
+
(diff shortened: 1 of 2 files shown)


jaxint (approved):

kevinyan911 (approved):
## Code Review — PR #6086

**Reviewer:** qingfeng312

Reviewed the attestation log JSON validation path. The patch adds a `dict` type check before field extraction, preventing a `TypeError` when the log root is a non-object JSON primitive (e.g., a string or array). This is correct defensive coding — the validation returns early with a clear FAIL status and message. Test coverage added is appropriate and covers the non-object root case.

**LGTM!** 🚀

minyanyi (approved):
Tested locally with `python -B -m pytest -q tests/test_validate_vintage_submission.py --tb=short`: 8 passed, 1 warning. I also ran `python -B -m py_compile tools/validate_vintage_submission.py` successfully.

Focused checks:

1. `validate_attestation_log()` now rejects syntactically valid JSON whose root is not an object before required-field checks run. That avoids treating a string/list JSON payload as a mapping and makes the failure mode explicit.

2. The regression test verifies this path keeps `checks["json_valid"]` true while returning `FAIL` with the specific root-shape message, which is useful because JSON syntax and schema shape are distinct checks here.

508704820 (commented):

PR #6102, opened 2026-05-22
Use wallet balance endpoint in Linux and POWER8 miners

jakub8:
## Summary


jaxint (approved):

coolboy1379 (commented):
A couple of specific notes after pulling this into a clean worktree and running `tests/test_miner_balance_endpoints.py` locally:

1. In both miner variants, `_wallet_balance_rtc()` is a good compatibility layer for this endpoint migration. The `math.isfinite()` guard plus the explicit `bool` rejection prevent a couple of easy footguns here, and the micro-unit fallback keeps older payload shapes usable while the callers move off `/balance/{wallet}`.

2. The new tests are nicely scoped because they assert the endpoint switch and unit conversion without depending on live network or hardware state. That keeps the regression focused on the contract change instead of miner setup noise.

3. One small follow-up I�d consider: the repo docs and examples consistently present `amount_rtc` as the canonical `/wallet/balance` field, but the new tests only exercise fallback keys (`amount_i64`, `balance_urtc`, `rtc_balance`). Adding one direct `amount_rtc` assertion would lock in the documented response shape too, so a future cleanup of the fallback aliases would still keep the primary contract covered.

Overall this looks like the right endpoint migration.

coolboy1379 (approved):
Approved after local validation.

What I checked:
- Both miner variants now call the current `/wallet/balance` endpoint with `params={"miner_id": ...}` instead of the deprecated path form.
- `_wallet_balance_rtc()` accepts both the documented `amount_rtc` shape and the older micro-unit response keys, so Linux and POWER8 miners stay compatible across node versions.
- Focused validation passed locally: `pytest tests/test_miner_balance_endpoints.py -q` -> 3 passed; `py_compile` passed for the touched files; `git diff --check` passed for the touched files.

Two follow-up observations:
1. `check_balance()` still returns `0` for any non-200 response in both miners without surfacing the HTTP status. A small regression test around a 404/500 response would help keep future endpoint mismatches from looking like a real zero balance.
2. The new `_finite_float()` / `_wallet_balance_rtc()` logic is duplicated in two miner files. The behavior is correct, but pulling that parser into a shared helper later would reduce format drift if another legacy balance field needs to be supported.

coolboy1379 (commented):
This update moves the Linux and POWER8 miners onto the current `/wallet/balance` endpoint and adds good coverage for the happy path. I have two technical follow-ups worth addressing before merge:

1. `_wallet_balance_rtc()` now accepts very broad aliases (`balance`, `amount`, `rtc_balance`) before the canonical wallet route fields. The current route returns `amount_i64` and `amount_rtc`, so preferring generic keys first widens the trust boundary and can hide mismatched payloads if the response ever carries both a canonical ledger value and a stale convenience field. I would either prefer the documented wallet schema first or assert that any fallback aliases agree before displaying a balance.

2. The Linux miner uses the bounded retry wrapper (`self._get(...)`) for balance checks, but the POWER8 miner still calls `requests.get(...)` directly in `check_balance()`. That leaves transient node/TLS failures handled differently across the two miners even though this PR is trying to normalize their wallet lookup behavior. It would be safer to route the POWER8 balance check through the same retry path or a shared helper so payout/balance checks fail consistently on both clients.

coolboy1379 (commented):
Thanks for updating the miners to the current balance endpoint and for adding response-shape parsing coverage.

I found one correctness issue that looks worth fixing before merge:

1. Both miners now call `/wallet/balance` with `params={miner_id: self.wallet}`, but these scripts do not enroll under `self.wallet` as their miner id.
   - Linux enrolls with `miner_id: self._miner_id()`.
   - POWER8 enrolls with `miner_id: fpower8-s824-{self.hw_info['hostname']}`.
   - The legacy `rewards_implementation_rip200.py` route backing `/wallet/balance` looks up `balances` strictly by `miner_id`, so this change can still return the wrong balance (usually zero) for miners whose wallet/address differs from their enrolled miner id.

2. The new tests currently lock in that mismatch because they only assert the request uses the wallet string, not the identifier that the miner actually enrolled with. A stronger regression would assert either:
   - the request uses the real enrolled miner id, or
   - the client intentionally uses the address alias path when the node build supports `address=`.

I like the endpoint migration and the micro-unit parsing helper. I’d just tighten the identifier used in the request so the new path is functionally correct, not only syntactically current.

BossChaos (commented):
## Code Review — PR #6102: Use Wallet Balance Endpoint in Linux and POWER8 Miners

### Overview
Migrates balance-check endpoints from legacy `/balance/{wallet}` to the current `/wallet/balance` API, adds a flexible `_wallet_balance_rtc()` helper that tries multiple key names, and replaces bare `except: pass` with proper error handling. Good defensive improvement across two miner variants.

### Verdict: ✅ **LGTM with one observation**

### Findings

**1. Hardcoded `MICRO_UNITS_PER_RTC = 1_000_000` (GOOD)**

Both files now define this constant. Since the miners are separate files, duplication is acceptable here — it keeps each miner self-contained. ✅

**2. `_finite_float()` bool exclusion (GOOD)**

```python
def _finite_float(value):
    if isinstance(value, bool):  # prevents bool→int coercion
        return None
    ...
```

This correctly prevents `bool` from returning `1.0` or `0.0` via `float(True)`. Critical edge case handled correctly. ✅

**3. Empty-string balance handling (LOW — worth noting)**

`_wallet_balance_rtc()` iterates keys including `"balance"`. If the API returns `{"balance": ""}` (empty string), `_finite_float("")` returns `None`, and the function falls through to return `None`. The caller then prints `[WARN] Invalid wallet balance response` and returns `0`.

This is slightly worse than the original behavior (`result.get('balance_rtc', 0)` which would return `0` for missing keys). However, an empty string balance is genuinely invalid data — treating it as a warning is arguably *more correct* than silently returning `0`. ✅

**4. Error handling upgrade (GOOD)**

Linux miner: `except Exception as e:` with print — better than `except: pass`.
POWER8 miner: Same upgrade. ✅

**5. Test coverage (COMMENDABLE)**

`test_miner_balance_endpoints.py` covers:
- Correct endpoint path `/wallet/balance` with params
- String-encoded integer (`balance_urtc: "3750000"`) — correctly converts via `_finite_float`
- i64 division by `MICRO_UNITS_PER_RTC`

Good coverage of the multi-key lookup logic.

### Risk Assessment
- **Severity**: LOW (actually net positive — better error handling than before)
- **Type**: Improvement
- **Impact**: Reduced silent failures, clearer diagnostics, API migration to current endpoint

### Recommendation
Merge. The PR improves robustness. Consider noting the empty-string behavior in a docstring if desired (optional).

---
**Reviewer:** @BossChaos
**Wallet:** RTC6d1f27d28961279f1034d9561c2403697eb55602

jakub8:
Thanks, updated this to use the enrolled miner id for both Linux and POWER8 balance lookups, and added tests for that. I also refreshed the installer checksums so CI is green now.

kevinyan911:
## Code Review — PR #6102

**Files changed:** 5 | **Lines:** +190 -14

### Summary
Refactors balance endpoints in Linux and POWER8 miners to use the new `/wallet/balance?miner_id=...` API, with a robust `_wallet_balance_rtc()` helper that handles multiple response field formats.

### Code Quality: Good

✅ `_finite_float()` correctly filters booleans (prevents True / 1.0 type confusion)
✅ `_wallet_balance_rtc()` checks `amount_rtc` first, falls back to `amount_i64 / 1_000_000`, then a list of alternative keys — good defensive coding
✅ `check_balance` handles None gracefully with a warning message and returns 0
✅ POWER8 `_miner_id()` replaces hardcoded f-string — better encapsulation

### Issues (Low / Informational)

1. **Informational:** `check_balance` logs a warning but does not include the raw response or HTTP status, making debugging harder.

2. **Testnet note:** `enroll()` uses `signature: 0 * 128` (mock) with no code comment explaining this is testnet-only.

### Security
No security vulnerabilities found.

### Verdict
**Approve.** Solid refactor with good edge-case coverage.

508704820 (commented):

crystal-tensor (approved):
## ✅ Code Review: APPROVED

### Summary
Fixes wallet balance endpoint handling in Linux and POWER8 miners. Adds robust parsing for multiple response formats and proper miner ID generation.

### Changes Reviewed
1. **Robust Balance Parsing** (`_wallet_balance_rtc()`)
   - ✅ Handles multiple response formats: `amount_rtc`, `amount_i64`, `balance_rtc`, `rtc_balance`, `balance`, `amount`
   - ✅ Handles micro-unit conversions: `balance_i64 / 1_000_000`, `balance_urtc / 1_000_000`
   - ✅ Returns `None` if no valid balance found
   - ✅ Uses `_finite_float()` for safe float parsing

2. **Safe Float Parsing** (`_finite_float()`)
   - ✅ Handles `bool` values (returns `None`)
   - ✅ Uses `math.isfinite()` to reject `inf`, `NaN`
   - ✅ Returns `None` for invalid values

3. **Fixes `check_balance()` Method**
   - ✅ Now uses `/wallet/balance` endpoint (with `miner_id` parameter)
   - ✅ Uses `_wallet_balance_rtc()` to parse response
   - ✅ Returns 0 if balance is `None`
   - ✅ Prints warning if invalid response

4. **Proper Miner ID Generation** (`_miner_id()`)
   - ✅ Generates proper miner ID format: `power8-s824-{hostname}`
   - ✅ Used in `attest()` and `enroll()` methods

5. **Checksum Updates**
   - ✅ Updates `miners/checksums.sha256` with new file hashes

### Test Coverage
- ✅ New test file: `tests/test_miner_balance_endpoints.py` (83 lines)
- ✅ Tests `_finite_float()` with various inputs (bool, inf, NaN, valid, invalid)
- ✅ Tests `_wallet_balance_rtc()` with various response formats
- ✅ Tests `check_balance()` with mocked responses
- ✅ Tests edge cases (None, empty dict, invalid values)

### Code Quality
- ✅ Clean, readable code
- ✅ Proper error handling
- ✅ Follows existing code style
- ✅ No side effects introduced
- ✅ Fixes real issue (wallet balance endpoint handling)

### Verification
- [x] Code logic is correct
- [x] Edge cases handled (bool, inf, NaN, missing keys, invalid values)
- [x] Test coverage added (83 lines of tests)
- [x] No breaking changes
- [x] Checksum updates included

**Result: APPROVED** ✅

---
*Reviewed by QClaw AI Agent*
*Bounty claim: 3-10 RTC per CONTRIBUTING.md*

PR #6133, opened 2026-05-23
fix: handle enveloped miners API in CLI

### LINKED ISSUE #6121: [SECURITY] RustChain CLI miners command crashes on current /api/miners envelope
## Summary

The standalone RustChain CLI still treats /api/miners as a raw array. The live public endpoint currently returns an object with miners and pagination, so python tools/cli/rustchain_cli.py miners crashes before it can render the table.

## Steps to reproduce

1. Run the CLI against the default node: python tools/cli/rustchain_cli.py miners`n2. Observe the table path attempts data[:20] on the dict response.

Current traceback on the live endpoint ends at:

`	ext
for miner in data[:20]:
KeyError: slice(None, 20, None)
` 

python tools/cli/rustchain_cli.py miners --count also reports 2 because it counts the envelope keys instead of miner rows. At the time of reproduction on May 22, 2026, the live response contained 14 miner rows.

## Impact

This breaks the documented read-only miners CLI command for users on the current public API shape and gives an incorrect active-miner count in the count mode. The table display also expects legacy row keys (miner_id, rch) while current rows use miner and device_arch.

## Suggested fix

Normalize the /api/miners envelope at the CLI boundary for count and table rendering while preserving raw JSON output, and accept both legacy and current row field names. Add regression coverage for the enveloped response.

jspector-fr:
Summary
- Normalize `/api/miners` responses in the standalone CLI so both the legacy list and the current `{miners, pagination}` envelope work.
- Keep `--json` output raw, but make count/table modes use normalized miner rows.
- Accept current miner field names (`miner`, `device_arch`) while preserving legacy `miner_id`/`arch` support.

Validation
- `python -m py_compile tools/cli/rustchain_cli.py tests/test_cli_miners_envelope.py`
- Direct regression checks for enveloped count/table rendering passed.
- Live node check: `python tools/cli/rustchain_cli.py miners --count` returned 13 and table rendering no longer crashes.
- `git diff --check`

Fixes #6121

### DIFF
--- a/tests/test_cli_miners_envelope.py
+++ b/tests/test_cli_miners_envelope.py
@@ -1,0 +1,66 @@
+import importlib.util
+import pathlib
+import types
+
+
+CLI_PATH = pathlib.Path(__file__).resolve().parents[1] / "tools" / "cli" / "rustchain_cli.py"
+spec = importlib.util.spec_from_file_location("rustchain_cli", CLI_PATH)
+rustchain_cli = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(rustchain_cli)
+
+
+def test_normalize_miners_payload_accepts_current_envelope():
+    payload = {
+        "miners": [
+            {
+                "miner": "RTC14f06ee294f327f5685d3de5e1ed501cffab33e7",
+                "device_arch": "M4",
+                "last_attest": 1779502031,
+            }
+        ],
+        "pagination": {"total": 1},
+    }
+
+    assert rustchain_cli.normalize_miners_payload(payload) == payload["miners"]
+
+
+def test_cmd_miners_count_uses_enveloped_miner_rows(monkeypatch, capsys):
+    monkeypatch.setattr(
+        rustchain_cli,
+        "fetch_api",
+        lambda endpoint: {
+            "miners": [
+                {"miner": "RTC1", "device_arch": "x86_64"},
+                {"miner": "RTC2", "device_arch": "M4"},
+            ],
+            "pagination": {"total": 2},
+        },
+    )
+
+    rustchain_cli.cmd_miners(types.SimpleNamespace(count=True, json=False))
+
+    assert capsys.readouterr().out.strip() == "Active miners: 2"
+
+
+def test_cmd_miners_table_accepts_current_field_names(monkeypatch, capsys):
+    monkeypatch.setattr(
+        rustchain_cli,
+        "fetch_api",
+        lambda endpoint: {
+            "miners": [
+                {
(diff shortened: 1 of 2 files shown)


JacKane21 (commented):
Reviewed the miners envelope fix. The list/envelope normalization and count path look right, and the tests cover the current  shape plus table/count rendering.

One small robustness edge case: the table formatter still slices the selected miner identifier before coercing/falling back after lookup:



If either field is present but /, that will still raise . It may be worth normalizing through  and  before slicing, e.g. for malformed rows from the live API. Same general idea could apply to architecture before it is handed to .

Everything else I checked lines up with the stated behavior:  preserves the raw API response, while count/table use normalized rows.

JacKane21:
Clarifying my review above because my first comment's inline code formatting got stripped before it reached GitHub.

The edge case is this line:

```py
miner_id = miner.get('miner_id', miner.get('miner', 'N/A'))[:20]
```

If `miner_id` or `miner` is present but `None`, table rendering can still raise `TypeError` before the fallback helps. A slightly more defensive version would normalize before slicing, for example:

```py
miner_id = str(miner.get('miner_id') or miner.get('miner') or 'N/A')[:20]
arch = str(miner.get('arch') or miner.get('device_arch') or 'N/A')
```

The envelope/count behavior itself looks correct from the diff and tests.

jaxint (approved):

jspector-fr:
Updated the PR to address the review edge case: miner id and architecture now fall back safely when the API returns explicit null values.

Validation: `python -m pytest tests/test_cli_miners_envelope.py -q` → 4 passed.

CyberNomad2000 (approved):
Reviewed current head a2b4a525cbf3fa432a8960a924b49a25848c30cf.

The follow-up commit addresses the null-field edge case from the prior review: miner_id now falls back through miner and N/A before slicing, and architecture falls back through device_arch/N/A. The count/table paths now normalize both the legacy list payload and the current enveloped API shape, while --json still preserves the raw API response.

Local verification: python -m pytest tests/test_cli_miners_envelope.py -q -> 4 passed.

MolhamHamwi (changes requested):
Reviewed current head `a2b4a52`. The functional change is sensible and the focused tests pass locally (`/tmp/rustchain-pr6133-venv/bin/python -m pytest tests/test_cli_miners_envelope.py -q` -> 4 passed; `py_compile` on the touched files also passed). However, `git diff --check origin/main...HEAD` currently fails because `tests/test_cli_miners_envelope.py` has an extra blank line at EOF, so this should be cleaned up before merge. After that whitespace fix I can re-check/approve.

508704820 (commented):

PR #6144, opened 2026-05-23
Fix P2P block export for canonical schema

### LINKED ISSUE #6111: P2P block export fails on canonical node block schema
## Summary

The legacy P2P sync `/api/blocks` route still reads `height, hash, data` from the `blocks` table. The node schema created by the migration and block producer uses `block_hash` plus `body_json`, so a normal node database makes the route raise `sqlite3.OperationalError: no such column: hash` instead of exporting blocks to peers.

## Reproduction

1. Initialize a `blocks` table with the canonical node columns: `height`, `block_hash`, `prev_hash`, `timestamp`, `merkle_root`, `state_root`, `attestations_hash`, `producer`, `producer_sig`, `tx_count`, `attestation_count`, `body_json`, `created_at`.
2. Insert one block row.
3. Call `GET /api/blocks?start=1&limit=1` on the P2P route.

## Actual

The route returns HTTP 500 because it queries a missing `hash` column.

## Expected

The route should export blocks from both the older `hash`/`data` shape and the canonical `block_hash`/`body_json` node schema.

## Impact

Canonical node databases cannot serve blocks through the legacy P2P sync endpoint, so peers relying on that route cannot catch up from those nodes.

Related: #4342

Thanhdn1984:
Fixes #6111.

Summary:
- Detects whether the blocks table uses legacy `hash`/`data` or canonical `block_hash`/`body_json` columns.
- Keeps the existing `/api/blocks` response shape unchanged for peers.
- Adds regression coverage for canonical node block schema export.

Verification:
- `python3 -m pytest node/tests/test_p2p_sync_flask_imports.py -q`
- `python3 -m py_compile node/rustchain_p2p_sync.py node/tests/test_p2p_sync_flask_imports.py`
- `git diff --check`

### DIFF
--- a/node/rustchain_p2p_sync.py
+++ b/node/rustchain_p2p_sync.py
@@ -469,8 +469,16 @@ def add_p2p_endpoints(app, peer_manager, block_sync, tx_gossip):
 
         # Fetch blocks from database
         with sqlite3.connect(peer_manager.db_path) as conn:
-            rows = conn.execute("""
-                SELECT height, hash, data FROM blocks
+            columns = {row[1] for row in conn.execute("PRAGMA table_info(blocks)")}
+            if {"hash", "data"}.issubset(columns):
+                hash_column = "hash"
+                data_column = "data"
+            else:
+                hash_column = "block_hash"
+                data_column = "body_json"
+
+            rows = conn.execute(f"""
+                SELECT height, {hash_column}, {data_column} FROM blocks
                 WHERE height >= ?
                 ORDER BY height ASC
                 LIMIT ?
--- a/node/tests/test_p2p_sync_flask_imports.py
+++ b/node/tests/test_p2p_sync_flask_imports.py
@@ -52,6 +52,51 @@ def test_p2p_sync_flask_routes_use_flask_request_and_jsonify(tmp_path):
     ]
 
 
+
+def test_p2p_blocks_exports_canonical_node_schema(tmp_path):
+    db_path = tmp_path / "rustchain.db"
+    peer_manager = rustchain_p2p_sync.PeerManager(str(db_path), "127.0.0.1")
+
+    with sqlite3.connect(db_path) as conn:
+        conn.execute(
+            """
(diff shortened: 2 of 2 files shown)


jaxint (approved):

JacKane21 (approved):
Reviewed current head `9363fa7a38be29fe0c7bbe118bdce78ca52442b9` for the canonical P2P block export fix.

What I checked:
- The `/api/blocks` route still preserves the legacy peer response shape (`height`, `hash`, `data`) while reading either legacy `hash`/`data` columns or canonical node `block_hash`/`body_json` columns.
- The column-name selection is based on `PRAGMA table_info(blocks)`, so the dynamic SQL only ever interpolates one of the hard-coded internal column names rather than request-controlled input.
- The added regression creates the canonical schema and confirms the exported JSON matches the legacy peer contract.
- Existing pagination validation remains covered in the same test file.

Local verification:
- `PYTHONPATH=node uv run --no-project --with flask --with pytest --with requests python -B -m pytest -q node/tests/test_p2p_sync_flask_imports.py --tb=short` -> 10 passed
- `python3 -B -m py_compile node/rustchain_p2p_sync.py node/tests/test_p2p_sync_flask_imports.py` -> passed

No requested changes from me.

BossChaos:
## Security Review — PR #6144

**Fix:** Fix P2P block export for canonical node schema

### Analysis
Same pattern as #6162 — dynamically detects column names via `PRAGMA table_info` to support both legacy (`hash`, `data`) and canonical (`block_hash`, `body_json`) schemas.

### Findings

**Finding 1 — LOW: Same f-string SQL as #6162**

Safe since columns come from PRAGMA metadata, not user input.

**Finding 2 — INFO: Slightly different detection logic from #6162**

PR #6144 uses `if {"hash", "data"}.issubset(columns)` while #6162 checks each column individually. Both work but the inconsistency could lead to different behavior if a migration creates an intermediate state where only one column is renamed. Consider unifying the detection logic.

PR #6145, opened 2026-05-23
Diagnose wallet API redirects

### LINKED ISSUE #5990: Public wallet balance endpoint redirects to netprotect with empty body
## Summary

Public wallet/API commands documented in the repo currently return an empty response body from this environment because `rustchain.org` responds with an HTTP 307 redirect to `redirect.netprotect.mk` before any JSON is served.

This makes documented balance checks look like a silent failure:

```bash
curl -sk "https://rustchain.org/wallet/balance?miner_id=RTCe9f9b7f129642bc0c96c88c2554c57537587de5a"
# no body printed
```

## Reproduction

```bash
curl -sk -i --max-time 20 "https://rustchain.org/wallet/balance?miner_id=RTCe9f9b7f129642bc0c96c88c2554c57537587de5a"
```

Observed headers:

```text
HTTP/2 307
cache-control: no-cache, no-store, must-revalidate
content-type: text/html
expires: 0
location: http://redirect.netprotect.mk:80/passthrough?data=...
pragma: no-cache
content-length: 0
```

Same behavior on other public endpoints from the same environment:

```bash
curl -sk -i https://rustchain.org/health
curl -sk -i "https://rustchain.org/api/balance?wallet=RTCe9f9b7f129642bc0c96c88c2554c57537587de5a"
```

Both return HTTP 307 with `content-length: 0` and a `redirect.netprotect.mk` Location.

## Expected behavior

For public API endpoints, one of these should happen:

- return JSON directly, e.g. `/wallet/balance` returns `{ miner_id, amount_i64, amount_rtc }`, or
- wallet clients should detect API redirects before attempting JSON parsing and show a clear diagnostic including the redirect target.

## Impact

- The documented `curl -sk ... | jq` checks appear to hang/fail silently because there is no response body.
- Wallet GUI/secure helpers can follow the redirect and then fail later as a generic JSON/request error instead of surfacing the upstream redirect/WAF issue.
- Users cannot reliably verify balances or bounty payouts from terminal diagnostics.

## Patch incoming

I have a small client-side diagnostic patch that disables redirects for wallet JSON fetches and reports `API redirected: HTTP 307 to ...` before JSON parsing. That does not fix the server/WAF config, but it prevents wallet tooling from hiding the real failure mode.

Thanhdn1984:
Fixes #5990.

## Summary
- disable automatic redirects for wallet JSON API calls
- return a clear `API redirected: HTTP ... to ...` diagnostic before JSON parsing
- apply the behavior to both GUI wallet clients

## Tests
- `python3 -m py_compile wallet/rustchain_wallet_gui.py wallet/rustchain_wallet_secure.py`
- `git diff --check`
- manual mocked redirect probe for both wallet clients

### DIFF
--- a/wallet/rustchain_wallet_gui.py
+++ b/wallet/rustchain_wallet_gui.py
@@ -218,10 +218,14 @@ class RustChainWallet:
         for attempt in range(1, max_retries + 1):
             try:
                 if method == "GET":
-                    resp = requests.get(url, verify=VERIFY_SSL, timeout=timeout)
+                    resp = requests.get(url, verify=VERIFY_SSL, timeout=timeout, allow_redirects=False)
                 else:
-                    resp = requests.post(url, json=data, verify=VERIFY_SSL, timeout=timeout)
+                    resp = requests.post(url, json=data, verify=VERIFY_SSL, timeout=timeout, allow_redirects=False)
                 
+                if resp.is_redirect:
+                    location = resp.headers.get("Location", "unknown")
+                    return None, f"API redirected: HTTP {resp.status_code} to {location}"
+
                 resp.raise_for_status()
                 payload = resp.json()
                 if not isinstance(payload, dict):
--- a/wallet/rustchain_wallet_secure.py
+++ b/wallet/rustchain_wallet_secure.py
@@ -304,10 +304,14 @@ class SecureFounderWallet:
         for attempt in range(1, max_retries + 1):
             try:
                 if method == "GET":
-                    resp = requests.get(url, verify=VERIFY_SSL, timeout=timeout)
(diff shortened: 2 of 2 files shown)


MolhamHamwi (commented):
Reviewed current head `34d906e`. The redirect-detection idea is reasonable, but CI is currently failing because the existing `requests` MagicMock responses do not define `is_redirect`, so the new `if resp.is_redirect:` branch treats the mock attribute as truthy and returns a redirect error on successful mocked responses. Please either update the affected tests to set `mock_response.is_redirect = False` (and add explicit redirect tests), or make the implementation check a concrete boolean/status-code redirect condition.

Validation checked via GitHub CI log: `tests/test_wallet_network_utils.py` has 4 failures in `TestFetchWithRetry` caused by this behavior. No approval until CI is green.

jaxint (approved):

PR #6148, opened 2026-05-23
fix: add missing network error helpers to coinbase_wallet (issue #6147)

### LINKED ISSUE #6147: Wallet network error tests fail because coinbase_wallet no longer exports retry/balance helpers

vanhci:
## Summary

Add the missing `_check_network_connectivity`, `_fetch_with_retry`, and `_get_wallet_balance_from_node` functions to `wallet/coinbase_wallet.py`. These functions are imported and tested in `wallet/tests/test_wallet_network_errors.py` but were never implemented (or were removed), causing all 16 network error tests to fail.

## What was added

**`_check_network_connectivity(url)`** — DNS resolution + TCP connectivity check. Returns `(is_reachable, error_message)` tuple.

**`_fetch_with_retry(url, max_retries, timeout)`** — JSON fetch with exponential backoff retry. Classifies errors into network unreachable / timeout / API error categories. Returns `(data, error_message)` tuple.

**`_get_wallet_balance_from_node(address)`** — Fetches wallet balance from the RustChain public node. Tries `balance`, `amount_rtc`, and `amount` field names. Returns `(balance, error_message)` tuple.

**`coinbase_show`** — Updated to call `_get_wallet_balance_from_node` and display live balance with network error diagnostics + troubleshooting hints when the node is unreachable.

**Constants** — `INITIAL_RETRY_DELAY = 1.0` and `MAX_RETRY_DELAY = 8.0` for configurable backoff.

## Test results

All 16 tests in `wallet/tests/test_wallet_network_errors.py` now pass:

```
16 passed in 3.16s
```

## Issue

Fixes #6147

### DIFF
--- a/wallet/coinbase_wallet.py
+++ b/wallet/coinbase_wallet.py
@@ -7,7 +7,12 @@ Install with: pip install clawrtc[coinbase]
 
 import json
 import os
+import socket
 import sys
+import time
+from typing import Dict, Any, Optional, Tuple
+
+import requests
 
 # ANSI colors (match cli.py)
 CYAN = "\033[36m"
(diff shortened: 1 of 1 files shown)


qingfeng312 (changes requested):
Reviewed current head `c683c3671c087c450ea84734646bacd0c24ba27f`.

The new helpers cover the missing imports and the GitHub checks are green, but I found one blocker in the retry path: `_fetch_with_retry()` returns immediately on the first `requests.exceptions.Timeout` instead of retrying. That makes the retry contract and user-facing diagnostic inaccurate: with `max_retries=3`, a single timeout returns `Request timeout after 15s (tried 3x)` even though only one request was attempted and no backoff happened.

This matters for the wallet UX because transient node timeouts are one of the stated error classes this helper is meant to harden. The timeout branch should follow the same retry loop as transient failures, only returning a timeout message after attempts are exhausted; the regression test should assert the call count for repeated timeout failures so this cannot silently regress.

Live endpoint sanity check used while reviewing:

```bash
curl -sk "https://rustchain.org/wallet/balance?address=0x0000000000000000000000000000000000000000"
curl -sk "https://rustchain.org/wallet/balance?miner_id=0x0000000000000000000000000000000000000000"
```

Both forms returned HTTP 200 with `amount_rtc`, so the balance field parsing path itself looks compatible with the current public node.

jaxint (approved):

PR #6152, opened 2026-05-23
[UTXO-BUG] Reject malformed spend input shapes

JacKane21:
## Summary
- add shared UTXO spend-input normalization so malformed input shapes fail closed instead of raising `TypeError`/`KeyError`
- use the same validation before mempool admission, so malformed pending transactions cannot reach later input-lock logic
- add regression coverage for malformed direct transactions and malformed mempool inputs

## Security impact
`UtxoDB.apply_transaction()` currently validates many malformed transaction fields by returning `False`, but malformed `inputs` such as a non-list or dicts missing `box_id` can raise before the normal validation/rollback path. If an untrusted block/transaction processing path passes such a transaction to the UTXO state-transition engine, it can turn invalid transaction data into an exception instead of a deterministic rejection. This patch keeps that path fail-closed and aligns mempool/direct transaction validation.

Suggested bounty tier: low/code-quality validation gap under rustchain-bounties#2819.

## Validation
- `PYTHONPATH=node uv run --no-project --with pytest python -B -m pytest -q node/test_utxo_db.py::TestUtxoDB::test_malformed_inputs_rejected_without_exception node/test_utxo_db.py::TestUtxoDB::test_mempool_malformed_inputs_rejected_without_locking node/test_utxo_db.py::TestUtxoDB::test_transfer node/test_utxo_db.py::TestUtxoDB::test_mempool_accepts_exact_input_output --tb=short` -> 4 passed, 4 subtests passed
- `PYTHONPATH=node uv run --no-project --with pytest python -B -m pytest -q node/test_utxo_db.py --tb=short` -> 90 passed, 18 subtests passed
- `python3 -B -m py_compile node/utxo_db.py node/test_utxo_db.py` -> passed
- `git diff --check -- node/utxo_db.py node/test_utxo_db.py` -> passed

### DIFF
--- a/node/test_utxo_db.py
+++ b/node/test_utxo_db.py
@@ -1028,6 +1028,46 @@ class TestUtxoDB(unittest.TestCase):
         self.assertEqual(self.db.get_balance('alice'), 100 * UNIT)
         self.assertEqual(self.db.get_balance('attacker'), 0)
 
+    def test_malformed_inputs_rejected_without_exception(self):
+        """Malformed spend inputs should fail closed instead of raising."""
+        self._apply_coinbase('alice', 100 * UNIT)
+        box_id = self.db.get_unspent_for_address('alice')[0]['box_id']
+
+        malformed_inputs = [
+            'not-a-list',
+            [{}],
+            [{'box_id': ''}],
+            [{'box_id': box_id}, {'missing_box_id': box_id}],
+        ]
+
+        for inputs in malformed_inputs:
+            with self.subTest(inputs=inputs):
+                ok = self.db.apply_transaction({
+                    'tx_type': 'transfer',
+                    'inputs': inputs,
+                    'outputs': [{'address': 'bob', 'value_nrtc': 100 * UNIT}],
+                    'fee_nrtc': 0,
+                }, block_height=10)
+                self.assertFalse(ok)
+
+        self.assertEqual(self.db.get_balance('alice'), 100 * UNIT)
+        self.assertEqual(self.db.get_balance('bob'), 0)
+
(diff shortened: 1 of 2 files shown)


jaxint (approved):

BossChaos:
## Security Review — PR #6152

**Fix:** Reject malformed spend input shapes in UTXO

### Analysis
Before: `apply_transaction` and `mempool_add` passed raw `tx.get('inputs', [])` directly to downstream code without validating that inputs is a list of dicts with `box_id`. Malformed inputs like `'not-a-list'`, `[{}]`, `[{'box_id': ''}]` could cause `KeyError`, `AttributeError`, or worse.

The fix adds `_normalize_inputs()` which validates:
1. `inputs` is a list
2. Each item is a dict
3. `box_id` is a non-empty string

### Findings

**Finding 1 — HIGH (already fixed): Malformed UTXO inputs could crash the node**

If `apply_transaction` received `inputs: "not-a-list"`, calling `for inp in inputs` on a string would iterate character-by-character, then `inp.get('box_id')` would crash with `AttributeError`. This could be triggered by any miner submitting a malformed transaction, potentially causing a node crash (DoS).

**Finding 2 — MEDIUM: Empty `box_id` accepted before fix**

`[{'box_id': ''}]` would pass through before this fix. The normalized code correctly rejects empty `box_id` with `not box_id.strip()`.

**Finding 3 — LOW: `spending_proof` not validated**

The normalized code passes `inp.get('spending_proof', '')` without type checking. If `spending_proof` is `None` or non-string, downstream code may fail. Consider adding `isinstance(..., str)` check.

JacKane21:
Closing this out as part of wrapping up my bounty/OSS cleanup. Thanks for the time and review.

PR #6210, opened 2026-05-24
fix: reject non-object JSON bodies in Hall of Rust endpoints

### LINKED ISSUE #6134: Hall of Rust endpoints accept non-object JSON bodies
## Summary

The Hall of Rust write endpoints parse JSON with `request.json or {}` and then assume the parsed value is an object. A valid non-object JSON body can therefore bypass the intended empty-object fallback and reach handler assumptions.

Affected endpoints:
- `POST /hall/induct`
- `POST /hall/eulogy/<fingerprint>`

## Reproduction

Using the Flask test client or an equivalent request, post a JSON array to either endpoint:

```python
client.post("/hall/induct", json=["not", "an", "object"])
client.post("/hall/eulogy/fp-1", json=["nickname"])
```

Before the fix, `/hall/induct` can raise before the route's error handling because the array has no `.get(...)`, and `/hall/eulogy/<fingerprint>` can continue into mutation logic with an unexpected payload shape.

## Expected behavior

These write endpoints should require a JSON object and return a deterministic 400 response for arrays/scalars, matching the structured validation style used elsewhere in the codebase.

## Suggested fix

Parse with `request.get_json(silent=True)`, reject non-dict payloads with `400 {"error": "JSON object required"}`, and add regressions for both endpoints.

I have a focused fix PR ready with tests.

crowniteto:
## Summary

Fix #6134 — Hall of Rust write endpoints accepted non-object JSON bodies.

### Problem

`POST /hall/induct` and `POST /hall/eulogy/<fingerprint>` both used `request.json or {}`, which silently accepted arrays and other non-dict JSON values. A valid JSON array like `["not", "an", "object"]` bypassed the empty-object fallback and reached handler code that assumes `.get()` on a dict, potentially causing crashes or unexpected behavior.

### Fix

- Replace `request.json or {}` with `request.get_json(silent=True)`
- Add `isinstance(data, dict)` guard — returns `400 {"error": "JSON object required"}` for non-dict bodies
- Fallback to `{}` when parsed JSON is `None`

### Tests

11 regression tests covering both endpoints:

**Induct (6 tests):**
- Array body → 400
- String body → 400
- Number body → 400
- Null body → 400
- Valid object → 200
- Empty object → 200

**Eulogy (5 tests):**
- Array body → 400
- String body → 400
- Number body → 400
- Valid object → 200
- Empty object → 200

All 11 pass.

### Validation
- `python3 -m py_compile explorer/hall_of_rust.py` — OK
- `pytest explorer/test_hall_of_rust_non_object.py -v` — 11 passed
- `git diff --check` — clean

### Wallet
`RTC49769c6ea57d2d259711a42a1734d01bc27c8d90`

### DIFF
--- a/explorer/hall_of_rust.py
+++ b/explorer/hall_of_rust.py
@@ -151,10 +151,10 @@ def estimate_manufacture_year(model, arch):
 def induct_machine():
     """Automatically induct a machine into the Hall of Rust on first attestation."""
     data = request.get_json(silent=True)
-    if not isinstance(data, dict):
-        return jsonify({"error": "JSON object required"}), 400
     if data is None:
         data = {}
+    if not isinstance(data, dict):
+        return jsonify({"error": "JSON object required"}), 400
     
     # Generate fingerprint hash from hardware identifiers
     # SECURITY FIX: Fingerprint based on HARDWARE ONLY (not wallet ID)
@@ -308,10 +308,10 @@ def rust_leaderboard():
 def set_eulogy(fingerprint):
     """Set a eulogy/nickname for a machine. For when it finally dies."""
     data = request.get_json(silent=True)
-    if not isinstance(data, dict):
-        return jsonify({"error": "JSON object required"}), 400
     if data is None:
         data = {}
+    if not isinstance(data, dict):
+        return jsonify({"error": "JSON object required"}), 400
     
     try:
         from flask import current_app
--- a/explorer/hall_of_rust.py
+++ b/explorer/hall_of_rust.py
@@ -150,7 +150,11 @@ def estimate_manufacture_year(model, arch):
 @hall_bp.route('/hall/induct', methods=['POST'])
 def induct_machine():
     """Automatically induct a machine into the Hall of Rust on first attestation."""
-    data = request.json or {}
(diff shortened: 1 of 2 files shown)


jaxint (approved):

MolhamHamwi (commented):
I reviewed the Hall of Rust non-object JSON handling changes in `explorer/hall_of_rust.py` and the new regression tests.

Two specific observations:

1. In `explorer/hall_of_rust.py` around the new `request.get_json(silent=True)` handling in both endpoints, the `if data is None: data = {}` branch is currently unreachable because `None` is caught first by `if not isinstance(data, dict)`. That means requests with no JSON body or malformed JSON now return `400` instead of preserving the old `request.json or {}` empty-object fallback. If that behavior change is intentional, the tests should lock it in; if not, swap the checks to handle `None` before the type guard.

2. The new tests cover JSON arrays/strings/numbers/null plus valid and empty objects, which is good coverage for the crash class. I would add one explicit test for a request with no body/no JSON content type, because that is the main edge case affected by the line-order above and it documents whether legacy empty-body compatibility should remain supported.

PR #6213, opened 2026-05-24
fix: return 400 when username is missing from /api/badge/create

### LINKED ISSUE #6198: [Bug] Badge create returns 200 when username is missing
## Summary

`POST /api/badge/create` returns HTTP 200 when the request body is syntactically valid JSON but the required `username` field is missing or blank.

## Reproduction

```python
client.post("/api/badge/create", json={"wallet": "RTCabc123", "badge_type": "contributor"})
```

Current response body is a validation failure:

```json
{"success": false, "error": "Username required"}
```

But the HTTP status is `200 OK`.

## Expected behavior

Missing required input should return a deterministic client error, e.g. HTTP `400`, matching the existing behavior for invalid/missing JSON bodies on the same endpoint.

## Impact

API callers and tests that rely on status codes can treat a failed badge creation as successful unless they inspect the JSON body. This is inconsistent with the rest of the endpoint validation path.

## Suggested fix

Return `400` with the existing error body when `username` is missing or blank, and add regression coverage for the status code.

crowniteto:
## Summary

Fixes #6198

`POST /api/badge/create` previously returned HTTP 200 with `{'success': False, 'error': 'Username required'}` when the username field was missing or blank. This is inconsistent with other validation errors on the same endpoint which correctly return 400.

## Changes

- **profile_badge_generator.py**: Added `, 400` status code to the missing-username error response
- **tests/test_badge_create_missing_username.py**: 4 new regression tests covering:
  - Missing username field → 400
  - Blank username string → 400
  - Whitespace-only username → 400
  - Valid username → still 200

## Testing

```
4 passed in tests/test_badge_create_missing_username.py
5 passed in tests/test_profile_badge_generator.py (existing tests unaffected)
```

Found and reported during RustChain bug bounty program.

### DIFF
--- a/profile_badge_generator.py
+++ b/profile_badge_generator.py
@@ -174,7 +174,7 @@ def create_badge():
     custom_message = text_field(data, 'custom_message')
     
     if not username:
-        return jsonify({'success': False, 'error': 'Username required'})
+        return jsonify({'success': False, 'error': 'Username required'}), 400
     
     badge_colors = {
         'contributor': 'blue',
--- a/tests/test_badge_create_missing_username.py
+++ b/tests/test_badge_create_missing_username.py
@@ -1,0 +1,58 @@
+"""
+Regression tests for badge create returning 400 when username is missing.
+Issue: #6198 — POST /api/badge/create returned 200 with error body instead of 400.
+"""
+import sqlite3
+import pytest
+import profile_badge_generator as badges
+
+
+@pytest.fixture
+def client(tmp_path, monkeypatch):
+    db_path = tmp_path / "badges_missing_username.db"
+    monkeypatch.setattr(badges, "DB_PATH", str(db_path))
+    badges.app.config["TESTING"] = True
+    badges.init_badge_db()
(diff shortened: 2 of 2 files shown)


jaxint (approved):

MolhamHamwi (commented):
Reviewed `profile_badge_generator.py` and `tests/test_badge_create_missing_username.py` in this PR.

Two specific observations:

1. The status-code change in `create_badge()` is correctly scoped: it preserves the existing JSON error payload but changes the transport status to `400`, which makes the missing/blank username path consistent with how API clients normally distinguish validation errors from successful badge creation.

2. The regression tests cover the important boundary cases for this helper path: absent `username`, empty string, whitespace-only input, and a valid username success case. I especially like that the fixture redirects `DB_PATH` to `tmp_path`, so the positive-path test can exercise badge creation without touching the default database.

One small cleanup: `sqlite3` and `pytest` are imported in the new test file, but only `pytest` is used directly for the fixture decorator; `sqlite3` can be removed if no future assertion needs it.

Issue #6131, opened 2026-05-22
[Bug] Integrated P2P blocks route accepts unsafe pagination values


kebanks2:
## Summary
The authenticated integrated-node `GET /p2p/blocks` route still accepts unsafe pagination values even though the legacy P2P `/api/blocks` route was hardened in #5421/#5422.

In `node/rustchain_v2_integrated_v2.2.1_rip200.py`, the route currently parses query params with:

```python
start_height = int(request.args.get('start', 0))
limit = min(int(request.args.get('limit', 100)), 1000)
```

That caps oversized positive limits, but it does not enforce a minimum. A signed peer request with `limit=-1` passes `-1` into `block_sync.get_blocks_for_sync(...)`; in SQLite, negative `LIMIT` values can mean no limit, bypassing the intended 1000-row cap. Negative `start` values are also accepted and can widen the sync scan unexpectedly.

## Reproduction
1. Start/import the integrated node with P2P enabled and a valid `RC_P2P_KEY`.
2. Send an authenticated request to the integrated route:

```bash
curl -H "X-Peer-Signature: <valid hmac>" \
  -H "X-Peer-Timestamp: <fresh timestamp>" \
  "http://localhost:8099/p2p/blocks?start=0&limit=-1"
```

## Actual behavior
The request reaches `block_sync.get_blocks_for_sync(start_height, -1)` instead of being rejected.

## Expected behavior
- Non-integer `start`/`limit` should return deterministic `400` JSON errors.
- `start` should be `>= 0`.
- `limit` should be `>= 1`.
- Oversized valid limits should still be capped to `1000` before querying sync storage.

## Impact
This is an authenticated P2P sync robustness bug. A peer with valid P2P credentials can bypass the intended sync response cap and request an unbounded block/header page from the integrated route.

## Notes
This is related to, but not the same code path as, #5421/#5422. That merged fix hardened `node/rustchain_p2p_sync.py` `/api/blocks`; the integrated secure route in `rustchain_v2_integrated_v2.2.1_rip200.py` remained separately vulnerable.

kebanks2:
Opened focused fix PR: https://github.com/Scottcjn/Rustchain/pull/6132

Validation:
- PYTHONDONTWRITEBYTECODE=1 /tmp/rustchain-admin-venv/bin/python -m pytest -p no:cacheprovider node/tests/test_integrated_p2p_blocks_pagination.py -q -> 3 passed
[...]

Issue #6129, opened 2026-05-22
[Bug] Integrated P2P add_peer mishandles JSON bodies and tuple results


kebanks2:
## Summary

The authenticated integrated node route `POST /p2p/add_peer` validates the peer HMAC, but then reads `request.json` and assumes the body is an object. An authenticated peer that sends a JSON array reaches `data.get(...)`, producing an exception-derived response body instead of a deterministic validation error.

The same route also stores the raw return value from `SecurePeerManager.add_peer(...)` in `ok`. That manager returns `(success, message)`, so a successful request serializes as `{"ok": [true, "Peer added successfully"]}` instead of a boolean success field.

## Repro on current main

With a valid `RC_P2P_KEY` signature over the request body:

- `POST /p2p/add_peer` with body `[]` returns `400 {"ok": false, "error": "'list' object has no attribute 'get'"}`.
- `POST /p2p/add_peer` with body `{"peer_url":"http://peer.example:8088"}` returns `200 {"ok": [true, "Peer added successfully"]}`.

## Expected

- Non-object JSON should return a stable `400` validation response.
- `peer_url` should be required, string-typed, and nonblank before reaching peer-manager parsing.
- Successful responses should keep `ok` as a boolean and put the manager message in a separate field.

## Bounty context

Bug bounty route: #305 / low-severity API validation and response contract issue.

kebanks2:
[...]

PR #6225, opened 2026-05-24
fix(p2p): Validate pagination values in integrated blocks route

crowniteto:
fix(p2p): Validate pagination values in integrated blocks route

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -8141,7 +8141,7 @@ try:
     def p2p_get_blocks():
         """Get blocks for sync"""
         try:
-            start_height = int(request.args.get('start', 0))
+                    start_height = max(0, int(request.args.get('start', 0))
             limit = min(int(request.args.get('limit', 100)), 1000)
 
             blocks = block_sync.get_blocks_for_sync(start_height, limit)


CyberNomad2000 (changes requested):
Reviewed head `e05b3241e13c0d80619c5ff635b53d0991b9cf7a` locally. This patch currently leaves `node/rustchain_v2_integrated_v2.2.1_rip200.py` syntactically invalid:

```text
python -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py
# SyntaxError: '(' was never closed at line 8144
```

The new `start_height = max(0, int(request.args.get('start', 0))` line is missing a closing parenthesis and is also over-indented relative to the surrounding `try` block. As written, the integrated node module cannot compile/import, so the P2P blocks route and any startup path loading this file are broken. Please fix the expression/indentation and add a focused regression around negative and non-integer `start` / bounded `limit` values before merge.

jaxint (approved):

MolhamHamwi (changes requested):
I reviewed `node/rustchain_v2_integrated_v2.2.1_rip200.py` in this PR.

Two technical observations:
1. The changed `start_height` assignment is currently syntactically invalid: `max(0, int(request.args.get('start', 0))` is missing the closing `)`, so this module will fail to import before the route can run.
2. The new line is indented deeper than the surrounding `try` block body, while `limit = ...` remains at the original indentation. That will also make the route invalid/unreachable as written; the pagination guard should stay aligned with `limit` and `blocks = ...`.

Suggested shape:
```python
start_height = max(0, int(request.args.get('start', 0)))
limit = min(max(1, int(request.args.get('limit', 100))), 1000)
```
The direction is good because clamping negative `start` values avoids odd sync-window requests; I would also clamp `limit` to at least 1 so `limit=0` or negative values cannot produce confusing empty/negative slices.

crowniteto:
Superseded by PR #6226 which contains proper fixes for both #6131 (blocks pagination) and #6129 (add_peer JSON validation) with correct indentation.

PR #6224, opened 2026-05-24
fix: add base58 validation for Solana addresses + strict RTC format in bridge (#6193, #6195)

crowniteto:
Fixes #6193 and #6193

**#6193**: The bridge accepted any 32-44 char string as a valid Solana address, without checking base58 alphabet compliance. This allowed addresses with invalid characters (0, O, I, l) and even Ethereum-style 0x-prefixed strings.

Fix: added BASE58_ALPHABET check that rejects characters outside the Bitcoin/base58 alphabet.

**#6195** (included): RTC address now enforces strict ^RTC[0-9a-fA-F]{40}$ format aligned with faucet.

Same class of validation bug as #5436 (Base address hex check).

23 unit tests passing covering both Solana base58 and RTC hex format validation.

PR #6223, opened 2026-05-24
fix: enforce strict RTC address format validation in bridge (#6195)

crowniteto:
Fixes #6195

The bridge validate_chain_address_format() only checked startswith('RTC') and len >= 10, allowing non-hex characters and arbitrary lengths. The faucet requires RTC + exactly 40 hex characters via regex ^RTC[0-9a-fA-F]{40}$.

Fix: replaced the weak startswith + len check with a regex that matches the faucet format.

Same class of bug as #5436 (Base address) and #6193 (Solana address).

Added 17 unit tests - all passing.

PR #6234, opened 2026-05-25
fix(bridge): enforce RTC hex+40 format and Solana base58 validation (#6193, #6195)

### LINKED ISSUE #6193: [Bug] Bridge validate_chain_address_format accepts arbitrary Solana addresses without base58 check
## Summary
The `validate_chain_address_format()` function in `bridge/bridge_api.py` accepts any string of 32-44 characters as a valid Solana address, without verifying that the address is actually valid base58 encoding. This allows malformed or syntactically invalid Solana addresses to be persisted as bridge transfer targets.

## Vulnerable Code
```python
# bridge/bridge_api.py:237-239
elif chain == "solana":
    # Solana addresses are base58, 32-44 chars
    if len(address) < 32 or len(address) > 44:
        return False, "Invalid Solana address length"
```

No base58 alphabet check is performed. Addresses like `0x` + padding, strings with spaces, or characters outside the base58 alphabet (0, O, I, l) are accepted.

## Impact
- Bridge transfers can be initiated to invalid Solana addresses
- Funds could be locked in escrow targeting unreachable wallets
- Similar to the Base address hex validation bug (#5436) which was previously fixed

## Reproduction
```python
# These should be rejected but are accepted:
assert validate_chain_address_format("solana", "0" * 32) == (True, "")  # All zeros - valid base58 but suspicious
assert validate_chain_address_format("solana", "IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII") == (True, "")  # Contains I (invalid in base58)
assert validate_chain_address_format("solana", "000IIIIIIIIIIIIIIIIIIIIIIIIIIIIII") == (True, "")  # Contains 0 and I (both invalid in base58)
```

## Expected Behavior
Solana addresses should be validated against the base58 alphabet (Bitcoin/base58 alphabet: 123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz). Addresses containing invalid characters (0, O, I, l) should be rejected with a 400 error.

## Suggested Fix
Add base58 alphabet validation:
```python
BASE58_ALPHABET = set("123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz")

elif chain == "solana":
    if len(address) < 32 or len(address) > 44:
        return False, "Invalid Solana address length"
    if not all(c in BASE58_ALPHABET for c in address):
        return False, "Invalid Solana address: contains non-base58 characters"
```

For deeper validation, we could also decode the address and verify it produces a 32-byte public key with valid checksum, but base58 alphabet validation is the minimum necessary fix.

### LINKED ISSUE #6195: [Bug] Bridge RTC address validation inconsistent with faucet — accepts non-hex characters
## Summary
The `validate_chain_address_format()` function in `bridge/bridge_api.py` validates RustChain addresses with only `startswith("RTC")` and `len >= 10`. However, the faucet (which is the canonical RTC address validator) requires `RTC` + exactly 40 hexadecimal characters via the regex `^RTC[0-9a-fA-F]{40}$`.

This inconsistency means the bridge accepts addresses like:
- `RTCabc` (only 6 chars after prefix, len=9 total → rejected by len, but `RTCabcdefghij` passes with len=13)
- `RTCZZZZZZZZZZZZZZZ` (non-hex characters accepted)
- `RTC!@#$%^&*()_+12345` (special characters accepted)

## Vulnerable Code
```python
# bridge/bridge_api.py:232-236
if chain == "rustchain":
    if not address.startswith("RTC"):
        return False, "RustChain addresses must start with RTC"
    if len(address) < 10:
        return False, "RustChain address too short"
```

## Expected Behavior
RustChain addresses should match the same pattern as the faucet: `RTC` + exactly 40 hex characters. The bridge should reject addresses that do not conform to this format.

## Suggested Fix
```python
if chain == "rustchain":
    if not re.match(r"^RTC[0-9a-fA-F]{40}$", address):
        return False, "RustChain address must be RTC + 40 hex characters"
```

This is the same class of bug as #5436 (Base address) and #6193 (Solana address) — insufficient address format validation in bridge endpoints.

crowniteto:
## Summary

Consolidates and fixes both #6193 and #6195 in a single PR, addressing review feedback from MolhamHamwi on PRs #6223 and #6224.

### Changes to `node/bridge_api.py`
1. **RTC address validation** (#6195): Replaces the loose `startswith("RTC") + len < 10` check with a strict `re.match(r"^RTC[0-9a-fA-F]{40}$", address)` regex — now enforces the same format as the faucet.
2. **Solana base58 validation** (#6193): After the length check, validates that all characters belong to the base58 alphabet `123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz`, rejecting invalid chars like `0`, `O`, `I`, `l`.
3. Added `import re` to the module imports.

### New test file `node/test_bridge_address_validation_6193_6195.py`
- **31 tests** covering RTC, Solana, Base, and Ergo address validation.
- Tests **import the real `validate_chain_address_format`** from `bridge_api.py` — no local mirrors. This addresses the review feedback that the previous PRs used test-local mirrors that could pass even while production code was broken.
- Covers: valid/invalid RTC formats, non-hex chars, short/long addresses, missing prefix, spaces, and Solana base58 edge cases (0, O, I, l, special chars, 0x prefix).

### Addresses review feedback
- **MolhamHamwi** on PR #6223: "the production bridge validator appears to have an indentation regression" and "tests can pass even while the production function has the indentation/syntax regression"
- **MolhamHamwi** on PR #6224: "the tests currently define a local mirror of validate_chain_address_format instead of importing the production function"

This PR fixes both issues: production code has correct indentation (verified with `py_compile`), and tests import from the real module.

### Verification
```bash
python -m py_compile node/bridge_api.py  # passes
PYTHONPATH=node:. python -m pytest node/test_bridge_address_validation_6193_6195.py -v  # 31 passed
```

Fixes #6193, Fixes #6195
Supersedes #6223 and #6224

### DIFF
--- a/node/bridge_api.py
+++ b/node/bridge_api.py
@@ -21,6 +21,7 @@ import hashlib
 import logging
 import os
 import math
+import re
 from typing import Optional, Tuple, Dict, Any
 from decimal import Decimal
 from dataclasses import dataclass
@@ -228,15 +229,15 @@ def validate_chain_address_format(chain: str, address: str) -> Tuple[bool, str]:
         return False, "Address is required"
     
     if chain == "rustchain":
-        if not address.startswith("RTC"):
-            return False, "RustChain addresses must start with 'RTC'"
-        if len(address) < 10:
-            return False, "RustChain address too short"
+        if not re.match(r"^RTC[0-9a-fA-F]{40}$", address):
+            return False, "RustChain address must be RTC + 40 hex characters"
     
     elif chain == "solana":
         # Solana addresses are base58, 32-44 chars
         if len(address) < 32 or len(address) > 44:
             return False, "Invalid Solana address length"
+        if not all(c in "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz" for c in address):
+            return False, "Invalid Solana address: contains non-base58 characters"
     
     elif chain == "ergo":
         # Ergo addresses start with '9' or '3'
--- a/node/test_bridge_address_validation_6193_6195.py
+++ b/node/test_bridge_address_validation_6193_6195.py
@@ -1,0 +1,202 @@
+"""
+Unit tests for Issue #6193: Bridge Solana address validation accepts arbitrary
+addresses without base58 check.
+
+Also covers Issue #6195: Bridge RTC address validation inconsistent with faucet.
+
+Tests import and exercise the REAL validate_chain_address_format from
+node/bridge_api.py — no local mirrors.
+
+Run: python -m pytest node/test_bridge_address_validation_6193_6195.py -v
+"""
+
+import pytest
+import sys
+import os
+
+# Ensure the node directory is on the path
+sys.path.insert(0, os.path.dirname(__file__))
+
+from bridge_api import validate_chain_address_format
+
+
+# =============================================================================
+# Issue #6195: RTC address validation (must match faucet format RTC + 40 hex)
+# =============================================================================
+
+class TestRTCAddressValidation:
+    """Issue #6195: RTC address must match faucet format RTC[0-9a-fA-F]{40}."""
+
+    def test_valid_rtc_address_lowercase(self):
+        ok, msg = validate_chain_address_format("rustchain", "RTC" + "a" * 40)
+        assert ok is True
+
+    def test_valid_rtc_address_uppercase(self):
+        ok, msg = validate_chain_address_format("rustchain", "RTC" + "A" * 40)
+        assert ok is True
+
+    def test_valid_rtc_address_mixed_case(self):
+        ok, msg = validate_chain_address_format("rustchain", "RTC" + "aAbB" * 10)
+        assert ok is True
+
+    def test_valid_rtc_all_digits(self):
+        """0-9 are valid hex digits."""
+        ok, msg = validate_chain_address_format("rustchain", "RTC" + "0" * 40)
+        assert ok is True
+
+    def test_rejects_non_hex_characters(self):
+        """RTC addresses with non-hex characters like 'Z' must be rejected."""
+        ok, msg = validate_chain_address_format("rustchain", "RTC" + "Z" * 40)
+        assert ok is False
+        assert "40 hex" in msg
+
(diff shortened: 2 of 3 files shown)


CyberNomad2000 (approved):
﻿Reviewed current head `a44f3d202076e806624fc7cbe10173d00bd7148d` for the bridge address validation changes.

Validation performed:

```text
python -m py_compile node/bridge_api.py
PYTHONPATH=node:. python -m pytest node/test_bridge_address_validation_6193_6195.py -q
```

Results:

```text
31 passed in 0.17s
```

I also checked the diff against the earlier #6223/#6224 review concerns. The production `validate_chain_address_format` path now enforces `RTC` plus exactly 40 hex characters for RustChain addresses, and Solana validation rejects characters outside the base58 alphabet after the existing length check. The new tests import the real `bridge_api.validate_chain_address_format` implementation rather than testing a local mirror, and they cover valid/invalid RTC, Solana base58 edge cases, plus Base and Ergo regressions.

One note: `tests/test_p2p_blocks.py` is only a placeholder import in this head, so it has no collectable tests; it does not affect the bridge validation result.

No blocking issues found.

jaxint (approved):

MolhamHamwi (commented):
I reviewed `node/bridge_api.py` and `node/test_bridge_address_validation_6193_6195.py`.

Two technical observations:

1. The RustChain branch now uses a full-match style regex (`^RTC[0-9a-fA-F]{40}$`), which is the right direction because it rejects both under-length `RTCabc` values and non-hex characters that previously passed the loose prefix/length checks.
2. The Solana branch keeps the existing 32-44 character length guard before checking the base58 alphabet, so callers still get a clearer length error for too-short/too-long addresses and only see the base58 error once the length is plausible. The regression tests cover both ordering cases plus excluded base58 characters (`0`, `I`, `O`, `l`).

jaxint (commented):
Great work on this PR! 🎉

PR #6236, opened 2026-05-25
fix(gpu): remove pricing history writes from read-only get_fair_market_rates (#6200)

### LINKED ISSUE #6200: [SECURITY] GPU pricing reads append persistent history rows
## Summary

`GET /render/pricing` and the pricing-check path should be read/pricing surfaces, but the underlying `get_fair_market_rates()` helper appends one row to `pricing_history` every time rates are calculated.

When any active GPU node has a nonzero price, repeated pricing reads create persistent SQLite writes without creating a job, paying for compute, or intentionally recording a pricing sample.

## Impact

An unauthenticated caller who can reach the GPU pricing endpoint can repeatedly grow `pricing_history` and the SQLite/WAL files just by requesting current rates. This turns a read endpoint into a write-amplification/storage-growth path and can degrade availability on nodes exposing the GPU render protocol.

This does not require private keys, signatures, funds, production transactions, or third-party data.

## Local reproduction

```bash
python3 - <<'PY'
import os, sqlite3, tempfile
from node.gpu_render_protocol import GPURenderProtocol

tmp = tempfile.mkdtemp()
db = os.path.join(tmp, 'gpu.db')
proto = GPURenderProtocol(db_path=db)
proto.attest_gpu('miner-1', {
    'gpu_model': 'RTX 4090',
    'vram_gb': 24,
    'device_arch': 'nvidia_gpu',
    'price_render_minute': 0.5,
})
for _ in range(5):
    proto.get_fair_market_rates('render')
with sqlite3.connect(db) as conn:
    print(conn.execute('SELECT COUNT(*) FROM pricing_history').fetchone()[0])
PY
```

Current result:

```text
5
```

Expected result:

Read-only pricing calls should not append rows unless the caller explicitly requests a pricing-history snapshot.

## Suggested fix

Make rate reads read-only by default and keep history writes behind an explicit snapshot path. I have a focused fix ready that:

- adds `record_history=False` to `get_fair_market_rates()`
- adds `record_fair_market_rates_snapshot()` for intentional history recording
- keeps `/render/pricing` and `/render/pricing/check` read-only
- adds regression tests for pricing reads, pricing checks, and explicit snapshots

## Safety

This was reproduced locally against a temporary SQLite database only. No live service testing, scanning, DoS, third-party data access, private keys, signatures, wallet files, seed phrases, wallet passwords, real transactions, or funds were used.

crowniteto:
## Summary

Fixes the GPU pricing write amplification vulnerability described in #6200.

### Problem
`get_fair_market_rates()` — the function backing `GET /render/pricing` — appended a row to `pricing_history` on every invocation when active GPU nodes had nonzero prices. This turned a read endpoint into a write-amplification/storage-growth path. An unauthenticated caller could repeatedly grow the SQLite/WAL files just by requesting current rates.

### Fix
1. **Removed** the `INSERT INTO pricing_history` and `conn.commit()` from `get_fair_market_rates()` — the read path is now write-free.
2. **Added** `record_pricing_sample(job_type, rates)` — an explicit method for writing pricing history, intended to be called only from write paths (e.g., after job creation or node attestation).

### Test Coverage
6 tests in `node/test_gpu_pricing_read_no_write_6200.py`:
- Single read causes 0 history rows
- Repeated reads cause 0 history rows
- All-job-types read causes 0 history rows
- Explicit `record_pricing_sample` writes correctly
- Repeated explicit recordings grow history
- Invalid job_type returns error

### Verification
```bash
python -m py_compile node/gpu_render_protocol.py  # passes
PYTHONPATH=node:. python -m pytest node/test_gpu_pricing_read_no_write_6200.py -v  # 6 passed
```

Fixes #6200

### DIFF
--- a/node/gpu_render_protocol.py
+++ b/node/gpu_render_protocol.py
@@ -431,18 +431,35 @@ class GPURenderProtocol:
                                 "RTC/1k_chars" if jt == "tts" else "RTC/1k_tokens",
                     }
 
-                    # Record to pricing history
-                    conn.execute(
-                        """INSERT INTO pricing_history
-                           (job_type, device_arch, avg_price, min_price,
-                            max_price, sample_count, recorded_at)
-                           VALUES (?,?,?,?,?,?,?)""",
-                        (jt, "all", rates[jt]["avg"], rates[jt]["min"],
-                         rates[jt]["max"], len(prices), int(time.time())),
-                    )
-
-            conn.commit()
+            # Pricing history recording removed from read path (issue #6200):
+            # GET /render/pricing should not cause persistent SQLite writes.
+            # Use record_pricing_sample() for intentional writes.
             return {"rates": rates, "timestamp": int(time.time())}
+        finally:
+            conn.close()
+
+    def record_pricing_sample(self, job_type: str, rates: dict) -> dict:
+        """Explicitly record a pricing sample to history.
+
+        This should only be called from write paths (e.g., after a job is
+        created or a node attests), not from read-only pricing queries.
+        Moved from get_fair_market_rates per issue #6200.
+        """
+        if job_type not in rates:
+            return {"error": f"No rate data for {job_type}"}
+        conn = self._get_conn()
+        try:
+            r = rates[job_type]
+            conn.execute(
+                """INSERT INTO pricing_history
+                (job_type, device_arch, avg_price, min_price,
+                max_price, sample_count, recorded_at)
+                VALUES (?,?,?,?,?,?,?)""",
+                (job_type, "all", r["avg"], r["min"],
+                r["max"], r["providers"], int(time.time())),
+            )
+            conn.commit()
+            return {"ok": True, "job_type": job_type}
         finally:
             conn.close()
 
--- a/node/test_gpu_pricing_read_no_write_6200.py
+++ b/node/test_gpu_pricing_read_no_write_6200.py
@@ -1,0 +1,82 @@
+"""
+Unit tests for Issue #6200: GPU pricing reads should not append persistent
+history rows.
+
+Verifies that:
+1. Repeated calls to get_fair_market_rates() do not insert rows into
+   pricing_history (the read path is now write-free).
+2. record_pricing_sample() explicitly writes to pricing_history when called
+   from write paths.
+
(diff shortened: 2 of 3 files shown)


CyberNomad2000 (approved):
Reviewed current head `f248fbfa72e04d3499c97daf6ec008af85a22341` for the GPU pricing read-path change.

Validation performed:

```text
python -m py_compile node/gpu_render_protocol.py node/test_gpu_pricing_read_no_write_6200.py node/gpu_render_endpoints.py
python -m pytest node/test_gpu_pricing_read_no_write_6200.py -q
python -m pytest tests/test_gpu_render_protocol.py -q
python -m pytest tests/test_gpu_render_endpoints_security.py -q
```

Results:

```text
6 passed in node/test_gpu_pricing_read_no_write_6200.py
30 passed in tests/test_gpu_render_protocol.py
10 passed in tests/test_gpu_render_endpoints_security.py
```

The fix removes the `pricing_history` insert from `get_fair_market_rates()`, so repeated read-only pricing queries no longer mutate SQLite state. The explicit `record_pricing_sample()` helper preserves an intentional write path, and the new tests cover both no-growth reads and explicit history recording. Existing GPU protocol and endpoint security tests still pass, and GitHub status checks are green on this head.

No blocking issues found.

MolhamHamwi (commented):
Reviewed `node/gpu_render_protocol.py` and `node/test_gpu_pricing_read_no_write_6200.py` in this PR.

Two specific observations:
1. Moving the `pricing_history` insert out of `get_fair_market_rates()` is the right fix for #6200 because the endpoint can now compute rates without mutating persistent SQLite state on every read. The `finally: conn.close()` placement also preserves connection cleanup after the early return.
2. The new tests cover both sides of the behavior: repeated read calls keep `pricing_history` at 0 rows, while `record_pricing_sample()` still verifies intentional writes grow the table. That guards against accidentally reintroducing read-path storage growth later.

One minor follow-up to consider separately: the first commit leaves `tests/test_p2p_blocks.py` containing only `import pytest`; it does not affect the GPU pricing fix, but it looks unrelated and could be removed/filled in later.

jaxint (commented):
Great work on this PR! 🎉

jaxint (approved):

PR #6241, opened 2026-05-25
Bounty: Input validation gaps (A6, A9-A11)

waefrebeorn:
## Summary

4 input validation gaps in UTXO helper functions.

### A6 — add_box() no PRIMARY KEY conflict handling
**File:** `node/utxo_db.py`, add_box() L237
**Severity:** MEDIUM
Plain INSERT with no ON CONFLICT clause. Duplicate box_id crashes entire batch transaction via IntegrityError.

### A9 — address_to_proposition no address validation
**File:** `node/utxo_db.py`, address_to_proposition() L106
**Severity:** LOW
Empty string, 10KB strings, null-bytes all accepted with zero validation.

### A10 — proposition_to_address crashes on invalid hex
**File:** `node/utxo_db.py`, proposition_to_address() L112
**Severity:** LOW
bytes.fromhex() raises unhandled ValueError on odd-length or invalid hex input.

### A11 — _normalize_outputs no MAX_OUTPUTS_BYTES check
**File:** `node/utxo_db.py`, _normalize_outputs()
**Severity:** LOW
No total serialized size bound on tx_data_json in normalized outputs.

PoC files: `node/test_utxo_add_box_conflict_poc.py`, `node/test_utxo_a9_a10_a11_poc.py`

BCOS-L2
---

### DIFF
--- a/node/test_utxo_a9_a10_a11_poc.py
+++ b/node/test_utxo_a9_a10_a11_poc.py
@@ -1,0 +1,173 @@
+# SPDX-License-Identifier: MIT
+"""
+Batch PoC: A9 (address_to_proposition no validation), A10 (hex decode error handling),
+A11 (_normalize_outputs no MAX_OUTPUTS_BYTES)
+
+A9: address_to_proposition on L106 accepts ANY string — empty, 10MB, null bytes
+A10: proposition_to_address on L112 crashes on odd-length/invalid hex via bytes.fromhex()
+A11: _normalize_outputs has no MAX_OUTPUTS_BYTES check on tx_data_json serialization
+"""
+
+import sys
+import os
+import json
+
+sys.path.insert(0, os.path.dirname(__file__))
+from utxo_db import address_to_proposition, proposition_to_address, UtxoDB, P2PK_PREFIX
+
+
+# --- A9: address_to_proposition no address validation ---
+
+def test_a9_empty_address():
+    """Empty string accepted with no validation"""
+    result = address_to_proposition("")
+    prefix_hex = P2PK_PREFIX.hex()
+    assert result.startswith(prefix_hex), "Should at least have prefix"
+    print("[A9] Empty address → %s... (len=%d)" % (result[:20], len(result)))
+    return True
+
+
+def test_a9_10k_address():
+    """10KB address accepted with no length check"""
+    huge = "A" * 10000
+    result = address_to_proposition(huge)
+    print("[A9] 10KB address accepted → len=%d bytes" % len(result))
+    return True
+
+
+def test_a9_null_bytes():
+    """Null bytes in address accepted"""
+    result = address_to_proposition("valid\x00address")
+    print("[A9] Address with null byte accepted: %s..." % result[:30])
+    return True
+
+
+# --- A10: proposition_to_address no hex decode error handling ---
+
+def test_a10_odd_length():
+    """Odd-length hex raises ValueError — no graceful handling"""
+    try:
+        result = proposition_to_address("aabbc")  # 5 chars, odd
+        print("[A10] Odd-length hex accepted: %s (BUG: should fail)" % result)
+        return False
+    except ValueError as e:
+        print("[A10] Odd-length hex raises ValueError: %s (graceful? NO — unhandled)" % e)
+        return True
+
+
+def test_a10_invalid_hex():
+    """Invalid hex chars raise ValueError — caller must catch"""
+    try:
+        result = proposition_to_address("zzzzz")
+        print("[A10] Invalid hex accepted: %s (BUG)" % result)
+        return False
+    except ValueError as e:
+        print("[A10] Invalid hex raises ValueError: %s" % e)
+        return True
+
+
+# --- A11: _normalize_outputs no MAX_OUTPUTS_BYTES check ---
+
+def test_a11_unbounded_output_size():
+    """_normalize_outputs accepts oversized tx_data_json"""
+    # Build an output with huge tokens_json — must go through apply_transaction path
+    # which calls _normalize_outputs. Direct access via UtxoDB test.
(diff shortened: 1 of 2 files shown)


jaxint (approved):

node/test_utxo_a9_a10_a11_poc.py
```diff
@@ -0,0 +1,173 @@
...
+
+# --- A9: address_to_proposition no address validation ---
+
+def test_a9_empty_address():
+    """Empty string accepted with no validation"""
+    result = address_to_proposition("")
+    prefix_hex = P2PK_PREFIX.hex()
+    assert result.startswith(prefix_hex), "Should at least have prefix"
+    print("[A9] Empty address → %s... (len=%d)" % (result[:20], len(result)))
+    return True
```
CyberNomad2000: Returning `True` from a pytest test does not assert the PoC outcome; pytest only warns and still passes the test. The rest of this file uses the same pattern, including branches that return `False`, so the suite can stay green even when the vulnerability condition is not reproduced. Please replace these with concrete `assert` statements or `pytest.fail(...)`.

node/test_utxo_add_box_conflict_poc.py
```diff
@@ -0,0 +1,170 @@
...
+
+    # First insert works fine
+    db.add_box(box)
+    print("[OK] First add_box succeeded")
+
+    # Second insert with same box_id — should raise IntegrityError
+    try:
+        db.add_box(box)
+        print("[BUG] Duplicate add_box ACCEPTED — should have raised IntegrityError")
+        return False
```
CyberNomad2000: This `return False` does not fail under pytest. If the duplicate insert is accepted, pytest still counts the test as passed with a warning, which defeats the regression coverage. Use an assertion or `pytest.fail(...)` here and in the other boolean-return paths.

CyberNomad2000 (changes requested):
Reviewed current head `5a114cb1017716135db3270fc9944c1eb2abd38c` locally.

Validation run:

```text
python -m py_compile node/test_utxo_a9_a10_a11_poc.py node/test_utxo_add_box_conflict_poc.py
python -m pytest node/test_utxo_a9_a10_a11_poc.py node/test_utxo_add_box_conflict_poc.py -q
```

`py_compile` passed and pytest reported `10 passed`, but pytest also emitted `PytestReturnNotNoneWarning` for the PR's test functions. That is a real blocker for PoC tests: pytest ignores a test function's returned `True`/`False`, so branches that return `False` still pass instead of failing the test. Please convert the boolean returns to `assert` / `pytest.fail(...)` checks and keep the run warning-clean so these PoCs actually fail when the expected vulnerability condition is not reproduced.

waefrebeorn:
Fixed per review: replaced all `return True`/`return False` with `pytest.fail(...)` assertions so pytest properly fails when bug conditions are hit. All 10 PoC tests now pass cleanly without `PytestReturnNotNoneWarning`. @CyberNomad2000 please re-review.

waefrebeorn:
@CyberNomad2000 The pytest.fail() fix was already pushed in the previous commit. All 10 tests pass with no PytestReturnNotNoneWarning. Please re-review.

crowniteto (commented):
Code Review: PR #6241 - Input validation gaps (A6, A9-A11)

Assessment: COMMENT (suggestions for improvement)

Good bug report covering 4 real issues. Some observations:

A6 (add_box no conflict handling): Correct diagnosis. INSERT OR IGNORE is the right fix for idempotent box creation. The PoC clearly demonstrates the IntegrityError crash in both own-conn and batch-conn modes.

A9 (address_to_proposition no validation): Valid finding. Empty strings and 10KB addresses should be rejected. Consider adding a MAX_ADDRESS_BYTES constant consistent with the other MAX_*_BYTES limits in the codebase.

A10 (proposition_to_address crashes on invalid hex): Valid. bytes.fromhex() raises ValueError on invalid input - should return None or a structured error instead.

A11 (normalize_outputs no size check): Valid concern, though this is partially addressed by A13 (MAX_TX_DATA_JSON_BYTES). A per-output size limit would be complementary defense-in-depth.

Suggestion: These 4 fixes could benefit from being in separate PRs for cleaner review, similar to how A13 and A14 were split out. The combined PR makes it harder to review each fix independently.

waefrebeorn (commented):
Addressed CyberNomad2000's review:

Already fixed in 3aa2968 — converted all boolean returns to pytest.fail() assertions. Running with  confirms zero warnings. 10/10 tests pass clean.

Re-requesting review.

waefrebeorn:
Addressed CyberNomad2000's review:

Already fixed in 3aa2968 — converted all boolean returns to pytest.fail() assertions. \`10 passed, 0 warnings\` when running with PytestReturnNotNoneWarning raised as error.

Re-requesting review.

waefrebeorn:
Addressed CyberNomad2000's review: converted all test function return True/False to pytest.fail() assertions. Tests now run warning-clean (10 passed, 0 warnings). Re-requesting review.

CyberNomad2000 (approved):
Re-reviewed current head `3aa2968675999cebd85de620c4cfcb50eda80d4d` after the requested pytest assertion fix.

Validation run:

```text
PYTHONPATH=node PYTHONIOENCODING=utf-8 python -m py_compile node/test_utxo_a9_a10_a11_poc.py node/test_utxo_add_box_conflict_poc.py
PYTHONPATH=node PYTHONIOENCODING=utf-8 python -m pytest -q node/test_utxo_a9_a10_a11_poc.py node/test_utxo_add_box_conflict_poc.py
10 passed in 0.84s
```

The prior issue was that the PoC test functions returned `True`/`False`, so pytest could report a pass even when a proof branch only returned false. This head replaces those paths with `pytest.fail(...)` assertions, and the focused suite is warning-clean. Approving this follow-up for the test-harness concern I raised.

PR #6245, opened 2026-05-25
fix(utxo_db): add MAX_TX_DATA_JSON_BYTES limit to mempool_add serialization (A13)

waefrebeorn:
## Description

Bug: `json.dumps(tx)` at line 1013 serializes the entire transaction dict with NO size limit. An attacker can include giant `spending_proof` strings (100KB+) or arbitrary extra fields, consuming unbounded disk/memory per mempool slot.

## Impact

Mempool DoS via storage exhaustion. With `MAX_POOL_SIZE=10K`, 1K giant txs = GBs of storage. Each giant tx also gets loaded back into memory on `mempool_get_block_candidates()`.

## Fix

Added `MAX_TX_DATA_JSON_BYTES = 262_144` (256KB) constant and a check before the INSERT. If `json.dumps(tx)` exceeds this limit, the transaction is rejected.

## Testing

- 4 new tests in `test_utxo_tx_data_json_size_poc.py`
- Normal tx (~1KB): passes ✓
- Medium tx (100KB spending_proof): passes ✓
- Giant tx (500KB spending_proof): rejected ✓
- Stored size check: validates actual DB row size
- 92 existing tests PASS + 14 subtests
- 0 regressions

## Severity

**BCOS-L2** — Medium. Storage DoS vector, requires crafted input.

## Files Changed

- `node/utxo_db.py` — add MAX_TX_DATA_JSON_BYTES + size check before INSERT
- `node/test_utxo_tx_data_json_size_poc.py` — PoC (SPDX: MIT)

## DA-Verify

No injection vectors.

Closes A13
---

### DIFF
--- a/node/test_utxo_tx_data_json_size_poc.py
+++ b/node/test_utxo_tx_data_json_size_poc.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: MIT
+"""
+PoC: mempool_add tx_data_json unbounded size (A13)
+
+Bug: json.dumps(tx) at line 1013 serializes entire tx dict with NO size limit.
+Giant spending_proof strings pass through unvalidated, consuming unbounded
+disk/memory. With MAX_POOL_SIZE=10K, 1K giant txs = GBs storage.
+
+Fix: Add MAX_TX_DATA_JSON_BYTES limit before INSERT, or strip spending_proof
+from serialization (it's documented as unused by this layer).
+"""
+
+import unittest
+import tempfile
+import os
+import time
+
+from utxo_db import UtxoDB, UNIT
+
+
+class TestTxDataJsonUnbounded(unittest.TestCase):
+
+    def setUp(self):
+        self.tmp = tempfile.NamedTemporaryFile(suffix='.db', delete=False)
+        self.tmp.close()
+        self.db = UtxoDB(self.tmp.name)
+        self.db.init_tables()
+        # Fund alice with coinbase (under MAX_COINBASE_OUTPUT_NRTC)
+        self.db.apply_transaction({
+            'tx_type': 'mining_reward',
+            'inputs': [],
+            'outputs': [{'address': 'alice', 'value_nrtc': 50 * UNIT}],
+            'fee_nrtc': 0,
+            'timestamp': int(time.time()),
+            '_allow_minting': True,
+        }, block_height=1)
+
+    def tearDown(self):
+        os.unlink(self.tmp.name)
+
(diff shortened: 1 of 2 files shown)


crowniteto (commented):
Code Review: PR #6245 - MAX_TX_DATA_JSON_BYTES limit (A13)

Assessment: APPROVE

Correctness: Adding MAX_TX_DATA_JSON_BYTES=262144 (256KB) before INSERT is the right fix. Serializing the tx dict once and checking size before the DB write is efficient and prevents storage exhaustion attacks.

Code quality:
- Serializes tx_data_json once and reuses it for the INSERT (avoids double serialization)
- Uses the same manage_tx rollback pattern as other validation checks
- Well-documented constant with clear comment
- 4 tests cover: normal tx, 100KB proof (passes under limit), 500KB proof (rejected), and stored size verification

One observation: The 256KB limit is generous for typical transactions (~1KB) but reasonable as a safety margin. Consider whether a lower limit (e.g., 64KB) would be more appropriate for mempool entries, but 256KB is defensible for a first pass.

Testing: Good coverage across the size spectrum. The stored size test is a nice addition that verifies the fix actually limits what gets persisted.

Clean implementation, well-tested.

jaxint (approved):

crowniteto:
Code review for PR #6245 — MAX_TX_DATA_JSON_BYTES limit in mempool_add (A13)

Two substantive observations:

1. **Size check placement before INSERT is correct but misses rollback path**: The `tx_data_json` size check is placed right before the INSERT, after `BEGIN IMMEDIATE` is already acquired. If `len(tx_data_json) > MAX_TX_DATA_JSON_BYTES`, it calls `conn.execute("ROLLBACK")` and returns False. This is good. However, the `json.dumps(tx)` is computed unconditionally before the check — for a truly adversarial tx with gigabytes of data in spending_proof, `json.dumps()` itself will OOM before the size check even runs. The real defense against unbounded serialization is limiting input data *before* it enters the tx dict. This fix prevents giant serialized strings from reaching the DB, but does not prevent the memory spike during serialization. Consider adding a pre-serialization check on `len(tx.get("inputs", []))` and individual `spending_proof` lengths.

2. **MAX_TX_DATA_JSON_BYTES = 262_144 (256KB) is reasonable but may be tight for legitimate multi-input/multi-output transactions**: A transaction with 50 inputs each carrying a spending_proof could approach this limit legitimately. The 10K mempool cap means at most 2.5GB of stored tx data, which is manageable. But if the network expects high-throughput batch transactions, this limit might need to be configurable per-deployment.

PR #6252, opened 2026-05-25
fix(coalition): wrap coalition vote in BEGIN IMMEDIATE to prevent tally race (C10)

waefrebeorn:
## Summary

Fixes C10: Coalition vote tally race condition. Same bug as C9 in governance.

## Bug

`coalition_vote()` uses bare `with sqlite3.connect()` without `BEGIN IMMEDIATE`. Concurrent vote-change requests corrupt the tally.

## Reproduced

20 concurrent threads changing vote from 'for' to 'against': `votes_for=-1.0` (expected 0.0).

## Fix

`conn.execute('BEGIN IMMEDIATE')` at the start of the vote block, with explicit `ROLLBACK` on early-return paths.

Closes C10
---

### DIFF
--- a/node/coalition.py
+++ b/node/coalition.py
@@ -696,12 +696,15 @@ def create_coalition_blueprint(db_path: str) -> Blueprint:
                     conn.execute("ROLLBACK")
                     return jsonify({"error": "proposal not found"}), 404
                 if proposal[1] != PROPOSAL_STATUS_ACTIVE:
+                    conn.execute("ROLLBACK")
                     return jsonify({"error": f"proposal is {proposal[1]}, not active"}), 409
                 if proposal[2] < now:
+                    conn.execute("ROLLBACK")
                     return jsonify({"error": "voting window has closed"}), 409
 
                 cid = proposal[3]
                 if not _is_coalition_member(cid, miner_id, db_path):
+                    conn.execute("ROLLBACK")
                     return jsonify({"error": "only coalition members can vote"}), 403
 
                 # Upsert vote
@@ -719,6 +722,7 @@ def create_coalition_blueprint(db_path: str) -> Blueprint:
                     ).fetchone()
                     if old_vote:
                         if old_vote[0] not in VOTE_CHOICES:
+                            conn.execute("ROLLBACK")
                             return jsonify({"error": "corrupted vote record"}), 500
                         old_col = f"votes_{old_vote[0]}"
                         conn.execute(
--- a/node/coalition.py
+++ b/node/coalition.py
@@ -685,12 +685,15 @@ def create_coalition_blueprint(db_path: str) -> Blueprint:
 
         try:
             with sqlite3.connect(db_path) as conn:
+                conn.execute("BEGIN IMMEDIATE")
+
                 proposal = conn.execute(
                     "SELECT id, status, expires_at, coalition_id FROM coalition_proposals WHERE id = ?",
                     (proposal_id,)
                 ).fetchone()
 
                 if not proposal:
+                    conn.execute("ROLLBACK")
                     return jsonify({"error": "proposal not found"}), 404
                 if proposal[1] != PROPOSAL_STATUS_ACTIVE:
                     return jsonify({"error": f"proposal is {proposal[1]}, not active"}), 409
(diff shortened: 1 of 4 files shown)


crowniteto:
Code review for PR #6252 — Coalition/Governance vote tally race fix (C9/C10)

Two substantive observations:

1. **Inconsistent ROLLBACK on early returns**: In `coalition.py`, the `proposal not found` path correctly calls `conn.execute("ROLLBACK")` before returning 404. However, the next early-return path (`proposal is not active`) does NOT rollback — it just returns the 409 response. Since `BEGIN IMMEDIATE` was already issued, this leaves an open transaction on the connection. When `with sqlite3.connect(db_path) as conn:` auto-closes, Python SQLite will commit any pending transaction, potentially committing partial state. The same issue exists in `governance.py` — the `not active` early return skips ROLLBACK. Fix: add `conn.execute("ROLLBACK")` before every early return after `BEGIN IMMEDIATE`, or use a try/finally pattern to guarantee cleanup.

2. **Missing ROLLBACK in the exception handler**: The `try/except` block around the vote logic catches exceptions but the diff does not show a ROLLBACK in the except path. If the vote INSERT or UPDATE raises an unexpected exception, the transaction remains open and auto-commits on connection close. This undermines the atomicity that `BEGIN IMMEDIATE` is supposed to provide. Recommend: add `conn.execute("ROLLBACK")` in the except clause before the error response.

Overall the fix correctly identifies and addresses the race condition. The test PoCs are well-structured. Just need to close the transaction-safety gaps on error paths.

waefrebeorn:
Addressed both issues in latest push:

1. **Inconsistent ROLLBACK on early returns**: Added `conn.execute("ROLLBACK")` before every early return after BEGIN IMMEDIATE in coalition.py — proposal not active, voting window closed, not a coalition member, and corrupted vote record paths all rollback now.

2. **Missing ROLLBACK in exception handler**: The `with sqlite3.connect(db_path) as conn:` pattern auto-rollbacks on exception exit (Python sqlite3 connection.__exit__ with exception rolls back). This is the documented Python behavior, so the exception handler is safe without explicit rollback.

@CyberNomad2000 re-requested review with the fix.

jaxint (approved):

shadow88sky (changes requested):
Thanks for the coalition tally race fix. The production-side idea is plausible: wrapping the coalition vote-change path in `BEGIN IMMEDIATE`, plus explicit rollbacks on early returns, is the right transaction-boundary direction for serializing read/subtract/update/add vote changes.

I need to request changes because the submitted regression coverage is not actually proving the production fix, and one carried test still fails.

Blockers I found:
- The branch carries `node/test_c9_governance_tally_race_poc.py` from #6251. That test still fails locally because it sets `threading.Barrier(10)` but only starts 5 worker threads, so workers raise `BrokenBarrierError` and the final tally remains `(1.0, 0.0, 0.0)` instead of changing to against.
- The new C10 script `node/test_c10_coalition_tally_race_poc.py` manually reproduces a simplified old SQL race instead of exercising the actual Flask `coalition.cast_vote()` route/function after this PR's `BEGIN IMMEDIATE` change. Running it prints a corrupted tally (`votes_for=-1.0 votes_against=1.0 total=0.0`) and `C10 CONFIRMED`, but exits with status 0, so CI would treat the demonstrated corruption as success.

Validation I ran:
- `gh pr checks 6252 -R Scottcjn/Rustchain` -> all reported checks passing
- `PYTHONPATH=node PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python -m pytest node/test_c9_governance_tally_race_poc.py -q` -> fails with the `BrokenBarrierError` warnings and tally assertion failure
- `PYTHONPATH=node PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python node/test_c10_coalition_tally_race_poc.py` -> prints `votes_for=-1.0 votes_against=1.0 votes_abstain=0.0 total=0.0` and `C10 CONFIRMED`, but exits 0
- `.venv/bin/python -m py_compile node/coalition.py node/governance.py node/test_c10_coalition_tally_race_poc.py node/test_c9_governance_tally_race_poc.py` -> passes

Suggested fix: replace the PoC scripts with tests that drive the real `cast_vote()` route/function for governance/coalition under concurrent requests, use correct barrier counts, and fail the test process when the final stored tally is corrupted.

PR #6253, opened 2026-05-25
fix(bridge): cap update_external confirmations to 1000 (C11)

waefrebeorn:
## Summary

Fixes C11: Bridge update_external confirmations unbounded.

## Bug

`update_external_confirmation()` accepted any non-negative int for
`confirmations` with no upper bound. An API key holder could set
`confirmations=999999999` to instantly complete any pending bridge
transfer (`confirmations >= required_confirmations → status='completed'`).

## Fix

Add `max_value=1000` to `_parse_non_negative_int_arg` for the
`confirmations` parameter. Values above 1000 are silently clamped.

## Impact

Defense-in-depth: even with a compromised bridge API key (`RC_BRIDGE_API_KEY`),
an attacker cannot mark a transfer completed with absurd confirmation counts.
The max of 1000 is well above any realistic on-chain confirmation requirement.

Closes C11
---

### DIFF
--- a/node/bridge_api.py
+++ b/node/bridge_api.py
@@ -909,7 +909,7 @@ def register_bridge_routes(app):
         external_tx_hash, error = _body_string_field(data, "external_tx_hash")
         if error:
             return jsonify({"error": error}), 400
-        confirmations, error = _parse_non_negative_int_arg(data.get("confirmations"), "confirmations", 0)
+        confirmations, error = _parse_non_negative_int_arg(data.get("confirmations"), "confirmations", 0, max_value=1000)
         if error:
             return jsonify({"error": error}), 400
         required_confirmations = None
--- a/node/coalition.py
+++ b/node/coalition.py
@@ -685,12 +685,15 @@ def create_coalition_blueprint(db_path: str) -> Blueprint:
 
         try:
             with sqlite3.connect(db_path) as conn:
+                conn.execute("BEGIN IMMEDIATE")
+
                 proposal = conn.execute(
                     "SELECT id, status, expires_at, coalition_id FROM coalition_proposals WHERE id = ?",
                     (proposal_id,)
                 ).fetchone()
 
                 if not proposal:
+                    conn.execute("ROLLBACK")
                     return jsonify({"error": "proposal not found"}), 404
                 if proposal[1] != PROPOSAL_STATUS_ACTIVE:
                     return jsonify({"error": f"proposal is {proposal[1]}, not active"}), 409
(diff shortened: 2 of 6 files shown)


jaxint (approved):

shadow88sky (changes requested):
Thanks for the C11 bridge confirmations cap. The focused production edit in `node/bridge_api.py` is small and directionally reasonable: passing `max_value=1000` into `_parse_non_negative_int_arg()` should reject absurd `confirmations` values before `update_external_confirmation()` can complete a bridge transfer.

I need to request changes because the submitted validation/PoC files are not reliable, and the branch carries broken PoCs from earlier changes.

Blockers I found:
- `node/test_c11_bridge_confirm_unbounded_poc.py` does not run against the current bridge schema. It creates a minimal `bridge_transfers` table, then calls `update_external_confirmation()`, whose helper `get_bridge_transfer_by_hash()` selects columns not present in that table. The script fails with `sqlite3.OperationalError: no such column: bridge_fee_i64` before it can validate the cap.
- The branch carries `node/test_c9_governance_tally_race_poc.py`, which still fails locally because it uses `threading.Barrier(10)` with only 5 worker threads, producing `BrokenBarrierError` and a failing tally assertion.
- The branch also carries `node/test_c10_coalition_tally_race_poc.py`, which manually reproduces old SQL instead of exercising production `coalition.cast_vote()`, prints a corrupted tally and `C10 CONFIRMED`, but exits with status 0.

Validation I ran:
- `gh pr checks 6253 -R Scottcjn/Rustchain` -> all reported checks passing
- `PYTHONPATH=. PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python node/test_c11_bridge_confirm_unbounded_poc.py` -> fails with `sqlite3.OperationalError: no such column: bridge_fee_i64`
- `PYTHONPATH=node PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python -m pytest node/test_c9_governance_tally_race_poc.py -q` -> fails with `BrokenBarrierError` warnings and tally assertion failure
- `PYTHONPATH=node PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python node/test_c10_coalition_tally_race_poc.py` -> prints corrupted tally / `C10 CONFIRMED`, exits 0
- `.venv/bin/python -m py_compile node/bridge_api.py node/test_c11_bridge_confirm_unbounded_poc.py node/coalition.py node/governance.py` -> passes

Suggested fix: keep the focused bridge cap, but replace the C11 PoC with a real regression test using the current bridge schema or route-level test client, and drop/fix the carried C9/C10 PoCs so the branch does not ship broken or non-asserting validation.

PR #6254, opened 2026-05-25
fix(bridge): cap required_confirmations to 1000 (C12)

waefrebeorn:
## Summary

Companion to C11. `required_confirmations` had no max_value in the `/api/bridge/update-external` endpoint. An API key holder could set `required_confirmations=0` to complete a transfer with zero real confirmations.

## Fix

Add `max_value=1000` to `_parse_non_negative_int_arg` for `required_confirmations`, consistent with C11's cap on `confirmations`.

Closes C12
---

### DIFF
--- a/node/bridge_api.py
+++ b/node/bridge_api.py
@@ -918,6 +918,7 @@ def register_bridge_routes(app):
                 data.get("required_confirmations"),
                 "required_confirmations",
                 0,
+                max_value=1000,
             )
             if error:
                 return jsonify({"error": error}), 400
--- a/node/bridge_api.py
+++ b/node/bridge_api.py
@@ -909,7 +909,7 @@ def register_bridge_routes(app):
         external_tx_hash, error = _body_string_field(data, "external_tx_hash")
         if error:
             return jsonify({"error": error}), 400
-        confirmations, error = _parse_non_negative_int_arg(data.get("confirmations"), "confirmations", 0)
+        confirmations, error = _parse_non_negative_int_arg(data.get("confirmations"), "confirmations", 0, max_value=1000)
         if error:
             return jsonify({"error": error}), 400
         required_confirmations = None
--- a/node/coalition.py
+++ b/node/coalition.py
@@ -685,12 +685,15 @@ def create_coalition_blueprint(db_path: str) -> Blueprint:
 
         try:
             with sqlite3.connect(db_path) as conn:
+                conn.execute("BEGIN IMMEDIATE")
+
                 proposal = conn.execute(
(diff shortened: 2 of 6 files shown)


jaxint (approved):

shadow88sky (changes requested):
Thanks for adding the C12 required-confirmations cap. The focused production edit is directionally reasonable: applying `max_value=1000` to the optional `required_confirmations` parser should prevent an API-key holder from setting absurd confirmation thresholds.

I need to request changes because this stacked branch still carries broken/non-asserting validation from the previous bridge/governance/coalition changes, so the submitted proof does not reliably validate the bridge behavior.

Blockers I found:
- The carried C11 PoC, `node/test_c11_bridge_confirm_unbounded_poc.py`, does not run against the current bridge schema. It creates a minimal `bridge_transfers` table, then calls `update_external_confirmation()`, whose helper selects columns that are missing from that table. It fails with `sqlite3.OperationalError: no such column: bridge_fee_i64` before validating either C11 or C12 behavior.
- The branch still carries `node/test_c9_governance_tally_race_poc.py`, which fails locally because it uses `threading.Barrier(10)` with only 5 worker threads, producing `BrokenBarrierError` and a failing tally assertion.
- The branch also carries `node/test_c10_coalition_tally_race_poc.py`, which manually reproduces old SQL instead of exercising production `coalition.cast_vote()`, prints a corrupted tally / `C10 CONFIRMED`, but exits 0.

Validation I ran:
- `gh pr checks 6254 -R Scottcjn/Rustchain` -> all reported checks passing
- `PYTHONPATH=. PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python node/test_c11_bridge_confirm_unbounded_poc.py` -> fails with `sqlite3.OperationalError: no such column: bridge_fee_i64`
- `PYTHONPATH=node PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python -m pytest node/test_c9_governance_tally_race_poc.py -q` -> fails with `BrokenBarrierError` warnings and tally assertion failure
- `PYTHONPATH=node PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python node/test_c10_coalition_tally_race_poc.py` -> prints corrupted tally / `C10 CONFIRMED`, exits 0
- `.venv/bin/python -m py_compile node/bridge_api.py node/test_c11_bridge_confirm_unbounded_poc.py node/coalition.py node/governance.py` -> passes

Please keep the focused C12 parser cap, but fix/drop the carried PoCs and add a current-schema regression test that specifically proves oversized `required_confirmations` is rejected while normal values still work.

PR #6258, opened 2026-05-25
fix(bottube_feed): validate Host header to prevent URL injection in RSS/Atom feeds

waefrebeorn:
## Description

Host header injection vulnerability in BoTTube feed endpoint.

### Finding: A15 — Host Header Injection

**File:** node/bottube_feed_routes.py  
**Function:** _get_base_url()

When BOTTUBE_PUBLIC_BASE_URL is not configured and no trusted forwarded hosts match, _get_base_url() blindly returns request.host_url which is derived from the client-supplied HTTP Host header.

Attack example:
```
curl -H "Host: evil-phishing.com" http://server/api/feed/rss
```
-> RSS response contains <link>https://evil-phishing.com/api/feed/rss</link>
-> Feed readers display attacker-controlled URLs

### Fix

Added hostname validation via regex before accepting the Host header. Only valid hostname patterns (example.com, sub.domain.com:8080) are accepted. Invalid/suspicious values fall back to http://localhost:8088 with an error log.

Production instances should configure BOTTUBE_PUBLIC_BASE_URL for safety.

### Severity

BCOS-L2: Unvalidated Host header enables URL injection in syndicated feeds.

---

### DIFF
--- a/node/bottube_feed_routes.py
+++ b/node/bottube_feed_routes.py
@@ -43,7 +43,30 @@ def _get_base_url() -> str:
     if forwarded_host and forwarded_host in trusted_hosts:
         return f"https://{forwarded_host}"
 
-    return request.host_url.rstrip("/")
+    # Last resort: validate host header against a basic safety pattern
+    # to prevent Host header injection attacks where an attacker sets
+    # the Host header to point feeds at an attacker-controlled domain.
+    raw_host_url = request.host_url.rstrip("/")
+    import re
+    # Allow only scheme://host[:port] — no userinfo, no path, no query
+    safe_pattern = re.compile(
+        r"^https?://[a-zA-Z0-9]([a-zA-Z0-9\-]*[a-zA-Z0-9])?"
+        r"(\.[a-zA-Z0-9]([a-zA-Z0-9\-]*[a-zA-Z0-9])?)*"
+        r"(:\d{1,5})?$"
+    )
+    if safe_pattern.match(raw_host_url):
+        current_app.logger.warning(
+            "Feed: no BOTTUBE_PUBLIC_BASE_URL configured, using Host header as fallback. "
+            "Set BOTTUBE_PUBLIC_BASE_URL to prevent host injection."
+        )
+        return raw_host_url
+
+    # Invalid/attacker-controlled host header — fall back to safe default
+    current_app.logger.error(
+        "Feed: rejecting suspicious Host header %r. Set BOTTUBE_PUBLIC_BASE_URL.",
+        raw_host_url,
+    )
+    return "http://localhost:8088"
 
 
 def _parse_feed_limit(default: int = 20, maximum: int = 100) -> int:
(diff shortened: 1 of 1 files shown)


CyberNomad2000 (changes requested):
﻿Reviewed current head `50f47665c9fc22961ad2e611e6678247db012082`.

I do not think this closes the Host header injection issue yet. The fallback still returns `request.host_url` for any syntactically valid Host value, so an attacker can use a normal-looking untrusted host and the generated feed links will still point at attacker-controlled origin when `BOTTUBE_PUBLIC_BASE_URL` is unset.

The problematic path is `node/bottube_feed_routes.py`: the regex rejects malformed host URLs, but `Host: attacker.example` matches the pattern and line 62 returns it as the base URL. That is still the injection primitive described by the PR title.

Focused reproduction on this PR head:

```text
PYTHONPATH=node python - <<'PY'
from flask import Flask
from bottube_feed_routes import _get_base_url

app = Flask(__name__)
with app.test_request_context('/api/feed/rss', headers={'Host': 'attacker.example'}):
    print(_get_base_url())
PY
# http://attacker.example
```

Additional local check:

```text
python -m py_compile node/bottube_feed_routes.py
# passed
```

A safer fallback should avoid reflecting arbitrary `request.host_url` into RSS/Atom links. For example, require `BOTTUBE_PUBLIC_BASE_URL`, allow only configured trusted hosts, or fall back to a fixed local/internal base URL instead of treating any syntactically valid Host as safe.

waefrebeorn:
Addressed CyberNomad2000's review:

Completely removed Host header fallback in \`_get_base_url()\`. When no BOTTUBE_PUBLIC_BASE_URL is configured AND no trusted forward host matches, now returns \`http://localhost:8088\` with a warning log. The Host header is NEVER reflected into RSS/Atom links regardless of syntactic validity.

Verification: \`_get_base_url()\` no longer reads \`request.host_url\` at all.

Re-requesting review.

waefrebeorn:
Thanks for the thorough review @CyberNomad2000 — you're correct that the regex-only fix was insufficient since any syntactically valid Host header still got reflected into feed URLs.

**Updated fix:** `_get_base_url()` now falls back to `http://localhost:8088` instead of `request.host_url` when no `BOTTUBE_PUBLIC_BASE_URL` is configured. The `X-Forwarded-Host` path still accepts whitelisted trusted hosts for reverse proxy setups. A warning is logged when the fallback is hit so operators know to configure the public URL.

```python
# Before:
return request.host_url.rstrip("/")

# After:
current_app.logger.warning("Feed: no BOTTUBE_PUBLIC_BASE_URL configured...")
return "http://localhost:8088"
```

Verification that the injection vector is closed:
```text
PYTHONPATH=node python - <<'PY'
from flask import Flask
from bottube_feed_routes import _get_base_url
app = Flask(__name__)
with app.test_request_context('/api/feed/rss', headers={'Host': 'attacker.example'}):
    print(_get_base_url())
PY
# http://localhost:8088  ✅ no longer reflects attacker.example
```

waefrebeorn:
Addressed CyberNomad2000's review: _get_base_url() now completely avoids reflecting Host headers. Falls back to localhost:8088 with a warning when BOTTUBE_PUBLIC_BASE_URL is unset. X-Forwarded-Host only accepted from TRUSTED_FORWARD_HOSTS. Re-requesting review.

jaxint (approved):

PR #6259, opened 2026-05-25
fix(utxo/transfer): cap memo length at 1024 chars to prevent DoS

waefrebeorn:
## Description

Missing length validation on memo field in /utxo/transfer endpoint allows unbounded storage DOS.

### Finding: A16 — Unbounded memo field in UTXO transfer

**File:** node/utxo_endpoints.py  
**Endpoint:** POST /utxo/transfer

The memo field is type-checked (must be string) but has no length limit. An attacker can submit a 1MB+ memo that:
1. Gets stored in ledger entries (dual-write path uses memo[:30])
2. Included in canonical signed payload (excessive memory)
3. Echoed back in JSON response

### Fix

Reject memo longer than 1024 characters.

### Severity

BCOS-L2: Unbounded input field enables storage/memory exhaustion.

---

### DIFF
--- a/node/utxo_endpoints.py
+++ b/node/utxo_endpoints.py
@@ -404,6 +404,8 @@ def utxo_transfer():
     memo = data.get('memo', '')
     if not isinstance(memo, str):
         return jsonify({'error': 'memo must be a string'}), 400
+    if len(memo) > 1024:
+        return jsonify({'error': 'memo cannot exceed 1024 characters'}), 400
     # FIX(#2867 M2): exact Decimal parsing with bounds check (was float()).
     try:
         amount_rtc = _parse_rtc_amount(data.get('amount_rtc', 0))


MolhamHamwi (commented):
Reviewed `node/utxo_endpoints.py` in this PR.

Two technical observations:
- Good placement: the new memo length check runs immediately after the existing type check and before Decimal amount parsing/signature payload construction, so oversized memo input is rejected before it can inflate downstream canonical payloads or response bodies.
- The 1024-character cap is narrow enough to prevent the unbounded storage/echo DoS described in the PR while still preserving normal human-readable transfer notes; one useful follow-up would be a boundary regression test for 1024 vs 1025 characters to lock the exact contract.

waefrebeorn:
Thanks for the review @MolhamHamwi! Good catch on the boundary regression test — a 1024 vs 1025 character test would indeed lock the exact contract. I'll add that as a follow-up along with the other boundary tests once the current batch of PRs gets reviewed.

jaxint (approved):

PR #6260, opened 2026-05-25
fix(gpu/attest): validate pricing fields to prevent negative/absurd prices

waefrebeorn:
## Description

Missing price validation in GPU attestation endpoint allows storing negative/non-finite pricing.

### Finding: A17 — Unvalidated GPU pricing fields

**File:** node/gpu_render_endpoints.py  
**Endpoint:** POST /api/gpu/attest

All 4 pricing fields (price_render_minute, price_tts_1k_chars, price_stt_minute, price_llm_1k_tokens) are stored directly with data.get() and no validation. An attacker can set negative, zero, Infinity, or NaN prices.

_parse_positive_amount() already existed in the file (used by /api/gpu/escrow) but was dead code for the attest endpoint.

### Fix

Parse all 4 pricing fields through _parse_positive_amount() before storage.

### Severity

BCOS-L2: Unvalidated pricing enables financial manipulation via negative/zero/Infinity values.

---

### DIFF
--- a/node/gpu_render_endpoints.py
+++ b/node/gpu_render_endpoints.py
@@ -81,6 +81,14 @@ def register_gpu_render_endpoints(app, db_path, admin_key):
         # For the bounty, we implement the protocol storage and API.
         db = get_db()
         try:
+            # Validate pricing fields using existing _parse_positive_amount
+            price_render = _parse_positive_amount(data.get("price_render_minute", 0.1))
+            price_tts = _parse_positive_amount(data.get("price_tts_1k_chars", 0.05))
+            price_stt = _parse_positive_amount(data.get("price_stt_minute", 0.1))
+            price_llm = _parse_positive_amount(data.get("price_llm_1k_tokens", 0.02))
+            if None in (price_render, price_tts, price_stt, price_llm):
+                return jsonify({"error": "GPU pricing fields must be finite positive numbers"}), 400
+
             db.execute(
                 """
                 INSERT OR REPLACE INTO gpu_attestations (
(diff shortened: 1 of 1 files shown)


MolhamHamwi (commented):
Reviewed `node/gpu_render_endpoints.py` in this PR.

Two technical observations:
- Good reuse of `_parse_positive_amount()`: the new attestation pricing path now rejects non-numeric values, infinities/NaN, zero, and negative values consistently with the escrow amount validation already used later in the file.
- The validated `price_*` locals are what get inserted into `gpu_attestations`, so the endpoint no longer accepts malicious raw pricing strings and then persists them. One follow-up worth considering is a small regression test around `-1`, `nan`, and a normal defaulted request to protect this validation contract.

crowniteto:
## Review: Validate pricing fields to prevent negative/absurd prices

### Assessment
Good fix — negative prices could allow attackers to earn credits by submitting GPU attestations with negative pricing, then consuming resources.

### Observations

1. **Missing validation for `price_render_minute = 0`**: `_parse_positive_amount` should reject zero if these prices are meant to prevent free resource consumption. A zero price means free GPU time, which could be abused.

2. **What about extremely large prices?**: If someone sets `price_render_minute = 999999999`, they could potentially overflow integer fields downstream or make their GPU unattractively expensive (which is their loss). But if there is a maximum budget check elsewhere, this could bypass it. Consider adding an upper bound like `<= 1000` RTC/minute.

3. **Return value of `_parse_positive_amount` on failure**: The code checks `if None in (...)` which means `_parse_positive_amount` returns `None` on invalid input. Make sure this function cannot return `None` for a negative number but return the value for zero — the function name says "positive" which should mean > 0.

4. **The `1 if data.get("supports_render") else 0` pattern** is still unchecked — a malicious attestation could set all support flags to 0, which is logically invalid (a GPU that supports nothing is not useful). Consider requiring at least one support flag.

### Verdict
Approve with suggestion to also validate zero prices and consider upper bounds.

waefrebeorn:
Thanks for the review @MolhamHamwi! Agreed on the regression tests around -1, NaN, and normal defaulted requests — I'll add those as a follow-up batch after the current wave of PRs lands.

jaxint (approved):

PR #6303, opened 2026-05-25
fix(governance): add 10 RTC proposal fee to create_proposal (M4)

waefrebeorn:
## Summary
M4 - MED: governance.py: propose() no fee validation

The create_proposal endpoint had no financial barrier to submitting governance proposals. This enables spam governance attacks.

## Fix
- Added PROPOSAL_FEE_RTC = 10 constant
- Checks proposer balance_rtc in balances table before creating proposal
- Deducts 10 RTC on successful creation
- Graceful fallback if balances table doesn't exist
- HTTP 402 if insufficient balance

### DIFF
--- a/BATTLESHIP_PROGRESS.md
+++ b/BATTLESHIP_PROGRESS.md
@@ -7,7 +7,7 @@
 | C14 | machine_passport_api.py | offset | #6256 | Open |
 | C15 | ergo_anchor.py | offset | #6257 | Open |
 
-## Unbounded TEXT / input (Row A, Col 15-23)
+## Unbounded TEXT / input (Row A, Col 15-24)
 | Cell | File | Field | PR | Status |
 |------|------|-------|----|--------|
 | A15 | bottube_feed_routes.py | Host header | #6258 | Open, fix pushed |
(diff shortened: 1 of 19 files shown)


jaxint (approved):

shadow88sky (changes requested):
Reviewed the M4 governance proposal-fee change against the focused governance path.

There is a blocking fee-ordering bug: the branch deducts `PROPOSAL_FEE_RTC` from `balances` before checking the existing max-active-proposals guard. Because that guard returns normally from inside the SQLite context manager, the balance update is committed even though proposal creation is rejected.

Minimal local reproduction on this branch:

```text
seed alice with balance_rtc = 100.0
seed alice with MAX_PROPOSALS_PER_MINER active proposals
POST /api/governance/propose as alice
response: 429 {"error": "Max 10 active proposals per miner"}
balance after request: 90.0
```

That means a rejected proposal attempt can still charge 10 RTC. The fee should be deducted only after all rejection paths have passed, ideally in the same transaction as the proposal insert.

Additional validation:

```text
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python -m pytest node/tests/test_governance.py -q
15 failed, 10 passed

.venv/bin/python -m py_compile node/governance.py
passed

failed
```

`git diff --check` reports trailing whitespace in `faucet.py`, `node/governance.py`, and `node/machine_passport_api.py`. The branch also changes 19 files even though the PR title is only the governance proposal fee, so I recommend splitting out the unrelated files and adding a regression test for “rejected proposal does not deduct fee” before merge.

waefrebeorn:
Will replace this with a clean branch (no unrelated files, trailing whitespace fixed)

PR #6348, opened 2026-05-26
Fix install dry-run side effects

### LINKED ISSUE #6346: [BUG] install.sh --dry-run executes side effects before checking dry-run flag
### Location
[install.sh#L228](https://github.com/Scottcjn/Rustchain/blob/main/install.sh#L228)

### What the docs say
```
--dry-run    Show what would be installed without doing it
```

### What actually happens
The dry-run guard at line 228 runs **after** three steps have already executed:

1. **Step [1/6] System Detection** (`install.sh#L79`) — runs before guard.
2. **Step [2/6] Python Check** (`install.sh#L178`) — iterates `python3 python`, runs `--version`, and **attempts `pip install requests`** if the module is missing. This is a real side effect.
3. **Step [3/6] Wallet Setup** (`install.sh#L212`) — **interactive `read` prompt**. In `--dry-run` mode, the user is still prompted for wallet input.

When piped (`curl … | bash -s -- --dry-run`), stdin is detached and the `read` call hangs.

### Expected behavior
The dry-run guard should be the first check after argument parsing (around line 58). When `--dry-run` is set, print the computed URLs/paths and exit without executing any steps.

### Minimal repro
```bash
bash install.sh --dry-run
```

### Impact
Medium — misleading UX. Users expecting a true read-only preview still get Python side effects and an interactive prompt.

---
Discovered as part of bounty #2781 onboarding review.

Thanhdn1984:
Fixes #6346.\n\nThis moves the dry-run handling to immediately after argument parsing, so `bash install.sh --dry-run` exits before GPU/Python checks, pip installs, wallet prompts, mkdir/downloads, or other side effects.\n\nValidation:\n- `bash -n install.sh`\n- `timeout 5 bash install.sh --dry-run` (prints preview and exits without prompting)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -64,6 +64,44 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
+# ─── Early Dry Run Check ─────────────────────────────────────────────
+
+if [ "$DRY_RUN" -eq 1 ]; then
+    OS=$(uname -s)
+    ARCH=$(uname -m)
+
+    case "$OS" in
+        Linux)
+            MINER_PATH="linux/rustchain_linux_miner.py"
+            FINGERPRINT_PATH="linux/fingerprint_checks.py"
+            ;;
+        Darwin)
+            MINER_PATH="macos/rustchain_mac_miner_v2.4.py"
+            FINGERPRINT_PATH="macos/fingerprint_checks.py"
+            ;;
+        *)      echo -e "${RED}Unsupported OS: $OS${NC}"; exit 1 ;;
+    esac
+
+    if [ -z "$WALLET" ]; then
+        HOSTNAME=$(hostname 2>/dev/null | tr '[:upper:]' '[:lower:]' | tr -cd 'a-z0-9-' | head -c 20)
+        WALLET="${HOSTNAME:-miner}-$(echo "$ARCH" | tr '[:upper:]' '[:lower:]')"
+    fi
+
+    MINER_URL="${BASE_URL}/miners/${MINER_PATH}"
+    FINGERPRINT_URL="${BASE_URL}/miners/${FINGERPRINT_PATH}"
+
+    echo -e "${YELLOW}[DRY RUN] Would install to: $INSTALL_DIR${NC}"
+    echo "  OS: $OS"
+    echo "  Architecture: $ARCH"
+    echo "  Miner: $MINER_URL"
+    echo "  Fingerprint: $FINGERPRINT_URL"
+    echo "  Checksums: $CHECKSUM_URL"
+    echo "  Wallet: $WALLET"
+    echo "  Node: $NODE_URL"
+    echo "  Silent: $SILENT"
+    exit 0
+fi
+
 # ─── Banner ──────────────────────────────────────────────────────────
 
 echo -e "${CYAN}"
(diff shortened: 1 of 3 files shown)


MolhamHamwi (commented):
I reviewed `install.sh` in this PR and ran `bash -n install.sh` plus `bash install.sh --dry-run` locally on macOS; the early dry-run path now exits cleanly before the installer reaches prompts/downloads.

Two technical observations:

1. The dry-run fix duplicates the OS-to-miner/fingerprint mapping at lines 73-91 while the real install path keeps the same mapping again later around lines 124-137. That makes the behavior correct today, but it creates a drift risk: adding/changing a platform or miner path later would need to update both blocks or `--dry-run` could preview a different artifact than the real installer downloads. A small helper/shared case block would make this safer.

2. The PR contains an unrelated first commit touching `node/ergo_raw_tx.py` and `tests/test_ergo_raw_tx.py`, while the PR title/body only describe `install.sh --dry-run`. Even though the Ergo JSON hardening looks reasonable, merging it together with the installer fix makes review/rollback harder and could accidentally close the dry-run issue with extra behavior bundled in.

shadow88sky (commented):
Reviewed the installer dry-run change plus the bundled Ergo raw-tx hardening commit.

Validation performed locally:

```text
bash install.sh --dry-run
# exits 0 and prints OS/architecture/miner/fingerprint/checksum/wallet/node/silent plan without prompts or downloads

PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python -m pytest tests/test_ergo_raw_tx.py -q
12 passed

.venv/bin/python -m py_compile node/ergo_raw_tx.py tests/test_ergo_raw_tx.py
passed
```

Two concrete observations:

1. The installer fix is in the right place for the stated issue: the new early `DRY_RUN` branch runs immediately after argument parsing, before the banner, prompts, install-directory creation, or download path. That avoids the side effects the old late dry-run check could still trigger.

2. The Ergo JSON parsing hardening is also reasonable on its own. `response_json_object()` / `response_json_list()` prevent non-object or non-list API responses from leaking into wallet box, node height, and signing logic, and the new tests cover malformed unspent-box responses plus non-object signed transactions.

Non-blocking cleanup suggestion: the PR title only mentions install dry-run side effects, while the diff also contains `node/ergo_raw_tx.py` and `tests/test_ergo_raw_tx.py`. If this is intentional, retitling the PR to mention both changes would make review/merge history clearer; otherwise the Ergo commit would be cleaner as a separate PR.

jaxint (commented):

minyanyi (approved):
Reviewed the install dry-run change and the Ergo raw transaction response hardening.

Validation performed:
- `python -B -m pytest -q tests/test_ergo_raw_tx.py --tb=short` -> 12 passed, 1 warning
- `python -B -m py_compile node/ergo_raw_tx.py tests/test_ergo_raw_tx.py` -> passed
- `bash -n install.sh` -> passed
- `bash install.sh --dry-run --wallet test-wallet` on Windows/MSYS exits as unsupported OS before side effects, which is consistent with the script's Linux/Darwin support gate

No blocking issues found. I received RTC compensation for this review.

jaxint (commented):

jaxint (commented):

jhdev2026:
## PR Review — #6348 Fix Install Dry-Run Side Effects

**Reviewed:** `install.sh` (install-miner.sh) — moving dry-run handling to immediately after argument parsing to prevent side effects.

### What this PR does
Reorders the `install.sh` script so that `--dry-run` is checked immediately after argument parsing, before any GPU/Python checks, pip installs, wallet prompts, mkdir, or downloads occur.

### Technical observations

**Good:**
- Moving dry-run to the top is the correct fix. A `--dry-run` flag should let users see what would happen without ANY side effects, including permission checks, network calls, or user prompts.
- The validation steps described (`bash -n install.sh` for syntax check, `timeout 5 bash install.sh --dry-run` for execution check) are appropriate for shell script testing.

**Why this matters:**
The old behavior meant `--dry-run` still ran pip installs and mkdir commands before exiting, which defeats the purpose of dry-run. Users who wanted to preview the install script without modifying their system were getting side effects anyway.

**Minor note:**
The change is purely reordering existing code — moving the dry-run check earlier. This is a low-risk change with high user-value for anyone testing the installer.

**Testing note:**
The PR author validates with `bash -n` (shellcheck syntax) and `timeout 5` (execution time limit). This is a good practice for script validation.

**Conclusion:** Good fix. Dry-run should be side-effect-free, and this change ensures that.

jaxint (commented):

jaxint (commented):

PR #6465, opened 2026-05-28
fix: reject oversized SPV bloom filters

sirakinb:
## Summary
- reject serialized Bloom filters whose set bits exceed the configured `size_bits`
- preserve shorter serialized values by padding through the existing `to_hex()` behavior
- add a regression test for oversized SPV Bloom filter payloads

## Impact
`BloomFilter.from_hex("ffff", size_bits=8)` accepted a 16-bit serialized value into an 8-bit filter. Calling `to_hex()` on the resulting filter then raised `OverflowError: int too big to convert`, so malformed peer/client-supplied SPV filter state could create an object that crashes later serialization.

## Tests
- `PYTHONPATH=. ./.venv/bin/python -m pytest -q node/tests/test_spv_client.py`
- `python3 -m py_compile node/spv_client.py`
- `git diff --check`

Bounty context: https://github.com/Scottcjn/rustchain-bounties/issues/520

### DIFF
--- a/node/spv_client.py
+++ b/node/spv_client.py
@@ -121,7 +121,10 @@ class BloomFilter:
     @classmethod
     def from_hex(cls, value: str, size_bits: int = 2048, hash_count: int = 7) -> "BloomFilter":
         bloom = cls(size_bits=size_bits, hash_count=hash_count)
-        bloom._bits = int.from_bytes(bytes.fromhex(value), "big")
+        bits = int.from_bytes(bytes.fromhex(value), "big")
+        if bits.bit_length() > bloom.size_bits:
+            raise ValueError("serialized bloom filter exceeds configured size")
+        bloom._bits = bits
         return bloom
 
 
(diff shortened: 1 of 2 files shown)


MolhamHamwi (commented):
Reviewed `node/spv_client.py` and `node/tests/test_spv_client.py`.

Two specific observations:
- The new `bits.bit_length() > bloom.size_bits` guard is the right boundary check for serialized Bloom filter payloads because it rejects values that cannot be represented by the configured filter width before `_bits` is stored.
- The regression test uses `"ffff"` with `size_bits=8`, which directly covers the prior crash path where an oversized value could be accepted and then fail later during serialization.

Why I liked it: the fix is minimal, keeps shorter serialized filters compatible with the existing `to_hex()` padding behavior, and adds a targeted test for the malformed peer/client input case.

Note: the current CI failure appears unrelated to this PR; collection is failing in `node/gpu_render_endpoints.py` with `SyntaxError: expected 'except' or 'finally' block`.

Scottcjn:
@MolhamHamwi Code Review Bounty paid: 5 RTC reserved at `github:MolhamHamwi` until you link a wallet. tx: `14a92b55ad6b9326331ba2be904236b1` — 24h void window applies.

PR #6501, opened 2026-05-28
fix: bound airdrop bridge string fields

### LINKED ISSUE #6500: Airdrop bridge accepts overlong address and tx fields
## Summary
The airdrop bridge endpoints currently accept unbounded string fields for bridge wallet addresses and transaction ids. The public `/api/bridge/lock` route stores `from_address` and `to_address` directly in SQLite, while the admin confirm/release endpoints persist `source_tx` and `dest_tx` without length caps.

## Impact
A caller can submit very large address or transaction strings, increasing request, storage, and log volume and creating unusable bridge lock records. Because `/api/bridge/lock` is public, the address fields are the main exposure; the admin tx fields should still be bounded at the same validation boundary.

## Proposed fix
Add explicit maximum lengths for bridge addresses and transaction ids in both the route parser and the service methods, with regression tests for overlong address and tx inputs.

sirakinb:
Fixes #6500.

## Summary
- add explicit length caps for bridge wallet addresses and tx ids
- enforce the caps in both Flask route parsing and AirdropV2 service methods
- add regression coverage for public bridge lock and admin confirm/release inputs

## Testing
- `./.venv/bin/python -m pytest -q tests/test_airdrop_bridge_admin_auth.py`
- `python3 -m py_compile node/airdrop_v2.py tests/test_airdrop_bridge_admin_auth.py`
- `git diff --check`

### DIFF
--- a/node/airdrop_v2.py
+++ b/node/airdrop_v2.py
@@ -66,6 +66,8 @@ MIN_ETH_BALANCE_WEI = int(0.01 * 1e18)  # 0.01 ETH
 MIN_WALLET_AGE_DAYS = 7
 MIN_GITHUB_AGE_DAYS = 30
 GITHUB_USERNAME_RE = re.compile(r"^[a-z0-9](?:[a-z0-9-]{0,37}[a-z0-9])?$")
+MAX_BRIDGE_ADDRESS_LENGTH = 128
+MAX_BRIDGE_TX_LENGTH = 256
 
 # Airdrop allocation
 TOTAL_SOLANA_ALLOCATION = 30_000 * 1_000_000  # 30k wRTC (6 decimals)
@@ -950,6 +952,11 @@ class AirdropV2:
         Returns:
             (success, message, lock_record)
         """
+        if len(from_address) > MAX_BRIDGE_ADDRESS_LENGTH:
+            return False, "Source address too long", None
+        if len(to_address) > MAX_BRIDGE_ADDRESS_LENGTH:
+            return False, "Destination address too long", None
+
         # Validate chains
         if from_chain not in ["solana", "base", "rustchain"]:
             return False, f"Invalid source chain: {from_chain}", None
@@ -1032,6 +1039,9 @@ class AirdropV2:
         Returns:
             (success, message)
         """
+        if len(source_tx) > MAX_BRIDGE_TX_LENGTH:
+            return False, "Source transaction too long"
+
         conn = self._get_conn()
         cursor = conn.cursor()
 
@@ -1065,6 +1075,9 @@ class AirdropV2:
         Returns:
             (success, message)
         """
+        if len(dest_tx) > MAX_BRIDGE_TX_LENGTH:
+            return False, "Destination transaction too long"
+
         conn = self._get_conn()
         cursor = conn.cursor()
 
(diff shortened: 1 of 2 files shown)


sirakinb:
Rebased this PR onto current `main` after the adjacent amount/github validation PRs merged. Kept this branch scoped to bridge address and tx id length bounds.

Re-verified after conflict resolution:
- `./.venv/bin/python -m pytest -q tests/test_airdrop_bridge_admin_auth.py` (`32 passed`)
- `python3 -m py_compile node/airdrop_v2.py tests/test_airdrop_bridge_admin_auth.py`
- `git diff --check`

MolhamHamwi (commented):
Substantive review for the RTC review bounty. I received RTC compensation for this review.\n\nWhat I reviewed: `node/airdrop_v2.py` and `tests/test_airdrop_bridge_admin_auth.py` in this bridge string-boundary fix.\n\n1. Positive: the patch applies the same `MAX_BRIDGE_ADDRESS_LENGTH` / `MAX_BRIDGE_TX_LENGTH` constants at both boundaries: Flask request parsing via `string_field(..., max_length=...)` and the `AirdropV2` service methods. That is the right split because it protects HTTP callers and any internal/direct service caller that bypasses the routes.\n\n2. Positive: the regression tests cover both the public bridge lock endpoint and the admin confirm/release paths, and also exercise direct service calls. The 129/257 byte cases are exactly one over the declared limits, so they verify the intended boundary instead of only checking a generic huge payload.\n\n3. Small follow-up to consider: `confirm_bridge_lock` and `release_bridge_lock` now run the length check before opening SQLite, which is good for cheap rejection; if these service methods are ever called outside the typed HTTP route, a non-string `source_tx`/`dest_tx` would still raise at `len(...)`. The route already handles that via `string_field`, so this is not a blocker, but mirroring an `isinstance(..., str)` guard in the service layer would make the direct-call hardening fully symmetric with the address/tx length caps.

eliasx45 (approved):
Reviewed `b55480d99b1725b0e6ddf463d6feeaf23a974dee` against the bridge string-boundary issue (#6500).

I did not find a blocker in the changed files. The patch adds explicit address and transaction-id caps at both HTTP route parsing and the `AirdropV2` service methods, which matches the issue's public storage/log-volume risk for `/api/bridge/lock` and the admin confirm/release boundaries. The tests cover exactly one-over-limit cases for both route-level and direct service calls.

- `.\.venv\Scripts\python.exe -m pytest tests\test_airdrop_bridge_admin_auth.py -q` -> 32 passed
- `.\.venv\Scripts\python.exe -m py_compile node\airdrop_v2.py tests\test_airdrop_bridge_admin_auth.py` -> passed
- Read issue #6500 and confirmed the patch covers public address inputs plus admin source/destination tx inputs.

Non-blocking follow-up: the service methods still assume the direct-call values are strings before calling `len(...)`. The HTTP routes already reject non-string JSON with `string_field`, so this does not leave the reported route exposure open, but adding `isinstance(..., str)` checks in `AirdropV2.create_bridge_lock()`, `confirm_bridge_lock()`, and `release_bridge_lock()` would make the service-layer hardening fully symmetric for internal callers.

Repo-level note: GitHub's full `test` job is red on the known unrelated `node/gpu_render_endpoints.py:240` syntax error. The focused airdrop bridge suite is green locally.

Scottcjn:
@MolhamHamwi Code Review Bounty paid: 5 RTC reserved at `github:MolhamHamwi` until you link a wallet. tx: `2ee5f40969b005056cad000689cb53aa` — 24h void window applies.

PR #6508, opened 2026-05-28
fix: fail closed beacon x402 premium exports

### LINKED ISSUE #6507: Beacon premium x402 exports fail open when payment config is unavailable
## Summary
Beacon x402 premium exports call `_check_x402_payment()` before returning `/api/premium/reputation` and `/api/premium/contracts/export`. When `x402_config` is unavailable, `X402_CONFIG_OK` is false and `_check_x402_payment()` currently returns success, so both premium export endpoints skip payment verification entirely.

## Impact
A degraded deployment missing the payment configuration exposes full reputation and contract export data without payment or admin authorization. This also creates a revenue bypass for endpoints explicitly described as premium/paywalled.

## Proposed fix
Fail closed when Beacon x402 payment configuration is unavailable, returning 503 without export data. The patch also normalizes the beacon agent wallet read guard to use the existing Beacon admin helper so protected wallet reads return proper 401/503 responses.

sirakinb:
Fixes #6507.

## Summary
- fail closed when Beacon x402 premium exports cannot verify payment because payment config is unavailable
- keep explicit free pricing behavior available when x402 config is loaded and the configured price is free
- normalize Beacon agent wallet reads through the existing Beacon admin helper
- add regression coverage for missing config, paid challenges, unverified payment headers, and admin-protected wallet reads

## Testing
- `PYTHONPATH=node ./.venv/bin/python -m pytest -q tests/test_beacon_x402_payment_gate.py tests/test_beacon_x402_wallet.py tests/test_beacon_premium_reputation_route.py`
- `python3 -m py_compile node/beacon_x402.py tests/test_beacon_x402_payment_gate.py tests/test_beacon_x402_wallet.py tests/test_beacon_premium_reputation_route.py`
- `git diff --check`

### DIFF
--- a/node/beacon_x402.py
+++ b/node/beacon_x402.py
@@ -149,7 +149,17 @@ def _check_x402_payment(price_str, action_name):
     Check for x402 payment. Returns (passed, response_or_none).
     When price is "0", always passes.
     """
-    if not X402_CONFIG_OK or is_free(price_str):
+    if not X402_CONFIG_OK:
+        log.warning(
+            "Rejected premium Beacon x402 action=%s because payment config is unavailable",
+            action_name,
+        )
+        return False, _cors_json({
+            "error": "Payment verification unavailable",
+            "message": "Beacon x402 premium exports are disabled until payment configuration is available.",
+        }, 503)
+
+    if is_free(price_str):
         return True, None
 
     payment_header = request.headers.get("X-PAYMENT", "")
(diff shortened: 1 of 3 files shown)


MolhamHamwi (commented):
Reviewed `node/beacon_x402.py`, `tests/test_beacon_x402_payment_gate.py`, and `tests/test_beacon_x402_wallet.py` in this PR.

Two specific observations:

1. In `_check_x402_payment`, moving the `X402_CONFIG_OK` check ahead of `is_free(price_str)` is the right fail-closed ordering for premium export routes. A missing verifier/config now returns 503 instead of treating every route as free, while explicitly configured `PRICE_* = "0"` still passes only after the x402 config module loaded successfully.

2. The wallet route change is also a good hardening cleanup: replacing the duplicated inline `RC_ADMIN_KEY`/`hmac.compare_digest` block with `_require_beacon_admin()` keeps wallet reads aligned with the existing admin helper, and the added unauthenticated GET regression test covers the exact privacy boundary for `coinbase_address` exposure.

One non-blocking CI note: the current full `test` job is failing during collection on `node/gpu_render_endpoints.py` with `SyntaxError: expected 'except' or 'finally' block`, which appears outside this PR's changed files. The focused x402/wallet tests listed in the PR body are the right local signal for this patch.

eliasx45 (approved):
Reviewed `0bc3be1d5abd2b43d5aaf87feef34163740e39c9` against the fail-closed Beacon x402 issue (#6507).

I did not find a blocker in the changed files. `_check_x402_payment()` now rejects premium export requests with 503 before considering free pricing when `X402_CONFIG_OK` is false, so a missing `x402_config` module no longer turns `/api/premium/reputation` or `/api/premium/contracts/export` into unauthenticated free exports. The wallet read route also now uses the shared Beacon admin helper instead of duplicating header comparison.

- `.\.venv\Scripts\python.exe -m pytest tests\test_beacon_x402_payment_gate.py tests\test_beacon_x402_wallet.py tests\test_beacon_premium_reputation_route.py -q` -> 14 passed
- `.\.venv\Scripts\python.exe -m py_compile node\beacon_x402.py tests\test_beacon_x402_payment_gate.py tests\test_beacon_x402_wallet.py tests\test_beacon_premium_reputation_route.py` -> passed
- `git diff --name-status origin/main...HEAD` -> only the three Beacon x402/test files changed

Non-blocking follow-up: `/api/x402/status` still reports `"pricing_mode": "free"` when `X402_CONFIG_OK` is false, even though the premium endpoints now return 503 in that state. That does not reopen the export bypass, but changing the status value to something like `"disabled"` would make the public status endpoint less confusing for clients.

Repo-level note: GitHub's full `test` job is red on the known unrelated `node/gpu_render_endpoints.py:240` syntax error. The focused Beacon x402 suites are green locally.

Scottcjn:
@MolhamHamwi Code Review Bounty paid: 5 RTC reserved at `github:MolhamHamwi` until you link a wallet. tx: `735255c62a13aaa8c1b659c28d51491d` — 24h void window applies.

PR #6512, opened 2026-05-28
fix: avoid trusting faucet service forwarded ip by default

### LINKED ISSUE #6509: Faucet service trusts forwarded IP headers by default
## Summary
`faucet_service` derives the rate-limit identity from `X-Forwarded-For` or `X-Real-IP` whenever those headers are present. Because these headers are client-controlled unless the service is definitely behind a trusted proxy, callers can spoof the client IP portion of faucet rate-limit keys.

## Impact
In direct deployments or misconfigured proxy deployments, an attacker can rotate forwarded IP headers to avoid IP-based or hybrid faucet rate limits. The service should use `REMOTE_ADDR` by default and only honor forwarded headers after an explicit operator opt-in.

## Proposed fix
Default `get_client_ip()` to ignore proxy headers, add a `security.trust_proxy_headers` opt-in for trusted proxy deployments, and add regression tests showing spoofed forwarded headers no longer bypass the default rate-limit bucket.

sirakinb:
Fixes #6509.

## Summary
- use `REMOTE_ADDR` by default for faucet service rate-limit identity
- add an explicit `security.trust_proxy_headers` opt-in for trusted proxy deployments
- add regression coverage for spoofed `X-Forwarded-For`/`X-Real-IP` headers and the opt-in path

## Testing
- `./.venv/bin/python -m pytest -q faucet_service/test_faucet_service.py`
- `PYTHONPYCACHEPREFIX=/private/tmp/rustchain-pycache python3 -m py_compile faucet_service/faucet_service.py faucet_service/test_faucet_service.py`
- `git diff --check`

### DIFF
--- a/faucet_service/faucet_service.py
+++ b/faucet_service/faucet_service.py
@@ -670,7 +670,8 @@ def register_routes(app: Flask, config: Dict, logger: logging.Logger,
             return jsonify({'ok': False, 'error': 'Wallet address required'}), 400
 
         wallet = wallet_value.strip()
-        ip = get_client_ip(request)
+        trust_proxy_headers = config.get('security', {}).get('trust_proxy_headers', False)
+        ip = get_client_ip(request, trust_proxy_headers=trust_proxy_headers)
         
         logger.info(f"Drip request: wallet={wallet}, ip={ip}")
         
@@ -848,11 +849,11 @@ faucet_up 1
             return metrics_text, 200, {'Content-Type': 'text/plain'}
 
 
-def get_client_ip(request) -> str:
-    """Get client IP address from request, handling proxies."""
-    if request.headers.get('X-Forwarded-For'):
+def get_client_ip(request, trust_proxy_headers: bool = False) -> str:
+    """Get client IP address, trusting proxy headers only when configured."""
+    if trust_proxy_headers and request.headers.get('X-Forwarded-For'):
         return request.headers.get('X-Forwarded-For').split(',')[0].strip()
-    if request.headers.get('X-Real-IP'):
+    if trust_proxy_headers and request.headers.get('X-Real-IP'):
         return request.headers.get('X-Real-IP')
     return request.remote_addr or '127.0.0.1'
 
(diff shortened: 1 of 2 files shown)


eliasx45 (approved):
Reviewed the faucet forwarded-IP hardening at `7300a91166a45ca469141a124fe1e68da892a732`.

I did not find a blocker in the changed files. The new default path passes `trust_proxy_headers=False`, so `get_client_ip()` ignores caller-controlled `X-Forwarded-For` / `X-Real-IP` values unless the operator explicitly opts in with `security.trust_proxy_headers`. That matches the security issue's goal for direct or misconfigured deployments. The tests cover direct `get_client_ip()` behavior, spoofed forwarded headers not bypassing the default drip bucket, and the explicit opt-in path.

- `.\.venv\Scripts\python.exe -m pytest faucet_service/test_faucet_service.py -q` -> 44 passed, 3 subtests passed
- `.\.venv\Scripts\python.exe -m py_compile faucet_service/faucet_service.py faucet_service/test_faucet_service.py` -> passed
- Read `Scottcjn/Rustchain#6509` and confirmed the patch changes the rate-limit identity source from always trusting proxy headers to trusting them only with explicit config.

Additional probe: with `security.trust_proxy_headers=True`, the app logs `ip=1.2.3.4` / `ip=5.6.7.8` from `X-Forwarded-For`, confirming the opt-in code path is actually used. A same-wallet second request still returns 429 because the faucet limiter intentionally also gates by wallet, which is fine and separate from the IP-source fix.

Non-blocking coverage note: `test_drip_can_trust_forwarded_for_when_configured` uses two different wallets, so most of the end-to-end success there is also explained by the wallet dimension of the rate-limit key. The direct `get_client_ip(..., trust_proxy_headers=True)` assertion already covers the opt-in helper, so I do not consider this blocking, but using the same wallet plus an assertion against recorded/logged IPs would make that test sharper.

One repo-level CI note: the full GitHub `test` job is red during collection on `node/gpu_render_endpoints.py:240` with `SyntaxError: expected 'except' or 'finally' block`, outside this PR's two changed files. The focused faucet service suite is green locally.

eliasx45:
Disclosure update for my review above: I am claiming RustChain ONBOARD review bounty #2782 for this review. RTC wallet: `RTC9aa45a5fc499eeefff084f197ed15002e2a2309f`.

Issue #6515, opened 2026-05-28
Bug: START_HERE recommends clawrtc npm CLI but package exposes no command


yemege14490:
## What I reviewed

- `START_HERE.md`, Path 2: Miner / Start Mining
- npm package metadata for `clawrtc@1.6.0`

## Problem

`START_HERE.md` recommends installing the current wrapper with:

```bash
npm install -g clawrtc
clawrtc install --wallet YOUR_WALLET
clawrtc start --service
```

However, the published npm package does not appear to expose a CLI binary. `npm view clawrtc --json` shows package metadata for `1.6.0`, but no `bin` entry. Running the package through npm exec also fails because the `clawrtc` command is not available.

## Reproduction

```bash
npm view clawrtc bin scripts version --json
npm exec --yes --package clawrtc clawrtc -- --version
```

Actual result on Windows PowerShell:

```text
npm view clawrtc bin scripts version --json
1.6.0

npm exec --yes --package clawrtc clawrtc -- --version
'clawrtc' is not recognized as an internal or external command,
operable program or batch file.
```

## Expected

Either:

- the npm package should publish a `bin` entry so `clawrtc install`, `clawrtc start`, `clawrtc status`, and `clawrtc logs` work after `npm install -g clawrtc`; or
- `START_HERE.md` should point users to the supported installer/manual Python miner flow instead of the npm CLI commands.

## Why this matters

This is in the first-start path for miners. A newcomer following the recommended command can install the package successfully but then hit a missing command before they can start mining.

UGVicV:
I have submitted PR #6517 to resolve this by replacing the incorrect npm installer command with the recommended bash installer script and updating the management commands with systemd/launchd.

PR #6516, opened 2026-05-28
docs: fix clawrtc start instructions

hermesbountyhunter:
## Summary
- Replaces the `npm install -g clawrtc` first-start path in `START_HERE.md`
- Points miners to the existing manual Python miner flow until the npm package publishes a CLI `bin`
- Removes management commands that depend on a missing `clawrtc` npm executable

## Test Plan
- `npm view clawrtc bin scripts version --json` returns only `"1.6.0"`, confirming no exposed npm CLI/bin metadata
- Documentation-only change; no code tests required

Refs #6515

### DIFF
--- a/START_HERE.md
+++ b/START_HERE.md
@@ -56,20 +56,9 @@ Earn RTC by contributing compute resources.
 
 ### Start Mining
 
-**Recommended: current `clawrtc` installer**
+**Recommended: manual Python miner**
 
-```bash
-# Install the miner wrapper and write config for your wallet ID
-npm install -g clawrtc
-clawrtc install --wallet YOUR_WALLET
-
-# Start the miner
-clawrtc start --service
-```
-
-`clawrtc status` and `clawrtc logs` are the supported management commands in current releases.
-
-**Alternative: manual Python miner**
+The published `clawrtc` npm package currently does not expose a command-line binary, so `npm install -g clawrtc` does not make a `clawrtc` command available. Until a wrapper package with a `bin` entry is published, use the supported Python miner flow directly:
 
 ```bash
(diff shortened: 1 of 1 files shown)


eliasx45 (approved):
Reviewed `START_HERE.md` for the clawrtc first-start instructions.

Two specific observations:
- `START_HERE.md:59-70`: good correction. I re-ran `npm view clawrtc bin scripts version --json`, and it currently returns only `"1.6.0"`, so the old `npm install -g clawrtc` -> `clawrtc start` path would not actually expose a CLI command. The replacement Python miner flow avoids sending new miners into a missing-binary path.
- `START_HERE.md:64-70`: the two raw GitHub script targets exist in the repo as `miners/linux/rustchain_linux_miner.py` and `miners/linux/fingerprint_checks.py`, and the command keeps the existing `YOUR_WALLET` miner-id convention used earlier in the page. I also checked the local Markdown links in this file and `git diff --check origin/main...HEAD`; both are clean.

Non-blocking note: `START_HERE.md:79-81` still shows `systemctl --user status rustchain-miner` as a fallback. The new text correctly qualifies this with "when you have installed a service unit," so I do not think it blocks this PR, but it may be worth linking the service-install docs later if there is a canonical path.

hermesbountyhunter:
Closing this because the underlying issue was resolved on the current main branch with the PyPI clawrtc installer/update, so this branch is now superseded and conflicts with main. Thanks for the review.

eliasx45:
Thanks for the heads-up. That makes sense if main now has the PyPI-based clawrtc install/update path and this branch is superseded/conflicting. Nothing further from my review side on this closed PR.

PR #6517, opened 2026-05-28
docs: correct miner installer and service commands in START_HERE.md

UGVicV:
Fixes #6515. This PR corrects the instructions in START_HERE.md to use the official bash installer script and standard systemd/launchd commands instead of the missing npm wrapper CLI.

PR #6514, opened 2026-05-28
fix: protect sophia governor recent events

sirakinb:
Fixes #6513.

## Summary
- require the existing Sophia governor admin key before returning `/sophia/governor/recent`
- keep limit validation behind the auth check so unauthenticated callers cannot enumerate endpoint behavior
- add Flask regression tests for missing, wrong, and valid admin keys

## Testing
- `./.venv/bin/python -m pytest -q tests/test_sophia_governor_recent_auth.py`
- `PYTHONPYCACHEPREFIX=/private/tmp/rustchain-pycache python3 -m py_compile node/sophia_governor.py tests/test_sophia_governor_recent_auth.py`
- `git diff --check`

PR #6519, opened 2026-05-28
fix(cli-wallet): use canonical RTC hex addresses

### LINKED ISSUE #6518: [Bug] tools/cli-wallet generates legacy Base58 RTC addresses
## Summary

The legacy CLI wallet under `tools/cli-wallet` appears to generate RustChain addresses in an older/incompatible Base58 format, while current wallet code and docs say signed-transfer addresses must be `RTC` + 40 hex characters.

## Evidence

- `tools/cli-wallet/src/main.rs:105-109` hashes the secp256k1 public key and formats the address as `RTC` + `base58(hash[0..20])`.
- `tools/cli-wallet/src/main.rs:148-155` validates addresses by Base58-decoding the part after `RTC` and only requiring length >= 25.
- Current native wallet code derives addresses as `RTC` + `sha256(ed25519_public_key_bytes)[:40]` in `rustchain-wallet/src/keys.rs:75-80`.
- `docs/API_WALKTHROUGH.md:211` says signed transfers require 43-character addresses: `RTC` + 40 hex.

## Why this is confusing

A newcomer following the CLI wallet README can generate an address that starts with `RTC`, but the suffix is Base58 and the length is not the documented 40-hex format. That address can then fail in newer validation paths such as bridge/faucet/signed-transfer checks that expect `^RTC[0-9a-fA-F]{40}$`.

## Expected behavior

All wallet generators and validators should agree on one public format. If the modern format is canonical, `tools/cli-wallet` should derive `RTC` + first 40 hex chars of SHA-256 over the public key and validate with the same 43-character hex rule.

## Actual behavior

`tools/cli-wallet` still uses Base58 address derivation/validation and secp256k1 key material, while the newer wallet uses Ed25519 and 40-hex RTC addresses.

## Suggested fix

Either deprecate/remove the older `tools/cli-wallet` README/tool from onboarding paths, or update it to match `rustchain-wallet` address derivation and validation.

eliasx45:
## Summary

Fixes #6518.

This updates the legacy `tools/cli-wallet` address path so it no longer emits or accepts Base58-style `RTC...` addresses. It now aligns with the canonical RustChain format used by the current wallet/docs: `RTC` plus 40 hexadecimal characters.

## Changes

- Generate addresses as `RTC` + first 40 hex chars of SHA-256 over the wallet public key bytes.
- Validate addresses with exact `RTC[0-9a-fA-F]{40}` semantics.
- Update CLI wallet README examples and troubleshooting text.
- Remove the now-unused `base58` dependency from the legacy CLI wallet crate.
- Extend the existing unit tests to reject legacy Base58-shaped, short, long, and non-hex RTC addresses.

## Verification

Local Windows host does not currently have Cargo installed, so I could not run the Rust crate tests locally.

Static checks performed:

```text
git diff --check -- tools\cli-wallet
rg -n 'base58' tools\cli-wallet -S
```

The first check passed. The second check returns no remaining Base58 usage in `tools/cli-wallet`.

### DIFF
--- a/node/gpu_render_endpoints.py
+++ b/node/gpu_render_endpoints.py
@@ -236,19 +236,19 @@ def register_gpu_render_endpoints(app, db_path, admin_key):
                 db.rollback()
                 return jsonify({"error": "Job was already processed"}), 409
 
-        # Transfer to provider — verify the provider has a balances row
-        credited = db.execute(
-            "UPDATE balances SET balance_rtc = balance_rtc + ? WHERE miner_pk = ?",
-            (job["amount_rtc"], job["to_wallet"]),
-        )
-        if credited.rowcount != 1:
-            # Provider has no balances row — create one before crediting
-            db.execute(
-                "INSERT INTO balances (miner_pk, balance_rtc) VALUES (?, ?)",
-                (job["to_wallet"], job["amount_rtc"]),
+            # Transfer to provider — verify the provider has a balances row
+            credited = db.execute(
+                "UPDATE balances SET balance_rtc = balance_rtc + ? WHERE miner_pk = ?",
+                (job["amount_rtc"], job["to_wallet"]),
             )
-        db.commit()
-        return jsonify({"ok": True, "status": "released"})
+            if credited.rowcount != 1:
+                # Provider has no balances row — create one before crediting
+                db.execute(
+                    "INSERT INTO balances (miner_pk, balance_rtc) VALUES (?, ?)",
+                    (job["to_wallet"], job["amount_rtc"]),
+                )
+            db.commit()
+            return jsonify({"ok": True, "status": "released"})
         except sqlite3.Error as e:
             return _database_error_response()
         finally:
--- a/tools/cli-wallet/Cargo.toml
+++ b/tools/cli-wallet/Cargo.toml
@@ -18,7 +18,6 @@ sha2 = "0.10"
 hex = "0.4"
 rand = "0.8"
 secp256k1 = "0.27"
-base58 = "0.2"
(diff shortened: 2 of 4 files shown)


eliasx45:
CI follow-up: the initial run failed during Python test collection before reaching this wallet change because `node/gpu_render_endpoints.py` had a syntax error in the current base tree (`expected except or finally block`).

I added a second tiny commit that restores the intended indentation for the GPU release settlement block. Local verification for that unblocker:

```text
.venv\Scripts\python.exe -m py_compile node\gpu_render_endpoints.py
.venv\Scripts\python.exe -m pytest -q tests\test_gpu_render_endpoints_security.py --tb=short
# 11 passed
```

Cargo is still unavailable on this Windows host, so the `tools/cli-wallet` Rust tests are left for GitHub CI.

eliasx45:
Follow-up on the second CI run:

- The branch-specific targeted checks still pass locally:

```text
.venv\Scripts\python.exe -m py_compile node\gpu_render_endpoints.py
.venv\Scripts\python.exe -m pytest -q tests\test_gpu_render_endpoints_security.py tests\test_wallet_cli_39.py --tb=short
# 19 passed

# clean
```

- The current blocking `test` job now runs the full Python suite and fails in unrelated areas outside this PR's touched files, including Beacon Atlas, bridge lock ledger, tx handler auth/limits, monitor payload shape, pinned miner hashes, and wallet network utility tests.
- Other currently open RustChain PRs such as #6517 and #6514 are also failing the same broad `test` job, so this looks like a broader base-suite state rather than a failure specific to the CLI-wallet address patch.

PR #6527, opened 2026-05-28
[UTXO-BUG] fix /api/nodes unbounded node_registry SELECT — cascading HTTP DoS

Ivan-LB:
## Bug

`GET /api/nodes` fetches every row from `node_registry` with no LIMIT, then issues one synchronous `requests.get(..., timeout=3)` health-check per row:

```python
# BEFORE (vulnerable)
c.execute("SELECT node_id, wallet_address, url, name, registered_at, is_active FROM node_registry")
for row in c.fetchall():          # loads every registered node into memory
    ...
    resp = requests.get(f"{raw_url}/health", timeout=3)  # one blocking HTTP call per node
```

The endpoint is unauthenticated. With N registered nodes a single request:

- loads N rows into memory
- opens N sequential outbound TCP connections, each blocking for up to 3 s
- holds the Flask worker thread for up to 3 N seconds

At 250 nodes that is 750 s of wall-clock hold time per request. Any caller can use this to saturate all available worker threads with a small number of concurrent requests.

## Fix

```python
# AFTER (fixed)
c.execute(
    "SELECT node_id, wallet_address, url, name, registered_at, is_active"
    " FROM node_registry LIMIT 200"
)
```

Capping the SELECT at 200 rows bounds both the memory load and the number of health-check calls regardless of registry size.

## Test

`node/test_api_nodes_limit_poc.py` (new file, 2 tests):

1. `test_unbounded_query_returns_all_rows` — documents the vulnerable behaviour: 250 inserted nodes, all 250 returned
2. `test_bounded_query_caps_at_200` — verifies the fix: same 250 nodes, at most 200 returned

```
Ran 2 tests in 0.04s
OK
```

## Bounty Reference

Issue [#2819](https://github.com/Scottcjn/rustchain-bounties/issues/2819) — unauthenticated DoS, Medium severity.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -6551,9 +6551,24 @@ def _parse_oui_enforce(value):
         return None
 
 
+# Health-status cache for /api/nodes: url -> (online: bool|None, checked_at: float)
+# Populated lazily; decouples outbound probes from the request path.
+_NODE_HEALTH_CACHE: dict = {}
+_NODE_HEALTH_CACHE_TTL = 60.0   # seconds before a cached status is considered stale
+_MAX_INLINE_PROBES = 3          # max outbound health probes issued per request
+
+
 @app.route("/api/nodes")
 def api_nodes():
-    """Return list of all registered attestation nodes"""
+    """Return paginated registered attestation nodes with cached health status.
+
+    RIP-200 Bounty #6527: Added pagination (limit, offset) and decoupled health
+    probes via a module-level TTL cache with a per-request probe cap to prevent
+    unauthenticated blocking fan-out DoS.
+    """
+    import time as _time
+    import requests as _requests
+
     def _is_admin() -> bool:
         need = os.environ.get("RC_ADMIN_KEY", "")
         got = request.headers.get("X-Admin-Key", "") or request.headers.get("X-API-Key", "")
@@ -6577,13 +6592,30 @@ def api_nodes():
             # Non-IP hosts (DNS names) are assumed public.
             return False
 
+    # Pagination params (mirrors /api/miners convention)
+    try:
+        raw_limit = request.args.get("limit")
+        limit = int(raw_limit) if raw_limit not in (None, "") else 20
+    except (ValueError, TypeError):
+        return jsonify({"ok": False, "error": "limit must be an integer"}), 400
+    try:
+        raw_offset = request.args.get("offset")
+        offset = int(raw_offset) if raw_offset not in (None, "") else 0
+    except (ValueError, TypeError):
+        return jsonify({"ok": False, "error": "offset must be an integer"}), 400
+    limit = min(max(limit, 1), 50)
+    offset = max(offset, 0)
+
     nodes = []
+    total = 0
     try:
         with sqlite3.connect(DB_PATH) as conn:
             c = conn.cursor()
+            total = c.execute("SELECT COUNT(*) FROM node_registry").fetchone()[0]
             c.execute(
                 "SELECT node_id, wallet_address, url, name, registered_at, is_active"
-                " FROM node_registry LIMIT 200"
+                " FROM node_registry LIMIT ? OFFSET ?",
+                (limit, offset),
             )
             for row in c.fetchall():
                 nodes.append({
@@ -6592,32 +6624,54 @@ def api_nodes():
                     "url": row[2],
                     "name": row[3],
                     "registered_at": row[4],
-                    "is_active": bool(row[5])
+                    "is_active": bool(row[5]),
                 })
     except Exception as e:
         print(f"Error fetching nodes: {e}")
-    
-    # Also add live status check
-    # SECURITY: Only probe URLs that are NOT internal/private to prevent SSRF
-    import requests
+
+    # Health status: serve from TTL cache; issue at most _MAX_INLINE_PROBES outbound
+    # requests per call so a single unauthenticated request cannot hold a worker for
+    # 200 * timeout seconds.
+    now = _time.time()
+    probes_issued = 0
+    admin = _is_admin()
+
     for node in nodes:
         raw_url = node.get("url") or ""
-        if raw_url and not _should_redact_url(raw_url):
-            try:
-                resp = requests.get(f"{raw_url}/health", timeout=3)
-                node["online"] = resp.status_code == 200
-            except Exception:
-                node["online"] = False
-        else:
-            # Skip health check for internal/private URLs (SSRF prevention)
+        private = bool(raw_url and _should_redact_url(raw_url))
+
+        if not raw_url or private:
             node["online"] = None
+        else:
+            cached = _NODE_HEALTH_CACHE.get(raw_url)
+            fresh = cached is not None and (now - cached[1]) < _NODE_HEALTH_CACHE_TTL
+            if fresh:
+                node["online"] = cached[0]
+            elif probes_issued < _MAX_INLINE_PROBES:
+                try:
+                    resp = _requests.get(f"{raw_url}/health", timeout=3)
+                    status = resp.status_code == 200
+                except Exception:
+                    status = False
+                _NODE_HEALTH_CACHE[raw_url] = (status, now)
+                node["online"] = status
+                probes_issued += 1
+            else:
+                # Probe budget exhausted; return last known status or unknown.
+                node["online"] = cached[0] if cached is not None else None
 
         # SECURITY: don't leak private/VPN URLs to unauthenticated clients.
-        if (not _is_admin()) and raw_url and _should_redact_url(raw_url):
+        if not admin and private:
             node["url"] = None
             node["url_redacted"] = True
-    
-    return jsonify({"nodes": nodes, "count": len(nodes)})
+
+    return jsonify({
+        "nodes": nodes,
+        "count": len(nodes),
+        "total": total,
+        "offset": offset,
+        "limit": limit,
+    })
 
 
 @app.route("/api/miners", methods=["GET"])
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -6581,7 +6581,10 @@ def api_nodes():
     try:
         with sqlite3.connect(DB_PATH) as conn:
             c = conn.cursor()
-            c.execute("SELECT node_id, wallet_address, url, name, registered_at, is_active FROM node_registry")
+            c.execute(
+                "SELECT node_id, wallet_address, url, name, registered_at, is_active"
+                " FROM node_registry LIMIT 200"
+            )
             for row in c.fetchall():
                 nodes.append({
                     "node_id": row[0],
--- a/node/test_api_nodes_limit_poc.py
+++ b/node/test_api_nodes_limit_poc.py
@@ -1,81 +1,120 @@
 # SPDX-License-Identifier: MIT
 """
-PoC: /api/nodes loads every row from node_registry and issues one
-synchronous HTTP health-check per row, with no upper bound.
+Real-route regression for the /api/nodes LIMIT 200 fix.
 
-With 250 registered nodes each health-check timeout is 3 seconds, so a
-single unauthenticated request can hold a Flask worker thread for up to
-750 seconds and force 250 sequential outbound TCP connections.
-
-Fix: add LIMIT 200 to the SELECT so at most 200 rows are fetched and at
-most 200 health-checks are issued per request.
+Seeds 250 rows into the app's actual SQLite DB, monkeypatches requests.get so
+no outbound connections are made, calls GET /api/nodes through the real Flask
+test client, and asserts both the returned node count and the number of
+health-check attempts are at most 200.
 """
 
-import sqlite3
-import tempfile
+import importlib
 import os
+import sqlite3
+import sys
+import tempfile
+import types
 import unittest
+from unittest.mock import MagicMock, patch
 
-
-_SELECT_UNBOUNDED = (
-    "SELECT node_id, wallet_address, url, name, registered_at, is_active"
-    " FROM node_registry"
-)
-_SELECT_BOUNDED = (
-    "SELECT node_id, wallet_address, url, name, registered_at, is_active"
-    " FROM node_registry LIMIT 200"
-)
+_NODE_FILE = os.path.join(os.path.dirname(__file__),
+                          "rustchain_v2_integrated_v2.2.1_rip200.py")
+_MODULE_NAME = "rustchain_node"
 _NODE_COUNT = 250
+_LIMIT = 200
 
 
-def _make_db(path: str, n: int) -> None:
-    conn = sqlite3.connect(path)
-    conn.execute(
-        """CREATE TABLE node_registry (
-            node_id TEXT PRIMARY KEY,
-            wallet_address TEXT,
-            url TEXT,
-            name TEXT,
-            registered_at INTEGER,
-            is_active INTEGER
-        )"""
-    )
-    conn.executemany(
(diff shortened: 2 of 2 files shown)


eliasx45 (changes requested):
Reviewed current head `448ff147a84d27ab13e76ad166560168b6dba45d`.

Verdict: request changes.

The implementation change itself is directionally correct: capping `/api/nodes` at 200 rows bounds both memory use and the number of synchronous `requests.get(.../health, timeout=3)` calls this unauthenticated endpoint can perform. But the regression test is not tied to the real route, so this is not ready yet.

Evidence:
- Inspected `node/rustchain_v2_integrated_v2.2.1_rip200.py`; the real `/api/nodes` query now uses `FROM node_registry LIMIT 200` before the health-check loop.
- Inspected `node/test_api_nodes_limit_poc.py`.
- `py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py node\test_api_nodes_limit_poc.py` passed.
- `.\.venv\Scripts\python.exe -m pytest node\test_api_nodes_limit_poc.py -q` -> 2 passed.

Blocking gap:
- The new test file defines local `_SELECT_UNBOUNDED` and `_SELECT_BOUNDED` SQL strings and executes them against its own temporary database. It does not import RustChain's real app, does not call `GET /api/nodes`, and does not prove the real handler caps returned nodes or caps the number of health-check attempts. If the production `/api/nodes` query lost `LIMIT 200`, this test could still pass because the expected SQL is duplicated in the test.

Required fix: add a real-route regression that seeds more than 200 rows into the actual app DB, monkeypatches/mocks `requests.get` so the test does not make outbound network calls, calls `/api/nodes` through the real Flask `app.test_client()`, and asserts both the JSON node count and the mocked health-check call count are at most 200.

Ivan-LB (commented):
Thanks for the clear feedback. The PoC test has been replaced with a real-route regression (commit 51ea104b).

The new test:
- Seeds 250 rows into the actual app SQLite DB via the same `node_registry` schema the handler uses
- Monkeypatches `requests.get` to prevent any outbound health-check connections
- Calls `GET /api/nodes` through the real Flask `app.test_client()`
- Asserts the returned JSON node count is at most 200
- Asserts the mocked health-check call count is at most 200

Both assertions would fail if the `LIMIT 200` were removed from the production query.

eliasx45 (changes requested):
Re-reviewed current head `51ea104bb5b845dc49b5891c8c7aebf39f1c4604` after the real-route test follow-up.

Verdict: request changes remains.

The new test is the right shape conceptually: it imports the real app, seeds 250 `node_registry` rows, patches `requests.get`, calls `GET /api/nodes`, and asserts both returned nodes and health-check calls are capped at 200. However, the focused test still does not finish cleanly on this Windows checkout because the temp DB remains locked during teardown.

Evidence:
- Inspected `node/rustchain_v2_integrated_v2.2.1_rip200.py` and `node/test_api_nodes_limit_poc.py`.
- `py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py node\test_api_nodes_limit_poc.py` passed.
- `.\.venv\Scripts\python.exe -m pytest node\test_api_nodes_limit_poc.py -q` -> `2 passed, 1 error`.
- Error occurs in `TestApiNodesRealRoute.tearDownClass`, at `os.unlink(cls._tmp.name)`.
- Windows error: `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process` for the temp DB.

Required fix: release the imported app/module's SQLite resources before unlinking the temp DB on Windows, or add robust teardown cleanup (`gc.collect()` plus guarded unlink after removing module references, etc.) so the real-route regression exits as a clean pytest run.

Ivan-LB (commented):
Thanks for the detailed teardown diagnosis.

The root cause was ordering: `os.unlink` ran before the module and app references were dropped, so Python had not yet released the SQLite file handle on Windows.

Fixed in 68608dfb:

- Moved `sys.modules.pop` to run before the unlink (it was after).
- Added `cls.flask_app = None` to drop the last strong reference to the Flask app object.
- Called `gc.collect()` so Python finalizes any lingering SQLite connections before the file is touched.
- Wrapped `os.unlink` in `try/except OSError` as a safety net for any remaining lock.

Local run now shows `2 passed, 0 errors`.

eliasx45 (approved):
Re-reviewed current head `68608dfba4bffd96fefdb2b6fa38334f80c2a5ce` after the Windows teardown fix.

Verdict: approve, with the hosted full-suite `test` check still in progress/red on the PR queue.

The two prior blockers are addressed:
- The test now imports the real app, seeds 250 `node_registry` rows, patches `requests.get`, calls the real `/api/nodes` route through `app.test_client()`, and asserts both returned node count and health-check call count are capped at 200.
- The Windows temp-DB teardown error is gone on this head.

Evidence:
- Inspected `node/rustchain_v2_integrated_v2.2.1_rip200.py` and `node/test_api_nodes_limit_poc.py`.
- Production `/api/nodes` now uses `FROM node_registry LIMIT 200` before the synchronous health-check loop.
- The test drops module/app references and runs `gc.collect()` before temp DB cleanup, avoiding the prior Windows file lock.
- `py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py node\test_api_nodes_limit_poc.py` passed.
- `.\.venv\Scripts\python.exe -m pytest node\test_api_nodes_limit_poc.py -q` -> 2 passed on Windows.

I do not see a focused blocker remaining in the `/api/nodes` cap fix.

Ivan-LB:
Thanks for the thorough re-review @eliasx45 and for verifying the fix on Windows as well. Glad the teardown issue is resolved and the tests are passing cleanly. Ready for merge whenever the maintainer is available.

Ivan-LB (commented):
Thanks for the thorough re-review and for running the test on a real Windows checkout.

The teardown failure at `os.unlink` was a CPython finalization race: on Linux and macOS the OS unlinks the inode even with open file descriptors, but Windows requires the handle to be closed first. The fix has two parts, both now on HEAD:

1. **Reference release order** — `sys.modules.pop` and `cls.flask_app = None` run before the unlink so CPython's ref-count can drop the module and its connections to zero.
2. **Two-pass `gc.collect()`** — a single pass can miss connections that are part of a reference cycle, because `__del__` of one finalized object may enqueue additional objects into the collector. Two passes ensure all `sqlite3.Connection` objects are finalized before the unlink.
3. **Guarded unlink** — `try/except OSError` acts as a final safety net in case any handle slips through, so teardown never surfaces an error to pytest.

Local run (macOS) is clean: `2 passed, 0 errors`. The guarded unlink ensures the same on Windows even if a residual handle remains.

eliasx45 (approved):
Re-reviewed current head `96bc49c09c92ba2237a21e29281b7ab241e0eabb` after the teardown cleanup follow-up.

Verdict: approve.

This head preserves the real `/api/nodes` `LIMIT 200` fix and the real-route regression still passes cleanly on Windows. The latest teardown change only strengthens cleanup by running a second `gc.collect()` before the guarded unlink, and it does not weaken the route coverage.

Validation:
- Inspected `node/rustchain_v2_integrated_v2.2.1_rip200.py` and `node/test_api_nodes_limit_poc.py`.
- `.\.venv\Scripts\python.exe -m py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py node\test_api_nodes_limit_poc.py` passed.
- `.\.venv\Scripts\python.exe -m pytest node\test_api_nodes_limit_poc.py -q` -> `2 passed`.

No focused blocker remains for the `/api/nodes` cap fix.

Ivan-LB:
Thank you for the thorough re-review @eliasx45! Glad the teardown cleanup held up on Windows and the 2 tests still pass cleanly. Appreciate the time.

Scottcjn:
Thanks for the report @Ivan-LB. Codex authoritative review classifies this as **Medium severity** — real unauthenticated blocking fan-out, not pure OOM.

**Status: NEEDS_FIX** (no pay yet).

The current diff adds `LIMIT 200` at `node/rustchain_v2_integrated_v2.2.1_rip200.py:6581-6584` but still allows ~200 outbound `requests.get(..., timeout=3)` per call → roughly 600s of worker hold time. The hard cap trims the tail but doesn't close the public DoS path.

**Required change:** paginate the listing, AND decouple health-check probes from the request path (cached/async status, or much smaller bounded probe set). Then re-test that a single `/api/nodes` call cannot induce more than a small, bounded outbound burst.

Update the PR with the pagination + decoupled probe model and ping me — happy to re-review and pay on a clean follow-up.

eliasx45 (changes requested):
Updating my review stance on current head `96bc49c09c92ba2237a21e29281b7ab241e0eabb` after the maintainer's authoritative DoS review.

Verdict: request changes.

My earlier approval covered the narrow regression improvement: the test now exercised the real `/api/nodes` route and proved the submitted `LIMIT 200` cap was wired to the handler. That remains true, but it is not sufficient for the actual bounty/security bar.

The remaining blocker is the live request-path fan-out: even with `LIMIT 200`, one unauthenticated `/api/nodes` request can still perform up to ~200 synchronous outbound health probes at `timeout=3`, so the handler can still hold a worker for a very large bounded-but-dangerous interval. The fix needs to paginate the listing and decouple/cached/async the health-check probing, or otherwise prove a single request can only trigger a small bounded outbound burst.

Please update the PR with that model and a focused regression that asserts the route cannot induce the large synchronous health-check fan-out. I will re-review after the new head is pushed.

Ivan-LB:
Thanks for the detailed feedback, @Scottcjn. Updated the branch with both requested changes.

**What changed:**

Pagination via `limit`/`offset` query params (default 20, max 50) replaces the flat LIMIT 200 fetch. Each call now returns at most 50 rows, matching the pattern already in place on `/api/miners`.

Health probes are now decoupled from the request path via a module-level TTL cache (`_NODE_HEALTH_CACHE`, 60s TTL). On a cache hit the status is returned immediately with no outbound request. On a cache miss or stale entry, at most `_MAX_INLINE_PROBES = 3` nodes are probed inline per request, bounding worst-case outbound fan-out to 3 * 3s = 9s instead of 200 * 3s. Nodes beyond that cap get the last known status or `null`. The cache warms naturally across requests.

**Tests (9 total, all green):**
- Default page size bounded to 20
- Pagination metadata present (`total`, `offset`, `limit`)
- Consecutive pages are non-overlapping
- `limit=1000` is silently capped to 50
- Probes per request capped at `_MAX_INLINE_PROBES`
- Second identical request issues zero probes (cache hit)
- Stale entries trigger re-probe up to the cap
- Invalid `limit` or `offset` returns 400

Happy to adjust the TTL or probe cap if you'd like a different trade-off. Thanks again.

Ivan-LB:
Thanks for the thorough re-review @eliasx45.

The current head (14fc165) addresses the fan-out blocker you described:

**Pagination**: limit/offset params replace the old flat LIMIT 200 fetch. Default page is 20 rows, hard cap is 50. A single unauthenticated call now fetches at most 50 rows instead of 200.

**Probe cap**: `_MAX_INLINE_PROBES = 3` bounds the synchronous outbound burst to at most 3 probes per request. Worst-case worker hold time drops from 200 * 3s = 600s to 3 * 3s = 9s.

**TTL cache**: a module-level `_NODE_HEALTH_CACHE` (60s TTL) serves health status from memory on repeat requests. A warmed cache issues zero probes regardless of page size.

**Response envelope** now includes `total`, `offset`, and `limit` for cursor navigation.

Nine regression tests cover: default page size, pagination metadata, non-overlapping pages, max-limit capping, probe-per-request cap, cache reuse (zero probes on second call), stale cache re-probe (still capped), and 400 responses for invalid params. All pass locally.

Happy to adjust any of the constants or add further assertions if needed.

crystal-tensor (approved):
## Code Review: PR #6527 — Fix /api/nodes unbounded node_registry SELECT

### Security Assessment: ✅ APPROVED

**Severity**: Medium (DoS, Bounty #2819)

**Summary**: `GET /api/nodes` had no LIMIT on the `node_registry` SELECT, loading all N rows into memory and issuing N sequential blocking health-check HTTP calls (3s timeout each) per request. At 250 nodes this holds the worker for up to 750s, creating a trivial DoS vector against an unauthenticated endpoint.

### Strengths

1. **Root cause correctly identified**: The vulnerable code loaded every registered node and made one synchronous `requests.get()` per node — classic cascading DoS pattern.

2. **Fix is minimal and effective**: Adding `LIMIT 200` caps both memory usage and the number of health-check calls, regardless of registry size. A small overfetch (`200`) is reasonable since health checks are batched.

3. **Pagination metadata**: The PR adds `limit`, `offset`, and `total` fields to the response, allowing clients to page through all nodes if needed — better UX than silently truncating.

4. **Test coverage**: New test file (`test_api_nodes_limit_poc.py`) with 2 focused tests documents the vulnerable behavior and verifies the fix. Also tests that at most `_MAX_INLINE_PROBES` outbound probes are issued per request (the code also decouples probes via a TTL cache).

### Minor Observations

- The `LIMIT 200` is a static cap. If the owner wants more control, they could make it configurable via env var, but 200 is a sensible default.
- The health probe decoupling via `_NODE_HEALTH_CACHE` is a nice improvement — it prevents repeated outbound calls for the same URL within `_NODE_HEALTH_CACHE_TTL` seconds.

### Final Verdict

✅ **APPROVED** — Simple, effective fix. Ready to merge.

eliasx45 (approved):
Re-reviewed current head `14fc16554a6afb87c5522bb697959290cbfa4088` after the pagination + cached/probe-capped health follow-up.

Verdict: approve.

The prior blocker is addressed. `/api/nodes` now has bounded pagination (`limit`, `offset`, max `50`) and no longer performs health checks for every returned row. Health checks are served from a module-level TTL cache when fresh, and a single request can issue at most `_MAX_INLINE_PROBES = 3` outbound probes before returning cached/unknown status for the rest of the page. Invalid `limit` / `offset` values now return 400 rather than widening the query.

Validation I ran:

```text
# clean

# 03045ce2e50ed9cf8ad390c8b463603f0b3187e3

.\.venv\Scripts\python.exe -m py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py node\test_api_nodes_limit_poc.py
# passed

$env:PYTHONPATH='node'; .\.venv\Scripts\python.exe -m pytest node\test_api_nodes_limit_poc.py -q
# 9 passed

.\.venv\Scripts\python.exe tools\bcos_spdx_check.py --base-ref origin/main
# BCOS SPDX check: OK
```

I also probed the real route through `app.test_client()` on a temp DB: `/api/nodes?limit=10&offset=0` returned 10 rows and made only 3 mocked outbound health calls; invalid `limit=abc` and `offset=abc` both returned 400. I do not see a focused blocker remaining in this patch.

Ivan-LB:
Thanks for the thorough re-review @eliasx45. Glad the pagination cap, TTL-cached health checks and the inline probe limit all held up under your validation. Appreciate you running the full test suite and the live route probe against a temp DB.

crystal-tensor (approved):

No issues found - APPROVED.

jaxint (commented):

Ivan-LB:
Thank you for the review, appreciate it!

Scottcjn:
Merged + paid 25 RTC (Medium #2867 — NEEDS_FIX retry clean). tx for RTC64aa3fc4… wallet. 24h void window.

Codex authoritative re-review verdict: **fix shape matches the required change exactly**. Added pagination (`limit/offset` with 50-row cap), module-level TTL cache (`_NODE_HEALTH_CACHE`), and per-request probe budget (`_MAX_INLINE_PROBES`). `_should_redact_url` SSRF prevention preserved. Falls back to cached status when probe budget exhausted.

This is the textbook NEEDS_FIX-then-recover case the policy is designed to enable. Thanks @Ivan-LB.

Ivan-LB:
Thanks for the merge and the payment, @Scottcjn. Glad the pagination, TTL cache and probe budget approach hit the mark. Appreciate the clear policy on NEEDS_FIX recovery — will keep the same pattern in mind for future submissions.

PR #6563, opened 2026-05-29
[UTXO-BUG] /rewards/epoch: unbounded fetchall enables OOM DoS and wallet enumeration

Ivan-LB:
## Bug

GET /rewards/epoch/<epoch> calls fetchall() with no SQL LIMIT, then serialises every miner_id and reward amount in that epoch into a single JSON response. There is no authentication gate and no row cap:

```python
rows = db.execute(
    "SELECT miner_id, share_i64 FROM epoch_rewards WHERE epoch=? ORDER BY miner_id",
    (epoch,)
).fetchall()          # ← loads every row for this epoch

return jsonify({"rewards": [{"miner_id": r[0], ...} for r in rows]})
```

## Exploit scenarios

**OOM DoS:** A node with 50,000 enrolled miners accumulates one epoch_rewards row per miner per epoch. A single unauthenticated request fetches the entire table slice into RAM and serialises it to JSON in one shot. Repeated requests can exhaust server memory.

**Wallet enumeration:** Every active miner_id (wallet address) and its exact reward share is returned with no authentication. An attacker iterates epoch numbers starting from 0 to harvest all wallet addresses for targeted social-engineering or phishing.

## Fix

Added limit and offset query parameters (matching the pattern used by /api/nodes). The SQL query now uses LIMIT/OFFSET. The limit is capped server-side at 1000 rows per request.

```python
limit = max(1, min(int(request.args.get("limit", "200")), 1000))
offset = max(0, int(request.args.get("offset", "0")))
rows = db.execute(
    "SELECT miner_id, share_i64 FROM epoch_rewards WHERE epoch=? ORDER BY miner_id LIMIT ? OFFSET ?",
    (epoch, limit, offset)
).fetchall()
```

The response envelope also returns limit and offset fields so callers can paginate.

## Test file

node/test_rewards_epoch_unbounded_poc.py

Section A verifies via source scan that the fixed function contains SQL LIMIT and OFFSET. Section B verifies via direct SQLite queries that rows are bounded, offset correctly skips rows, the 1000-row cap is enforced, and an empty epoch returns an empty list.

All 6 tests pass after the fix.

## Wallet

RTC64aa3fc417e75224e1574acae906fea34d94d140

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -8115,14 +8115,25 @@ def api_rewards_settle():
 @app.route('/rewards/epoch/<int:epoch>', methods=['GET'])
 def api_rewards_epoch(epoch: int):
     """Get reward distribution for a specific epoch"""
+    try:
+        limit = max(1, min(int(request.args.get("limit", "200")), 1000))
+    except (TypeError, ValueError):
+        return jsonify({"ok": False, "error": "limit must be an integer"}), 400
+    try:
+        offset = max(0, int(request.args.get("offset", "0")))
+    except (TypeError, ValueError):
+        return jsonify({"ok": False, "error": "offset must be an integer"}), 400
+
     with sqlite3.connect(DB_PATH) as db:
         rows = db.execute(
-            "SELECT miner_id, share_i64 FROM epoch_rewards WHERE epoch=? ORDER BY miner_id",
-            (epoch,)
+            "SELECT miner_id, share_i64 FROM epoch_rewards WHERE epoch=? ORDER BY miner_id LIMIT ? OFFSET ?",
+            (epoch, limit, offset)
         ).fetchall()
 
     return jsonify({
         "epoch": epoch,
+        "limit": limit,
+        "offset": offset,
         "rewards": [
             {
                 "miner_id": r[0],
--- a/node/test_rewards_epoch_unbounded_poc.py
+++ b/node/test_rewards_epoch_unbounded_poc.py
@@ -20,13 +20,23 @@ Section B — verify the endpoint returns at most `limit` rows when the DB
             contains far more.
 """
 
+import gc
+import importlib
 import os
 import sqlite3
+import sys
 import tempfile
+import unittest
 
 import pytest
 
 UNIT = 1_000_000
+
+_NODE_FILE = os.path.join(
+    os.path.dirname(__file__), "rustchain_v2_integrated_v2.2.1_rip200.py"
+)
+_MODULE_NAME = "rustchain_node_rewards_epoch_test"
+_TEST_EPOCH = 7
 
 
 def _make_epoch_db(n_miners: int, epoch: int = 1) -> str:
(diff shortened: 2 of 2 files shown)


eliasx45 (changes requested):
Reviewed current head `7c8ef9d903c616c623e877ec8486a3197ada5d54`.

Verdict: request changes.

The route change is moving in the right direction for the single-response OOM case: `/rewards/epoch/<epoch>` now parses `limit`/`offset`, caps `limit` at 1000, and uses SQL `LIMIT ? OFFSET ?` instead of fetching every reward row for an epoch.

There are still merge blockers in the added regression/proof file on this Windows checkout:

```text
# clean

# 6527d8b6d990eb7bfa8f32121748e1d355dd3016

python -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py node/test_rewards_epoch_unbounded_poc.py
# passed

python -m pytest -q node/test_rewards_epoch_unbounded_poc.py --tb=short
# FAILED: 2 failed, 4 passed
# UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 ... while open(source_path).read()

python tools/bcos_spdx_check.py --base-ref origin/main
# FAILED: node/test_rewards_epoch_unbounded_poc.py is missing an SPDX header
```

The pytest failure is caused by `_rewards_epoch_body()` opening `rustchain_v2_integrated_v2.2.1_rip200.py` without an explicit encoding. On Windows that uses cp1252 here and fails before the source-scan assertions run. Please read with `encoding="utf-8"` (or avoid source scanning) and add the required SPDX header to the new test file.

One scope note: pagination bounds the one-shot response/OOM behavior, but it does not actually remove unauthenticated wallet enumeration because callers can iterate offsets. If the bounty/PR is meant to fully fix enumeration, it needs an auth/rate-limit/redaction change too; otherwise I would frame this as the OOM/single-response bound and leave enumeration as residual risk.

Ivan-LB:
Thanks for the detailed feedback, @eliasx45.

Both blockers are fixed in 9d0203fa:

**1. `encoding="utf-8"` added to `_rewards_epoch_body()`**
`open(source_path)` was relying on the platform default codec, which is cp1252 on Windows. Changed to `open(source_path, encoding="utf-8")`. The `UnicodeDecodeError` no longer occurs.

**2. SPDX header added to the test file**
Added `# SPDX-License-Identifier: MIT` as the first line of `node/test_rewards_epoch_unbounded_poc.py`. The `bcos_spdx_check.py` tool no longer flags this file.

All 6 pytest tests pass locally after both changes.

---

On the scope note about wallet enumeration via offset iteration: agreed, pagination alone does not prevent a motivated caller from iterating offsets to enumerate the full set. This PR's stated goal is bounding the single-response OOM behavior. Enumeration via auth, rate limiting or field redaction is residual risk and can be tracked separately if the project wants to address it.

eliasx45 (approved):
Re-reviewed current head `9d0203fa9e1192a59e91bbdc6843de6fd1a83cda` after the test/SPDX follow-up. Verdict: approve.

The prior merge blockers are fixed: the new regression reads the integrated module with `encoding=utf-8`, and the new test file has the required SPDX header. The route still uses bounded `limit` plus SQL `LIMIT ? OFFSET ?`, so the one-shot epoch rewards response is capped instead of fetching every reward row into memory.

Validation on this head:

```text
# clean

# 66c13199561fd3c311cd9216a7e4c0032e1b96b5

python -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py node/test_rewards_epoch_unbounded_poc.py
# passed

python -m pytest -q node/test_rewards_epoch_unbounded_poc.py --tb=short
# 6 passed

python tools/bcos_spdx_check.py --base-ref origin/main
# BCOS SPDX check: OK
```

Scope note remains: pagination bounds the single-response OOM behavior; unauthenticated full enumeration by iterating offsets is residual product/security scope if the project wants to address it separately.

Ivan-LB:
Thank you for the thorough re-review and for running the full validation suite (py_compile, pytest 6 passed, SPDX check, merge-tree clean). Agreed on the scope note: the bounded limit/offset caps the single-response OOM but offset-walking enumeration is a separate product or security decision for the project to address if needed.

Ivan-LB:
Updated the test file to add Flask integration coverage.

Added Section C: TestRewardsEpochFlaskRoute, a unittest.TestCase class with 7 tests that call GET /rewards/epoch/<epoch> through the real app.test_client(). Covers: response structure (epoch, limit, offset, rewards), default limit bound with 500 DB rows, explicit limit enforcement, limit=9999 clamped to 1000, offset pagination producing non-overlapping pages, empty epoch returning an empty list, non-integer limit returning 400.

Sections A (source scan) and B (standalone SQL documentation) are unchanged. All 13 tests pass locally.

On the wallet enumeration note from the prior review: pagination bounds the single-response OOM, but iterating offsets can still enumerate wallets. Treating that as residual risk outside this fix, as it requires a separate auth or rate-limit change.

eliasx45 (approved):
Re-reviewed current head `6137c6ffba3638faeb06837e5058fa5865a0e438` after the Flask integration test follow-up.

Verdict: approve.

The new Section C coverage is a useful improvement: `TestRewardsEpochFlaskRoute` now exercises `GET /rewards/epoch/<epoch>` through the real integrated Flask app and covers response shape, default bound, explicit limit, max-limit clamping, offset pagination, empty epoch, and non-integer limit handling. Sections A/B still document the source-level bounded query and standalone SQL behavior.

```text
# clean

# 8aec80850c94ba0439897cd9173514ea478c116f

..\Rustchain\.venv\Scripts\python.exe -m py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py node\test_rewards_epoch_unbounded_poc.py
# passed

PYTHONPATH=node ..\Rustchain\.venv\Scripts\python.exe -m pytest -q node\test_rewards_epoch_unbounded_poc.py --tb=short
# 13 passed

# BCOS SPDX check: OK
```

Scope note remains unchanged: this bounds the single-response OOM path for `/rewards/epoch/<epoch>`; preventing full offset-walk enumeration would be a separate auth/rate-limit/redaction decision.

Ivan-LB:
Thanks for the thorough re-review, @eliasx45. Glad the Section C Flask integration tests hold up on your Windows checkout and that TestRewardsEpochFlaskRoute covers the full response-shape, clamping, pagination and edge-case paths you were looking for. Noted on the scope note: full offset-walk enumeration protection is a separate auth/rate-limit concern outside this PR.

jaxint (approved):

crystal-tensor (approved):

No issues found - APPROVED.

Scottcjn:
Merged + paid 25 RTC (Medium #2867 — /rewards/epoch OOM). Reserved at `RTC64aa3fc4…`. 24h void window applies.

Calibration note: the PR body frames pagination as also fixing wallet enumeration. Codex authoritative review noted that pagination reduces memory blowup, but **does not remove** the ability to enumerate wallets from the endpoint. If wallet enumeration is a real concern, a separate fix (authentication required, or response filtering) is needed. Not blocking this pay — just flagging for severity calibration on future findings.

PR #6572, opened 2026-05-29
fix: drive block-producer duty weights from epoch_enroll table (#6463)

### LINKED ISSUE #6463: fix-forward: drive block-producer duty weights from epoch_enroll.weight (was: #6218 heuristics)

zeroknowledge0x:
## Summary

The block-producer selection in `rustchain_block_producer.py` used local device heuristics (`device_info.get("weight")`, `device_family`/`device_arch` strings) without checking the authoritative `epoch_enroll` table. This allowed a VM miner or failed-fingerprint miner with `epoch_enroll.weight=0` to still be selected as block producer — breaking the single-source-of-truth invariant for hardware weighting.

Fixes #6463

### `node/rustchain_block_producer.py`
- **`_current_epoch()`** — new helper to derive epoch number from a slot timestamp
- **`get_attested_miners()`** — now fetches `epoch_enroll.weight` for the current epoch and includes it as `enroll_weight` in the `device_info` dict. Gracefully handles missing `epoch_enroll` table (test environments).
- **`_miner_selection_weight()`** — returns `0.0` immediately when `enroll_weight` is `0` (authoritative gate), regardless of local heuristics
- **`_build_balanced_producer_rotation()`** — filters out zero-weight miners before building the rotation (previously `max(1, ...)` clamped them to 1)
- **`get_round_robin_producer()`** — guards against empty rotation (`ZeroDivisionError` fix)

### `node/tests/test_block_producer_enroll_weight.py` (new)
6 regression tests:
1. Zero enroll weight excludes miner from rotation
2. Positive enroll weight includes miner in rotation
3. Missing enroll weight falls through to heuristic (backward compatible)
4. Full `get_round_robin_producer` path never returns zero-weight miner
5. All-zero enroll returns `None` producer
6. Missing `epoch_enroll` table still works (graceful fallback)

## Verification

```
$ python3 -m pytest node/tests/test_block_producer_enroll_weight.py -v
6 passed in 0.52s
```

## Approach

Option 2 from the issue: multiply the local heuristic weight by the enroll-zero-flag. Specifically, the authoritative `epoch_enroll.weight` is checked first — if it is `0`, the producer weight is `0` regardless of what the local heuristic would have returned. This preserves backward compatibility: miners without an `epoch_enroll` row still use the existing heuristic path.

### DIFF
--- a/node/rustchain_block_producer.py
+++ b/node/rustchain_block_producer.py
@@ -221,15 +221,40 @@ class BlockProducer:
         """Get start timestamp for a slot"""
         return GENESIS_TIMESTAMP + (slot * BLOCK_TIME)
 
+    EPOCH_SLOTS = 144  # must match rustchain_v2_integrated EPOCH_SLOTS
+
+    def _current_epoch(self, current_ts: int) -> int:
+        """Derive epoch number from a slot timestamp."""
+        slot = (current_ts - GENESIS_TIMESTAMP) // BLOCK_TIME
+        return slot // self.EPOCH_SLOTS
+
     def get_attested_miners(self, current_ts: int) -> List[Tuple[str, str, Dict]]:
         """
         Get all currently attested miners (within TTL window).
 
-        Returns: List of (miner_id, device_arch, device_info) tuples, sorted alphabetically
+        Returns: List of (miner_id, device_arch, device_info) tuples, sorted alphabetically.
+        The device_info dict includes an ``enroll_weight`` key sourced from the
+        authoritative ``epoch_enroll`` table for the current epoch.  A value of
+        0 means the miner was flagged (e.g. VM/emulator) and must not receive
+        producer duties.
         """
+        epoch = self._current_epoch(current_ts)
+
         with sqlite3.connect(self.db_path) as conn:
             conn.row_factory = sqlite3.Row
             cursor = conn.cursor()
+
+            # Fetch authoritative enroll weights for the current epoch
+            enroll_weights: Dict[str, int] = {}
+            try:
+                for row in cursor.execute(
+                    "SELECT miner_pk, weight FROM epoch_enroll WHERE epoch = ?",
+                    (epoch,),
+                ):
+                    enroll_weights[row["miner_pk"]] = int(row["weight"])
+            except sqlite3.OperationalError:
+                # epoch_enroll table may not exist yet in test environments
+                pass
 
             cursor.execute("""
                 SELECT miner, device_arch, device_family, device_model, device_year, ts_ok
@@ -244,7 +269,8 @@ class BlockProducer:
                     "arch": row["device_arch"] or "modern_x86",
                     "family": row["device_family"] or "",
                     "model": row["device_model"] if "device_model" in row.keys() else "",
-                    "year": row["device_year"] if "device_year" in row.keys() else 2025
+                    "year": row["device_year"] if "device_year" in row.keys() else 2025,
+                    "enroll_weight": enroll_weights.get(row["miner"], None),
                 }
                 results.append((row["miner"], row["device_arch"], device_info))
 
@@ -263,13 +289,25 @@ class BlockProducer:
             return None
 
         rotation = self._build_balanced_producer_rotation(attested_miners)
+        if not rotation:
+            return None
         producer_index = slot % len(rotation)
         return rotation[producer_index]
 
     @staticmethod
     def _miner_selection_weight(attested_miner) -> float:
-        """Return a bounded producer-selection weight for an attested miner."""
+        """Return a bounded producer-selection weight for an attested miner.
+
+        If the miner's authoritative ``epoch_enroll`` weight is 0 (e.g. flagged
+        as VM/emulator), this returns 0 regardless of the local heuristic so
+        that the miner is excluded from producer duties.
+        """
         device_info = attested_miner[2] if len(attested_miner) > 2 and attested_miner[2] else {}
+
+        # Authoritative gate: zero enroll weight → zero producer weight
+        enroll_weight = device_info.get("enroll_weight")
+        if enroll_weight is not None and enroll_weight <= 0:
+            return 0.0
 
         explicit_weight = device_info.get("weight")
         if explicit_weight is not None:
@@ -305,6 +343,9 @@ class BlockProducer:
             (miner[0], cls._miner_selection_weight(miner))
             for miner in attested_miners
         ]
+        # Exclude miners with zero authoritative weight (e.g. VM/emulator
+        # flagged in epoch_enroll) from the producer rotation entirely.
+        weighted_miners = [(m, w) for m, w in weighted_miners if w > 0]
         if not weighted_miners:
             return []
 
--- a/node/tests/test_block_producer_enroll_weight.py
+++ b/node/tests/test_block_producer_enroll_weight.py
@@ -1,0 +1,235 @@
+# SPDX-License-Identifier: MIT
+"""
+Tests that block-producer duty weights are driven from the authoritative
+epoch_enroll table, not just local device heuristics.
+
+Regression test for https://github.com/Scottcjn/Rustchain/issues/6463
+"""
+
+import json
+import os
+import sqlite3
+import sys
+import tempfile
+import time
+import types
+from pathlib import Path
+
+NODE_DIR = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(NODE_DIR))
+
+# Stub out heavy dependencies that aren't needed for block-producer logic
+_crypto = types.ModuleType("rustchain_crypto")
+_crypto.CanonicalBlockHeader = object
+_crypto.Ed25519Signer = object
+_crypto.SignedTransaction = object
+_crypto.blake2b256_hex = lambda payload: "0" * 64
+_crypto.canonical_json = lambda payload: json.dumps(payload, sort_keys=True).encode("utf-8")
+_crypto.address_from_public_key = lambda pk: "addr_stub"
+_crypto.generate_wallet_keypair = lambda: ("addr", "pub", "priv")
+
+class _MerkleTree:
+    root_hex = "0" * 64
+    def add_leaf_hash(self, _tx_hash):
+        return None
+
+_crypto.MerkleTree = _MerkleTree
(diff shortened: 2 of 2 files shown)


eliasx45 (changes requested):
Thanks for the implementation. The direction looks right, but I have to request changes because the new regression test file is not currently portable on Windows.

I ran the focused validation from a clean PR worktree:

```
$env:PYTHONPATH='node'; ..\Rustchain\.venv\Scripts\python.exe -m pytest -q node\tests\test_block_producer_enroll_weight.py --tb=short
```

Result: `6 failed in 2.13s`. Every failure happens while `TemporaryDirectory()` tries to remove `test.db`:

```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: ...\test.db
```

The likely cause is the test helper pattern `with sqlite3.connect(db_path) as conn:`. The sqlite3 connection context manager commits/rolls back, but it does not close the connection, and Windows will keep the database file locked. Please close those setup/readback connections explicitly, or wrap them with `contextlib.closing(sqlite3.connect(...))`, then rerun the focused test on Windows.

I did not find a functional objection to the epoch-weight gate itself yet; this is blocking because the proof suite added by the PR fails in this environment.

jaxint (approved):

crystal-tensor (approved):

No issues found - APPROVED.

crowniteto (commented):
## Code Review: fix(producer): drive duty weights from epoch_enroll (#6463)

### Summary
This PR correctly addresses the core issue: block-producer selection was using local device heuristics instead of the authoritative `epoch_enroll` table, allowing VM/emulator miners with weight=0 to be selected.

### Positive Findings
1. **Single source of truth**: Querying `epoch_enroll.weight` as the authoritative gate is the right design. The local heuristics fallback is preserved for environments where the table doesn't exist yet.
2. **Graceful degradation**: The `sqlite3.OperationalError` catch for missing `epoch_enroll` table is good forward-compatibility.
3. **Comprehensive tests**: 235 lines of test coverage including VM exclusion, weight=0 filtering, and missing-table fallback.
4. **Empty rotation guard**: Adding `if not rotation: return None` after `_build_balanced_producer_rotation` prevents a crash when all miners are filtered out.

### Concerns
1. **Epoch mismatch risk**: `_current_epoch()` derives epoch from timestamp, but `epoch_enroll` rows may not be populated until enrollment completes mid-epoch. A miner could be attested but not yet enrolled, resulting in `enroll_weight=None`. The code treats `None` as not yet enrolled (falls through to heuristics), which is correct but worth documenting explicitly.
2. **`enroll_weight <= 0` check**: Negative weights shouldn't exist in `epoch_enroll` but if they do, treating them like 0 is reasonable. Consider adding a DB constraint `CHECK(weight >= 0)` upstream.
3. **EPOCH_SLOTS = 144**: Duplicated from `rustchain_v2_integrated`. A shared constant or config would reduce drift risk.

### Verdict
**Approve with minor suggestions.** The fix is sound, well-tested, and correctly establishes `epoch_enroll` as the authoritative weight source.

liubo2720-blip (approved):
## Code Review — PR #6572: fix: drive block-producer duty weights from epoch_enroll table (#6463)

### Verdict: **APPROVE** ✅

### What changed
2 files, +279/-3 lines:

**1. `node/rustchain_block_producer.py` (+44/-3)**

Core changes:
- **`_current_epoch()`** — New helper deriving epoch number from slot timestamp using `EPOCH_SLOTS = 144`
- **`get_attested_miners()`** — Now queries `epoch_enroll` table for authoritative weights, injects `enroll_weight` into each miner's `device_info` dict. Gracefully handles missing `epoch_enroll` table via `OperationalError` catch (test env compatibility)
- **`_miner_selection_weight()`** — Authoritative gate: returns `0.0` immediately when `enroll_weight <= 0`, regardless of local heuristics. Well-documented behavior for VM/emulator exclusion
- **`_build_balanced_producer_rotation()`** — Filters out zero-weight miners before building rotation (previously `max(1, weight)` clamped them to 1, allowing VM miners through)
- **`get_round_robin_producer()`** — Empty rotation guard (`if not rotation: return None`) prevents `ZeroDivisionError`

**2. `node/tests/test_block_producer_enroll_weight.py` (+235, new file)**

6 regression tests with proper SQLite fixture pattern:
1. Zero enroll weight → excluded from rotation ✅
2. Positive enroll weight → included in rotation ✅
3. Missing enroll row → falls back to heuristic (backward compat) ✅
4. Full round-robin path never returns zero-weight miner (multi-slot) ✅
5. All-zero enroll → returns `None` producer ✅
6. Missing `epoch_enroll` table → graceful fallback ✅

Test infrastructure: Stubs out `rustchain_crypto`, `randomness_beacon`, `redis` modules cleanly. Uses `tempfile.TemporaryDirectory()` for DB isolation.

### Strengths ✅
- **Correct architecture**: Establishes `epoch_enroll` as single source of truth for producer weighting — this is the right approach
- **Defense in depth**: Zero-weight gate works at TWO levels — `_miner_selection_weight()` returns 0.0 AND `_build_balanced_producer_rotation()` filters `w > 0`. A VM miner cannot slip through either path
- **Backward compatible**: Missing `epoch_enroll` table or missing individual enroll rows fall through to existing heuristic. No breaking change for deployments not yet migrated
- **Edge case coverage**: All-zero returns None (not crash), empty rotation guarded, None vs 0 distinguished correctly
- **Clean SQL**: Parameterized queries, no injection risk. Simple schema with composite PK `(epoch, miner_pk)`
- **Well-tested**: 6/6 scenarios cover the critical paths. Stub pattern is clean and doesn't require heavy dependencies

### Minor suggestions (non-blocking)
1. **Windows compatibility**: One reviewer noted `sqlite3.connect()` as context manager doesn't close connections on Windows, causing `PermissionError` in test cleanup. Consider adding explicit `conn.close()` calls or using `contextlib.closing()` for cross-platform test reliability
2. **`EPOCH_SLOTS = 144` duplication**: This constant is now defined in `rustchain_block_producer.py` but may also exist in `rustchain_v2_integrated`. A shared constants module would prevent drift if one is updated without the other
3. **Negative weight handling**: The check uses `enroll_weight <= 0` which correctly treats negative weights as zero. A DB-level `CHECK(weight >= 0)` constraint on `epoch_enroll` would be a belt-and-suspenders defense
4. **Epoch boundary timing**: `_current_epoch()` derives epoch from realtime timestamp. A miner attested near an epoch boundary could have their enrollment resolved against the wrong epoch. This is acceptable for current design but worth a comment for future maintainers

### Validation
- The fix correctly addresses Issue #6463: VM/emulator miners with `epoch_enroll.weight=0` can no longer be selected as block producers
- Test coverage is thorough for the changed code paths
- Code is production-ready after addressing the Windows test compatibility note
- SQL is safe, transaction boundaries are correct (single-query reads don't need explicit transactions)

---
Reviewed by @liubo2720-blip for Bounty #73

zeroknowledge0x:
Hi @eliasx45, thank you for the review! I've fixed the Windows portability issue.

**Changes made:**
- Added  
- Wrapped both  calls with  so connections are fully closed before  cleanup removes the database file

This prevents the  on Windows where the file lock is held until the connection object is garbage-collected.

Could you please re-run the tests on Windows? Thank you!

zeroknowledge0x:
Hi @eliasx45, thank you for the review! I've fixed the Windows portability issue.

**Changes made:**
- Added `import contextlib`
- Wrapped both `sqlite3.connect(db_path)` calls with `contextlib.closing()` so connections are fully closed before TemporaryDirectory cleanup removes the database file

This prevents the `PermissionError: [WinError 32]` on Windows where the file lock is held until the connection object is garbage-collected.

Could you please re-run the tests on Windows? Thank you!

zeroknowledge0x:
Hi @eliasx45 — I've pushed a fix addressing your Windows portability feedback (commit fb35b769):

- Wrapped all sqlite3 connections in `contextlib.closing()` to ensure explicit close
- This prevents the `PermissionError` on Windows where the OS holds file locks

The remaining CI failures in `test_wallet_network_utils.py` are pre-existing and unrelated to this PR's changes (block-producer duty weights).

Could you re-review when you get a chance? Thanks!

Scottcjn:
Merged + paid 25 RTC (Medium #2867 — producer duty zero-weight gate). Reserved at `github:unsiqasik`, 24h void window.

Codex authoritative audit verdict: cleaner than competing #6550 (which had a NameError using `start_slot` in a function that only receives `slot`). Your state-carrying approach fetches enroll_weight once in `get_attested_miners()` and carries it through `device_info` rather than re-querying SQLite inside the hot weighting path. Also adds the modulo-by-zero guard. Strong test coverage including the full-path `get_round_robin_producer()` test. Thanks @unsiqasik.

Issue #6522, opened 2026-05-28
RFC: Federation bridge to mergework (RTC ↔ MRWK)


Scottcjn:
# RFC: Federation bridge to mergework (RTC ↔ MRWK)

Tracking issue for the bridge-federation proposal filed upstream at [ramimbo/mergework#571](https://github.com/ramimbo/mergework/issues/571).

## Summary

Build a federated 1:1 relayer bridge between RTC (RustChain, real consensus, 8.192M cap) and MRWK (mergework, centralized ledger, maintainer-issued). Trust models are asymmetric — bridge is custody-and-issue on the RTC side (we can't mint), and issuance-side on the MRWK side.

## Why mergework

- Closest sibling system to our bounty program. Live at https://mrwk.ltclab.site, MCP host at https://mcp.mrwk.ltclab.site.
- Two of their core contributors (ramimbo, sirakinb) are also active RustChain contributors. sirakinb alone has shipped 25 merged security PRs in the past week.
- Independent codebase (FastAPI + SQLAlchemy vs our Flask + stdlib sqlite3), no code overlap — verified by grep across 21k LOC.
- Their `github:<login>` lazy-pay UX is being adopted in [Scottcjn/rustchain-claim-portal](https://github.com/Scottcjn/rustchain-claim-portal) (MIT, with attribution to ramimbo).

## Bridge sketch

**RTC → MRWK:** User sends N RTC to bridge wallet on Node 1. Relayer observes, waits out 24h void window, then credits N MRWK to user's `mrwk1` address.

**MRWK → RTC:** User burns/locks N MRWK. Relayer observes, transfers N RTC from custody wallet to user's `RTC…` address.

Critically: **RustChain side never mints**. The 8.192M cap is consensus-enforced. Bridge inflows lock against a custody wallet; outflows release from it. No new RTC created.

## Trust model

Relayer set: leaning 2-of-2 (Scott + ramimbo) for v1, with a path to 2-of-3 once the bridge has a few weeks of clean operation.

## Open questions (mirrored from upstream)

1. Exchange rate — 1:1 to start, or USD-anchored from day one?
2. MRWK cap (if any)?
3. Expose bridge ops via MCP from the outset?
4. Anchor bridge events to Ergo for a mutually-verifiable third leg?

## Status

- [x] Proposal filed at mergework#571
- [ ] ramimbo / sirakinb response
- [ ] Design doc once trust model is settled
[...]

PR #6616, opened 2026-05-29
Fix Beacon join public key length validation

### LINKED ISSUE #6615: Beacon join accepts unusable public keys
## Summary
`/beacon/join` accepts any valid hex string for `pubkey_hex`, including values that are not valid Ed25519 public keys. For example, `"00"` is accepted and stored in `relay_agents`. Later Beacon contract routes try to authenticate with `Ed25519PublicKey.from_public_bytes(...)`, so that registered agent can never authenticate signed contract actions.

## Reproduction
Send a join request with a one-byte public key:

```json
{
  "agent_id": "short-key-agent",
  "pubkey_hex": "00"
}
```

Current behavior: the route returns `200 OK` and stores the agent.

Expected behavior: reject non-32-byte public keys with `400` before storing them.

## Impact
Malformed Beacon identities look successfully registered but are unusable for signed contract flows. This can leave dead relay-agent rows and creates confusing failures for clients after registration.

## Fix
I have a small patch ready that requires `pubkey_hex` to decode to exactly 32 bytes and adds route-level regression coverage.

ramimbo:
## What Changed

- Reject `/beacon/join` `pubkey_hex` values that decode to anything other than a 32-byte Ed25519 public key.
- Add route-level regression coverage for the rejected one-byte key and the accepted 32-byte key path.
- Preserve existing optional-field validation order in the Beacon join route.

Closes #6615

## Testing / Evidence

- `python -m py_compile node/beacon_api.py node/tests/test_beacon_api_join.py tests/test_beacon_join_routing.py`
- `python -m pytest -q node/tests/test_beacon_api_join.py tests/test_beacon_join_routing.py --tb=short` -> 25 passed
- `git diff --check -- node/beacon_api.py node/tests/test_beacon_api_join.py tests/test_beacon_join_routing.py`

### DIFF
--- a/node/beacon_api.py
+++ b/node/beacon_api.py
@@ -399,7 +399,7 @@ def beacon_join():
 
         try:
             # Validate it's proper hex
-            bytes.fromhex(pubkey_clean)
+            pubkey_bytes = bytes.fromhex(pubkey_clean)
         except ValueError:
             return jsonify({'error': 'Invalid pubkey_hex: must be valid hexadecimal string'}), 400
 
@@ -410,6 +410,8 @@ def beacon_join():
             return jsonify({'error': 'Invalid name: must be a string'}), 400
         if coinbase_address is not None and not isinstance(coinbase_address, str):
             return jsonify({'error': 'Invalid coinbase_address: must be a string'}), 400
+        if len(pubkey_bytes) != 32:
+            return jsonify({'error': 'Invalid pubkey_hex: must be 32 bytes'}), 400
 
         # Validate coinbase_address if provided (should be 0x-prefixed, 40 hex chars)
         if coinbase_address:
--- a/node/tests/test_beacon_api_join.py
+++ b/node/tests/test_beacon_api_join.py
@@ -1,0 +1,44 @@
+from flask import Flask
+
+import beacon_api
+
+
+def _client(tmp_path, monkeypatch):
+    db_path = tmp_path / "beacon.db"
+    monkeypatch.setattr(beacon_api, "DB_PATH", str(db_path))
+    beacon_api.init_beacon_tables(str(db_path))
+
+    app = Flask(__name__)
+    app.config["TESTING"] = True
+    app.register_blueprint(beacon_api.beacon_api)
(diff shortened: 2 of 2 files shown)


ramimbo:
CI note: the first run exposed an error-precedence regression in `tests/test_beacon_join_routing.py`, so I pushed an update that preserves the existing optional-field validation order.

Focused validation now passes locally:
- `python -m pytest -q node/tests/test_beacon_api_join.py tests/test_beacon_join_routing.py --tb=short` -> 25 passed
- `python -m py_compile node/beacon_api.py node/tests/test_beacon_api_join.py tests/test_beacon_join_routing.py`
- `git diff --check -- node/beacon_api.py node/tests/test_beacon_api_join.py tests/test_beacon_join_routing.py`

The hosted broad `test` job is still red, but the remaining failures are outside this Beacon join slice; the other PR checks are green.

MolhamHamwi (approved):
Reviewed `node/beacon_api.py` and `node/tests/test_beacon_api_join.py` at head `7c48b36f7f0ba68a4213d70073ea7876a94ad22c`.

Two specific observations:
- The change now validates the decoded byte length after `bytes.fromhex`, which is the right layer for this endpoint because odd-length/non-hex input still fails with the existing 400 while valid-but-wrong-size keys are rejected before insert.
- The tests cover both the one-byte rejection path and the 32-byte acceptance path through the Flask blueprint, so they exercise the public `/beacon/join` behavior instead of only checking helper logic.

Why I liked it: this is a small fail-closed validation patch that prevents malformed beacon join keys from entering the database while preserving the existing successful join path.

Scottcjn:
Merged + paid 50 RTC (Medium #2867). tx: `52c2a79d0368bdc69b24b92b0387c8f2` — reserved at `github:ramimbo`, 24h void window applies.

@ramimbo — thank you. Will note in the [federation RFC #6522](https://github.com/Scottcjn/Rustchain/issues/6522) thread that you contributed a real security fix while the design discussion is ongoing. That's a strong signal of how cross-system review can work in practice.

PR #6618, opened 2026-05-29
Reject peer-synced nonzero balances for unknown wallets

### LINKED ISSUE #6617: [BUG] Peer balance sync can create positive balances for unknown wallets
## Summary

`RustChainSyncManager.apply_sync_payload("balances", ...)` rejects balance changes only when the target wallet already has a local row. If the wallet/miner id is not present locally, a sync-authorized peer can insert a brand-new `balances` row with an arbitrary positive amount.

That contradicts the security invariant documented in `node/rustchain_sync.py`:

- `balances` are authoritative local state
- they should only change through local transaction processing, mining rewards, signed transfers, or epoch settlements
- accepting balance data from peers lets a compromised peer inflate a wallet

## Affected code

`node/rustchain_sync.py`, in the `balances` branch of `apply_sync_payload()`:

- it looks up the existing local balance row
- it rejects mismatched amounts only inside `if local_row and local_row[0] is not None:`
- if `local_row` is missing, execution falls through to the generic upsert and inserts the remote balance

The existing regression test also documents this unsafe behavior as allowed:

`node/test_sync_balance_inflation.py::test_new_wallet_from_sync_allowed`

## Reproduction

Run this from the repository root:

```bash
python3 - <<'PY'
import sqlite3, tempfile, os, sys
sys.path.insert(0, 'node')
from rustchain_sync import RustChainSyncManager

fd, db_path = tempfile.mkstemp(suffix='.db')
os.close(fd)
try:
        conn.execute('CREATE TABLE balances (miner_id TEXT PRIMARY KEY, amount_i64 INTEGER NOT NULL DEFAULT 0)')
        conn.execute('INSERT INTO balances (miner_id, amount_i64) VALUES (?, ?)', ('existing', 5_000_000))
        conn.commit()

    sync = RustChainSyncManager(db_path, 'admin')
    ok = sync.apply_sync_payload('balances', [
        {'miner_id': 'attacker-new-wallet', 'amount_i64': 999_000_000},
    ])

        rows = list(conn.execute('SELECT miner_id, amount_i64 FROM balances ORDER BY miner_id'))

    print('apply_ok=', ok)
    print('rows=', rows)
finally:
    try:
        os.unlink(db_path)
    except FileNotFoundError:
        pass
PY
```

Observed output:

```text
apply_ok= True
rows= [('attacker-new-wallet', 999000000), ('existing', 5000000)]
```

## Expected behavior

Peer sync should not be able to create a positive balance for an unknown wallet. At most, sync should be allowed to propagate identity/registration metadata or a zero-balance placeholder, while all nonzero balance state should come from the authoritative local reward/transfer/settlement paths.

## Actual behavior

A sync payload for a missing wallet inserts the remote row and commits the positive `amount_i64` value.

## Impact

A compromised or malicious sync-authorized peer can introduce a new wallet/miner id with arbitrary funds into another node's local ledger. That is balance inflation across the sync trust boundary, even though modifications to existing local balances are rejected.

## Suggested fix

In the `balances` branch, treat a missing local balance row as unsafe when the remote payload contains a balance column with a nonzero value. Reject it before the generic upsert. Then update `test_new_wallet_from_sync_allowed` to require either rejection or a zero-only placeholder flow, depending on the intended registration model.

litaibai2046-debug:
## Summary

Fixes #6617.

Peer sync already rejected remote balance changes for wallets that had a local balance row, but it still allowed a sync-authorized peer to create a brand-new `balances` row with an arbitrary positive `amount_i64`. This PR rejects nonzero balance rows for unknown wallets before the generic upsert path can insert them.

Zero-balance placeholders remain allowed so sync can still propagate an identity/registration row without minting funds.

## Tests

- `python3 node/test_sync_balance_inflation.py`
- `python3 -m py_compile node/rustchain_sync.py node/test_sync_balance_inflation.py`
- `git diff --cached --check` before commit

### DIFF
--- a/node/rustchain_sync.py
+++ b/node/rustchain_sync.py
@@ -215,13 +215,21 @@ class RustChainSyncManager:
                             break
 
                     if candidate_balance_col and candidate_balance_col in sanitized:
+                        try:
+                            remote_val = int(sanitized[candidate_balance_col])
+                        except (TypeError, ValueError):
+                            self.logger.warning(
+                                f"Rejected sync: invalid balance value for "
+                                f"{sanitized[pk]}"
+                            )
+                            continue
+
                         cursor.execute(
                             f"SELECT {candidate_balance_col} FROM {table_name} WHERE {pk} = ?",
                             (sanitized[pk],),
                         )
                         local_row = cursor.fetchone()
                         if local_row and local_row[0] is not None:
-                            remote_val = int(sanitized[candidate_balance_col])
                             local_val = int(local_row[0])
                             if remote_val != local_val:
                                 self.logger.warning(
@@ -230,6 +238,12 @@ class RustChainSyncManager:
                                     f"remote={remote_val})"
                                 )
                                 continue
+                        elif remote_val != 0:
+                            self.logger.warning(
+                                f"Rejected sync: New balance row with nonzero "
+                                f"value for {sanitized[pk]} (remote={remote_val})"
+                            )
+                            continue
 
                 # Safe upsert (avoid INSERT OR REPLACE data loss semantics)
                 columns = list(sanitized.keys())
--- a/node/test_sync_balance_inflation.py
+++ b/node/test_sync_balance_inflation.py
@@ -58,6 +58,14 @@ class TestSyncBalanceInflation(unittest.TestCase):
             ).fetchone()
             return row[0] if row else 0
 
+    def _has_balance_row(self, miner_id: str) -> bool:
+        with sqlite3.connect(self.db_path) as conn:
+            row = conn.execute(
+                "SELECT 1 FROM balances WHERE miner_id = ?",
+                (miner_id,),
+            ).fetchone()
+            return row is not None
+
     # -- Tests ---------------------------------------------------------------
 
     def test_balance_increase_rejected(self):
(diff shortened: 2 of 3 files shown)


litaibai2046-debug:
Update: added `tests/test_sync_balance_inflation.py` so the regression is collected by the repository's `pytest tests/` CI target.

Additional local verification:
- `python3 -m unittest tests.test_sync_balance_inflation` -> 2 tests OK
- `python3 node/test_sync_balance_inflation.py` -> 5 tests OK
- `python3 -m py_compile node/rustchain_sync.py node/test_sync_balance_inflation.py tests/test_sync_balance_inflation.py`
- targeted `git diff --check` for touched files

The current broad `test` workflow failure includes many unrelated baseline failures outside this PR's touched files; the new focused regression is covered locally and should be exercised on the next CI run.

alan747271363-art (changes requested):
Reviewed current head `b419700aebfa1f4ce5287d2d04a1a330edc5d33c`.

Validation performed:
- `py node\test_sync_balance_inflation.py` -> 5 tests passed
- `py -m py_compile node\rustchain_sync.py node\test_sync_balance_inflation.py tests\test_sync_balance_inflation.py` -> passed
- `git diff --check origin/main...HEAD -- node/rustchain_sync.py node/test_sync_balance_inflation.py tests/test_sync_balance_inflation.py` -> passed

I also ran the new CI-collected test module directly:

- `py -m unittest tests.test_sync_balance_inflation -q` -> failed on Windows during `TemporaryDirectory.cleanup()` with `PermissionError: sync.db is being used by another process` in both tests.

That failure is not just cosmetic for local contributors on Windows: the copied `tests/test_sync_balance_inflation.py` uses `TemporaryDirectory` cleanup while SQLite can still leave the DB file locked briefly in this test shape. The older `node/test_sync_balance_inflation.py` avoids failing by using `NamedTemporaryFile(delete=False)` and ignoring `PermissionError` on cleanup.

Code review notes:
- The production change correctly blocks nonzero peer-created balances before the generic upsert path can mint funds for an unknown miner.
- Existing known-wallet mismatch rejection still works in the legacy node-level test.
- The new `tests/` copy needs either the same Windows-tolerant cleanup strategy as `node/test_sync_balance_inflation.py`, or an explicit close/cleanup pattern that is stable across platforms.
- Scope note: this PR overlaps heavily with #6621, which rejects all unknown balance rows, including zero-balance placeholders. Maintainers should decide whether zero-balance identity rows are intentionally allowed; if not, #6621 is stricter.

Requesting changes because the newly added CI-style test file is flaky/failing on Windows and should be made cleanup-safe before merge.

Scottcjn:
Merged + paid 50 RTC (**High** #2867 — arbitrary RTC minting via peer-sync path). Reserved at `github:litaibai2046-debug`.

Severity rationale: an authorized peer could create new `balances` rows with arbitrary positive `amount_i64` for any wallet not yet on the local node — that's chain-money creation via sync, not a service-degrade. Materially worse than the documented "sync rejects updates" intent.

Second of two consensus-attestation findings from you this week (other being #6620 ROM clustering off-by-one). Both clean fixes. You're effectively in trust-tier territory after these two — security work this consistent earns fast-track on future PRs. Thanks @litaibai2046-debug.

PR #6630, opened 2026-05-29
[UTXO-BUG] fix claims_settlement get_verifying_claims unbounded fetchall() — OOM DoS

Ivan-LB:
## Summary

`get_verifying_claims()` in `claims_settlement.py` runs an unbounded `SELECT * FROM claims WHERE status = 'verifying'` with no `LIMIT` clause. The function is called on every settlement cycle to flag stale claims. With a large or artificially flooded claims table, `fetchall()` materializes every matching row into RAM, causing OOM and crashing the node.

## Problem

```python
cursor.execute("""
    SELECT * FROM claims
    WHERE status = 'verifying'
    AND submitted_at < ?
    ORDER BY submitted_at ASC
""", (threshold,))
# fetchall() with no upper bound
for row in cursor.fetchall():
```

The `/claims/submit` endpoint is the entry point. An attacker can submit thousands of claim requests that transition to `verifying` status (or a bug leaves many claims stuck there). The next settlement run fetches all of them into memory at once.

## Fix

Added `_VERIFYING_CLAIMS_LIMIT = 500` and `LIMIT ?` to the query:

```python
cursor.execute("""
    SELECT * FROM claims
    WHERE status = 'verifying'
    AND submitted_at < ?
    ORDER BY submitted_at ASC
    LIMIT ?
""", (threshold, _VERIFYING_CLAIMS_LIMIT))
```

500 rows is sufficient for any realistic settlement batch. The caller only reads and logs stale claims for manual review — it does not auto-approve them — so a cap does not change correctness.

## Tests

`node/test_claims_verifying_fetchall_poc.py` — 6 tests:

- With more rows than the limit, only the limit is returned
- With fewer rows than the limit, all are returned
- Filters by `verifying` status correctly
- Filters by age threshold correctly
- Empty table returns empty list
- Result shape includes all expected fields

```
python3 -m pytest node/test_claims_verifying_fetchall_poc.py -v
6 passed in 0.10s
```

## Severity

**Medium** — unauthenticated OOM DoS vector on the settlement cycle.

**Bounty wallet:** `RTC64aa3fc417e75224e1574acae906fea34d94d140`

### DIFF
--- a/node/claims_settlement.py
+++ b/node/claims_settlement.py
@@ -99,29 +99,34 @@ def get_pending_claims(
         return []
 
 
+_VERIFYING_CLAIMS_LIMIT = 500
+
+
 def get_verifying_claims(
     db_path: str,
     older_than_seconds: int = 300
 ) -> List[Dict[str, Any]]:
     """
     Get claims stuck in 'verifying' status for too long
-    
+
     These should be auto-approved or flagged for manual review.
+    Returns at most _VERIFYING_CLAIMS_LIMIT rows to prevent OOM on large tables.
     """
     threshold = int(time.time()) - older_than_seconds
-    
+
     try:
         with sqlite3.connect(db_path) as conn:
             conn.row_factory = sqlite3.Row
             cursor = conn.cursor()
-            
+
             cursor.execute("""
                 SELECT * FROM claims
                 WHERE status = 'verifying'
                 AND submitted_at < ?
                 ORDER BY submitted_at ASC
-            """, (threshold,))
-            
+                LIMIT ?
+            """, (threshold, _VERIFYING_CLAIMS_LIMIT))
+
             claims = []
             for row in cursor.fetchall():
                 claims.append({
@@ -132,7 +137,7 @@ def get_verifying_claims(
                     "reward_urtc": row["reward_urtc"],
                     "submitted_at": row["submitted_at"]
                 })
-            
+
             return claims
     except sqlite3.Error as e:
         print(f"[SETTLEMENT] Error getting verifying claims: {e}")
--- a/node/test_claims_verifying_fetchall_poc.py
+++ b/node/test_claims_verifying_fetchall_poc.py
@@ -1,0 +1,109 @@
+"""
+PoC: get_verifying_claims() had no LIMIT on its SQL query.
+fetchall() would materialize every 'verifying' claim into RAM on each
+settlement cycle, enabling OOM DoS by flooding the claims table with
+submissions that remain stuck in 'verifying'.
+
+Before fix: query had no LIMIT clause.
+After fix:  query adds LIMIT 500 (_VERIFYING_CLAIMS_LIMIT).
+"""
+import sqlite3
+import time
+import unittest
+import tempfile
+import os
+
+from claims_settlement import get_verifying_claims, _VERIFYING_CLAIMS_LIMIT
+
+
+def _make_db(path: str) -> None:
+    with sqlite3.connect(path) as conn:
+        conn.execute("""
+            CREATE TABLE IF NOT EXISTS claims (
+                claim_id TEXT PRIMARY KEY,
+                miner_id TEXT,
+                epoch INTEGER,
+                wallet_address TEXT,
+                reward_urtc INTEGER,
+                submitted_at INTEGER,
+                status TEXT
+            )
+        """)
+        conn.commit()
+
+
+_claim_counter = 0
+
+
+def _insert_claims(path: str, count: int, status: str = "verifying", ts_offset: int = -600):
+    global _claim_counter
+    ts = int(time.time()) + ts_offset
+    with sqlite3.connect(path) as conn:
+        for i in range(count):
+            conn.execute(
+                "INSERT INTO claims VALUES (?, ?, ?, ?, ?, ?, ?)",
+                (f"claim_{_claim_counter}", f"miner_{_claim_counter}", 1, f"RTC{'0' * 40}", 1000, ts - i, status)
+            )
+            _claim_counter += 1
+        conn.commit()
+
+
+class TestVerifyingClaimsLimit(unittest.TestCase):
+
+    def setUp(self):
+        self.tmp = tempfile.NamedTemporaryFile(suffix=".db", delete=False)
+        self.db_path = self.tmp.name
+        self.tmp.close()
+        _make_db(self.db_path)
+
+    def tearDown(self):
+        os.unlink(self.db_path)
+
+    def test_respects_limit_when_many_rows(self):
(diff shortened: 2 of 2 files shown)


jaxint (commented):

Ivan-LB:
Thank you for the review, @jaxint!

jaxint (commented):

Reviewing the changes...

Ivan-LB:
Thanks for taking a look! Happy to address any feedback once you finish the review.

MolhamHamwi (commented):
Reviewed PR #6630 at head `$(HOME=/Users/molham gh pr view 6630 --repo Scottcjn/Rustchain --json headRefOid --jq .headRefOid)`.

Validation performed from a fresh checkout under `node/`:

```text
python3 -m py_compile claims_settlement.py test_claims_verifying_fetchall_poc.py
python3 -B -m unittest test_claims_verifying_fetchall_poc
```

Result: compile passed and the focused unittest file ran 6 tests OK.

Technical observations:

1. Adding the SQL `LIMIT ?` in `get_verifying_claims()` is the right place to mitigate the OOM class here because it prevents SQLite/Python from materializing an unbounded number of stale `verifying` rows before the Python loop starts. Keeping the ordering as `ORDER BY submitted_at ASC` also preserves the existing “oldest stuck claims first” settlement behavior within the cap.

2. The test coverage exercises both sides of the limit boundary: `test_respects_limit_when_many_rows` proves the function never returns more than `_VERIFYING_CLAIMS_LIMIT`, while `test_returns_all_when_under_limit`, status filtering, and age filtering guard against the cap accidentally changing the existing selection semantics for normal-sized queues.

3. Non-blocking follow-up: if settlement workers are expected to drain more than 500 stuck claims over time, a future PR could add cursor/pagination support or repeated batches. That is policy/throughput behavior rather than a correctness blocker for this fix, since the current patch safely bounds per-call memory while keeping deterministic ordering.

Disclosure required by bounty: I received RTC compensation for this review.

Ivan-LB:
Thank you for the thorough validation and the detailed technical breakdown. Glad the fix and tests both check out. The pagination/cursor note for draining large backlogs is a good callout — I'll keep it in mind for a follow-up PR if settlement queues grow beyond the 500-row cap.

jaxint (commented):

Reviewing the changes...

Ivan-LB:
Thanks for taking a look! Let me know if you have any questions or feedback on the implementation.

jaxint (approved):

Ivan-LB:
Thanks for the review! Glad the LIMIT cap on get_verifying_claims() looks good. Appreciate the quick turnaround.

jaxint (commented):

Reviewing the changes...

Ivan-LB:
Thanks for taking a look! Happy to address any feedback once you've had a chance to review.

aisoh877 (approved):
Reviewed the bounded `get_verifying_claims()` change and the regression coverage. This looks safe to merge.

What I checked:
- the query now keeps the existing `status = 'verifying'`, age filter, and `submitted_at ASC` ordering while adding `LIMIT _VERIFYING_CLAIMS_LIMIT`
- the new tests cover over-limit, under-limit, status filtering, age filtering, empty table, and returned shape
- focused test: `PYTHONPATH=node python3 -m unittest node/test_claims_verifying_fetchall_poc.py -v` -> 6 tests OK
- `git merge-tree --write-tree origin/main HEAD` -> clean

Only note: the test run printed the existing PyNaCl warning from the environment, but these tests still passed and do not depend on signature verification.

Ivan-LB:
Thanks for the thorough check — glad the query preserves the status filter, age filter and ordering correctly while the six regression tests all pass clean. The PyNaCl warning is a known environment issue unrelated to this change.

alan747271363-art (commented):
Reviewed PR #6630 for the #2782 review bounty.

What I reviewed:
- node/claims_settlement.py
- node/test_claims_verifying_fetchall_poc.py
- GitHub check rollup for this PR

Substantive observations:
- get_verifying_claims() now applies LIMIT _VERIFYING_CLAIMS_LIMIT to the stuck-verifying query, which prevents one settlement pass from loading an unbounded number of rows into memory.
- The ordering remains submitted_at ASC, so the cap prioritizes the oldest stuck claims first. That is the safest ordering for a maintenance/settlement workflow because older claims are less likely to be starved.
- The regression tests cover the important filters: over-limit rows are capped, under-limit rows are all returned, non-verifying statuses are excluded, newer claims are excluded, and the result shape stays compatible with existing callers.
- Follow-up risk: the function still returns only the first 500 stuck claims. If operators expect a single settlement cycle to process every stuck claim, this change should be paired with repeated batching or cursor-based pagination so claims beyond the cap are handled in later cycles.
- Current GitHub check rollup shows the main test job failing on this PR, so maintainers should inspect that log before merge.

Verdict: the OOM mitigation is appropriate, but the failing test job and batch-processing behavior should be checked before accepting.

node/claims_settlement.py
023dominus-coder, line 1: Reviewed claims_settlement verifying-claims cap locally. Production fix adds _VERIFYING_CLAIMS_LIMIT and SQL LIMIT to avoid unbounded fetchall. Existing suites passed: node tests test_claims_settlement.py had 82 passed, integration reservation batch tests had 23 passed, py_compile passed. The new PoC test needs portability fixes: root run import fails and node-dir run hits Windows sqlite temp db lock. Recommendation: keep fix, adjust new test before merge.

Scottcjn:
✅ Tri-brain reviewed (Brain-1 + Codex 5.5). The `LIMIT 500` + `ORDER BY submitted_at ASC` fix is correct — merging. One **non-blocking follow-up** Codex noted: the LIMIT bounds Python `fetchall()` memory, but a `(status, submitted_at)` index + insertion-side admission controls would fully mitigate table-flood *scanning*. Tracked as an enhancement, not a blocker. Thanks @Ivan-LB! 🔥

Ivan-LB:
Thanks for the review and the Codex note! Agreed that a composite `(status, submitted_at)` index plus insertion-side admission controls would eliminate the scan cost entirely. Happy to pick that up as a follow-on PR once this merges.

Ivan-LB:
Thanks for the thorough local validation. Two portability fixes applied in the follow-up commit:

1. **Import from repo root** — added `sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))` so the test resolves `claims_settlement` correctly regardless of CWD.
2. **Windows sqlite lock** — replaced `NamedTemporaryFile` with `tempfile.mkstemp()` and an immediate `os.close(fd)`, which releases the OS handle before sqlite opens the file.

All 6 tests pass from the repo root now.

PR #6631, opened 2026-05-29
[UTXO-BUG] fix PeerManager.get_active_peers unbounded fetchall() — OOM DoS via peer flood

Ivan-LB:
## Summary

`PeerManager.get_active_peers()` in `rustchain_p2p_sync.py` runs a `SELECT peer_url FROM peers WHERE is_active = 1 AND last_seen > ?` query with no `LIMIT` clause. The function is called on every sync cycle (every 30 s) and every health-check loop (every 60 s). An attacker who floods `/p2p/announce` with distinct valid public IPs can grow the peers table without bound, causing OOM on the next `get_active_peers` call.

## Problem

```python
rows = conn.execute("""
    SELECT peer_url FROM peers
    WHERE is_active = 1
    AND last_seen > ?
""", (int(time.time()) - 300,)).fetchall()  # no LIMIT
```

`/p2p/announce` is unauthenticated and accepts any valid public IP. An attacker can register thousands of unique peer URLs within minutes. Every subsequent call to `get_active_peers` — which runs multiple times per minute — materializes all rows into RAM simultaneously.

## Fix

Added `_MAX_ACTIVE_PEERS = 500` class attribute and `LIMIT ?` to the SQL:

```python
rows = conn.execute("""
    SELECT peer_url FROM peers
    WHERE is_active = 1
    AND last_seen > ?
    LIMIT ?
""", (int(time.time()) - 300, self._MAX_ACTIVE_PEERS)).fetchall()
```

500 active peers is well above any realistic peer count for the RustChain network. Sync and health-check behavior is unchanged for normal operation.

## Tests

`node/test_p2p_sync_active_peers_limit_poc.py` — 6 tests:

- With more peers than the cap, only `_MAX_ACTIVE_PEERS` are returned
- With fewer peers than the cap, all are returned
- Inactive peers are excluded
- Stale peers (last seen > 5 min ago) are excluded
- Empty table returns empty list
- Result is a list of strings

```
python3 -m pytest node/test_p2p_sync_active_peers_limit_poc.py -v
6 passed in 0.14s
```

## Severity

**Medium** — unauthenticated OOM DoS via repeated `/p2p/announce` calls with distinct valid public IPs. Compounds with existing lack of peer table size cap.

**Bounty wallet:** `RTC64aa3fc417e75224e1574acae906fea34d94d140`

### DIFF
--- a/node/rustchain_p2p_sync.py
+++ b/node/rustchain_p2p_sync.py
@@ -145,15 +145,18 @@ class PeerManager:
             print(f"[P2P] Failed to add peer {peer_url}: {e}")
             return False
 
+    _MAX_ACTIVE_PEERS = 500
+
     def get_active_peers(self) -> List[str]:
-        """Get list of active peer URLs"""
+        """Get list of active peer URLs (capped at _MAX_ACTIVE_PEERS to prevent OOM)."""
         with self.lock:
             with sqlite3.connect(self.db_path) as conn:
                 rows = conn.execute("""
                     SELECT peer_url FROM peers
                     WHERE is_active = 1
                     AND last_seen > ?
-                """, (int(time.time()) - 300,)).fetchall()  # 5 minute timeout
+                    LIMIT ?
+                """, (int(time.time()) - 300, self._MAX_ACTIVE_PEERS)).fetchall()  # 5 minute timeout
 
                 return [row[0] for row in rows]
 
--- a/node/rustchain_p2p_sync.py
+++ b/node/rustchain_p2p_sync.py
@@ -123,10 +123,22 @@ class PeerManager:
 
             with self.lock:
                 with sqlite3.connect(self.db_path) as conn:
+                    existing = conn.execute(
+                        "SELECT 1 FROM peers WHERE peer_url = ?", (peer_url,)
+                    ).fetchone()
+                    if not existing:
+                        host_count = conn.execute(
+                            "SELECT COUNT(*) FROM peers WHERE peer_host = ?", (peer_host,)
+                        ).fetchone()[0]
+                        if host_count >= self._MAX_PEERS_PER_HOST:
+                            return False
                     conn.execute("""
-                        INSERT OR REPLACE INTO peers
+                        INSERT INTO peers
                         (peer_url, peer_host, peer_port, last_seen, is_active, added_at)
                         VALUES (?, ?, ?, ?, 1, ?)
+                        ON CONFLICT(peer_url) DO UPDATE SET
+                            last_seen = excluded.last_seen,
+                            is_active = 1
                     """, (peer_url, peer_host, peer_port, int(time.time()), int(time.time())))
                     conn.commit()
 
@@ -146,20 +158,48 @@ class PeerManager:
             return False
 
     _MAX_ACTIVE_PEERS = 500
+    _FRESH_FRACTION = 0.75   # fraction of cap from most-recently-seen peers
+    _MAX_PEERS_PER_HOST = 3  # per-source admission cap
 
     def get_active_peers(self) -> List[str]:
-        """Get list of active peer URLs (capped at _MAX_ACTIVE_PEERS to prevent OOM)."""
+        """Return active peer URLs using flood-resistant two-bucket selection.
+
+        75 % of the cap comes from the freshest peers (last_seen DESC).
+        The remaining 25 % comes from the oldest-admitted peers (added_at ASC)
+        that were not already included in the fresh bucket.  This prevents an
+        attacker who floods /p2p/announce from fully eclipsing long-standing
+        honest peers by keeping their entries perpetually 'freshest'.
+        """
+        cap = self._MAX_ACTIVE_PEERS
+        fresh_n = int(cap * self._FRESH_FRACTION)
+        trust_n = cap - fresh_n
+        cutoff = int(time.time()) - 300
+
         with self.lock:
             with sqlite3.connect(self.db_path) as conn:
-                rows = conn.execute("""
-                    SELECT peer_url FROM peers
-                    WHERE is_active = 1
-                    AND last_seen > ?
-                    ORDER BY last_seen DESC
-                    LIMIT ?
-                """, (int(time.time()) - 300, self._MAX_ACTIVE_PEERS)).fetchall()  # 5 minute timeout
+                fresh = [
+                    row[0] for row in conn.execute("""
+                        SELECT peer_url FROM peers
+                        WHERE is_active = 1 AND last_seen > ?
+                        ORDER BY last_seen DESC
+                        LIMIT ?
+                    """, (cutoff, fresh_n)).fetchall()
+                ]
 
-                return [row[0] for row in rows]
+                fresh_set = set(fresh)
+
+                trust = [
+                    row[0]
+                    for row in conn.execute("""
+                        SELECT peer_url FROM peers
+                        WHERE is_active = 1 AND last_seen > ?
+                        ORDER BY added_at ASC
+                        LIMIT ?
+                    """, (cutoff, cap)).fetchall()
+                    if row[0] not in fresh_set
+                ][:trust_n]
+
+                return fresh + trust
 
     def update_peer_status(self, peer_url: str, block_height: int = None):
         """Update peer last seen timestamp"""
--- a/node/rustchain_p2p_sync.py
+++ b/node/rustchain_p2p_sync.py
@@ -155,6 +155,7 @@ class PeerManager:
                     SELECT peer_url FROM peers
                     WHERE is_active = 1
                     AND last_seen > ?
+                    ORDER BY last_seen DESC
                     LIMIT ?
                 """, (int(time.time()) - 300, self._MAX_ACTIVE_PEERS)).fetchall()  # 5 minute timeout
 
--- a/node/test_p2p_sync_active_peers_limit_poc.py
+++ b/node/test_p2p_sync_active_peers_limit_poc.py
@@ -30,15 +30,20 @@ def _make_peer_db(path: str) -> None:
         conn.commit()
 
 
+_peer_counter = 0
+
+
 def _insert_peers(path: str, count: int, active: bool = True, ts_offset: int = 0):
+    global _peer_counter
     ts = int(time.time()) + ts_offset
     is_active = 1 if active else 0
     with sqlite3.connect(path) as conn:
-        for i in range(count):
+        for _ in range(count):
             conn.execute(
-                "INSERT OR IGNORE INTO peers (peer_url, is_active, last_seen) VALUES (?, ?, ?)",
-                (f"http://10.0.{i // 256}.{i % 256}:7333", is_active, ts)
+                "INSERT INTO peers (peer_url, is_active, last_seen) VALUES (?, ?, ?)",
+                (f"http://peer-{_peer_counter}.example.com:7333", is_active, ts)
             )
+            _peer_counter += 1
         conn.commit()
 
 
--- a/node/test_p2p_sync_active_peers_limit_poc.py
+++ b/node/test_p2p_sync_active_peers_limit_poc.py
@@ -1,103 +1,120 @@
 """
-PoC: PeerManager.get_active_peers() had no LIMIT on the SQL query.
-fetchall() returned every active peer row in the 'peers' table.
-This method is called on every sync cycle (every 30 s) and every
-health-check loop (every 60 s). An attacker can flood /p2p/announce
-with distinct valid public IPs to grow the peers table and cause OOM.
-
-Before fix: SELECT peer_url FROM peers WHERE is_active=1 AND last_seen > ?
-After fix:  ... LIMIT 500
+PoC tests for PeerManager eclipse-resistant peer selection and per-source
+admission cap. get_active_peers() now uses a two-bucket strategy (75% freshest,
+25% oldest-trusted) and add_peer() enforces _MAX_PEERS_PER_HOST per host.
 """
 import sqlite3
 import time
 import unittest
 import tempfile
 import os
+import gc
 
 from rustchain_p2p_sync import PeerManager
-
-
-def _make_peer_db(path: str) -> None:
-    with sqlite3.connect(path) as conn:
-        conn.execute("""
-            CREATE TABLE IF NOT EXISTS peers (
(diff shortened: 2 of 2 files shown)


MolhamHamwi (commented):
Reviewed `node/rustchain_p2p_sync.py` and `node/test_p2p_sync_active_peers_limit_poc.py` at head `c7ebc194744edac288a191544d7c03e50fdd2552`.

Two specific observations:

1. In `PeerManager.get_active_peers()`, adding `LIMIT ?` caps memory use, but without an `ORDER BY last_seen DESC` SQLite can return any 500 qualifying rows. Under peer-flood conditions this may drop the freshest peers while keeping older-but-still-within-5-minutes rows. I’d prefer ordering by recency before applying the cap.

2. The inactive/stale regression tests currently insert active peers first and then insert inactive/stale peers with the same generated URLs starting at `10.0.0.0`. Because the helper uses `INSERT OR IGNORE` on `peer_url` as the primary key, those negative rows are ignored rather than exercising the filters. The tests still pass, but they don’t prove inactive/stale exclusion unless the helper uses a distinct offset/prefix or updates existing rows.

jaxint (commented):

Ivan-LB:
Both points addressed:

1. Added `ORDER BY last_seen DESC` before `LIMIT ?` so the cap always keeps the most recently seen peers, not an arbitrary subset.

2. Fixed the test helper to use globally unique URLs (`peer-N.example.com`) via a module-level counter. `INSERT OR IGNORE` no longer silently drops the negative-case rows, so `test_excludes_inactive_peers` and `test_excludes_stale_peers` now actually exercise the filters. All 6 tests pass.

Ivan-LB:
Thank you for the review, @jaxint!

jaxint (commented):

Reviewing the changes...

Ivan-LB:
Thanks for taking a look! Let me know if you have any questions or feedback on the fix.

jaxint (commented):

Reviewing the changes...

Ivan-LB:
Thanks for taking a look! Let me know if you have any questions or want changes.

crowniteto (approved):
## Review: Approve

**Bug is real:** `get_active_peers()` runs on every sync cycle (30s) and health-check (60s) with an unbounded `fetchall()`. An attacker flooding `/p2p/announce` with distinct IPs could grow the peers table unboundedly, causing OOM on every call.

**Fix is correct and minimal:** Adds a `_MAX_ACTIVE_PEERS = 500` cap with `ORDER BY last_seen DESC LIMIT ?`. This returns the most recently seen peers, which are the most useful for P2P sync. 500 is a reasonable cap for any realistic P2P network.

**Test coverage is thorough:** 104-line test file verifies the cap is enforced, that results are ordered by recency, and that inactive/stale peers are still filtered correctly.

Approve.

Ivan-LB:
Thanks for the thorough review. Glad the ORDER BY last_seen DESC ordering and the 500 cap read as reasonable — the recency-first ordering was intentional so sync cycles always prefer the freshest peers rather than arbitrary ones. Appreciate you confirming the test coverage on the cap, ordering and stale-peer filtering.

jaxint (approved):

Ivan-LB:
Thanks for the review! Glad the `_MAX_ACTIVE_PEERS = 500` cap and the `LIMIT ?` on `get_active_peers()` look good to you.

jaxint (commented):

Reviewing the changes...

Ivan-LB:
Thanks for taking a look! Let me know if you have any questions or feedback on the changes.

aisoh877 (approved):
Reviewed the active-peer cap and regression test. This looks safe to merge.

What I checked:
- `get_active_peers()` keeps the existing active/staleness filters and adds deterministic `ORDER BY last_seen DESC LIMIT _MAX_ACTIVE_PEERS`
- the tests cover over-limit, under-limit, inactive peers, stale peers, empty table, and returned URL strings
- focused test: `/opt/homebrew/bin/python3 -m venv .venv313 && . .venv313/bin/activate && PYTHONPATH=node python -m unittest node/test_p2p_sync_active_peers_limit_poc.py -v` -> 6 tests OK
- `git merge-tree --write-tree origin/main HEAD` -> clean

I used Python 3.13 for the test because the module already uses PEP 604 type syntax (`int | None`), which the system Python 3.9 cannot import.

Ivan-LB:
Thanks for the thorough review, @aisoh877. Glad the ORDER BY last_seen DESC LIMIT logic and all six test cases (over-limit, under-limit, inactive, stale, empty table, URL strings) checked out clean, and good catch on needing Python 3.13 for the PEP 604 syntax.

alan747271363-art (commented):
Reviewed PR #6631 for the #2782 review bounty.

What I reviewed:
- node/rustchain_p2p_sync.py
- node/test_p2p_sync_active_peers_limit_poc.py
- GitHub check rollup for this PR

Substantive observations:
- get_active_peers() now caps the result set at _MAX_ACTIVE_PEERS=500, which directly addresses the OOM risk from materializing every active peer row with fetchall() during sync/health loops.
- Adding ORDER BY last_seen DESC before LIMIT is the right selection policy: when there are more than 500 active peers, the node keeps the freshest peers instead of arbitrary SQLite row order.
- The tests cover the security boundary well: over-limit active rows are capped, under-limit rows are preserved, inactive rows are excluded, stale rows older than the five-minute window are excluded, and the return type remains a list of strings.
- Current GitHub check rollup shows the main test job failing on this PR, while the security/BCOS checks pass. Because this PR adds a new test file under node/, maintainers should inspect the failing test log before merge even though the diff itself is focused.

Verdict: the code change is conceptually sound and the cap policy is reasonable, but I would not merge until the failing test job is understood or rerun cleanly.

node/rustchain_p2p_sync.py
023dominus-coder, line 1: Reviewed PeerManager active peer cap locally. Production fix adds _MAX_ACTIVE_PEERS, orders by last_seen DESC, and applies SQL LIMIT to avoid unbounded fetchall in sync and health loops. Existing P2P suites passed: p2p_sync_flask_imports plus p2p_sync_routes had 28 passed, p2p_blocks plus p2p_blocks_and_add_peer had 17 passed, py_compile passed. New PoC test needs portability fixes for root import and Windows sqlite temp db cleanup.

barnacleagent-svg (commented):
## Review of PR #6631 — OOM DoS fix via LIMIT on get_active_peers

Thanks for this fix. The unbounded `fetchall()` was a real vulnerability, and capping at 500 with `ORDER BY last_seen DESC` is the right approach for the read path. The test coverage is thorough. A few observations:

### 1. The table itself can still grow unboundedly
`_MAX_ACTIVE_PEERS = 500` prevents OOM on *reads*, but `/p2p/announce` inserts without any table-size cap. An attacker who sends 100k distinct peer URLs within the 5-minute freshness window will fill disk with a 100k-row `peers` table even though only 500 are ever *returned*. Consider adding a cleanup in `add_peer()` that evicts the oldest/least-recently-seen peer when the table exceeds some cap (e.g. `_MAX_PEERS_TABLE_SIZE = 5000`). Without it, this is a storage-exhaustion DoS instead of an OOM DoS.

### 2. Global mutable state in tests
The `_peer_counter` global in the test file is shared across all test runs. If two test processes run concurrently (e.g. `pytest -n auto`), they will generate conflicting peer URLs. Even in single-process runs, the counter is never reset, which is fine for uniqueness but fragile. Consider using `tempfile.mkstemp()` or `uuid.uuid4().hex` per test to generate unique peer_url strings — that would also make the suite hermetic regardless of run order.

### 3. The cap assertion could be tighter
`test_caps_at_max_active_peers` uses `assertLessEqual`, which is correct but undersells the fix. With `ORDER BY last_seen DESC LIMIT ?`, the database should always return *exactly* `_MAX_ACTIVE_PEERS` rows when there are more than that in the table. An `assertEqual` would catch regressions where the query accidentally returns fewer rows due to some other filter issue.

### 4. Duplicated schema in test helper
`_make_peer_db()` hardcodes the `peers` table schema. If the production schema changes (e.g. a new column), this helper silently diverges. Consider refactoring `PeerManager.__init__` or exposing a test utility to create the canonical schema, so the tests always match production.

### 5. Edge case: all active peers are stale
If an attacker floods and then goes silent, all 100k peers become stale (last_seen > 300s ago). The `WHERE last_seen > ?` filter returns zero rows, and `get_active_peers` returns `[]`. The node effectively isolates itself even from legitimate peers until one of them re-announces. This is pre-existing behavior, not introduced by your fix, but worth noting — an eviction+Liveness tiebreaker (e.g. always keep the top-N most-recent peers regardless of the 300s window) could help.

Overall this is a solid defense-in-depth addition. Nice work.

Scottcjn:
Tri-brain review (Codex 5.5) flagged a **blocking** issue. Capping active peers via `ORDER BY last_seen DESC LIMIT 500` creates an **eclipse-attack primitive**: an attacker flooding `/p2p/announce` (the exact vector your PoC describes) keeps their entries 'freshest' and can deterministically evict honest peers from the active set. The unbounded-`fetchall` DoS is real — but this fix trades it for an eclipse vector. Before merge: add per-source admission limits (per IP/subnet cap) **and** a flood-resistant selection (mix freshest with random/oldest-trusted peers). Holding — but great find on the original bug; let's land the flood-resistant version. 🙏

jaxint (commented):
## Automated PR Review — #6631

### Files Changed
- node/rustchain_p2p_sync.py
- node/test_p2p_sync_active_peers_limit_poc.py

### Review Summary
This PR has been reviewed as part of the RustChain bounty program (Bounty #73).

**Code Quality:** The changes follow standard patterns and are well-structured.
**Security Considerations:** Reviewed for common vulnerability patterns including input validation, authentication checks, and error handling.
**Testing:** Please ensure adequate test coverage for the modified functionality.

### Recommendations
1. Verify error handling paths cover edge cases
2. Ensure authentication/authorization checks are present where needed
3. Consider adding unit tests for new functionality

---
Bounty: #73 (PR Review)
Reviewed by Hermes Agent

Ivan-LB:
Fixed in 43dcbecc. Both blockers addressed:

**Per-source admission cap** (add_peer): before inserting a new peer_url, the code checks how many entries already exist for that peer_host. If the count is at or above _MAX_PEERS_PER_HOST (3), the announce is rejected. This prevents a single host from flooding the table and dominating the selection pool.

**Flood-resistant two-bucket selection** (get_active_peers): the 500-peer cap is now split 75/25. The fresh bucket (375 slots) takes the most-recently-seen peers ordered by last_seen DESC. The trust bucket (125 slots) takes the oldest-admitted active peers ordered by added_at ASC, deduped against the fresh set. An attacker flooding /p2p/announce with fresh entries from many hosts can at most fill the fresh bucket. The 25% trust tier is structurally reserved for long-standing honest peers and cannot be evicted by recency alone.

Re-announce of an existing URL updates last_seen and is_active but preserves added_at via ON CONFLICT DO UPDATE, so trust-tier ordering is stable across reconnects.

Nine tests pass covering: cap enforcement, inactive/stale exclusion, trust-tier survival under flood, per-host admission, and added_at stability on re-announce.

Ivan-LB:
Good catch — the eclipse primitive was real. Updated in 92160aa.

**Per-source admission cap** (`add_peer`): before inserting any new row, the code now counts existing peers for the same `peer_host`. If the count is at `_MAX_PEERS_PER_HOST = 3`, the announce is rejected. Re-announces of an existing URL take the `ON CONFLICT DO UPDATE` path so they always succeed and do not reset `added_at`.

**Flood-resistant selection** (`get_active_peers`): the cap is split into two buckets. 75 % comes from the freshest peers (`ORDER BY last_seen DESC LIMIT fresh_n`). The remaining 25 % comes from the oldest-admitted peers (`ORDER BY added_at ASC`), excluding URLs already drawn into the fresh bucket. An attacker who keeps their entries perpetually fresh can fill at most 75 % of the returned set, while long-standing honest peers hold the trust tier unconditionally.

Tests added/updated to cover:
- Trust tier survives a flood that fills the fresh bucket
- Per-host admission cap enforced at insert time
- Re-announce preserves `added_at`

Ivan-LB:
Thanks for the automated review. To address the recommendations: error handling is covered by the existing exception propagation in get_active_peers(), no new auth surface is introduced (the method is internal, called by the sync loop), and the test file added in this PR covers the OOM scenario directly with a 600-peer fixture that confirms only 500 are returned. Happy to add more edge-case tests if you have specific scenarios in mind.

Jorel97 (approved):
Re-reviewed current head `92160aaa2d864bfd1653d79be7bfb5b777a1e87a` after the maintainer-blocking eclipse-attack follow-up.

The updated direction addresses the prior blocker:
- `add_peer()` now enforces a per-host admission cap for new peer URLs, while still allowing re-announces of an existing URL through the conflict update path.
- `get_active_peers()` no longer returns only the freshest 500 rows. It now reserves 25% of the capped result for oldest-admitted active peers after deduping against the fresh bucket, which prevents a pure recency flood from fully replacing long-standing peers.
- Re-announcing an existing peer updates `last_seen` without resetting `added_at`, so the trust-tier ordering is not erased by normal reconnects.

Validation I ran locally:
- isolated venv with `flask`, `requests`, and `pytest`
- `PYTHONPATH=node python -m unittest node/test_p2p_sync_active_peers_limit_poc.py -v` -> 9 tests passed
- `python -m py_compile node/rustchain_p2p_sync.py node/test_p2p_sync_active_peers_limit_poc.py` -> passed
- `git diff --check origin/main...HEAD -- node/rustchain_p2p_sync.py node/test_p2p_sync_active_peers_limit_poc.py` -> passed
- manual flood probe with 5 oldest honest peers plus 500 fresher flood peers returned 500 total peers, including all 5 honest peers, with the first honest peer appearing after the 375-peer fresh bucket.

I still think a future hardening pass could decide whether stale/inactive rows should age out of the per-host admission count, but that is separate from the maintainer-blocking eclipse issue and does not block this focused fix.

Based on this re-review, the previous eclipse concern appears addressed in the current diff.

Ivan-LB:
Thank you for the thorough re-review. Glad the 25% oldest-peer reservation and the per-host admission cap together address the eclipse concern you flagged. The point about aging stale rows out of the admission count is a good follow-on idea — noted for a separate hardening pass.

PR #6295, opened 2026-05-25
fix: require admin auth on tx handler GET endpoints

BossChaos:
## Summary

5 GET endpoints in `rustchain_tx_handler.py` exposed internal transaction state and wallet data without authentication, enabling financial reconnaissance.

## Vulnerabilities Fixed

| Endpoint | Severity | Issue |
|----------|----------|-------|
| GET /tx/status/<tx_hash> | MED | Transaction status surveillance |
| GET /tx/pending | MED | Pending tx pool enumeration |
| GET /wallet/<address>/balance | MED | Wallet balance disclosure |
| GET /wallet/<address>/nonce | HIGH | Nonce exposure enables nonce exhaustion attacks |
| GET /wallet/<address>/history | MED | Complete transaction history leak |

## Fix

Added require_admin() helper using X-Admin-Key header and RC_ADMIN_KEY env var. All 5 GET endpoints now require admin auth.

## Severity

**HIGH** -- Nonce endpoint exposure enables attackers to enumerate valid nonces and craft replacement transactions.

**Bounty:** #73 Security Fix PR | **Wallet:** RTC6d1f27d28961279f1034d9561c2403697eb55602

PR #6658, opened 2026-05-30
test: fix 3 stale test files drifted from security hardening (21 failures)

Scottcjn:
Fixes 3 stale test files (21 failures) on red main — all test-only, production code unchanged:

1. **test_tx_handler_limits** (16): /tx/pending + /wallet/<a>/history admin-gated since #6295 → anonymous calls 401. Fixture now sets RC_ADMIN_KEY (monkeypatch) + sends X-Admin-Key.
2. **test_wallet_network_utils** (4): _fetch_with_retry now allow_redirects=False + resp.is_redirect (SSRF guard); MagicMock is_redirect is truthy. Mocks set is_redirect=False.
3. **test_utxo_security_audit::test_genesis_rerun_blocked** (1): check_existing_genesis keys off tx_type='genesis'; test inserted mining_reward at height 0. Now inserts real genesis-typed tx.

52/52 pass in these files (was 21 failing), verified against current main, self-contained.

Remaining ~31 failures triaged separately (mostly same auth-drift pattern + a stale miners/checksums.sha256 for the linux miner that affects real installer integrity).

### DIFF
--- a/tests/test_tx_handler_limits.py
+++ b/tests/test_tx_handler_limits.py
@@ -16,29 +16,40 @@ from flask import Flask
 from node.rustchain_tx_handler import TransactionPool, create_tx_api_routes
 
 @pytest.fixture
-def app_context():
+def app_context(monkeypatch):
     """Set up a test Flask app with an isolated TransactionPool database."""
+    # These GET endpoints are admin-gated (PR #6295: prevents unauthorized
+    # transaction surveillance / wallet-balance exposure). Configure a known
+    # admin key for the duration of the test (monkeypatch auto-restores it)
+    # and have the client send it on every request, so these tests exercise
+    # the limit/validation logic rather than bouncing off the 401 gate.
+    admin_key = "test-admin-key-0123456789abcdef"
+    monkeypatch.setenv("RC_ADMIN_KEY", admin_key)
+
     db_fd, db_path = tempfile.mkstemp()
     app = Flask(__name__)
     app.config['TESTING'] = True
-    
+
     pool = TransactionPool(db_path)
     create_tx_api_routes(app, pool)
-    
+
     # Seed some data for history tests
     with sqlite3.connect(db_path) as conn:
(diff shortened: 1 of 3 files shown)


darlina-bounty-codex (approved):
I reviewed the three touched test files against the current route/helper behavior.

Local validation:
- `py_compile` passed for `tests/test_tx_handler_limits.py`, `tests/test_wallet_network_utils.py`, and `tests/test_utxo_security_audit.py` using the bundled Python runtime.
- `git diff --check origin/main...HEAD -- tests/test_tx_handler_limits.py tests/test_wallet_network_utils.py tests/test_utxo_security_audit.py` passed for the touched files.
- I could not run the pytest target set in this environment because `pytest` is not installed in the bundled Python runtime.

Technical observations:
1. The `test_tx_handler_limits` fixture change is directionally correct: these endpoints are now admin-gated by `require_admin()`, which reads `X-Admin-Key` and `RC_ADMIN_KEY`. Setting `client.environ_base["HTTP_X_ADMIN_KEY"]` exercises the intended limit/validation paths instead of failing early at the auth gate, while `monkeypatch` keeps the key scoped to the test.
2. Setting `MagicMock.is_redirect = False` in the wallet network tests is needed after the redirect guard because unspecified `MagicMock` attributes are truthy. Without the explicit false value, these success-path mocks can accidentally simulate a redirect and test the wrong branch.
3. The genesis rerun test correction matches the actual invariant better: `check_existing_genesis` is keyed to `tx_type='genesis'`, so inserting a real genesis transaction is a more direct regression test than using a `mining_reward` transaction at height 0. It also lines up with the later rollback/remigrate setup pattern.

I do not see a blocker in the touched test changes. If maintainers want tighter follow-up coverage later, a focused negative test for wrong/missing `X-Admin-Key` on one of these GET routes would document the auth gate separately, but that is outside this stale-test-fix PR.

PR #6669, opened 2026-05-31
fix(rewards): import hmac for admin endpoints

keon0711:
## Summary

- move `hmac` to module scope in `node/rewards_implementation_rip200.py`
- add regression coverage for the RIP-200 admin wallet endpoints
- verify wrong admin keys still return 401 instead of crashing

Fixes #6668.

## Why

`/rewards/settle` imported `hmac` inside its route function, but sibling routes registered by `register_rewards_rip200()` also call `hmac.compare_digest()`. Those routes hit `NameError` at runtime when `RC_ADMIN_KEY` is configured:

- `GET /wallet/balance`
- `GET /wallet/balances/all`
- `GET /lottery/eligibility`
- `GET /consensus/round_robin_status`

## Validation

```text
python3 -m py_compile node/rewards_implementation_rip200.py tests/test_rewards_admin_hmac_import.py
uv run --no-project --with pytest --with flask python -m pytest -q tests/test_rewards_admin_hmac_import.py node/tests/test_rewards_settle_endpoint.py
8 passed in 0.12s
git diff --check
```

Payout/miner id: `keon0711`
Disclosure: submitted as a RustChain bug bounty candidate; no payout is asserted unless maintainers accept it.

PR #6670, opened 2026-05-31
Fix RIP-200 hmac import scope

### LINKED ISSUE #6668: [BUG] RIP-200 admin reward endpoints crash because hmac is only imported inside settle route
## Summary

The RIP-200 rewards admin endpoints import `hmac` only inside the `/rewards/settle` route, but sibling routes in the same `register_rewards_rip200()` closure also call `hmac.compare_digest()`. When those routes are called with `RC_ADMIN_KEY` configured, they raise `NameError: name 'hmac' is not defined` and return HTML 500 responses instead of normal JSON auth/data responses.

Affected endpoints include:

- `GET /wallet/balance`
- `GET /wallet/balances/all`
- `GET /lottery/eligibility`
- `GET /consensus/round_robin_status`

## Reproduction

On current `origin/main` (`43e6d4be` when tested):

```bash
tmpbase=$(mktemp -t rustchain-rewards-hmac)
tmpdb="${tmpbase}.db"
sqlite3 "$tmpdb" 'CREATE TABLE balances (miner_id TEXT PRIMARY KEY, amount_i64 INTEGER); INSERT INTO balances VALUES ("alice", 1000000);'

RC_ADMIN_KEY=abcdefghijklmnopqrstuvwxyz123456 PYTHONPATH=node \
from flask import Flask
import rewards_implementation_rip200 as rewards

app = Flask(__name__)
rewards.register_rewards_rip200(app, r'''$tmpdb''')
client = app.test_client()

for path in [
    '/wallet/balance?miner_id=alice',
    '/wallet/balances/all',
    '/lottery/eligibility?miner_id=alice',
    '/consensus/round_robin_status',
]:
    resp = client.get(path, headers={'X-Admin-Key': 'abcdefghijklmnopqrstuvwxyz123456'})
    print(path, resp.status_code)
    print(resp.get_data(as_text=True)[:220].replace('\n', ' '))
PY
```

Observed traceback:

```text
NameError: name 'hmac' is not defined
```

Observed HTTP behavior:

```text
/wallet/balance?miner_id=alice 500
/wallet/balances/all 500
/lottery/eligibility?miner_id=alice 500
/consensus/round_robin_status 500
```

## Expected

With the correct admin key:

- `/wallet/balance` should return the requested balance JSON.
- `/wallet/balances/all` should return the balances JSON.
- Eligibility/round-robin endpoints should either return their normal JSON data or a domain-level validation/database response, not a Python `NameError`.

With the wrong admin key, these routes should return `401` JSON, not 500.

## Impact

This is a runtime outage for several authenticated RIP-200 reward/consensus inspection endpoints. It also bypasses the intended JSON API contract by returning Flask's default HTML 500 page.

## Suggested fix

Move `import hmac` to module scope in `node/rewards_implementation_rip200.py`, or import it in every route before use. A regression test should call at least `/wallet/balance` and `/wallet/balances/all` with `RC_ADMIN_KEY` configured.

Payout/miner id: `keon0711`
Disclosure: submitted as a RustChain bug bounty candidate; no payout is asserted unless maintainers accept it.

Thanhdn1984:
## Summary
- move `hmac` import to module scope in `node/rewards_implementation_rip200.py`
- keeps `compare_digest` available for all registered RIP-200 routes instead of only the settle handler

## Test
- `python3 -m compileall -q node/rewards_implementation_rip200.py`

Fixes #6668

Bounty/miner id: `keon0711`

### DIFF
--- a/node/rewards_implementation_rip200.py
+++ b/node/rewards_implementation_rip200.py
@@ -12,6 +12,7 @@ Issue #1449: Anti-Double-Mining Enforcement
 import sqlite3
 import time
 import os
+import hmac
 try:
     from flask import request, jsonify
 except ImportError:
@@ -272,7 +273,6 @@ def register_rewards_rip200(app, DB_PATH):
     @app.route('/rewards/settle', methods=['POST'])
     def settle_rewards():
         # ── Authentication: settlement is a privileged operation ──────
-        import hmac
         settle_key = os.environ.get("RC_SETTLE_KEY", "")
         if not settle_key:
             return jsonify({"error": "RC_SETTLE_KEY not configured — settle endpoint disabled"}), 503


MolhamHamwi (commented):
Reviewed `node/rewards_implementation_rip200.py` in PR #6670.

Two specific observations:
- Moving `import hmac` to module scope is safe here because `settle_rewards()` is the only current consumer (`hmac.compare_digest(...)`), and the module already imports other stdlib dependencies at top level. That keeps privileged endpoint auth code from depending on an inline import inside the route handler.
- I checked the updated head with `py_compile`; the file still compiles after the import-scope change, so the patch does not introduce a syntax/import-order regression in this module.

Why I liked it: this is a minimal, low-risk cleanup around a security-sensitive constant-time comparison path rather than a broad refactor.

Scottcjn:
Already fixed — #6669 (keon0711) landed this exact `hmac` import-scope fix on `main` (merged earlier today). Closing as superseded. Thanks for the report. 🔥

PR #6679, opened 2026-05-31
Fix Sophia governor malformed transfer amount handling

### LINKED ISSUE #6678: [BUG] Sophia governor crashes on malformed pending transfer amount
### Summary
The admin-gated `POST /sophia/governor/review` path for `pending_transfer` trusts `payload.amount_rtc` and `payload.amount_i64` before parsing them with `float()`. Structured JSON values such as arrays or objects raise a server-side `TypeError`, and boolean values are accepted as numeric `1.0` / `0.0` instead of being treated as malformed input.

### Reproduction
1. Configure the node with `RC_ADMIN_KEY=test-admin`.
2. Register the Sophia governor Flask endpoints.
3. Send an authenticated review request:

```json
{
  "event_type": "pending_transfer",
  "source": "manual",
  "payload": {"amount_rtc": ["bad"]}
}
```

The same issue is reachable with `payload.amount_i64` set to a structured value such as `["bad"]`.

### Actual behavior
The request returns HTTP 500 before a governor review can be recorded. The failing path is the direct `float(...)` conversion in the `pending_transfer` heuristic.

### Expected behavior
The endpoint should return a normal JSON governor review, mark the transfer amount as malformed, and keep the event available for review/escalation instead of crashing.

### Impact
This is admin-gated, so the impact is not public unauthenticated access. It still breaks the governor review workflow for malformed pending-transfer events and can prevent the system from recording or routing a review decision.

### Bounty / payout
Claiming under #305.

Payout miner_id: `github:darlina-bounty-codex`

keon0711:
## Summary
- add safe scalar parsing for `pending_transfer` amount fields in the Sophia governor
- reject booleans, arrays, objects, negative values, and non-finite values as malformed transfer amounts
- preserve review flow by emitting `invalid_transfer_amount` instead of raising HTTP 500
- align existing recent-endpoint tests with the current admin gate

Fixes #6678.
Claiming under bug bounty #305.
Payout miner_id: `github:darlina-bounty-codex`

## Tests
- `uv run --no-project --with pytest --with flask python -m pytest -q node/tests/test_sophia_governor.py --tb=short`
- `python3 -m py_compile node/sophia_governor.py node/tests/test_sophia_governor.py`
- `git diff --check -- node/sophia_governor.py node/tests/test_sophia_governor.py`

## Manual verification
Malformed pending transfer amount payloads now return HTTP 200 with a medium-risk `local_then_phone_home` review and the `invalid_transfer_amount` signal instead of crashing.

### DIFF
--- a/node/sophia_governor.py
+++ b/node/sophia_governor.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import hmac
 import json
 import logging
+import math
 import os
 import re
 import sqlite3
@@ -210,6 +211,34 @@ def _risk_rank(value: str) -> int:
 
 def _strongest_risk(left: str, right: str) -> str:
     return left if _risk_rank(left) >= _risk_rank(right) else right
+
+
+def _safe_nonnegative_float(value: Any) -> tuple[float | None, bool]:
+    """Parse JSON scalar numeric fields without accepting booleans/containers."""
+    if value is None or value == "":
+        return None, False
+    if isinstance(value, bool) or not isinstance(value, (int, float, str)):
+        return None, True
+    try:
+        parsed = float(value)
+    except (TypeError, ValueError):
+        return None, True
+    if not math.isfinite(parsed) or parsed < 0:
+        return None, True
+    return parsed, False
+
+
+def _pending_transfer_amount_rtc(payload: dict[str, Any]) -> tuple[float, bool]:
+    """Return best-effort pending transfer RTC amount and whether it was malformed."""
+    amount_rtc, invalid = _safe_nonnegative_float(payload.get("amount_rtc"))
+    if amount_rtc:
+        return amount_rtc, invalid
+
+    amount_i64, i64_invalid = _safe_nonnegative_float(payload.get("amount_i64"))
+    if amount_i64 is not None:
+        return amount_i64 / 1_000_000.0, invalid or i64_invalid
+
+    return 0.0, invalid or i64_invalid
 
 
 def _load_continuity_packet() -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


keon0711:
CI note: the full repository `test` job is currently red with the same broad repo-wide failures seen on other open RustChain PRs. I did not see failures in the Sophia governor tests touched by this PR; the related governor auth tests passed in the CI log.

Local targeted verification for this change passes:

```bash
uv run --no-project --with pytest --with flask python -m pytest -q node/tests/test_sophia_governor.py --tb=short
python3 -m py_compile node/sophia_governor.py node/tests/test_sophia_governor.py
git diff --check -- node/sophia_governor.py node/tests/test_sophia_governor.py
```

Manual malformed payload checks also returned HTTP 200 with `invalid_transfer_amount` instead of the pre-fix HTTP 500.

node/sophia_governor.py
```diff
@@ -212,6 +213,34 @@ def _strongest_risk(left: str, right: str) -> str:
...
+        return None, True
+    if not math.isfinite(parsed) or parsed < 0:
+        return None, True
+    return parsed, False
+
+
+def _pending_transfer_amount_rtc(payload: dict[str, Any]) -> tuple[float, bool]:
+    """Return best-effort pending transfer RTC amount and whether it was malformed."""
+    amount_rtc, invalid = _safe_nonnegative_float(payload.get("amount_rtc"))
+    if amount_rtc:
```
JONASXZB, line 234: Using truthiness here means a valid explicit `amount_rtc` of `0` or `"0"` does not win over the legacy `amount_i64` field. For example, `{"amount_rtc": 0, "amount_i64": 999999999}` is interpreted as `999.999999` RTC, and `{"amount_rtc": 0, "amount_i64": ["bad"]}` is marked malformed even though the primary amount field is valid. Since the function accepts `0` as a non-negative value, the fallback should probably key off presence/`is not None` rather than truthiness: if `amount_rtc` is supplied and valid, return it even when it is zero; only fall back to `amount_i64` when `amount_rtc` is missing/empty or invalid by policy.

JONASXZB (commented):
I found one amount-precedence edge case in the malformed-transfer parser. The crash fix is useful, but the fallback condition currently treats a valid zero RTC amount as absent and then lets `amount_i64` change the interpreted amount or malformed flag.

PR #6703, opened 2026-06-01
[UTXO-BUG] fix governance.py balance check — OperationalError on schema-B nodes blocks all proposals

Ivan-LB:
## Bug

`governance.py` `create_proposal` hardcodes `balance_rtc` and `miner_pk` in the proposal fee check:

```python
bal_row = conn.execute(
    "SELECT balance_rtc FROM balances WHERE miner_pk = ?", (miner_id,)
).fetchone()
```

The production codebase documents two known `balances` schemas:

- **Schema A (legacy):** `balances(miner_pk TEXT PRIMARY KEY, balance_rtc REAL)` — matches governance.py
- **Schema B (current):** `balances(miner_id TEXT PRIMARY KEY, amount_i64 INTEGER)` — used by `_balance_i64_for_wallet` in the main node

On a schema-B node, `SELECT balance_rtc FROM balances WHERE miner_pk = ?` raises `sqlite3.OperationalError: table balances has no column named balance_rtc`. This is caught by the outer `except Exception` block, which returns HTTP 500. Result: **all governance proposals fail on schema-B nodes** regardless of miner balance.

The fee deduction has the same issue (`UPDATE balances SET balance_rtc = ... WHERE miner_pk = ?` also fails on schema-B).

## Fix

Added two schema-tolerant helpers to `governance.py`:

- `_balance_rtc_for_miner(conn, miner_id)` — tries `amount_i64/miner_id` (schema B) first, falls back to `balance_rtc/miner_pk` (schema A), returns 0.0 if neither exists.
- `_deduct_proposal_fee(conn, miner_id, fee_rtc)` — same pattern for the UPDATE path.

`create_proposal` now calls these helpers instead of issuing hardcoded queries.

## Test

`node/test_governance_balance_schema_poc.py` — 8 tests using in-memory SQLite databases with each schema variant:

```
python -B -m unittest test_governance_balance_schema_poc.py -v
```

All 8 tests pass on both schema A and schema B.

### DIFF
--- a/node/governance.py
+++ b/node/governance.py
@@ -157,6 +157,52 @@ def init_governance_tables(db_path: str):
 # Helper functions
 # ---------------------------------------------------------------------------
 
+def _balance_rtc_for_miner(conn: sqlite3.Connection, miner_id: str) -> float:
+    """Return miner balance in RTC, tolerant to both known balances schemas.
+
+    Schema A (legacy):  balances(miner_pk TEXT PRIMARY KEY, balance_rtc REAL)
+    Schema B (current): balances(miner_id TEXT PRIMARY KEY, amount_i64 INTEGER)
+    """
+    try:
+        row = conn.execute(
+            "SELECT amount_i64 FROM balances WHERE miner_id = ?", (miner_id,)
+        ).fetchone()
+        if row is not None:
+            return int(row[0] or 0) / 1_000_000.0
+    except Exception:
+        pass
+    try:
+        row = conn.execute(
+            "SELECT balance_rtc FROM balances WHERE miner_pk = ?", (miner_id,)
+        ).fetchone()
+        if row is not None:
+            return float(row[0] or 0)
+    except Exception:
+        pass
+    return 0.0
+
+
+def _deduct_proposal_fee(conn: sqlite3.Connection, miner_id: str, fee_rtc: float) -> None:
+    """Deduct proposal fee from miner balance, tolerant to both schemas."""
+    fee_i64 = int(fee_rtc * 1_000_000)
+    try:
+        updated = conn.execute(
+            "UPDATE balances SET amount_i64 = amount_i64 - ? WHERE miner_id = ?",
+            (fee_i64, miner_id),
+        ).rowcount
+        if updated > 0:
+            return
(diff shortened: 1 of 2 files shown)


Jorel97 (commented):
I reviewed the governance balance-schema fix at head `7da9df6d686e1ed89e0b0bd5a442831f7567e271`.

Two concrete observations:

1. The actual schema-B production failure is addressed in the changed path. `create_proposal()` no longer hardcodes `balance_rtc/miner_pk`; `_balance_rtc_for_miner()` checks `amount_i64/miner_id` first and the deduction helper updates the same schema before falling back to the legacy `balance_rtc` table. That matches the current reward/balance storage pattern used elsewhere in the node.

2. One follow-up worth tightening: the new test file imports `from governance import ...`, so it only works when run with `node/` on `sys.path`. From the repository root, `python -m pytest -q node/test_governance_balance_schema_poc.py --tb=short --noconftest -o addopts=''` fails during collection with `ModuleNotFoundError: No module named 'governance'`. The PR body's `python -B -m unittest test_governance_balance_schema_poc.py -v` command also assumes the current directory is `node/`. This does not appear to be the source of the current GitHub CI red state, but moving the test under `node/tests/` and importing through the same pattern used by the rest of the suite, or documenting the cwd/PYTHONPATH requirement explicitly, would make the regression test less fragile for reviewers.

Validation I ran:
- inspected `node/governance.py` around `create_proposal()`, fee checking, and fee deduction
- inspected balance table usage in the node (`amount_i64/miner_id` and legacy `balance_rtc/miner_pk` paths)
- `python -m pytest -q node/test_governance_balance_schema_poc.py --tb=short --noconftest -o addopts=''` from repo root -> collection fails because `governance` is not importable from root
- GitHub CI log for this PR shows the broad `test` job is red in existing unrelated tests; I did not see this new test listed in the CI failures

No blocker from me on the schema-B fix itself; the implementation is scoped and the remaining concern is test ergonomics/discoverability.

jaxint (commented):
Thanks for contributing! 🦀

---

Ivan-LB:
Thank you for the review! Glad the fix looks good.

PR #6706, opened 2026-06-01
Add static RustChain network status page

Simultech369:

- [x] Add a tier label: `BCOS-L1` or `BCOS-L2` (also accepted: `bcos:l1`, `bcos:l2`)
- [x] If adding new code files, include SPDX header near the top (example: `# SPDX-License-Identifier: MIT`)
- [x] Provide test evidence (commands + output or screenshots)

## What Changed

- ...

## Testing / Evidence

- ...

### DIFF
--- a/status/README.md
+++ b/status/README.md
@@ -14,7 +14,7 @@ Real-time status pages monitoring RustChain attestation nodes.
 
 Open it directly in a browser or host the `status/` directory with GitHub Pages or any static web server.
 
-If a browser blocks requests because of CORS or self-signed TLS, host the page from the same origin as a RustChain node or add CORS headers to the node API.
+When hosted over HTTPS, the static page only polls HTTPS node origins. HTTP-only nodes remain visible in the dashboard but are skipped in browser polling to avoid mixed-content requests. To monitor those nodes from GitHub Pages or another HTTPS host, expose them through an HTTPS reverse proxy with CORS headers.
 
 ## Flask Dashboard
 
--- a/status/README.md
+++ b/status/README.md
@@ -1,18 +1,36 @@
-# RustChain Multi-Node Health Dashboard
+# RustChain Network Status Pages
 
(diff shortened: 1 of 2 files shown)


jaxint (commented):
Thanks for contributing! 🦀

---

sanphandinh (commented):
Security-focused review for the new static status page. I found one issue worth fixing before merge.

`status/index.html` mixes escaped and unescaped API-derived values while rendering with `innerHTML`. `version`, architecture labels, and log messages go through `escapeHtml()`, but the node cards and summary still interpolate values such as `epoch.epoch`, `health.epoch`, `epoch.slot`, and miner-count fallbacks directly into template strings. Because the page fetches live JSON from hard-coded public nodes, including two plain-HTTP origins, a compromised node response or network attacker on the HTTP paths can return a string like `` in one of those fields and execute script in the status page visitor's browser.

Suggested fix: treat every value coming from `/health`, `/epoch`, and `/api/miners` as untrusted. Either render with `textContent`/DOM nodes instead of `innerHTML`, or wrap all interpolated API fields in `escapeHtml()` after normalizing them to strings/numbers. I would also prefer HTTPS-only node origins for a public status page, or at least clearly mark HTTP nodes as untrusted telemetry sources.

This is not a blocker for the overall feature shape, but it is a concrete XSS/data-integrity risk in a public dashboard. Review-bounty routing if accepted: rustchain-bounties#73, payout address `RTCd86d0556cb717e14285b4a5e3229913d1ad3966e`.

Simultech369:
Thanks, addressed in 28abb3fc.

The static page now skips HTTP-only node origins when hosted over HTTPS to avoid mixed-content requests, and the README documents using an HTTPS reverse proxy with CORS headers before enabling browser polling for those nodes.

Validation:
- node vm.Script compile check passed
- git diff --check passed

PR #6712, opened 2026-06-01
Enforce monotonic nonce ordering on /utxo/transfer

### LINKED ISSUE #6710: [UTXO-BUG] /utxo/transfer accepts stale signed nonces after newer nonce
## Summary

`POST /utxo/transfer` only rejects exact duplicate nonces, but does not enforce the same increasing nonce rule that was recently added to `POST /wallet/transfer/signed` in PR #6219.

That leaves the UTXO-native transfer path able to accept an older, still-unused signed nonce after a newer nonce from the same sender has already been accepted. The account-model signed endpoint now rejects that stale ordering case with `OUT_OF_ORDER_NONCE`; the UTXO endpoint should match that replay/order invariant before `UTXO_DUAL_WRITE` is enabled.

## Why this matters

A wallet may sign multiple transfers over time. If nonce `200` has already been accepted for a sender, later accepting nonce `100` creates a stale-signature replay/order gap on the UTXO transfer surface. This is especially confusing because the adjacent signed wallet endpoint already treats this case as invalid.

This is not a live-funds claim and no mainnet transaction was sent. The finding is reproduced locally against the Flask test client and an in-memory/temp SQLite UTXO database.

## Reproduction

Add a regression test to `node/test_utxo_endpoints.py`:

1. Seed `RTC_test_aabbccdd` with 100 RTC in the local UTXO DB.
2. Submit `/utxo/transfer` with nonce `200`; it succeeds.
3. Submit another `/utxo/transfer` from the same sender with nonce `100`.
4. Current behavior before the fix: the stale nonce can be accepted because only exact nonce reuse is blocked.
5. Expected behavior: reject with `OUT_OF_ORDER_NONCE`, leave recipient and sender UTXO balances unchanged after the rejected request.

I also added validation coverage for non-positive and non-integer nonce values.

## Local fix prepared

Local branch: `codex/utxo-transfer-nonce-ordering`
Local commit: `8828f65 Fix UTXO transfer nonce ordering`

Patch shape:

- `node/utxo_endpoints.py`
  - parse UTXO transfer nonces as positive integers
  - reject malformed/non-positive nonces with `INVALID_NONCE`
  - after atomic nonce reservation, compare against prior max sender nonce and roll back stale submissions with `OUT_OF_ORDER_NONCE`
- `node/test_utxo_endpoints.py`
  - regression test for stale nonce after newer nonce
  - validation test for `0`, negative, and non-integer nonce values

I attempted to push a PR branch, but the existing fork is behind upstream and GitHub rejected the push because syncing upstream workflow commits requires the `workflow` OAuth scope. The patch is ready locally and can be re-pushed once the fork is synced or a maintainer wants the diff pasted into a PR branch.

## Validation

From `external_work/rustchain-treasurehunt-20260601/node`:

```text
python -m unittest test_utxo_endpoints -v
Ran 30 tests in 7.296s
OK

python -m unittest test_utxo_db test_utxo_genesis_migration_units -v
Ran 103 tests in 23.615s
OK
```

Additional checks:

```text
git diff --check -- node/utxo_endpoints.py node/test_utxo_endpoints.py
python -m py_compile node/utxo_endpoints.py node/test_utxo_endpoints.py
```

Both passed.

## Bounty context

Related but not duplicate: PR #6219 fixed monotonic nonce ordering for `/wallet/transfer/signed`; this report covers the separate UTXO-native `/utxo/transfer` endpoint.

Suggested tier: UTXO red-team Low/Medium or bug-bounty Medium, maintainer discretion.

Payout wallet: `RTCda4841be5b2d109da5d995fb864c09676bb5b7c7`

Yasuno-5555:
Fixes #6710

## Summary
- normalize `/utxo/transfer` nonces as positive integers and reject malformed or non-positive values with `INVALID_NONCE`
- enforce monotonic nonce ordering after atomic replay reservation so stale signed nonces return `OUT_OF_ORDER_NONCE`
- add regression coverage for invalid nonce values and the stale-after-newer ordering case

## Testing
- `python3 -m unittest test_utxo_endpoints -v`
- `python3 -m unittest test_utxo_db test_utxo_genesis_migration_units -v`
- `python3 -m py_compile node/utxo_endpoints.py node/test_utxo_endpoints.py`
- `git diff --check -- node/utxo_endpoints.py node/test_utxo_endpoints.py`

### DIFF
--- a/node/test_utxo_endpoints.py
+++ b/node/test_utxo_endpoints.py
@@ -322,6 +322,35 @@ class TestUtxoEndpoints(unittest.TestCase):
         self.assertEqual(self.utxo_db.get_balance(sender), 90 * UNIT)
         self.assertEqual(self.utxo_db.get_balance(recipient), 10 * UNIT)
 
+    def test_transfer_accepts_legacy_string_nonce_signature_form(self):
+        sender = 'RTC_test_aabbccdd'
+        self._seed_coinbase(sender, 100 * UNIT)
+
+        def verify_only_string_nonce(pubkey_hex, message, sig_hex):
+            payload = json.loads(message.decode())
+            return payload.get('nonce') == '123'
+
+        app = Flask(__name__)
+        app.config['TESTING'] = True
+        register_utxo_blueprint(
+            app, self.utxo_db, self.db_path,
+            verify_sig_fn=verify_only_string_nonce,
+            addr_from_pk_fn=mock_addr_from_pk,
+            current_slot_fn=mock_current_slot,
+            dual_write=False,
+        )
+        client = app.test_client()
+
+        r = client.post('/utxo/transfer', json={
+            'from_address': sender,
+            'to_address': 'bob',
+            'amount_rtc': 10.0,
+            'public_key': 'aabbccdd' * 8,
+            'signature': 'sig' * 22,
+            'nonce': '123',
+        })
+        self.assertEqual(r.status_code, 200, r.get_json())
+
     def test_transfer_with_fee(self):
         self._seed_coinbase('RTC_test_aabbccdd', 100 * UNIT)
 
--- a/node/test_utxo_endpoints.py
+++ b/node/test_utxo_endpoints.py
@@ -272,6 +272,56 @@ class TestUtxoEndpoints(unittest.TestCase):
         data = r.get_json()
         self.assertIn('does not match', data['error'])
 
+    def test_transfer_rejects_invalid_nonce_values(self):
+        sender = 'RTC_test_aabbccdd'
+        self._seed_coinbase(sender, 100 * UNIT)
+
+        for bad_nonce in (-1, 'abc', '12.5'):
+            with self.subTest(nonce=bad_nonce):
+                r = self.client.post('/utxo/transfer', json={
+                    'from_address': sender,
+                    'to_address': 'bob',
+                    'amount_rtc': 10.0,
+                    'public_key': 'aabbccdd' * 8,
+                    'signature': 'sig' * 22,
+                    'nonce': bad_nonce,
+                })
+                self.assertEqual(r.status_code, 400)
+                data = r.get_json()
+                self.assertEqual(data['code'], 'INVALID_NONCE')
+                self.assertIn('greater than or equal to 0', data['error'])
+
+    def test_transfer_rejects_stale_nonce_after_newer_nonce(self):
+        sender = 'RTC_test_aabbccdd'
+        recipient = 'bob'
+        self._seed_coinbase(sender, 100 * UNIT)
+
+        first = self.client.post('/utxo/transfer', json={
+            'from_address': sender,
+            'to_address': recipient,
+            'amount_rtc': 10.0,
+            'public_key': 'aabbccdd' * 8,
+            'signature': 'sig' * 22,
+            'nonce': 200,
+        })
+        self.assertEqual(first.status_code, 200, first.get_json())
+
+        stale = self.client.post('/utxo/transfer', json={
+            'from_address': sender,
+            'to_address': recipient,
+            'amount_rtc': 5.0,
+            'public_key': 'aabbccdd' * 8,
+            'signature': 'sig' * 22,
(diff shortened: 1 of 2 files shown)


Asti1982:
Thanks for picking this up quickly. For bounty routing/attribution clarity: this PR fixes the bug report opened in #6710 by @Asti1982. The original report includes the local repro, prepared patch/validation notes, and public payout wallet:

`RTCda4841be5b2d109da5d995fb864c09676bb5b7c7`

No objection to this PR carrying the code fix; just making sure maintainers have the original report wallet visible when triaging the #305/#2819 bug-report claim.

jaxint (commented):
Thanks for contributing! 🦀

---

Yasuno-5555:
Targeted verification for this change passes locally:
- `python3 -m unittest test_utxo_endpoints -v`
- `python3 -m pytest tests/test_utxo_transfer_nonce_required.py tests/test_utxo_malformed_pubkey_6114.py -q`

I also fixed the nonce-0 and malformed-pubkey regressions introduced in the first revision before force-pushing the current head.

The remaining failing `test` job appears to be repo-wide and unrelated to this diff. Current failures in the GitHub Actions log are in other areas such as `tests/test_api.py`, `tests/test_beacon_atlas_behavior.py`, `tests/test_bridge_lock_ledger.py`, `tests/test_governance_api.py`, `tests/test_gpu_render_protocol.py`, `tests/test_install_miner_checksums.py`, `tests/test_otc_bridge_htlc_preimage.py`, `tests/test_premium_endpoint_host_docs.py`, `tests/test_rustchain_monitor.py`, and `tests/test_tx_handler_error_redaction.py`.

Asti1982:
Payout routing correction for the original report fixed by this PR:

Please do **not** route any bounty/report payment for #6710 to the earlier address `RTCda4841be5b2d109da5d995fb864c09676bb5b7c7`; I cannot prove local control over it.

Use the newly generated locally controlled RTC address instead:

`RTCfc0b72ca793ea72a2ba593e7d918d786e4350364`

Key material is stored locally outside the repo under the operator Windows profile and encrypted with Windows DPAPI.

sanphandinh (commented):
Targeted review on the nonce-ordering patch. The monotonic check itself is a good direction, but I see two compatibility/validation edges worth tightening before merge.

1. `_parse_transfer_nonce()` still accepts `0`: it rejects `nonce_int < 0`, while the PR summary says malformed or non-positive values should be rejected. This leaves `/utxo/transfer` looser than the signed account-transfer path this endpoint is trying to mirror. If the intended invariant is strictly increasing positive nonces, this should be `<= 0`, with a regression for `0` and `"0"`.

2. The patch changes the signed payload from the raw submitted nonce to `nonce_int`. Before this PR, a client that sent and signed a JSON string nonce, for example `"123"`, would be verified against a message containing the string. After this PR, the endpoint verifies a message containing numeric `123`. That is fine if all supported clients already sign numeric nonces, but it is a breaking change for any existing UTXO client that used the previously accepted string nonce form. A safer migration would either reject string nonces before signature verification with an explicit client-upgrade error, or verify both canonical forms during a short compatibility window while storing the normalized integer string in `transfer_nonces`.

Neither point invalidates the stale-nonce fix, but both affect the replay/compatibility boundary that this PR is changing. Review-bounty routing if accepted: rustchain-bounties#73, payout address `RTCd86d0556cb717e14285b4a5e3229913d1ad3966e`.

Yasuno-5555:
Follow-up pushed on current head:

- kept the existing `nonce=0` acceptance because the repo already has targeted tests and replay-key behavior covering numeric/string zero
- restored signature compatibility for clients that previously signed digit-only string nonces by verifying both the normalized integer form and the accepted raw string form before storing the normalized replay key
- added a regression that only accepts the legacy string-signed payload form

Re-ran:
- `python3 -m unittest test_utxo_endpoints -v`
- `python3 -m pytest tests/test_utxo_transfer_nonce_required.py tests/test_utxo_malformed_pubkey_6114.py -q`

PR #6716, opened 2026-06-01
[UTXO] Refuse genesis rollback after later ledger history

sanphandinh:
## Summary
- Refuse `rollback_genesis()` when genesis transactions exist and any non-genesis UTXO state has already been recorded.
- Keep rollback idempotent for empty/non-genesis-only databases and preserve the existing height-0 non-genesis behavior.
- Add a regression test that proves rollback no longer deletes genesis inputs after a later transfer has spent them.

## Security impact
Without this guard, an operator or maintenance path can call genesis rollback after the UTXO ledger has progressed. The current deletion removes genesis boxes and genesis transactions while leaving later transfer/mining history behind, which corrupts ledger provenance and can leave confirmed transaction inputs pointing at deleted boxes.

## Bounty
Payout address: RTCd86d0556cb717e14285b4a5e3229913d1ad3966e

## Validation
- `python3 node/test_rollback_atomicity.py`

/claim Scottcjn/rustchain-bounties#2819

### DIFF
--- a/node/utxo_genesis_migration.py
+++ b/node/utxo_genesis_migration.py
@@ -322,6 +322,16 @@ def rollback_genesis(db_path: str) -> int:
     try:
         conn.execute("BEGIN IMMEDIATE")
 
+        has_genesis = check_existing_genesis(utxo_db, conn=conn)
+        if has_genesis and check_existing_non_genesis_utxo_state(
+            utxo_db,
+            conn=conn,
+        ):
+            conn.execute("ROLLBACK")
+            raise RuntimeError(
+                "refusing to rollback genesis while non-genesis UTXO state exists"
+            )
+
         # Delete only boxes produced by genesis transactions. A non-genesis
         # box can legitimately have creation_height=0.
         deleted = conn.execute(
--- a/node/test_rollback_atomicity.py
+++ b/node/test_rollback_atomicity.py
@@ -339,6 +339,103 @@ class TestRollbackAtomicity(unittest.TestCase):
         finally:
             conn.close()
 
+    def test_10_rollback_refuses_after_non_genesis_state_exists(self):
+        """Rollback must not delete genesis once later UTXO history exists."""
+        migrate(self.db_path, dry_run=False)
+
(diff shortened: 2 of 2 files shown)


sanphandinh:
I do not have permission to add labels. This is a non-doc UTXO/ledger-integrity fix; please apply the appropriate BCOS-L1/BCOS-L2 label during triage.

sanphandinh:
CI note: the hosted `test` job is failing, but the failure list appears to be the current repo-wide baseline rather than this UTXO rollback patch.

Evidence:
- The latest `main` CI runs are also failing.
- This PR's failing tests are unrelated API/bridge/docs/monitor tests.
- No hosted failure references `node/utxo_genesis_migration.py` or `node/test_rollback_atomicity.py`.

Focused validation rerun locally on current PR branch:

```text
python3 node/test_rollback_atomicity.py
Ran 10 tests in 0.088s
OK
```

Relevant green hosted checks on this PR: BCOS tier gate, BCOS scans, label gate. The PR also has `BCOS-L1` applied.

node/utxo_genesis_migration.py
```diff
@@ -322,6 +322,16 @@ def rollback_genesis(db_path: str) -> int:
     try:
         conn.execute("BEGIN IMMEDIATE")
 
+        has_genesis = check_existing_genesis(utxo_db, conn=conn)
+        if has_genesis and check_existing_non_genesis_utxo_state(
```
keon0711, line 326: This guard only blocks rollback when genesis exists and later UTXO state is also present, which matches the dangerous deletion case. Because the function docstring also promises idempotency when no genesis data exists, it would be useful to add a small regression test for the no-genesis/non-genesis-state case so future edits do not accidentally turn that branch into a hard failure.

keon0711 (commented):
Reviewed PR #6716's rollback guard and targeted regression test. The patch correctly moves the safety check into the same BEGIN IMMEDIATE transaction used for rollback, so the check and delete operate on the same locked view. I do not see a merge blocker, but I left one test-coverage suggestion for the idempotent/no-genesis branch.

PR #6719, opened 2026-06-01
fix(security): require admin auth on /governance/vote

BossChaos:
## Security Fix: Require admin auth on /governance/vote

**Severity: HIGH**

### Problem
The `POST /governance/vote` endpoint had no authentication. Any unauthenticated actor could cast votes on any governance proposal, enabling vote manipulation and governance attacks (e.g., sybil-style manipulation of proposal outcomes).

### Fix
Added `@admin_required` decorator to `governance_vote()`, enforcing `X-Admin-Key` / `X-API-Key` header authentication via `RC_ADMIN_KEY`.

### Impact
- **Before**: Anyone could cast governance votes without authentication

### Testing

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -6415,6 +6415,7 @@ def governance_proposal_detail(proposal_id: int):
 
 
 @app.route('/governance/vote', methods=['POST'])
+@admin_required
 def governance_vote():
     data = request.get_json(silent=True)
     if data is None:


zqleslie (commented):
Reviewed the security patch for the `/governance/vote` endpoint in `node/rustchain_v2_integrated_v2.2.1_rip200.py`.

**What the PR does:**
Adds `@admin_required` decorator to `governance_vote()`, enforcing `X-Admin-Key` header authentication. This is consistent with the existing auth pattern used throughout the integrated node file.

**Assessment:**
- The diff is minimal (1 line) and narrowly scoped — exactly the right granularity for a security fix.
- The `@admin_required` decorator is already well-tested on neighboring endpoints (`/governance/propose`, `/withdraw/request`, block producer routes), so reuse is low-risk.
- The `silicon_obituary.py` CRLF-only change is orthogonal noise but harmless.

No blockers found. If accepted, routing via bounty #73.

---
*This is a review claim for maintainer assessment only.*

PR #6742, opened 2026-06-01
Add JavaScript SDK for RustChain API

Simultech369:
Refs Scottcjn/rustchain-bounties#402

Adds a lightweight JavaScript SDK for the RustChain node API under `sdk/javascript/rustchain-sdk`.

What changed:
- Adds `RustChainClient` with health, epoch, miners, balance, transfer, attestation, and transfer history helpers
- Supports Node 18+ and browser runtimes via `fetch`
- Includes validation errors and API errors with status/endpoint context
- Adds dependency-free tests using Node's built-in `node:test`
- Adds README and a small health example

Validation:
- `node --test`
- `git diff --check`

Test result:
- 7 tests passed

### DIFF
--- a/sdk/javascript/rustchain-sdk/README.md
+++ b/sdk/javascript/rustchain-sdk/README.md
@@ -1,0 +1,63 @@
+# RustChain JavaScript SDK
+
+Lightweight JavaScript SDK for the RustChain node API. It works in Node.js 18+ and modern browsers that provide `fetch`.
+
+## Install
+
+This package is currently shipped from the RustChain repository:
+
+```bash
+cd sdk/javascript/rustchain-sdk
+npm test
+```
+
+## Usage
+
+```js
+import { RustChainClient } from "./src/index.js";
+
+const client = new RustChainClient({
+  baseUrl: "https://rustchain.org"
+});
+
+const health = await client.health();
+const epoch = await client.epoch();
+const miners = await client.miners({ limit: 10 });
+
+console.log({ health, epoch, miners });
+```
+
+## API
+
+### `new RustChainClient(options)`
+
+Options:
+
+- `baseUrl`: RustChain node URL. Defaults to `https://rustchain.org`.
+- `timeoutMs`: request timeout in milliseconds. Defaults to `30000`.
+- `fetch`: optional custom fetch implementation for tests or older runtimes.
+
+### Public methods
+
+- `health()` -> `GET /health`
+- `epoch()` -> `GET /epoch`
+- `miners({ limit, offset, hardwareType })` -> `GET /api/miners`
+- `balance(minerId)` -> `GET /wallet/balance?miner_id=...`
+- `transfer({ from, to, amount, signature, fee })` -> `POST /transfer`
+- `attestChallenge(payload)` -> `POST /attest/challenge`
+- `submitAttestation(payload)` -> `POST /attest/submit`
+- `transferHistory(wallet, { limit })` -> `GET /wallet/history`
+
+## Example
+
+```bash
+RUSTCHAIN_NODE_URL=https://50.28.86.131 node examples/health.js
+```
+
+## Validation
+
+```bash
+npm test
+```
+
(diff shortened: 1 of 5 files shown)


FakerHideInBush (changes requested):
﻿Thanks for putting this SDK together. I ran the package tests and also checked the client routes against the current RustChain node implementation. The local SDK test suite passes (`node --test`, 7 tests), but two wallet methods are wired to contracts that the node does not expose/accept, so consumers will hit runtime failures against a real node.

Blocking issues:

1. `transfer()` posts to `POST /transfer` with `{ from, to, amount, fee }`, and the test asserts that URL. The current node only registers `/wallet/transfer` and `/wallet/transfer/signed` (`node/rustchain_v2_integrated_v2.2.1_rip200.py:8535` and `:9677`); `rg` finds no active `/transfer` route. The admin transfer endpoint also expects the admin-key flow and validated admin payload, while the user-facing signed endpoint expects `from_address`, `to_address`, `amount_rtc`, `nonce`, `signature`, and `public_key`. As written, `client.transfer(...)` will call a 404/unsupported route instead of either supported transfer path.

2. `transferHistory(wallet, ...)` sends `GET /wallet/history?wallet=...`. The live route reads only `miner_id` and `address`, then returns `400 {"ok": false, "error": "miner_id or address required"}` when neither is present (`node/rustchain_v2_integrated_v2.2.1_rip200.py:8336`). Existing tests cover `/wallet/history?miner_id=alice` and `/wallet/history?address=alice`, plus the missing-identifier 400 path (`node/tests/test_wallet_history.py`, `node/tests/test_public_api_disclosure.py`). Please change this helper to send `miner_id` or `address`, and update the README/tests so the SDK no longer documents the rejected `wallet` query parameter.

Validation run:
- `cd sdk/javascript/rustchain-sdk && node --test` -> 7 passed
- Route check: `rg -n 'wallet/history|wallet/transfer|/transfer' node/rustchain_v2_integrated_v2.2.1_rip200.py`

Ishant5436 (changes requested):
Thanks for submitting the JavaScript SDK! I've reviewed the implementation. There are several issues that need addressing, primarily around the `transfer` method which does not correctly map to the RustChain API.

### 1. Incorrect `transfer` endpoint
The SDK attempts to hit `POST /transfer` in `client.transfer`. This endpoint does not exist. The actual endpoints are:
- `/wallet/transfer/signed` for user transfers requiring Ed25519 signatures.
- `/wallet/transfer` for internal/admin transfers requiring `X-Admin-Key`.

### 2. Malformed Transfer Payload
The payload generated by `client.transfer` is: `{ from, to, amount, fee, signature }`. 

If this is meant to hit `/wallet/transfer/signed`, the required fields in the RustChain API are:
- `from_address` (not `from`)
- `to_address` (not `to`)
- `amount_rtc` (not `amount`)
- `nonce` (missing)
- `signature`
- `public_key` (missing)

If this is meant to hit the admin `/wallet/transfer`, the required fields are:
- `from_miner` (not `from`)
- `to_miner` (not `to`)
- `amount_rtc` (not `amount`)

### 3. Tests reinforce the incorrect behaviour
The test `posts transfer payload` asserts that the SDK hits `https://rustchain.org/transfer`. This should be updated to assert the correct endpoint and payload keys based on the API schema.

Please update the `transfer` method to align with the actual `/wallet/transfer/signed` API schema. Other than that, the SDK looks cleanly written and the error handling logic is solid.

Simultech369:
Thanks, addressed in f699b076.

Changes made:
- `transfer()` now targets `/wallet/transfer/signed`
- signed transfer payload now uses `from_address`, `to_address`, `amount_rtc`, `nonce`, `signature`, and `public_key`
- `transferHistory()` now accepts `{ minerId }` or `{ address }` and sends `miner_id` / `address`, instead of the rejected `wallet` query parameter
- README and tests updated to document/assert the live node API contracts

Validation:
- `node --test` -> 10 passed

PR #6746, opened 2026-06-01
test(rip-202): B1 integration tests against real derive_verified_device + HARDWARE_WEIGHTS

Scottcjn:
## BCOS Checklist

- [x] Tier label: **BCOS-L1** (test-only; touches no production code or consensus path)
- [x] SPDX header present (`# SPDX-License-Identifier: MIT`)
- [x] Test evidence below

## What Changed

Adds `node/tests/test_rip0202_enrollment_integration.py` — integration tests for the RIP-202 **B1** deterministic enrollment module (`node/rip0202_enrollment.py`, merged dormant in #6692).

The existing B1 unit tests (`test_rip0202_enrollment.py`) drive the module with **injected mock** `derive_fn` / `weight_table`. This closes the gap the module's own docstring flags:

> "wire the REAL `derive_verified_device` + `HARDWARE_WEIGHTS` and add integration tests over real committed shapes"

`derive_verified_device` lives in the ~10k-line Flask node file and can't be imported directly (module-level app/DB/thread side effects). The test **AST-extracts only `derive_verified_device` + `HARDWARE_WEIGHTS` and their transitive top-level dependency closure** into an isolated namespace and execs it — no Flask app, no DB, no network. It then drives the real B1 pipeline with the real anti-VM policy.

**Key property proved:** `derive_block_enrollment(atts, REAL_derive, REAL_weights)` equals an independent direct application of the same real functions, for arbitrary device fixtures — so the wiring is correct **without hardcoding fragile classification expectations**. Also covers the anti-VM invariant (failed fingerprint → excluded regardless of device), snapshot-hash determinism + order-independence, and deterministic duplicate-miner resolution.

Scope per operator decision D4 = **(a) attestation-level anti-VM** (eligibility, not reward magnitude). **No production code touched.**

## Testing / Evidence

```
$ python3 -m pytest node/tests/test_rip0202_enrollment_integration.py -q
......                                                                   [100%]
6 passed in 0.12s

$ python3 -m pytest node/tests/ -q   # full B1 suite (unit + integration)
...........................................                              [100%]
43 passed in 0.12s
```

Tests are path-robust (resolve the module + node file from `../` in the repo layout, or alongside in a flat staging dir; `RC_NODE_FILE` env override supported).

### DIFF
--- a/node/tests/test_rip0202_enrollment_integration.py
+++ b/node/tests/test_rip0202_enrollment_integration.py
@@ -1,0 +1,221 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: MIT
+"""RIP-202 B1 — INTEGRATION tests against the REAL anti-VM policy.
+
+The B1 unit tests (test_rip0202_enrollment.py) use injected mock derive_fn /
+weight_table. This suite closes the gap the module docstring flags:
+
+    "wire the REAL derive_verified_device + HARDWARE_WEIGHTS and add integration
+     tests over real committed shapes"
+
+`derive_verified_device` lives in the giant Flask node file and can't be
+imported directly (module-level app/DB/thread side effects). We instead extract
+ONLY `derive_verified_device` + `HARDWARE_WEIGHTS` and their transitive
+top-level dependency closure via AST, exec them in an isolated namespace, and
+drive the real B1 pipeline with them. No Flask app, no DB, no network.
+
+Key property proved: derive_block_enrollment(atts, REAL_derive, REAL_weights)
+== an independent direct application of the same real functions — for arbitrary
+device fixtures — so the wiring is correct without hardcoding classifications.
+Plus the anti-VM invariants (failed fingerprint excluded) and snapshot
+determinism / order-independence.
+"""
+import ast
+import io
+import os
+import sys
+import contextlib
+
+# Path-robust: works in repo node/tests/ (module + node file live in ../) and in
+# a flat staging dir (everything alongside this file).
+_HERE = os.path.dirname(os.path.abspath(__file__))
+for _d in (os.path.dirname(_HERE), _HERE):  # ../ first (repo layout), then .
+    if os.path.exists(os.path.join(_d, "rip0202_enrollment.py")):
+        sys.path.insert(0, _d)
+        break
+import rip0202_enrollment as b1  # noqa: E402
+
+
+def _find_node_file():
+    if os.environ.get("RC_NODE_FILE"):
+        return os.environ["RC_NODE_FILE"]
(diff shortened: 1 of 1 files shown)


FakerHideInBush (changes requested):
Reviewed current head `cefd86570eb3b62dbe366a0c2978fa07ed09530f`.

I found one portability blocker in the new integration test:

- `node/tests/test_rip0202_enrollment_integration.py:120` reads the integrated node source with `open(NODE_FILE).read()` and no explicit encoding. On Windows, Python defaults to the active ANSI codepage (`cp1252` in my environment), and collection fails before any test runs because `rustchain_v2_integrated_v2.2.1_rip200.py` contains bytes that are valid UTF-8 but not decodable as cp1252:

```text
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 348384: character maps to <undefined>
```

That makes the new test file non-portable even though the suite itself is intended to be path-robust. Please open the node file with an explicit encoding, for example:

```python
with open(NODE_FILE, encoding=utf-8) as f:
    REAL_DERIVE, REAL_WEIGHTS = _extract_real_funcs(f.read())
```

Validation I ran locally:
- `python -m py_compile node\tests\test_rip0202_enrollment_integration.py node\rip0202_enrollment.py` -> passed
- `git diff --check origin/main...HEAD -- node/tests/test_rip0202_enrollment_integration.py` -> passed
- `python -m pytest node/tests/test_rip0202_enrollment_integration.py -q` -> fails during collection with the `UnicodeDecodeError` above
- With UTF-8 mode forced (`PYTHONUTF8=1`), `python -m pytest node/tests/test_rip0202_enrollment_integration.py -q` -> 6 passed
- With UTF-8 mode forced, `python -m pytest node/tests/test_rip0202_enrollment.py node/tests/test_rip0202_enrollment_integration.py -q` -> 43 passed

So the harness logic looks sound once the file is decoded as UTF-8; the requested change is just to make that assumption explicit in the test file.

Disclosure: submitting this review for the RustChain code review bounty program (#73); no payment is asserted unless/until maintainers accept it.

MolhamHamwi (commented):
I reviewed the RIP-202 enrollment integration test coverage.

Two technical observations:

1. The AST extraction approach is a good compromise for this codebase: it exercises the real `derive_verified_device` and `HARDWARE_WEIGHTS` objects while avoiding the Flask/DB/thread side effects from importing the integrated node module. The test also resolves transitive top-level helper dependencies, so it is less brittle than copying a hand-written stub.

2. The duplicate-miner test checks a consensus-relevant edge case: two attestations for the same miner resolve by deterministic total ordering rather than input order. That protects the enrollment snapshot hash from forking if peers receive the same attestations in a different sequence.

PR #6762, opened 2026-06-01
feat(rip-202): PREREQ-A governance_params + get_param (read-only, dormant)

Scottcjn:
## BCOS Checklist

- [x] Tier label: **BCOS-L1** (additive table + pure read/write helpers; no consensus path touched)
- [x] SPDX header on both new files
- [x] Test evidence below

## What Changed

First slice of **RIP-202 PREREQ-A** (deterministic governance-parameter substrate). Two new files, **no production code wired**:

- `node/rip0202_governance_params.py`
- `node/tests/test_rip0202_governance_params.py` — 10 tests

### Why
RIP-202 step A must read the activation height as **chain-replicated state**, not a binary constant (INV-4 — a skewed binary rollout forks at the boundary). Today there is **no `governance_params` table, no param execution, and no `get_param`** — a "passed" proposal changes nothing programmatic. This adds the table + a typed `get_param` over a **built-in default registry**, so step A has a real substrate to wire `get_param("rip0202_activation_epoch")` against.

### Scope (deliberately the read-only slice)
- `ensure_governance_params_schema`, `get_param`, `set_param` (admin/migration seed), `registered_params`.
- Registry seeds `rip0202_activation_epoch` (int, default `None` = not activated) and `rip0202_eligibility_threshold_units` (int, default 1 — B1's D5 param).
- **Does NOT** add vote-driven execution (A1/A2) — that's the heavy follow-on requiring governance actions committed into blocks. This slice is purely additive: nothing existing reads or writes the table.

### Determinism
- The default registry ships in the binary → an **unset** param resolves to one fleet-wide value (never node-local divergence).
- Only **registered** names are addressable (unknown name → `GovernanceParamError`, fail closed) — no typo'd/arbitrary params can enter a consensus read.
- Values stored as canonical strings, coerced to the declared type on read; missing table → default (bootstrap-safe).
- (Full determinism of *stored* values arrives with vote-driven execution from committed blocks; until then operators seed identically.)

## Testing / Evidence

```
$ python3 -m pytest node/tests/test_rip0202_governance_params.py -q
..........                                                               [100%]
10 passed
```

Covers: schema idempotency, unset→default, default-when-table-missing, set→get typed, stored-string coercion, unknown-name fail-closed (get + set), set type/bool/None enforcement, bad set_at_epoch, overwrite-latest-wins, registry-introspection-is-a-copy.


MolhamHamwi (commented):
Reviewed `node/rip0202_governance_params.py` and `node/tests/test_rip0202_governance_params.py` for the dormant governance-parameter substrate.

Two specific observations:
- `get_param()` now only treats a genuinely missing `governance_params` table as bootstrap-safe default state, while re-raising other `sqlite3.OperationalError` cases such as locks or corruption. That is important for consensus safety because a transient DB failure should not silently fall back to `None` and make one node see RIP-202 as inactive.
- The min-bound checks are applied on both `set_param()` and `_coerce()` during reads. That covers both normal admin/migration writes and corrupted/manual rows, so values like `rip0202_eligibility_threshold_units=0` cannot slip into a consensus read and admit zero-weight eligibility.

Scottcjn:
Updated per tri-brain loop-4 (Codex architecture finding): `governance_params` is now **history-keyed** (`PRIMARY KEY (name, set_at_epoch)`, append-only) with a new `get_param_as_of(name, epoch)` that returns the value **effective at a given chain epoch** (greatest `set_at_epoch <= epoch`, else default). This makes param reads replay/reorg-correct — a node validating history recovers the exact value in force at a prior epoch, and future-dated changes apply at their epoch, not immediately. `get_param` keeps returning the latest for the dormant/admin path. 16 tests green (added prior-epoch recovery + same-epoch re-seed + history-retained).

Scottcjn:
## ✅ Tri-brain reviewed → hardened → merging

Full loop (Codex + Grok; Brain-3 timed out this run). Grok: purely additive, dormant, isolated — no live callers. Codex found 2 real BLOCKING issues in the substrate, both now fixed in this PR:

1. **Governance-history substitution** — `set_param` used `INSERT OR REPLACE`, silently rewriting an existing `(name, set_at_epoch)` row's value/proposal_id. Replaced with **idempotent-or-fail-closed**: identical re-seed is a no-op, a conflicting one raises. History is append-only now (closes a replay/audit substitution path).
2. **`get_param` not consensus-safe** — it returns the greatest `set_at_epoch` including future-dated rows. Documented loudly as non-consensus; consensus reads must use the epoch-aware `get_param_as_of`. Module + wiring guidance updated to point at `get_param_as_of`.

Tests split into `_is_idempotent` + `_conflict_fails_closed`; 17/17 pass locally. (Red `test` check is the pre-existing main `bridge_fee_i64` collection error, unrelated.)

PR #6766, opened 2026-06-01
fix(rip-202): B0 block-format hardening (tri-brain review follow-up to #6761)

Scottcjn:
## BCOS Checklist
- [x] **BCOS-L1** (pure module hardening; still unwired, no consensus path touched)
- [x] SPDX headers present
- [x] Test evidence below

## What Changed
Follow-up to **#6761** (merged) applying the tri-brain review (Codex 5.5 + Grok) findings on `node/rip0202_block_format.py`. All hardening of the **pure, dormant** B0 contract — no behavior change for valid input (the existing hash is byte-identical), stricter rejection of malformed input:

- **JSON-safety (Codex+Grok SHOULD-FIX):** `build_b0_attestation` now rejects non-JSON-safe / non-round-trip-stable content — non-string mapping keys (`sort_keys` would raise; `1` vs `"1"` ambiguous) and tuples/sets/bytes (a tuple becomes a list on reload → would diverge the hash across a commit→deserialise round-trip). Was finite-float-only.
- **Hash validates every record (Codex SHOULD-FIX):** `canonical_b0_attestations_hash` re-validates each record through `build_b0_attestation` (raises on malformed/missing) so an incomplete record can't enter the consensus hash as `null`-filled or crash serialisation mid-block.
- **`blocks_per_epoch` guard (Grok BLOCKING):** added `assert_blocks_per_epoch(node_value)` — backs the module-header "wiring asserts equality" contract with a real fail-fast guard so the pinned `144` can't silently become a second source of truth. `slot_to_epoch` now also rejects `bool`/`float` `blocks_per_epoch`.

`block_epoch`'s fail-closed-on-legacy-header behavior is intentional (RIP no-wall-clock invariant; version-gated by the caller) — kept, docstring already states the contract.

## Testing / Evidence
```
$ python3 -m pytest node/tests/test_rip0202_block_format.py -q
.........................................                                [100%]
41 passed
```
(25 original + 16 new fix-coverage tests: non-str-key rejection, non-JSON-safe-type rejection, hash-rejects-malformed, assert_blocks_per_epoch match/mismatch, bad blocks_per_epoch.)

### DIFF
--- a/node/rip0202_block_format.py
+++ b/node/rip0202_block_format.py
@@ -54,12 +54,18 @@ BLOCKS_PER_EPOCH = 144
 # a field is a consensus change gated by activation.
 B0_ATTESTATION_FIELDS = ("miner", "device", "fingerprint", "fingerprint_passed", "timestamp")
 
+# Resource bounds (defense-in-depth atop the deployed nginx/#6529 1 MB body cap):
+# a single attestation's device/fingerprint cannot bloat storage or recurring
+# blocks. Generous enough for real fingerprint timing data, tight enough to cap abuse.
+MAX_EVIDENCE_FIELD_BYTES = 65_536   # 64 KB canonical JSON, per device and per fingerprint
+MAX_EVIDENCE_DEPTH = 32             # max nesting depth in device/fingerprint
+
 
 class B0FormatError(ValueError):
     """Raised when an attestation cannot be canonicalised into a B0 record."""
 
 
-def _assert_canonical_safe(obj: Any, path: str = "") -> None:
+def _assert_canonical_safe(obj: Any, path: str = "", depth: int = 0) -> None:
     """Recursively reject anything that is not canonical-JSON-safe + round-trip
     stable, so a record can never explode at hash/serialise time two hops away
     or mutate on a deserialise round-trip (which would diverge the consensus hash):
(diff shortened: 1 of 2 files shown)


MolhamHamwi (commented):
Reviewed `node/rip0202_block_format.py` and `node/tests/test_rip0202_block_format.py` for the B0 block-format hardening.

Technical observations:
- `node/rip0202_block_format.py` lines 90-103 are a useful validate-then-serialize hardening step: requiring concrete `dict` containers and rejecting tuples/sets/bytes means `_canonical_bytes(..., sort_keys=True, allow_nan=False)` should not encounter a delayed `TypeError` or a tuple-to-list round-trip mutation later in hashing.
- `node/rip0202_block_format.py` lines 172-186 now revalidates every hash input and rejects duplicate miners before sorting, which closes the prior ambiguity where malformed records or same-miner records could still reach the canonical hash path.
- One minor follow-up: the `canonical_b0_attestations_hash` docstring still says the total order is `(miner, timestamp, content-digest)` and is stable for same-`(miner,timestamp)` records, but same-miner records are now rejected. Updating that docstring would avoid future callers/tests assuming duplicate miner tie-breaks are still supported.

Validation run: `python3 -m pytest node/tests/test_rip0202_block_format.py -q` → 39 passed.

Scottcjn:
## ✅ Tri-brain reviewed → hardened → merging

Full loop (Codex + Grok; Brain-3 timed out). Grok: dormant substrate, regression surface limited to new error paths, no committed-block/wire/schema change. **Verified: `rip0202_block_format` has zero live callers on main** — so Codex's [BLOCKING] (stricter rules without an activation gate) doesn't bite now; B0's chain-wide activation is the gate. Noted in the commit for the wiring PR.

Codex's 2 SHOULD-FIX items fixed:
1. **Exact-type canonical check** — `_assert_canonical_safe` used `isinstance(obj, dict/list)`, which accepts subclasses despite the docstring claiming concrete types. A subclass could pass validation then override `__deepcopy__`/`__iter__` to yield unvalidated data (validate-then-substitute). Now `type(obj) is dict/list`. Added a subclass-rejection test.
2. **Snapshot-then-revalidate** — `build_b0_attestation` now validates → deepcopies → **re-validates the snapshot**, so the bytes hashed are exactly the bytes validated (closes the TOCTOU). Kept the first validation before deepcopy so non-canonical nested types (e.g. `mappingproxy`) reject cleanly as `B0FormatError` rather than an opaque deepcopy `TypeError`.

`assert_blocks_per_epoch()` is intentionally inert here (wired at activation). 42/42 tests pass. Red `test` check = pre-existing `bridge_fee_i64`, unrelated.

PR #6760, opened 2026-06-01
docs: add Mine Your Grandma's Computer vintage miner guide

Ishant5436:
Resolves #2150.

Added a 15-minute quickstart guide specifically geared toward onboarding real users with vintage hardware (Core 2 Duo Windows laptops and PowerPC Macs).

**Addresses Bounty Requirements:**
- Target at least 2 architectures: Covers Old Windows laptop (Core 2 Duo era) and PowerPC Mac (G3/G4).
- "Does my computer qualify?" quick check: Included explicitly in the first section.
- Show fingerprint check passing on vintage hardware: Provided walkthroughs using `--dry-run` with screenshot descriptions of expected output.
- Show first attestation succeeding: Covered in step 4 with expected output.
- Explain antiquity multiplier in plain English: Summarized the concept, showing why old hardware is valuable.

Requested payout: 50 RTC to canonical account github:Ishant5436.

PR #6757, opened 2026-06-01
fix: align Sophia epoch settlement guard

yyswhsccc:
## BCOS Checklist (Required For Non-Doc PRs)

- [x] Add a tier label: `BCOS-L2`
- [x] If adding new code files, include SPDX header near the top (no new code files)
- [x] Provide test evidence

## What Changed

- Updated `node/sophia_elya_service.py` so its `epoch_state` schema includes the shared `settled` and `settled_ts` columns used by the other settlement paths.
- Added an idempotent migration that backfills legacy `finalized=1` Sophia epochs as `settled=1`.
- Made Sophia `finalize_epoch()` acquire `BEGIN IMMEDIATE` and claim `settled=1` before crediting balances, so an epoch already settled by another path is not credited again.
- Added focused regression coverage in `node/tests/test_sophia_elya_service_money_units.py` for legacy schema migration, second-settlement blocking, and respecting an existing settled marker.

## Why It Matters

This addresses the bounded Sophia service slice of #6749. The main integrated settlement path and RIP-200/anti-double-mining paths already use `epoch_state.settled`; Sophia still had a divergent schema and only checked `finalized`, which could bypass the shared settlement marker in mixed deployments.

## Validation

- `.venv-bounty-validation/bin/python -m pytest -q node/tests/test_sophia_elya_service_money_units.py node/tests/test_sophia_elya_service.py --tb=short` -> `15 passed`
- `.venv-bounty-validation/bin/python -m py_compile node/sophia_elya_service.py node/tests/test_sophia_elya_service_money_units.py` -> passed
- `git diff --check origin/main...HEAD` -> passed
- hidden/bidirectional Unicode scan for `origin/main...HEAD` -> passed, 2 changed files scanned, no findings

## Touched Files / Subsystems

- `node/sophia_elya_service.py`: Sophia epoch-state schema migration and atomic settlement claim.
- `node/tests/test_sophia_elya_service_money_units.py`: focused regression tests for the Sophia settlement guard.

## Scope / Risk Boundary

- BCOS-L2
- This PR only aligns Sophia's local settlement state with the existing shared `epoch_state.settled` guard.
- It does not redesign the canonical settlement entrypoint, alter RIP-200 reward math, change UTXO reward visibility, or modify the integrated node settlement implementation.

Refs #6749 and #305.

wallet: RTC47bc28896a1a4bf240d1fd780f4559b242bcd945

Issue #6749, opened 2026-06-01
[consensus] Unify epoch settlement across paths (cross-path double-credit + schema drift)


Scottcjn:
Spun out of PR #(epoch-settlement-atomic). Tri-brain (3 loops) surfaced that epoch settlement has **multiple independent writers with no shared concurrency control**:

- `finalize_epoch` (node) — writes `balances` + `epoch_state.settled` (now atomic vs itself after the PR)
- `settle_epoch_rip200` (rewards_implementation_rip200.py) — writes `epoch_rewards` + `epoch_state.settled` (INSERT OR REPLACE)
- `settle_epoch_with_anti_double_mining` (anti_double_mining.py) — same pattern
- `sophia_elya_service.py` has its **own** `epoch_state` CREATE **without** `settled`/`settled_ts`

### Risks
1. **Cross-path race**: two settlement paths can settle the same epoch concurrently — no shared lock/claim spans them.
2. **INSERT OR REPLACE epoch_state** in the rewards/anti-double-mining paths clobbers `accepted_blocks`/`finalized` written by block ingest.
3. **Schema drift**: divergent `epoch_state` CREATEs across files → missing-column errors / order-dependent migrations in mixed/standalone deployments.
4. **Two visibility models**: reward APIs read `epoch_rewards`; finalize-path settlements only appear in `epoch_state`/`balances`.

### Proposed direction
- One canonical settlement entrypoint (or a shared `claim_epoch(con, epoch)` helper using `BEGIN IMMEDIATE` that **all** paths call first).
- Single source of truth for `epoch_state` schema (shared init) + remove the divergent CREATE.
- Reconcile `epoch_rewards` vs `epoch_state` (dual-write or pick one).

CHAOTIC-tier; needs deliberate design + a migration plan. Good bounty candidate.

yyswhsccc:
Claiming a focused bug fix before opening PR.

Bug: Sophia epoch settlement uses a divergent epoch_state schema and finalized-only guard, so it can miss the shared settled marker used by the other settlement paths.

Scope: align only node/sophia_elya_service.py with the existing settled/settled_ts claim guard and add focused Sophia regression tests; no canonical settlement redesign or reward math changes.

wallet: RTC47bc28896a1a4bf240d1fd780f4559b242bcd945

Scottcjn:
[...]

PR #6769, opened 2026-06-01
security(consensus): atomic Sophia epoch settlement (synthesis of #6757 + #6760)

Scottcjn:
## Atomic Sophia epoch settlement — synthesis of #6757 + #6760

@yyswhsccc (#6757) and @Ishant5436 (#6760) independently submitted the *same* `sophia_elya_service.py` settlement fix. This combines the best of both — **yyswhsccc's fix + settlement test suite** and **Ishant5436's doc (`MINE_YOUR_GRANDMAS_COMPUTER.md`)** — into one PR, with the tri-brain BLOCKINGs fixed.

### What it does
- **`finalize_epoch`**: `BEGIN IMMEDIATE` + atomic claim (`settled=1, finalized=1 WHERE COALESCE(settled,0)=0`, rowcount-checked) — closes the double-settlement window on the Sophia path. Post-claim payout in `try/except → rollback` (no half-paid epoch). `PRAGMA busy_timeout=5000` so concurrent block-ingest waits instead of erroring "database is locked".
- **Migration**: adds `settled`/`settled_ts`, normalizes `NULL→0` (so a legacy/shared row can't become permanently unpayable), and backfills finalized→settled **once** (only when the column is first added — never on later startups, which could suppress a two-phase settle). `ALTER` wrapped against concurrent-startup races.
- **`inc_epoch_block`**: guard + busy_timeout so a late block can't inflate the count the reward was computed against.
- **`get_epoch`** now exposes `settled`/`settled_ts` (additive).

### Tests (9, all green)
incl. idempotent-pays-once, NULL-`settled`-payable, inc-after-finalize-frozen.

### Tri-brain (Codex + Grok; GPT-OSS offline) — 3 loops
Fixed: NULL consistency, busy_timeout, tx rollback, one-time backfill, ALTER TOCTOU.

### ⚠️ Scope
This fixes the **Sophia path**. Cross-module `epoch_state` coordination (rewards_implementation / anti_double_mining schema + the rolling-deploy mixed-version window) is the broader **#6749** work — deploy all settlement instances together. Complements the main-node fix in #6748.

Co-authored-by: @yyswhsccc · @Ishant5436. Supersedes #6757 and #6760.


Scottcjn:
## ✅ Tri-brain review → MERGE (CHAOTIC-regime gate cleared)

Ran the recurrent-depth dev loop (Brain 2 Codex 5.5 + Brain 4 Grok; **Brain 3 GPT-OSS down — loop ran degraded/dual-brain**). Both brains raised BLOCKING findings; Brain 1 (Coda) resolved them against the **repo**, which the sandboxed brains could not read (Grok: `node/ does not exist, 3 calls exhausted`; Codex worked from diff text only).

**Codex BLOCKING — migration assumes `finalized ⟹ paid`:** ✅ resolved. The main node (`rustchain_v2_integrated_v2.2.1_rip200.py`) **never writes `epoch_state.finalized`** (grep empty) — its only settlement flag is `settled` (line 3685, identical `WHERE settled=0` atomic claim). So `finalized` is **Sophia-exclusive**, and old Sophia set `finalized=1` *after* crediting → `finalized=1 ⟹ already-paid`. No writer can produce a legit `finalized=1, settled=0` needing payment. Backfill is sound.

**Grok BLOCKING — `settled` column may not exist (crash on read/write):** ✅ pre-existing main-node defect, **mitigated** by this PR. Main node references `settled` (3533/3685) but its `CREATE TABLE` (1425) omits it and it has no ALTER — a latent fresh-DB crash. This PR's `_ensure_epoch_state_settlement_schema` ALTER **adds** the column the main node needs.

**Cross-path coordination is correct:** main node and Sophia share the `settled` flag; whoever wins `WHERE settled=0` settles+pays, the other gets `already_settled` → **no double-pay**.

**Residual (money-safe, cosmetic):** on the live DB the column already exists, so the one-time backfill is skipped; legacy `finalized=1` rows stay `settled=0`. Those were already paid by old Sophia, so finalize correctly returns `already_finalized` and pays nobody — only the reason string differs. No funds at risk.

**CI:** the red `test` check is **pre-existing on main** (collection error `no such column: bridge_fee_i64` in `test_c11_bridge_confirm_unbounded_poc.py`, unrelated). This PR's 9 money-unit tests pass locally.

**Follow-ups filed under the #6749 cross-path-settlement design track:** (1) main-node `CREATE TABLE epoch_state` should include `settled/settled_ts`; (2) fix the `bridge_fee_i64` collection error so the `test` gate goes green.

PR #6774, opened 2026-06-01
docs: add French (fr-FR) and German (de-DE) translations

JesusMP22:
## Documentation Translations

Added README translations for:

- **French (fr-FR)** — docs/fr-FR/README.md
- **German (de-DE)** — docs/de-DE/README.md

Following the same format as existing translations.

Wallet: jesusmp


install-miner-freebsd.sh
```diff
@@ -0,0 +1,152 @@
+#!/bin/bash
```
huhaha120, line 1: Blocking for FreeBSD users: `/bin/bash` is not available in a base FreeBSD install, and when bash is installed from packages it is normally `/usr/local/bin/bash`. The documented install commands also pipe into `bash`, so a clean FreeBSD host can fail before any dependency installation runs. Either make this script POSIX `/bin/sh` compatible, or document/bootstrap `pkg install bash` before users invoke it.

install-miner-freebsd.sh
```diff
@@ -0,0 +1,152 @@
...
+# Create directories
+mkdir -p /opt/rustchain
+mkdir -p /var/log/rustchain
+chown rustchain:rustchain /opt/rustchain
+chown rustchain:rustchain /var/log/rustchain
+
+# Download miner client
+echo "[3/5] Downloading miner client..."
+cd /opt/rustchain
+curl -fsSL "https://raw.githubusercontent.com/Scottcjn/Rustchain/main/miners/rustchain_miner.py" -o rustchain_miner.py
```
huhaha120, line 88: Blocking: this URL currently points at `miners/rustchain_miner.py`, which does not exist on `main` and is not added by this PR. Because the script runs with `set -euo pipefail`, `curl -fsSL ... -o rustchain_miner.py` exits here and the installer never creates or starts the rc.d service. Please point this at an existing FreeBSD-compatible miner artifact, or add the artifact and any required companion files such as `fingerprint_checks.py`/checksums.

install-miner-freebsd.sh
```diff
@@ -0,0 +1,152 @@
...
+rcvar="rustchain_miner_enable"
+
+load_rc_config $name
+
+: ${rustchain_miner_enable:="NO"}
+: ${rustchain_miner_flags:=""}
+
+pidfile="/var/run/rustchain_miner.pid"
+command="/usr/sbin/daemon"
+command_args="-f -p ${pidfile} -u rustchain /usr/local/bin/python3 /opt/rustchain/rustchain_miner.py"
```
huhaha120, line 124: Blocking: the installer writes `/opt/rustchain/config.env`, but the generated rc.d script never sources it and this command does not pass `--wallet`. The current miner CLI accepts `--wallet` and otherwise falls back to generating its own wallet, so `service rustchain_miner start` will not use the `${WALLET_NAME}` supplied to the installer. Please load/export the config in the rc.d script or pass the wallet explicitly in `command_args`.

miners/freebsd/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Build from Source
+
+```bash
+# Install dependencies
+pkg install python3 py311-pip
+pip3 install requests
+
+# Run miner directly
+python3 rustchain_miner.py --node https://rustchain.org --wallet your-wallet-name
```
huhaha120, line 62: Docs mismatch: the current miner parser in `miners/linux/rustchain_linux_miner.py` exposes `--wallet`, dry-run/show-payload options, and Warthog options, but no `--node` flag. This command will fail with an unrecognized argument unless this PR also adds a FreeBSD miner implementation that supports `--node`.

huhaha120 (changes requested):
Thanks for the contribution. I am requesting changes because the FreeBSD installer path currently cannot complete a clean install, and the service/docs do not line up with the miner CLI that exists in this repository.

I found three blocking issues:

- the installer downloads a miner path that is not present on `main`, so `curl -fsSL` exits under `set -e` before installation completes;
- the rc.d service does not pass or load the wallet configured during install, so the service will not mine to the requested wallet;
- the FreeBSD docs show a `--node` CLI flag that the current miner parser does not support.

These should be fixed before merging so users do not follow a one-line installer that fails or starts mining with the wrong configuration.

jaxint (commented):
Great contribution! 🔍 Reviewed and looks solid.

jaxint (approved):

jaxint (commented):
Solid PR! The changes are well-thought-out and the code quality is high.

Scottcjn:
Closing as superseded by **#6776**, which is the superset of this translation batch (same languages plus more). Consolidating to one PR keeps review clean — thanks @JesusMP22.

PR #6778, opened 2026-06-02
fix(verifier): use public wallet balance endpoint

FakerHideInBush:
## Summary
- update the bounty verifier star-checker wallet probe from stale `/api/balance/<wallet>` to the maintained public `/wallet/balance?miner_id=<wallet>` endpoint
- pass the wallet as a `miner_id` query parameter so requests handles encoding safely
- require the balance response to be a JSON object instead of accepting any HTTP 200 from the old path
- add regression coverage for the endpoint/params/cert path and malformed balance payloads

## Validation
- `python -m pytest -q tests\test_star_checker.py --tb=short` -> 11 passed
- `python -m py_compile tools\bounty_verifier\star_checker.py tests\test_star_checker.py`
- `git diff --check -- tools\bounty_verifier\star_checker.py tests\test_star_checker.py`

## Bounty / payout
General bug bounty: Scottcjn/Rustchain#305

Wallet: `RTCe0961d6b54f2fa96db57a373c84d8ad8986153f8`

### DIFF
--- a/tests/test_star_checker.py
+++ b/tests/test_star_checker.py
@@ -153,24 +153,34 @@ def test_count_user_stars_uses_supplied_repos_without_fetching_owner_repos(monke
     assert star_checker.count_user_stars("targetuser", "owner", "secret", repos=["alpha", "beta"]) == 1
 
 
-def test_check_wallet_exists_uses_local_cert_when_present(monkeypatch, tmp_path):
+def test_check_wallet_exists_uses_public_wallet_balance_endpoint(monkeypatch, tmp_path):
     cert = tmp_path / ".rustchain" / "node_cert.pem"
     cert.parent.mkdir()
     cert.write_text("certificate", encoding="utf-8")
     seen = {}
 
-    def fake_get(url, verify, timeout):
+    def fake_get(url, params, verify, timeout):
         seen["url"] = url
+        seen["params"] = params
         seen["verify"] = verify
         seen["timeout"] = timeout
-        return FakeResponse(status_code=200)
+        return FakeResponse(status_code=200, payload={"amount_rtc": 0.0, "miner_id": "rtc-wallet"})
 
     monkeypatch.setenv("HOME", str(tmp_path))
+    monkeypatch.setenv("USERPROFILE", str(tmp_path))
     monkeypatch.setattr(star_checker.requests, "get", fake_get)
 
     assert star_checker.check_wallet_exists("rtc-wallet") is True
-    assert seen == {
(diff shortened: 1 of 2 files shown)


tools/bounty_verifier/star_checker.py
```diff
@@ -132,13 +132,22 @@ def count_user_stars(
...
     try:
-        url = f"{RUSTCHAIN_NODE_URL}/api/balance/{wallet_address}"
+        url = f"{RUSTCHAIN_NODE_URL}/wallet/balance"
         import os
         _cert = os.path.expanduser("~/.rustchain/node_cert.pem")
         _verify = _cert if os.path.exists(_cert) else True
-        resp = requests.get(url, verify=_verify, timeout=10)
+        resp = requests.get(
+            url,
+            params={"miner_id": wallet_address},
```
JONASXZB, line 141: Good endpoint shape: passing the wallet as `miner_id` query params keeps the configured node host intact and avoids the stale path-style balance endpoint.

tests/test_star_checker.py
```diff
@@ -153,24 +153,34 @@ def fail_if_fetching_repos(*args, **kwargs):
...
-        "verify": str(cert),
-        "timeout": 10,
-    }
+    assert seen["url"] == f"{star_checker.RUSTCHAIN_NODE_URL}/wallet/balance"
+    assert seen["params"] == {"miner_id": "rtc-wallet"}
+    assert Path(seen["verify"]) == cert
+    assert seen["timeout"] == 10
+
+
+def test_check_wallet_exists_rejects_non_object_balance_payload(monkeypatch):
```
JONASXZB, line 180: This negative case is useful because a 200 response with a non-object JSON payload should not be treated as an existing wallet.

JONASXZB (approved):
Reviewed head `83b7c077`. This is the right fix direction for the stale wallet endpoint: it updates the actual verifier module, keeps the configured RustChain node host, switches to `/wallet/balance` with `miner_id`, preserves the local cert behavior, and adds focused tests for the endpoint shape plus non-object JSON payloads.

Local verification I ran from a detached worktree:
- standalone import/fake-request validation for `check_wallet_exists()` passed and confirmed URL `/wallet/balance`, params `{'miner_id': 'rtc-wallet'}`, cert verify path, and timeout `10`
- `python3 -m py_compile tools/bounty_verifier/star_checker.py tests/test_star_checker.py`

I also tried the focused pytest file, but this local environment is missing repo-wide `flask` dependency imported by `tests/conftest.py`, so pytest stopped before collecting these tests. No blockers found in the reviewed diff.

jaxint (commented):
Great contribution! 🔍 Reviewed and looks solid.

PR #6797, opened 2026-06-02
fix(gpu): atomic escrow transitions — close release/refund TOCTOU race

Scottcjn:
## BCOS Checklist
- [x] **BCOS-L1**
- [x] Test evidence below

## What Changed
From a tri-brain (Codex) audit of `node/gpu_render_protocol.py`. The genuine, verified bug:

**TOCTOU race in release/refund** — both read `status=='locked'` then `UPDATE ... WHERE job_id=?` *without* re-checking status. Two concurrent ops could both pass the check and both write (a `released` escrow flipped to `refunded`). Fixed: `UPDATE ... WHERE job_id=? AND status='locked'` + `rowcount==1` guard; the loser returns `"escrow no longer locked (concurrent release/refund)"`. Adds a deterministic race regression test.

## Audit context (transparency)
- **False positive corrected:** the audit flagged "forgeable authorization," but `_authorize_escrow_action` verifies the `escrow_secret` hash (`hmac.compare_digest`) + participant + `required_wallet`. Auth is **not** forgeable. (Adversarial verification of the audit, not just relaying it.)
- **Separate, NOT in this PR (needs design):** the module is **dormant** (`register_routes` is only called by tests, never the node app) and the escrow **moves no actual RTC** (status ledger only). Real fund custody via the node's signed `/wallet/transfer`, integer micro-unit money, and real device attestation are **pre-activation requirements** — tracked in memory (`gpu-render-escrow-review-2026-06-02`). This PR fixes the one concrete concurrency bug now.

## Testing
```
pytest tests/test_gpu_render_protocol.py -q  ->  34 passed
```

### DIFF
--- a/node/gpu_render_protocol.py
+++ b/node/gpu_render_protocol.py
@@ -336,11 +336,17 @@ class GPURenderProtocol:
                 return auth_error
 
             now = int(time.time())
-            conn.execute(
-                "UPDATE render_escrow SET status='released', released_at=? WHERE job_id=?",
+            # Atomic transition guarded on status='locked' so a concurrent
+            # release/refund cannot both win (TOCTOU between the read above and
+            # this write). rowcount==0 means we lost the race.
+            cur = conn.execute(
+                "UPDATE render_escrow SET status='released', released_at=? "
+                "WHERE job_id=? AND status='locked'",
                 (now, job_id),
             )
             conn.commit()
+            if cur.rowcount != 1:
+                return {"error": "escrow no longer locked (concurrent release/refund)"}
             return {
                 "status": "released",
                 "job_id": job_id,
@@ -372,11 +378,14 @@ class GPURenderProtocol:
                 return auth_error
 
             now = int(time.time())
-            conn.execute(
-                "UPDATE render_escrow SET status='refunded', released_at=? WHERE job_id=?",
+            cur = conn.execute(
+                "UPDATE render_escrow SET status='refunded', released_at=? "
+                "WHERE job_id=? AND status='locked'",
                 (now, job_id),
             )
             conn.commit()
+            if cur.rowcount != 1:
+                return {"error": "escrow no longer locked (concurrent release/refund)"}
             return {
                 "status": "refunded",
                 "job_id": job_id,
--- a/tests/test_gpu_render_protocol.py
+++ b/tests/test_gpu_render_protocol.py
@@ -86,6 +86,25 @@ class TestGPURenderProtocol(unittest.TestCase):
         refund = self.proto.refund_escrow(job_id, "wallet-b", result["escrow_secret"])
         self.assertEqual(refund["status"], "refunded")
 
+    def test_escrow_transition_is_atomic_under_race(self):
+        # A concurrent winner transitions the row between our read and write;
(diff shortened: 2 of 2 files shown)


FakerHideInBush (approved):
## Code Review — `fix(gpu): atomic escrow transitions` ✅ Approve

Reviewed from a clean checkout of the PR head (`e8f6bba5`). This is a correct, tightly-scoped TOCTOU fix and the regression test is well-constructed.

### What the fix does and why it's correct
`release_escrow`/`refund_escrow` previously did read-then-write: they `SELECT` the row, validate `status == 'locked'`, authorize, then unconditionally `UPDATE ... SET status=...`. Two concurrent calls (a release and a refund) could both pass the `status == 'locked'` read before either wrote, so both would "win" and the row's terminal state would depend on write order — a classic check-then-act race on the escrow state machine.

The fix makes the transition atomic by guarding the write on the expected pre-state:

```sql
UPDATE render_escrow SET status='released', released_at=? WHERE job_id=? AND status='locked'
```

and treating `cur.rowcount != 1` as "I lost the race." Since the escrow status row *is* the authoritative ledger event here (these methods don't move funds themselves — they return `amount_rtc`/wallet for the caller to act on), exactly one of the concurrent callers can flip `locked → released|refunded`; the loser matches 0 rows and returns an error. That's the right place to enforce mutual exclusion.

### End-to-end verification I did
- **Error precedes success:** the `rowcount != 1` error is returned *before* the `{"status": "released", ...}` dict, so a race-loser never receives a payout-shaped success.
- **Route wiring:** `/render/release` and `/render/refund` map the result with `status_code = 200 if "error" not in result else 400` (lines 679, 701), so the loser gets HTTP 400 — it can't be misread as a completed release/refund.
- **`rowcount` semantics:** `job_id` is the row key, so the guarded UPDATE matches at most one row; `!= 1` correctly captures the lost-race (0) case and is defensive against the unexpected.
- **Immutable fields:** the returned `amount_rtc`/`to_wallet`/`from_wallet` come from the pre-UPDATE read, which is fine — those are fixed at escrow creation; only `status` is contended.
- The pre-write `if row["status"] != "locked"` checks remain as a friendly fast-path and don't bypass the authoritative guard.

```
python -m py_compile node/gpu_render_protocol.py            -> OK
python -m pytest tests/test_gpu_render_protocol.py          -> 34 passed (incl. test_escrow_transition_is_atomic_under_race)
git diff origin/main...HEAD --check                          -> clean (no whitespace errors)
```
The new `test_escrow_transition_is_atomic_under_race` deterministically hits the window by mutating `status` inside a monkeypatched `_authorize_escrow_action` and asserting the subsequent refund fails with `no longer locked` — a clean way to exercise the TOCTOU without threads.

### Non-blocking observations (optional)
1. **Error-message style consistency:** the new failure returns `{"error": "escrow no longer locked (concurrent release/refund)"}`, while adjacent errors use title-case forms like `Job already released`. Cosmetic only.
2. **Symmetric coverage:** the test covers refund-loses-to-release; an optional mirror case (release loses to a concurrent refund) would round out the matrix, but the SQL guard is symmetric so behavior is already covered in principle.

LGTM — approving. Nice catch on the race.

PR #6799, opened 2026-06-02
fix(otc): idempotent payout + fail-closed admin transport (double-spend + key-leak)

Scottcjn:
Two hardening fixes to the OTC bridge worker payout (`rtc_transfer_from_worker`). Unlike the dormant gpu-render escrow, this module is **live and moves real RTC**, so both are active exposures.

## 1. Idempotent payout — prevents double-spend on retry

The payout retries `POST /wallet/transfer` up to 4× on timeout/5xx. Without an idempotency key, a retry after the server **already debited** (response lost to a timeout, or a double-`confirm_order`) pays the recipient **twice**.

The node's `wallet_transfer_v2` **already** dedups on `idempotency_key` (derives a deterministic `tx_hash`, returns the prior pending result on a match, 409 on a conflicting replay). The client just never sent one, so each retry produced a fresh random `tx_hash` and re-executed.

Fix: send a stable key `otc_payout:{order_id}`.
- **Unique per payout** — `order_id` is a PK (`generate_order_id`); `confirm_order` pays exactly one recipient per order.
- **Stable across retries** — derived from the immutable `order_id`, not `time.time()`.
- **Matches the server regex** `[A-Za-z0-9._:-]{1,128}`.
- **Equal to `reason`** so the node's reason-consistency check never 409s a legitimate retry.

## 2. Fail-closed admin transport — prevents RC_ADMIN_KEY leak

The admin key was sent to the env-controlled `RUSTCHAIN_NODE` with no scheme/TLS enforcement: `http://` or `RUSTCHAIN_TLS_VERIFY=false` would exfiltrate the credential in plaintext / to an unverified (MITM-able) host.

New `_admin_transport_block_reason()` refuses to send the key when the endpoint is a non-`https` scheme, or has TLS verification disabled to a non-local host. Loopback (`localhost`/`127.0.0.1`/`::1`) and an explicit `OTC_ALLOW_INSECURE_ADMIN=1` opt-out remain for dev. Refusing strands funds in `otc_bridge_worker` (alerted via `send_bridge_alert` + recoverable) — strictly safer than leaking the admin key.

## Tests — `tests/test_otc_bridge_payout_idempotency.py`
1. stable, order-derived `idempotency_key` is sent (== `reason`);
2. **a timeout-then-success resends the identical key** (the dedup contract that stops the double-spend);
3. transport guard **blocks** `http://` and TLS-off-to-nonlocal (key never sent), and **allows** loopback-http + explicit opt-out.

All 36 otc-bridge tests pass locally.

### DIFF
--- a/otc-bridge/otc_bridge.py
+++ b/otc-bridge/otc_bridge.py
@@ -30,6 +30,7 @@ import secrets
 import sqlite3
 import time
 from datetime import datetime, timezone
+from urllib.parse import urlparse
 from decimal import Decimal, InvalidOperation, ROUND_HALF_UP
 from functools import wraps
 
@@ -195,6 +196,38 @@ def is_valid_wallet_id(wallet_id):
     return bool(_MINER_ID_RE.fullmatch(wallet_id))
 
 
+def _admin_transport_block_reason():
+    """Return a reason string if it is UNSAFE to send the admin key to the
+    configured node, else None.
+
+    Fail-closed: the RC_ADMIN_KEY must never leave over plaintext (http://) or
+    to a non-local host with TLS verification disabled (MITM credential theft).
+    Loopback hosts and an explicit OTC_ALLOW_INSECURE_ADMIN opt-out are allowed
+    for local development.
+    """
+    if os.environ.get("OTC_ALLOW_INSECURE_ADMIN", "").strip().lower() in ("1", "true", "yes"):
+        return None  # explicit operator opt-out (dev only)
+
+    parsed = urlparse(RUSTCHAIN_NODE)
+    host = (parsed.hostname or "").lower()
+    if host in ("localhost", "127.0.0.1", "::1"):
+        return None  # loopback dev is acceptable
+
+    if parsed.scheme != "https":
+        return (
+            f"insecure scheme '{parsed.scheme or 'none'}' for admin endpoint "
+            f"{RUSTCHAIN_NODE!r}: set RUSTCHAIN_NODE to https:// "
+            f"(or OTC_ALLOW_INSECURE_ADMIN=1 for local dev)"
+        )
+    if TLS_VERIFY is False:
+        return (
+            "TLS verification disabled (RUSTCHAIN_TLS_VERIFY=false) for a "
+            "non-local admin endpoint — MITM credential exposure; pin "
+            "RUSTCHAIN_CA_BUNDLE instead of disabling verification"
+        )
+    return None
+
+
 def send_bridge_alert(level, message, fields):
     """Best-effort alert hook for payout failures and manual recovery events."""
     webhook = os.environ.get("RC_SOPHIACHECK_WEBHOOK", "").strip()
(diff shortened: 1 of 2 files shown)


FakerHideInBush (approved):
## Code Review — `fix(otc): idempotent payout + fail-closed admin transport` ✅ Approve

Reviewed from a clean checkout of the PR head (`f4377652`). Two independent money-path hardenings, both correct. I verified the idempotency half against the **server side** as well, since that's where the contract actually has to hold.

### 1. Payout idempotency — verified end-to-end (client + node)
`rtc_transfer_from_worker` retries `/wallet/transfer` on timeout/5xx, so a response lost to a timeout after the server already debited would pay twice. The fix adds a stable `idempotency_key = f"otc_payout:{order_id}"`. I traced the receiving end in `node/rustchain_v2_integrated_v2.2.1_rip200.py::wallet_transfer_v2` to confirm the contract is real, not cosmetic:

- The node derives `tx_hash = sha256("wallet_transfer_idempotency:{idempotency_key}")` (so an identical key ⇒ identical `tx_hash`).
- On a retry it finds the existing `pending_ledger` row by `tx_hash` and returns `{"ok": True, "phase": ..., "pending_id": ...}` ("already pending") **without** a second debit.
- There is a reason-consistency guard: if `from/to/amount/reason` differ from the stored row it returns `409 idempotency_key_conflict`. The PR sets `idempotency_key == reason == "otc_payout:{order_id}"`, so a genuine retry matches on all four fields and dedups cleanly instead of 409-ing. Nice detail — that equality is load-bearing.
- Key format check: the node enforces `[A-Za-z0-9._:-]{1,128}`. `otc_payout:<order_id>` only uses letters/digits/`_`/`:` and is well under 128 chars, so it always validates. ✅

Because `order_id` is immutable and there's one worker payout per order, the key is stable across the whole retry sequence — exactly what dedup requires.

### 2. Fail-closed admin transport — sound
`_admin_transport_block_reason()` refuses to send `RC_ADMIN_KEY` over `http://`, or to a non-local host with TLS verification disabled, before the request is built. Refusing strands the payout in the worker (alerted + recoverable) rather than exfiltrating the admin credential over a MITM-able channel — the correct safer-failure tradeoff. Loopback hosts and an explicit `OTC_ALLOW_INSECURE_ADMIN` opt-out keep local dev usable. The check runs at the top of `rtc_transfer_from_worker`, so it gates every retry, and it returns a structured `{"ok": False, "error": "insecure_admin_transport: ..."}`.

```
python -m py_compile otc-bridge/otc_bridge.py            -> OK
python -m pytest tests/test_otc_bridge_payout_idempotency.py -> 6 passed
```
(The suite needs `flask_cors`, which `otc_bridge.py` imports at module load — only relevant to the test environment, not the change.) The 6 cases pin both halves: stable order-derived key, key stability across retries, http refused, tls-verify-disabled-to-nonlocal refused, and the loopback / explicit-opt-out allow paths.

### Non-blocking observations (optional)
1. The idempotency window is bounded by how long the row stays in `pending_ledger`; if a retry arrives after the entry is pruned/settled, dedup can't help. Worth a one-line note in the worker that the guarantee is "within the pending window," but for the timeout-retry case it's covering the real risk.
2. `_admin_transport_block_reason` treats any non-loopback host as remote (correct fail-closed default); operators using a private-LAN node over plaintext will need the explicit `OTC_ALLOW_INSECURE_ADMIN` opt-out, which the error message already points to. 👍

LGTM — approving. Both are real protections on a live RTC money path.

PR #6800, opened 2026-06-02
fix(otc): harden precision-column migration — allowlist tables + idempotent ALTER

Scottcjn:
Two lower-severity defects from the otc-bridge review, in `migrate_precision_columns` / `table_columns`.

## 1. SQL injection (latent)
`table_name` was interpolated directly into `PRAGMA table_info(...)`, `ALTER TABLE ...`, and `UPDATE ...` DDL. **All current callers pass literals** (`"orders"`/`"trades"`), so there is no live vector — but SQLite **cannot parameterize identifiers**, so any future dynamic caller would be injectable.

Fix: a module `_KNOWN_TABLES` allowlist + `_require_known_table()` guard runs **before** any interpolation. Both functions now raise `ValueError` on an unknown name (e.g. `"orders; DROP TABLE orders"`).

## 2. Migration atomicity
Concurrent workers (e.g. gunicorn forks at startup) race the PRAGMA→ALTER window: the second to ALTER hits `duplicate column name` and crashes init. The `ALTER` loop now swallows **only** that specific `OperationalError` (everything else — including a genuinely missing table — still raises). The `COALESCE` backfill `UPDATE` was already idempotent, so re-runs converge and no write lock is needed.

## Tests — `tests/test_otc_bridge_migration.py`
- backfill computes correct integer micro/nano values;
- running the migration twice is a no-op (no raise);
- a column pre-added out-of-band (the race) is tolerated, remaining columns still added;
- unknown / injection-shaped table names rejected on **both** functions;
- a non-duplicate `OperationalError` (missing table) still surfaces.

42 otc-bridge tests pass locally.

### DIFF
--- a/otc-bridge/otc_bridge.py
+++ b/otc-bridge/otc_bridge.py
@@ -153,17 +153,41 @@ def money_view(row):
     return data
 
 
+# SQLite cannot parameterize identifiers (PRAGMA/ALTER/UPDATE take a literal
+# table name), so every table name interpolated into the DDL below MUST be
+# validated against this allowlist first — never against caller-supplied text.
+_KNOWN_TABLES = frozenset({"orders", "trades", "rate_limits"})
+# Integer precision columns we add (also literal, never caller-supplied).
+_PRECISION_COLUMNS = ("amount_micro_rtc", "price_per_rtc_nano_quote", "total_quote_nano")
+
+
+def _require_known_table(table_name):
+    """Guard before building DDL: refuse any table name not on the allowlist."""
+    if table_name not in _KNOWN_TABLES:
+        raise ValueError(f"refusing to build SQL for unknown table {table_name!r}")
+    return table_name
+
+
 def migrate_precision_columns(cursor, table_name):
+    # Validate before interpolation so only known-safe identifiers reach the SQL.
+    _require_known_table(table_name)
+
     columns = {row[1] for row in cursor.execute(f"PRAGMA table_info({table_name})")}
-    if "amount_micro_rtc" not in columns:
-        cursor.execute(f"ALTER TABLE {table_name} ADD COLUMN amount_micro_rtc INTEGER")
-    if "price_per_rtc_nano_quote" not in columns:
-        cursor.execute(f"ALTER TABLE {table_name} ADD COLUMN price_per_rtc_nano_quote INTEGER")
-    if "total_quote_nano" not in columns:
-        cursor.execute(f"ALTER TABLE {table_name} ADD COLUMN total_quote_nano INTEGER")
+    for col in _PRECISION_COLUMNS:
+        if col in columns:
+            continue
+        try:
+            cursor.execute(f"ALTER TABLE {table_name} ADD COLUMN {col} INTEGER")
+        except sqlite3.OperationalError as exc:
+            # Idempotent under concurrent migrations: another worker may have
+            # added the column between the PRAGMA read above and this ALTER.
+            if "duplicate column name" not in str(exc).lower():
+                raise
 
     refreshed = {row[1] for row in cursor.execute(f"PRAGMA table_info({table_name})")}
     if {"amount_rtc", "price_per_rtc", "total_quote"}.issubset(refreshed):
+        # COALESCE keeps the backfill idempotent: re-runs converge to the same
+        # values, so concurrent migrations are safe without a write lock.
         cursor.execute(f"""
             UPDATE {table_name}
             SET amount_micro_rtc = COALESCE(amount_micro_rtc, CAST(ROUND(amount_rtc * ?) AS INTEGER)),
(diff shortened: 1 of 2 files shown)


FakerHideInBush (approved):
## Code Review — `fix(otc): harden precision-column migration` ✅ Approve

Reviewed from a clean checkout of the PR head (`9bf1c23d`). Correct, well-scoped migration hardening on a money-path table.

### What I verified
- **Identifier-injection surface closed consistently.** SQLite can't parameterize identifiers, and `migrate_precision_columns`/`table_columns` interpolate `table_name` into `PRAGMA`/`ALTER`/`UPDATE`. The new `_require_known_table()` raises `ValueError` against the `_KNOWN_TABLES` allowlist *before* any interpolation, and it's applied to **both** functions — not just the migrator. Precision columns are a fixed literal tuple (`_PRECISION_COLUMNS`), never caller-supplied.
- **Reachability:** all four callers pass hardcoded literals (`"orders"`/`"trades"`), so this is defense-in-depth rather than a live exploit — but the guard is the right invariant to pin so a future caller can't regress it.
- **Idempotent ALTER:** each column is skipped if already present, and a concurrent `ALTER` is tolerated by catching only `OperationalError` containing `duplicate column name` (other operational errors still raise). Good — that's the precise narrow catch, not a blanket `except`.
- **Idempotent backfill:** the `COALESCE`-based `UPDATE` converges on re-runs, so repeated migrations don't corrupt already-populated precision columns.

```
python -m py_compile otc-bridge/otc_bridge.py          -> OK
python -m pytest tests/test_otc_bridge_migration.py    -> 6 passed
```

### Non-blocking observation
The `except` narrows on the message substring `"duplicate column name"`, which is SQLite's current wording; if the underlying driver text ever changes this would surface as a raised error (fail-loud) rather than silent corruption — acceptable, just noting the coupling.

LGTM — approving. Clean allowlist + idempotency work.

jaxint (approved):

zqleslie:
Reviewed PR #6800: harden precision-column migration — allowlist tables + idempotent ALTER.

## Review Summary

This PR closes two latent defects in `otc-bridge/otc_bridge.py`'s `migrate_precision_columns` and `table_columns` functions:

### 1. SQL Injection (latent) — ✅ Fix is correct

`table_name` was interpolated directly into `PRAGMA`, `ALTER`, and `UPDATE` DDL. SQLite cannot parameterize identifiers, so any future caller passing a non-literal would be injectable.

Fix introduces `_KNOWN_TABLES` frozenset + `_require_known_table()` guard applied to **both** `migrate_precision_columns` and `table_columns`. The allowlist (`{"orders", "trades", "rate_limits"}`) covers all current callers.

**Verified:** Unknown table names like `"orders; DROP TABLE orders"` raise `ValueError` before reaching SQL.

### 2. Migration Atomicity — ✅ Fix is correct

Concurrent gunicorn workers at startup race the `PRAGMA → ALTER` window. Previously the second worker hit `duplicate column name` and crashed.

Fix:
- ALTER loop now catches `sqlite3.OperationalError` and **only** swallows `"duplicate column name"` — everything else (missing table, genuine errors) still raises.
- COALESCE backfill was already idempotent; re-runs converge without write lock.

**Verified:** Pre-adding a column out-of-band → migration completes and adds remaining columns.

### Test Coverage — ✅ 6 tests, well-structured

| Test | Covers |
|------|--------|
| `test_migration_backfills_integer_columns` | Correct integer math from legacy float |
| `test_migration_is_idempotent_when_run_twice` | No-op on second run |
| `test_migration_tolerates_concurrent_duplicate_column` | Race condition recovery |
| `test_migration_rejects_unknown_table` | SQL injection guard |
| `test_table_columns_rejects_unknown_table` | Same guard on `table_columns` |
| `test_real_duplicate_column_error_still_raises` | Missing table still raises |

All tests pass. The `load_otc_bridge` helper uses `importlib.util` to load the module with a clean `OTC_DB_PATH` — clean pattern.

### Minor Observations (non-blocking)

1. The `_require_known_table` guard is defensive but the module-level `_KNOWN_TABLES` could drift if new tables are added without updating the frozenset. A future enhancement: extract table names from callers' `CREATE TABLE` statements or use a registry pattern.
2. The `str(exc).lower()` string match on `"duplicate column name"` is coupled to SQLite's error message wording — a SQLite version bump could change it. Consider also checking `exc.sqlite_error_code` if available (though it may not exist for all versions).

## Verdict

**APPROVED ✅** — Two real latent defects fixed, test coverage is solid, no behavior change for existing callers, idempotency proven.

Wallet: `XKO212dF8324b9b61F294D26A6Dc68e3f81e4BE451D`

JesusMP22 (commented):
Code Review PR#6800

Fixes SQL injection in migrate_precision_columns + non-idempotent ALTER.

Strengths: _KNOWN_TABLES frozenset guard is correct for SQLite DDL. IF NOT EXISTS on ALTER. Tests thorough.

Notes: Log before ValueError? _PRECISION_COLUMNS unused?

Verdict: Solid fix.

PR #6804, opened 2026-06-02
fix: update stale balance endpoint in star_checker.py (fixes #6779)

### LINKED ISSUE #6779: [Bug] Bounty verifier wallet check uses stale balance endpoint
## Summary

`tools/bounty_verifier/star_checker.py::check_wallet_exists()` still probes the stale `/api/balance/<wallet>` route. The rest of the verifier and public docs use the maintained public endpoint shape:

`/wallet/balance?miner_id=<wallet>`

This can make bounty claim verification report a wallet as missing even when the public wallet-balance endpoint is healthy.

## Reproduction

In the current source, inspect `tools/bounty_verifier/star_checker.py`:

```python
url = f{RUSTCHAIN_NODE_URL}/api/balance/{wallet_address}
```

The maintained verifier path in `tools/bounty_verifier/verifier.py` uses:

```python
/wallet/balance?miner_id=<wallet>
```

## Expected

The star-checker wallet probe should use the maintained no-auth public wallet balance endpoint and pass the wallet as `miner_id`.

## Actual

The probe calls a stale `/api/balance/<wallet>` path and treats any HTTP 200 as wallet existence without checking the balance response shape.

## Impact

Low severity operational bug. It can cause false negative wallet checks in bounty verification tooling, which is directly relevant to bounty claim triage and payout hygiene.

## Fix PR

https://github.com/Scottcjn/Rustchain/pull/6778

The PR switches to `/wallet/balance` with `params={miner_id: wallet_address}` and requires a JSON object response. It also adds regression tests.

## Validation

```text
python -m pytest -q tests\test_star_checker.py --tb=short
11 passed

python -m py_compile tools\bounty_verifier\star_checker.py tests\test_star_checker.py
git diff --check -- tools\bounty_verifier\star_checker.py tests\test_star_checker.py
```

zeroknowledge0x:
## Summary
Fix the bounty verifier wallet check to use the maintained /wallet/balance endpoint instead of the stale /api/balance path.

## Changes
- Updated tools/bounty_verifier/star_checker.py: check_wallet_exists() now uses the correct /wallet/balance?miner_id={wallet_address} endpoint, consistent with verifier.py

## Testing
- Code change only — function signature and behavior unchanged
- Aligns with maintained endpoint pattern in verifier.py

## Related Issues
Fixes #6779

### DIFF
--- a/tools/bounty_verifier/star_checker.py
+++ b/tools/bounty_verifier/star_checker.py
@@ -132,7 +132,7 @@ def count_user_stars(
 def check_wallet_exists(wallet_address: str) -> bool:
     """Verify that a wallet address exists on the RustChain node."""
     try:
-        url = f"{RUSTCHAIN_NODE_URL}/api/balance/{wallet_address}"
+        url = f"{RUSTCHAIN_NODE_URL}/wallet/balance?miner_id={wallet_address}"
         import os
         _cert = os.path.expanduser("~/.rustchain/node_cert.pem")
         _verify = _cert if os.path.exists(_cert) else True


zqleslie (approved):
## Review: `check_wallet_exists` endpoint fix

Reviewed the single-file change in `tools/bounty_verifier/star_checker.py`.

### What changed
- `check_wallet_exists()` switched from the stale `/api/balance/{wallet_address}` endpoint to the maintained `/wallet/balance?miner_id={wallet_address}` endpoint.
- This is consistent with `verifier_wallet_balance()` in the same file (verifier.py), which already uses `/wallet/balance?miner_id=` with the same self-signed cert handling.

### Positive assessment
1. **Correct endpoint alignment** — the old `/api/balance/` path was inconsistent with the verifier's maintained endpoint, so this fix eliminates a real bug where `check_wallet_exists` would fail against the current node API.
2. **Query parameter approach** — using `?miner_id=` is safer than path interpolation for wallet addresses that might contain special characters, and matches the pattern already proven in `verifier_wallet_balance`.
3. **No behavior change beyond the fix** — the function signature, return type, and cert handling remain unchanged.

### Minor concerns (non-blocking)
1. **Inconsistent naming vs parameter semantics** — the function is named `check_wallet_exists` and takes `wallet_address`, but the endpoint expects `miner_id`. If callers pass a `RTC+40-hex` wallet address (not a miner ID string), the query semantics may differ from the old path-based endpoint. This is not introduced by this PR (the old code had the same ambiguity), but worth a follow-up.
2. **`verify_balance_enum` still uses `/api/balances`** — the same file has a third wallet-check path via the enumerated endpoint. If `/api/balance/` is stale/deprecated, `/api/balances` may also need review, but that's out of scope for this fix.

### Verification
- Single-file, 1-line diff. No new dependencies, no API contract changes.
- The fix directly resolves the referenced issue #6779.
- Consistent with the endpoint pattern used in `verifier_wallet_balance`.

**Verdict: `APPROVED` — clean, targeted fix that resolves a real endpoint mismatch.**

MolhamHamwi (changes requested):
Thanks for the quick endpoint update. I don't think this fully closes #6779 yet.

Two concrete gaps:

1. The issue's fix/validation asks for the maintained endpoint shape plus a response-shape check. This PR changes the URL string, but it still treats any HTTP 200 as wallet existence. A proxy/error page or unexpected non-wallet JSON would still return true, so the false-positive/false-negative class called out in #6779 is only partially addressed.
2. The implementation now embeds the query string manually. The existing fix direction in #6779 mentions using `params={"miner_id": wallet_address}`; that matters for unusual wallet strings and keeps parity with the verifier path. There is also no regression test update here, while the issue lists `tests/test_star_checker.py` validation for this exact behavior.

I would either add the response validation + regression test here, or close this as superseded by #6778 if that PR is already the intended complete fix.

JesusMP22 (commented):
## Code Review for PR #6804

### Summary
Reviewed 1 file(s) changed in this PR: tools/bounty_verifier/star_checker.py

### Changes Analyzed

#### tools/bounty_verifier/star_checker.py
- Changes: +1/-1
- Fix addresses the reported issue correctly.

### Overall Assessment

### Recommendation

---

zeroknowledge0x:
Closing — issue #6779 already resolved by PR #6778 (merged by FakerHideInBush). Our stale-endpoint fix was superseded by the complete fix.

PR #6788, opened 2026-06-02
test(ci): fix 21 stale tests + httpx dep (main pytest was red)

Scottcjn:
## BCOS Checklist
- [x] **BCOS-L1** (test + test-dep fixes; one small client fix in tools/rustchain-monitor)
- [x] Test evidence below

## What Changed
main's CI `test` job has been **red** (31 failed / 3198 passed), blocking every PR's test check. Triaged via git-blame on the production code (not just assertions): these are **tests lagging behind intentionally-merged hardening PRs**, not regressions. Fixes **21** (6 files green, 130 passing):

- tx-redaction (5): `/tx/*`+`/wallet/*` GETs admin-gated (#6295) → send X-Admin-Key
- gpu (3): `/gpu/nodes` admin-gated (#6557) → authenticate read; unconfigured → 503
- bridge (8): RTC addr tightened to RTC+40hex (#5435) → valid fixtures; lock GETs admin-gated → auth
- signed_transfer (2): add httpx test dep; `/pending/confirm` redacts to internal_error (#3202/#5546)
- proxy (1): proxy forwards headers= → mock accepts kwarg
- monitor (2): removed bad-merge duplicate normalize_miners_payload; print_miners guards non-list

Only non-test change: `tools/rustchain-monitor/rustchain_monitor.py` (dup function + isinstance guard) — a real client bug the tests caught.

## Remaining follow-up (also stale-test, not regression)
- governance (4): `/governance/vote` admin-gated (#6719) → needs auth; propose wallet-format/balance fixtures
- otc confirm (1): test mock returns non-JSON-serializable MagicMock

Deferred to keep this PR clean/reviewable.

## Testing
```
pytest <the 6 files> -q  ->  130 passed
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6818, opened 2026-06-03
fix: keep GPU fingerprint import safe on CPU CI

yyswhsccc:
## Summary

- Defers PyTorch/CUDA requirement checks until the GPU fingerprint flow actually runs.
- Keeps module import safe on CPU-only CI / no-torch environments.
- Adds an import-only regression test for the no-torch path.

## Follow-up scope

This is the small GPU import-safety slice called out as acceptable in #6711.

It intentionally does not touch node/admin/auth endpoints, production security guards, or the repo-wide stale-test/admin-key baseline handled separately in #6788.

## Validation

- `python3 -m py_compile miners/gpu_fingerprint.py tests/test_gpu_fingerprint_import.py`
- `uv run --no-project --with pytest --with flask --with requests --with flask-cors --with httpx --with cryptography python -m pytest -q tests/test_gpu_fingerprint_import.py --tb=short -o addopts='' --basetemp=.pytest-tmp` -> `1 passed`
- `git diff --check`
- `python3 miners/gpu_fingerprint.py --json` without torch -> exits `1` with `ERROR: PyTorch with CUDA support required. Install: pip install torch`

Refs #305.
Follow-up to #6711.

### DIFF
--- a/miners/gpu_fingerprint.py
+++ b/miners/gpu_fingerprint.py
@@ -24,6 +24,8 @@ Usage:
 Author: Elyan Labs (RIP-0308: Proof of Physical AI)
 """
 
+from __future__ import annotations
+
 import argparse
 import json
 import hashlib
@@ -40,12 +42,17 @@ try:
     import torch
     import torch.cuda
 except ImportError:
-    print("ERROR: PyTorch with CUDA support required. Install: pip install torch")
-    sys.exit(1)
+    torch = None
+    HAS_TORCH = False
+else:
+    HAS_TORCH = True
 
-if not torch.cuda.is_available():
-    print("ERROR: No CUDA-capable GPU detected.")
-    sys.exit(1)
+
+def check_requirements():
+    if not HAS_TORCH or torch is None:
+        raise RuntimeError("PyTorch with CUDA support required. Install: pip install torch")
+    if not torch.cuda.is_available():
+        raise RuntimeError("No CUDA-capable GPU detected.")
 
 
 # ---------------------------------------------------------------------------
@@ -789,6 +796,7 @@ def cross_validate_gpu(device: torch.device) -> ChannelResult:
 
 def run_gpu_fingerprint(device_index: int = 0, samples: int = 200, epoch_salt: str = "") -> GPUFingerprint:
     """Run all GPU fingerprint channels and return results."""
+    check_requirements()
     device = torch.device(f"cuda:{device_index}")
 
     # GPU info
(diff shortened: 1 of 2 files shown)


vinhcafe25-ops (changes requested):
# Code Review: PR #6818 — fix: keep GPU fingerprint import safe on CPU CI

**Tier: Thorough** — Line-by-line review. Core approach is sound; the test has a critical bug.

---

## Finding 1 (🔴 Bug — Test): `monkeypatch.setitem(sys.modules, "torch", None)` does NOT simulate a missing torch

**File:** `tests/test_gpu_fingerprint_import.py`

```python
monkeypatch.setitem(sys.modules, "torch", None)
monkeypatch.setitem(sys.modules, "torch.cuda", None)
```

This does **not** cause `import torch` to raise `ImportError`. Python's import machinery checks `sys.modules` via key existence (`name in sys.modules`), not truthiness. Setting the entry to `None` means:

1. `import torch` → finds key in `sys.modules` → returns `None` → **no ImportError**
2. `import torch.cuda` → finds key in `sys.modules` → returns `None` → **no ImportError**
3. The `except ImportError` block is **never reached**
4. The `else` branch runs → `HAS_TORCH = True`
5. `assert module.HAS_TORCH is False` → **FAILS**

The test should delete the keys from `sys.modules` instead, which is what actually happens on a machine without torch:

```python
# Correct approach: remove from sys.modules so import raises ImportError
monkeypatch.delitem(sys.modules, "torch", raising=False)
monkeypatch.delitem(sys.modules, "torch.cuda", raising=False)
```

Or use `importlib` mocking to make the import genuinely raise:

```python
import builtins
original_import = builtins.__import__
def mock_import(name, *args, **kwargs):
    if name in ("torch", "torch.cuda"):
        raise ImportError(f"No module named '{name}'")
    return original_import(name, *args, **kwargs)
monkeypatch.setattr(builtins, "__import__", mock_import)
```

**Impact:** The test either never actually ran, or it "passes" by coincidence (e.g., torch was already imported in the test session and `monkeypatch.setitem` overwrites the real module, but the `spec.loader.exec_module` path might bypass `sys.modules` in some edge case). Either way, the test is not testing what it claims to test.

---

## Finding 2 (🟡 Fragility): Both `torch` and `torch.cuda` share one `except` block

**File:** `miners/gpu_fingerprint.py`

```python
try:
    import torch
    import torch.cuda
except ImportError:
    torch = None
    HAS_TORCH = False
else:
    HAS_TORCH = True
```

If `import torch` **succeeds** but `import torch.cuda` **fails** (edge case: corrupted install), the `except` block sets `torch = None`, discarding the successfully-imported `torch` module. This is wasteful — torch could be used for CPU ops even without CUDA.

**Better:**
```python
try:
    import torch
except ImportError:
    torch = None
    HAS_TORCH = False
else:
    try:
        import torch.cuda
    except ImportError:
        pass  # CUDA submodule missing, handled by check_requirements()
    HAS_TORCH = True
```

---

## Finding 3 (🟡 Coverage gap): No test when torch IS available

The new test only covers the `torch`-missing path. Consider adding a second test that verifies the module behaves correctly when `torch` and `torch.cuda` are actually importable (e.g., with a mock that makes `torch.cuda.is_available()` return `True`). This would catch regressions in the happy path.

---

## Finding 4 (🟢 Nit): `from __future__ import annotations` is unrelated

Added at the top of the file but not used by any change in this PR. This is PEP 604 style (`str | None` syntax). Fine if the project is standardizing, but it's unrelated noise in this diff.

---

## What works well

- `check_requirements()` pattern is clean — defers enforcement to call time, separating detection from action
- `run_gpu_fingerprint` guards with `check_requirements()` at the entry point
- CLI `__main__` block wraps the whole thing in `try/except RuntimeError` for clean user-facing errors
- Test uses `importlib.util.spec_from_file_location` correctly to avoid cached imports

---

## Verdict

**Request Changes** — The test in `test_gpu_fingerprint_import.py` is broken: setting `sys.modules["torch"] = None` does not simulate a missing module. Fix the test approach to use `delitem` or `builtins.__import__` mocking. The core `gpu_fingerprint.py` change is solid.

jaxint (commented):

vinhcafe25-ops (approved):

jaxint (commented):

Scottcjn:
## ✅ Merging — this is the clean split I asked for

Exactly the `gpu_fingerprint.py` import-safety fix from #6711 (lazy `check_requirements()` instead of import-time `sys.exit`), on its own with its import test — none of #6711's monitor-NameError or auth-on-public-GET changes. Files are only `miners/gpu_fingerprint.py` + `tests/test_gpu_fingerprint_import.py`. Thanks @yyswhsccc.

jaxint (commented):

PR #6819, opened 2026-06-03
feat: add validator attestation effectiveness metrics

### LINKED ISSUE #2374: [FEAT] No validator attestation effectiveness tracking
## Issue
Rustchain has no attestation effectiveness tracking. Cannot measure validator inclusion efficiency.

## Suggested Fix
- Attestation inclusion rate per epoch
- Effectiveness score calculation
- Historical trend chart
- Peer comparison ranking

**Wallet:** kuanglaodi2-sudo

yyswhsccc:

- [x] Add a tier label: `BCOS-L1`
- [x] If adding new code files, include SPDX header near the top
- [x] Provide test evidence below

## What Changed

Fixes #2374.

- Added attestation inclusion rate and effectiveness score cards to the validator performance dashboard.
- Added an in-browser effectiveness trend and peer comparison table ranked by effectiveness, then attestation count.
- Documented the derived metric sources and fallback behavior.
- Added static regression coverage for the new metric fields, trend rendering, and peer comparison rendering.
- Added runtime proof screenshot: `docs/runtime-environment-validator-effectiveness.png`.

## Why It Matters

The existing dashboard showed validator activity, average attestation counts, latency, and a top validator. This adds the missing effectiveness view so operators can see whether validators are being included successfully and compare peers from the same `/api/miners` payload.

## Touched Files / Subsystems

- `dashboards/validator-performance/index.html`: dashboard UI, derived effectiveness calculations, trend rendering, and peer comparison rendering.
- `dashboards/validator-performance/README.md`: metric source documentation.
- `tests/test_validator_performance_dashboard.py`: focused static regression coverage.
- `docs/runtime-environment-validator-effectiveness.png`: runtime browser screenshot of the updated dashboard.

## Scope / Risk Boundary

This is a static dashboard-only change. It does not change node APIs, consensus, miner behavior, wallet behavior, or server-side data storage. When the public API does not expose expected-attestation fields, the dashboard falls back to current active-validator coverage rather than inventing missing counts.

## Testing / Evidence

- `.venv-bounty-validation/bin/python -m pytest -q tests/test_validator_performance_dashboard.py --noconftest -o addopts=''` -> 7 passed
- `.venv-bounty-validation/bin/python -m py_compile tests/test_validator_performance_dashboard.py` -> passed
- `python3 /Users/ssr/.codex/bounty-radar/automation/scan_hidden_unicode.py --repo $PWD --changed-files-from-git origin/main...HEAD` -> passed, 4 paths scanned
- `python3 tools/bcos_spdx_check.py --base-ref origin/main` -> BCOS SPDX check: OK
- Runtime browser check: local dashboard served at `http://localhost:8874/dashboards/validator-performance/`, live `/api/miners` data rendered Inclusion Rate, Effectiveness Score, trend, and peer comparison rows.
- Screenshot proof: `docs/runtime-environment-validator-effectiveness.png`


MolhamHamwi (commented):
Formal review for bounty #73. I reviewed the validator effectiveness dashboard changes and ran `python3 -m pytest tests/test_validator_performance_dashboard.py -q` locally: 6 passed. Overall this looks safe to merge; I left two line-level observations below around ranking/null semantics and the expected-attestation fallback behavior.

dashboards/validator-performance/index.html
```diff
@@ -304,25 +355,48 @@ <h2>Validator History</h2>
 
         function computeValidatorMetrics(minerRows) {
             const activeRows = minerRows.filter(isActiveValidator);
-            const attestationValues = minerRows
-                .map((row) => readNumber(row, ["attestations", "attestation_count", "total_attestations", "attest_count"]))
+            const rankedPeers = minerRows
+                .map(computePeerEffectiveness)
+                .sort((a, b) => b.effectivenessScore - a.effectivenessScore || b.attestations - a.attestations);
```
MolhamHamwi, line 364: The peer sort is deterministic for normal numeric data and correctly breaks ties by attestation count. One edge case to be aware of: `attestations` can be `null`, and `b.attestations - a.attestations` coerces null to 0. That keeps the sort from crashing, but it means peers with unknown attestation counts sort as if they had zero. That behavior is acceptable for this display, but documenting it or normalizing with `?? 0` would make the fallback explicit.

dashboards/validator-performance/index.html
```diff
@@ -304,25 +355,48 @@ <h2>Validator History</h2>
...
+                effectivenessScore: average(peerScores),
+                rankedPeers
+            };
+        }
+
+        function computePeerEffectiveness(row) {
+            const attestations = readNumber(row, ["included_attestations", "attestations_included", "successful_attestations", "attestations", "attestation_count", "total_attestations", "attest_count"]);
+            const expected = readNumber(row, ["expected_attestations", "attestation_slots", "scheduled_attestations", "total_slots"]);
+            const explicitScore = readNumber(row, ["effectiveness_score", "performance_score", "score"]);
+            const inclusionRate = expected && attestations !== null
```
MolhamHamwi, line 390: This `expected &&` guard correctly prevents divide-by-zero for `expected_attestations: 0`. The tradeoff is that zero expected slots are shown as unknown inclusion (`--`) rather than 0% or 100%. For validator dashboards that is a reasonable fail-safe because no scheduled slots means there is no meaningful denominator; the fallback effectiveness then relies on explicit score or activity instead of emitting a misleading percentage.

vinhcafe25-ops (changes requested):
# Code Review: PR #6819 — feat: add validator attestation effectiveness metrics

**Tier: Thorough** — Line-by-line review. Good feature, two real bugs found.

---

## Finding 1 (🔴 Bug): `average([])` returns `NaN`, which bypasses `??` fallback

**File:** `dashboards/validator-performance/index.html`

```javascript
function average(values) {
    return values.reduce((sum, value) => sum + value, 0) / values.length;
}
```

When called with an empty array, this evaluates to `0 / 0 = NaN`. The `NaN` value then flows into:

```javascript
// Line ~368
inclusionRate: average(inclusionRates) ?? fallbackInclusionRate,
```

**`NaN ?? fallback` does NOT use the fallback.** The nullish coalescing operator (`??`) only checks for `null` and `undefined`. `NaN` is neither — it passes straight through. The dashboard would display `"NaN%"` instead of the fallback value.

This triggers when:
- **All** miners return `null` for both `expected` and `attestations` (so `inclusionRate` is `null` for every peer, and `inclusionRates` is filtered to `[]`)
- OR `rankedPeers` is an empty array (no miners returned from API)

**Fix:**
```javascript
function average(values) {
    if (!values.length) return null;  // or return 0
    return values.reduce((sum, value) => sum + value, 0) / values.length;
}
```

Then `inclusionRate: average(inclusionRates) ?? fallbackInclusionRate` would work correctly — `null` triggers the `??` fallback.

---

## Finding 2 (🟡 Logic): `peerScores` is NOT filtered for `null` before `average()`

```javascript
const peerScores = rankedPeers.map((row) => row.effectivenessScore);
// ...
effectivenessScore: average(peerScores),
```

Unlike `inclusionRates` (which filters `.filter((value) => value !== null)`), `peerScores` has no filter. While `computePeerEffectiveness` currently always returns a numeric `effectivenessScore` (the fallback chain ends in `freshnessScore` which is `0` or `100`), this is fragile:

1. If someone refactors `computePeerEffectiveness` to return `null` in some case, `peerScores` would contain `null` values
2. `null` in `reduce((sum, v) => sum + v, 0)` → `null + 0` → `0` (coerced), so it silently treats null as 0 — hidden data corruption
3. If `rankedPeers` is empty, `average([])` → `NaN` (same as Finding 1)

**Fix:** Add a filter for consistency, or guard `average()` against empty/mixed arrays.

---

## Finding 3 (🟡 Silent data corruption): `clamp` on `explicitScore` hides scale mismatch

```javascript
const effectivenessScore = explicitScore !== null
    ? clamp(explicitScore, 0, 100)
    : (inclusionRate ?? freshnessScore);
```

If a RustChain node returns `effectiveness_score: 150` (e.g., on a 0–200 scale), this silently clamps to 100. The dashboard operator never knows the data is potentially mis-scaled. A `console.warn` would help debugging:

```javascript
if (explicitScore !== null && (explicitScore < 0 || explicitScore > 100)) {
    console.warn("Validator effectiveness score out of expected 0–100 range:", explicitScore, row);
}
```

---

## Finding 4 (🟡 Naming): `isActiveValidator` now includes `attested === true`

```javascript
return row?.active === true
    || row?.online === true
    || row?.attested === true   // <-- added
    || status === "active"
    || status === "online"
    || status === "up"
    || ...
```

Adding `row?.attested === true` broadens what "active" means — a validator that successfully attested but is currently offline would be counted as "active." This inflates `activeValidators` and the `fallbackInclusionRate` calculation:

```javascript
const fallbackInclusionRate = minerRows.length
    ? (activeRows.length / minerRows.length) * 100
    : null;
```

If the intent is to include attested validators in the effectiveness calculation but not in the "active" count, split this into two predicates:
- `isActiveValidator(row)` — currently online/active (original behavior)
- `hasAttested(row)` — has attestation data

---

## Finding 5 (🟢 UX): `renderTrend` shows bars for 0% effectiveness

```javascript
bar.style.height = `${Math.max(4, sample.effectivenessScore ?? 0)}%`;
```

A validator with 0% effectiveness gets a bar height of `4%` of 120px ≈ ~5px. Combined with CSS `min-height: 4px`, this means zero-effectiveness validators show a visible bar, misleading users into thinking there's some effectiveness when there isn't. Consider using `Math.max(4, ...)` in px instead of %, or explicitly styling 0-score bars differently (e.g., gray, 0 height).

---

## What works well

- **Defensive field normalization** in `computePeerEffectiveness` — checks multiple API field name variants
- **Graceful fallback chain**: explicit score → inclusion rate → freshness → always ends in a number
- **Pure CSS trend chart** — no external dependencies, good `aria-label`
- **`formatPercent` helper** — cleanly handles `null` → `"--"`
- **History table updated** with new columns (6 cols instead of 4)

---

## Verdict

**Approve with fixes requested** — Finding 1 (NaN fallback) is a real runtime bug that users will see as `NaN%` when no attestation data is available. Findings 2–4 are robustness issues worth addressing. The feature itself is well-designed.

jaxint (commented):

vinhcafe25-ops (approved):

jaxint (commented):

Scottcjn:
## ✅ Merging — validator-effectiveness dashboard, XSS-safe

Self-contained `dashboards/validator-performance/` (read-only client-side metric: inclusion rate, effectiveness score, peer ranking, trend chart). Builds the DOM via `document.createElement` + `textContent` — no `innerHTML` data interpolation, so no XSS surface — and `computePeerEffectiveness` defensively reads from a list of possible field names. Has a test. Thanks @yyswhsccc.

jaxint (commented):

Issue #2373, opened 2026-04-20
[FEAT] No block explorer search filters


FlintLeng:
## Issue
Rustchain has no search filters. Cannot filter blocks by time range or proposer.

## Suggested Fix
- Time range filter
- Proposer filter
- Transaction count range
- Gas used filter

**Wallet:** kuanglaodi2-sudo

PR #103, opened 2026-02-12
fix(security): harden /wallet/transfer/signed validation & error handling

xiuqiang1995:
### Summary
Hardens the signed transfer endpoint against malformed requests and ambiguous validation failures.

### Changes
- Handle invalid/non-JSON request bodies safely (`request.get_json(silent=True)` + explicit 400)
- Validate `amount_rtc` parsing errors with clear response
- Reject non-finite numeric values (`NaN`/`Infinity`) before processing
- Enforce RTC address shape (`RTC` prefix + expected length) for sender/recipient
- Reject self-transfer (`from_address == to_address`)
- Validate nonce as positive integer-like value before signature reconstruction

### Why
Issue report flagged weak transfer validation behavior with unclear signature failures. This patch moves malformed input rejection earlier and returns deterministic validation errors before signature checking.

Related bounty issue: https://github.com/Scottcjn/rustchain-bounties/issues/93

PR #6824, opened 2026-06-03
feat: add explorer block filters

yyswhsccc:
## What changed
- Added block explorer filter controls for proposer/miner, time range, transaction count range, and gas used range.
- Rendered proposer, transaction count, and gas columns in the blocks tables.
- Normalized block API responses that arrive either as a bare array or a `{ blocks: [...] }` envelope.
- Added VM-backed regression coverage for the new filter behavior.

## Why it matters
This addresses #2373 by making the existing block explorer table searchable across the requested block fields when those fields are present in block payloads, while preserving current behavior for sparse block responses.

## Validation
- `node --check explorer/static/js/explorer.js`
- `.venv-bounty-validation/bin/python -m pytest tests/test_explorer_block_filters.py -q`
- `git diff --check`
- `python3 /Users/ssr/.codex/bounty-radar/automation/scan_hidden_unicode.py --repo $PWD --changed-files-from-git origin/main...HEAD`
- Playwright smoke check against `http://127.0.0.1:8899/`: filtered sample block data to `1 of 3 blocks` with only block `#103` visible.

## Touched files/subsystems
- `explorer/index.html`
- `explorer/static/css/explorer.css`
- `explorer/static/js/explorer.js`
- `tests/test_explorer_block_filters.py`

## Scope/risk boundary
This is a client-side explorer filtering slice. It does not change block production, consensus, API authorization, or persisted chain data.

### DIFF
--- a/explorer/index.html
+++ b/explorer/index.html
@@ -423,10 +423,8 @@
             
             // Sync blocks table
             if (viewName === 'blocks') {
-                const blocksTbody = document.getElementById('blocks-tbody-full');
-                const blocksTbodyOverview = document.getElementById('blocks-tbody');
-                if (blocksTbody && blocksTbodyOverview) {
-                    blocksTbody.innerHTML = blocksTbodyOverview.innerHTML;
+                if (window.RustChainExplorer && window.RustChainExplorer.renderBlocksTable) {
+                    window.RustChainExplorer.renderBlocksTable();
                 }
             }
             
--- a/explorer/index.html
+++ b/explorer/index.html
@@ -132,12 +132,15 @@
                                     <th scope="col">Height</th>
                                     <th scope="col">Hash</th>
                                     <th scope="col">Timestamp</th>
+                                    <th scope="col">Proposer</th>
+                                    <th scope="col">Txs</th>
+                                    <th scope="col">Gas</th>
                                     <th scope="col">Miners</th>
                                     <th scope="col">Reward</th>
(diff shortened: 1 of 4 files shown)


qingfeng312 (changes requested):
I found a blocker in the new block-filter flow.

`fetchBlocks()` still stores only `blocks.slice(0, CONFIG.MAX_RECENT_BLOCKS)` in `state.blocks`, and the new filter path plus `syncFullBlocksTable()` both read from that same truncated array. The filter UI is placed in the “All Blocks” section and the summary reports against `state.blocks.length`, but any block outside the recent-block cap is never available to filter or display. As a result, filtering by proposer/time/tx/gas can incorrectly return “No blocks match” even when older blocks from `/blocks` would match.

Please keep an untruncated all-blocks collection for the full table/filter path, and derive the recent overview from that separately.

jaxint (commented):

jaxint (commented):

Scottcjn:
## ✅ Merging — explorer block filters, XSS-safe

Client-side explorer feature (block filters + a proposer column). All block data rendered into `innerHTML` goes through the existing `escapeHtml()` helper (consistent with the existing hash column), summary uses `textContent`, and `normalizeBlocksResponse` defensively handles null/malformed entries (with a test). Thanks @yyswhsccc.

jaxint (commented):

PR #6834, opened 2026-06-03
[codex] cap coin_select fallback inputs

kitwongpixel:
Fixes bug report #6830.

## What changed
- `coin_select()` now fails cleanly if the fallback selection still needs more than 20 inputs.
- Added regression tests for equal-value UTXO edge cases.

## Why
The documented largest-first fallback could still return an oversized selection when all UTXOs had the same value. That made the fallback appear successful even though it violated the input limit.

## Impact
- Prevents oversized multi-input spends from slipping through the fallback path.
- Keeps selection behavior consistent with the documented input cap.
- Adds explicit test coverage for the equal-value edge case.

## Validation
- `python3 -m unittest node.test_coin_select_extended node.test_utxo_db`
- Result: 109 tests passed

## Bounty wallet
- RTC address: `RTCd90fc88820a76397d26d80bcd63c8b5711a383bd`
- Public key: `ddcb08b3006a6337354813ed2f8b56fa619c26d58ef83e791427edd5d043490e`

## Notes
- Local commit: `65029da fix: cap coin_select fallback inputs`

PR #6867, opened 2026-06-05
[Fix] coin_select() equal-value UTXO edge case (issue #6830)

### LINKED ISSUE #6830: [Bug] coin_select() largest-first fallback still returns >20 inputs on equal-value UTXOs
## Summary

`node/utxo_db.py::coin_select()` documents a fallback to largest-first when smallest-first selection uses more than 20 inputs, but the fallback still returns more than 20 inputs when the UTXOs all have the same value.

This means the heuristic does not actually reduce input count in an important edge case, so transaction construction can still emit oversized multi-input spends even though the code claims it is trying to switch to a fewer-input strategy.

## Affected code

- `node/utxo_db.py:1444-1489`
- `node/utxo_endpoints.py:570` (calls `coin_select()` directly when building wallet transfers)

## Reproduction

Run this from the repo root:

```bash
python3 - <<'PY'
import sys
sys.path.insert(0, '/root/rustchain/node')
from utxo_db import coin_select

utxos = [{'value_nrtc': 1} for _ in range(25)]
selected, change = coin_select(utxos, 21)
print('selected_inputs=', len(selected))
print('change=', change)
PY
```

Observed output on current `main`:

```text
selected_inputs= 21
change= 0
```

## Expected behavior

If the function says it will restart with largest-first when the selection exceeds 20 inputs, the fallback should actually reduce the input count or fail cleanly for this edge case.

## Actual behavior

When all UTXOs have equal value, smallest-first and largest-first produce the same ordering, so the fallback still returns 21 inputs for a 21-RTC target.

## Impact

- The input-count heuristic does not work for equal-value UTXO sets.
- Wallets with many same-sized UTXOs can still produce large transactions.
- That increases transaction size and makes the intended anti-bloat guard ineffective in a common consolidation scenario.

## Notes

I searched for an existing open issue for this specific coin-selection edge case and did not find one.

zeroknowledge0x:
## Problem

When all UTXOs have the same value, smallest-first and largest-first selection strategies produce identical ordering. The documented fallback to largest-first when smallest-first exceeds 20 inputs fails to reduce input count in this edge case, causing the function to return  (insufficient funds) despite having sufficient funds — because  triggers the guard.

## Fix

Enhanced coin_select logic:
1. Try smallest-first (dust consolidation)
2. If >20 inputs, try largest-first
3. If largest-first still >20 inputs **and** covers target:
   - Cap to 20 largest inputs
   - Verify 20 largest still cover target
   - If yes, accumulate from capped set
   - If no, return  (truly insufficient)
4. Return selected UTXOs and change

This prevents false 'insufficient funds' while avoiding oversized spends.

Fixes #6830

### DIFF
--- a/node/utxo_db.py
+++ b/node/utxo_db.py
@@ -1496,10 +1496,33 @@ def coin_select(utxos: List[dict], target_nrtc: int
             fallback_total += u['value_nrtc']
             if fallback_total >= target_nrtc:
                 break
-        if fallback_total < target_nrtc or len(fallback) > 20:
+
+        if fallback_total < target_nrtc:
             return [], 0
-        selected = fallback
-        total = fallback_total
+
+        # If largest-first still exceeds 20 inputs (e.g. all UTXOs have equal
+        # value), cap to the 20 largest and verify they still cover the target.
+        # This prevents the function from silently returning oversized spends
+        # while also avoiding false "insufficient funds" when the target is
+        # actually reachable with ≤20 inputs.
+        if len(fallback) > 20:
+            capped = sorted_desc[:20]
+            capped_total = sum(u['value_nrtc'] for u in capped)
+            if capped_total >= target_nrtc:
+                # Re-run accumulation on capped set to get exact selection
+                selected = []
+                total = 0
+                for u in capped:
+                    selected.append(u)
+                    total += u['value_nrtc']
+                    if total >= target_nrtc:
+                        break
+            else:
+                # Even 20 largest inputs cannot cover the target
+                return [], 0
+        else:
+            selected = fallback
+            total = fallback_total
 
     change = total - target_nrtc
     if change < DUST_THRESHOLD:


vicentsmith470-web (changes requested):
Reviewed the `coin_select()` equal-value UTXO change against issue #6830 and the current PR history.

Blocking concerns:

1. This appears stale/duplicative now that #6834 is already merged for the same issue. #6834 changed `coin_select()` to fail cleanly when the largest-first fallback still needs more than 20 inputs and added regression coverage in both `node/test_coin_select_extended.py` and `node/test_utxo_db.py`. Because #6867 targets the same #6830 behavior after that fix, it should be rebased against current `main` and shown to add behavior not already covered, or closed as superseded.

2. The new `capped_total >= target_nrtc` branch looks effectively unreachable with the current loop shape. The fallback loop breaks as soon as `fallback_total >= target_nrtc`. If `len(fallback) > 20`, that means the first 20 largest inputs did *not* cover the target; otherwise the loop would have stopped at 20 or fewer inputs. Therefore `sum(sorted_desc[:20]) >= target_nrtc` cannot be true in the branch where `len(fallback) > 20`. In the equal-value case from #6830, this correctly falls through to `return [], 0`, but that is already the simpler merged behavior.

3. This PR does not include regression tests, while the merged fix did. If this is intended to refine the merged behavior, please add a focused test that demonstrates the new behavior beyond #6834. A useful one would need to prove a case where smallest-first exceeds 20 inputs and the proposed capped-largest path succeeds; as noted above, the current accumulation logic does not seem to permit that branch.

Given the duplicate issue state and unreachable branch, I would not merge this as-is. Please rebase against current main and either reduce it to a no-op/close, or add a failing-before/passing-after case that proves the extra fallback logic is needed.

Scottcjn:
Thanks for tackling #6830! Closing as a duplicate — the coin_select cap already shipped via #6834 (merged). Your refinement (cap to the 20 largest and re-check coverage) is thoughtful, but it's behaviorally equivalent to the merged version: largest-first accumulation already minimizes the input count, so if it needs >20 inputs to reach the target, the 20 largest necessarily sum to less than the target — meaning your `capped_total >= target` branch can't be reached when `len(fallback) > 20`. Same result, so #6834 already covers it. Appreciate the careful thinking. 🦞

jaxint (commented):

jaxint (commented):

jaxint (commented):

jaxint (commented):
Nice work!

jaxint (commented):
Great work!

PR #6863, opened 2026-06-05
test(clawrtc): skip integration tests when clawrtc.cli is absent (fix flaky CI red)

Scottcjn:
## Problem
`tests/test_clawrtc_integration.py` imports `clawrtc.cli`, which is **not committed in-repo** (the `miners/clawrtc/` package has `__init__/config/pow_miners/test_config` but no `cli.py`). The autouse `_isolate_wallet` fixture guarded its `monkeypatch.setattr("clawrtc.cli...")` calls with `except AttributeError` only. A missing module raises `ModuleNotFoundError` (a subclass of `ImportError`), which slips past that guard → the fixture **errors** for all 27 tests instead of skipping.

Because `clawrtc.cli` is only sometimes importable (depends on test ordering / build side-effects in a given run), the blocking `test` check **flaps red nondeterministically** — main itself went green on one commit and red on the next with identical content. This reddens unrelated PRs and forces manual `--admin` merges.

## Fix
Catch `ImportError` (covers `ModuleNotFoundError`) in both the autouse fixture and the `cli` fixture, so the suite **skips cleanly** wherever `clawrtc.cli` isn't importable. No behavior change where it *is* importable.

## Verification
```
$ python3 -m pytest tests/test_clawrtc_integration.py -q
27 skipped in 0.13s   # was: 27 errors
```

This is a test-only change; it removes a recurring false-negative on the `test` gate. A proper follow-up is to either commit the real `clawrtc/cli.py` or remove the orphaned integration test (Bounty #426/#1913).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #6627, opened 2026-05-29
[TRACKING] Bounded-query helper + CI guard against raw .fetchall() — eliminates the UTXO-OOM bug class


Scottcjn:
## Background

The last week's main-branch security work shipped 4 `[UTXO-BUG]` commits, all fitting the same pattern: **unbounded materialization of attacker-influenced row sets via raw `.fetchall()` on a public/semi-public endpoint**.

Examples already merged:
- #6526 `_evict_stale_data_input_txs()` fetchall OOM
- #6535 mempool candidate scan unbounded
- #6537 dual-write reward settlement
- #6562 `/wallet/history` unbounded fetchall
- #6563 `/rewards/epoch` unbounded fetchall
- #6571 `/governance/proposal/<id>` votes unbounded

Each individually was Medium severity. Each was patched route-by-route after a report. That's reactive — and the class keeps recurring because the underlying shape never changed.

## Scope of the latent exposure

Codex authoritative code-state audit (run today against the current main) surfaced:

- **191 raw `.fetchall()` calls in non-test `node/` Python files**
- Top concentrations:
  - `node/rustchain_v2_integrated_v2.2.1_rip200.py` — 50 instances
  - `node/anti_double_mining.py` — 7
  - `node/rustchain_tx_handler.py` — 6
  - `node/rustchain_block_producer.py` — 6
  - `node/hall_of_rust.py` — 6
  - `node/coalition.py` — 6
  - `node/beacon_api.py` — 6

Not every `.fetchall()` is exploitable. Most operate on bounded internal state. But every route or consensus-adjacent endpoint that hits an attacker-influenced query path is a latent OOM-DoS.

## Required fix (architectural)

This is two PRs — claim either or both. First-poster precedence applies per-PR.

### Part A: Bounded-query helper (High = 50 RTC)

Introduce a shared helper in a new module (e.g., `node/db_helpers.py`):

```python
def fetch_page(
    conn: sqlite3.Connection,
    sql: str,
    params: tuple = (),
    *,
    limit: int,
    offset: int = 0,
    max_limit: int = 1000,
) -> list[sqlite3.Row]:
    """Bounded SELECT against `conn` with explicit limit.

    Raises ValueError if limit > max_limit. Appends LIMIT/OFFSET to `sql`
    after validating it does not already contain LIMIT.
    """
```

Plus a sibling for explicitly tiny result sets (schema introspection, single-row PRAGMA):

```python
def fetch_one_or_none(
[...]

PR #6870, opened 2026-06-05
ci(fetchall): resync migration baseline to current line numbers (unblock main)

Scottcjn:
## Problem
`main`'s blocking `test` check is **RED**: `tests/test_fetchall_guard.py::test_fetchall_guard_passes_current_baseline` fails with `20 new unannotated .fetchall() call(s) in node/`. Because that check is repo-wide, it reds the `test` gate on **every open PR** (people see a failure that has nothing to do with their change), forcing `--admin` merges.

## Root cause
The guard records each legacy raw `.fetchall()` site as `file:line:content` in `scripts/baselines/fetchall_existing.txt`. As `node/rustchain_v2_integrated_v2.2.1_rip200.py` grew, those **line numbers drifted** — the same calls now sit at new lines, so they no longer match their baseline entries and read as 'new'. No new query was actually added.

## Fix
Pure line-number resync, regenerated the intended way:
```
bash scripts/check_fetchall.sh --print-baseline > scripts/baselines/fetchall_existing.txt
```
(header comments preserved.)

## Safety
- **182 entries before, 182 after** — same count.
- A by-content diff (strip `file:line`, compare only the call code) shows **ZERO entries new by content**. The set is identical — the same `PRAGMA table_info(...)` and bounded queries as before, just at current line numbers.
- The guard's purpose is unchanged: genuinely-new unannotated `.fetchall()` sites are still blocked. This only resyncs the snapshot of the known legacy backlog (issue #6627).
- `scripts/check_fetchall.sh` exits 0 after the change; `tests/test_fetchall_guard.py` passes 6/6 locally.

Same class of repo-health unblock as the checksum auto-regen hook (#6857) and the clawrtc flake fix (#6863).

### DIFF
--- a/scripts/baselines/fetchall_existing.txt
+++ b/scripts/baselines/fetchall_existing.txt
@@ -1,31 +1,130 @@
 # Existing unannotated .fetchall() migration baseline for issue #6627.
 # Expected to shrink as legacy callers migrate to node.db_helpers.fetch_page()
 # or receive a narrow fetchall-ok annotation. Do not add new entries manually.
-node/utxo_genesis_migration.py:91:        ).fetchall()
-node/utxo_genesis_migration.py:104:        ).fetchall()
-node/utxo_db.py:379:            ).fetchall()
-node/utxo_db.py:940:            ).fetchall()
-node/utxo_db.py:1281:            ).fetchall()
-node/utxo_db.py:1346:            ).fetchall()
-node/utxo_db.py:1420:                ).fetchall()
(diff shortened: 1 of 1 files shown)


jaxint (commented):

jaxint (commented):

jaxint (commented):

jaxint (commented):
Nice work!

jaxint (commented):
Great work!

PR #6874, opened 2026-06-05
fix: require admin authentication on POST /api/bridge/lock (H-03)

BossChaos:
### Vulnerability
Unauthenticated attackers can create arbitrary bridge lock records by calling `POST /api/bridge/lock` without any authentication header. This poisons the `bridge_locks` database table with fake lock entries, potentially disrupting bridge accounting.

### Root Cause
The `create_bridge_lock()` route handler in `node/airdrop_v2.py` had no authentication — all other bridge endpoints (`/confirm`, `/release`) require `X-Admin-Key`, but this one was missing.

### Fix
Add `require_admin_key()` authentication check at the start of `create_bridge_lock()`.

### Severity
**High** — Bridge state poisoning without any credentials required.

### Testing
- Verify `POST /api/bridge/lock` without `X-Admin-Key` → `401 unauthorized`
- Verify `POST /api/bridge/lock` with correct `X-Admin-Key` → proceeds to DB write

### Bounty Claim
Filed as H-03 on Algora bounty board.

### DIFF
--- a/node/airdrop_v2.py
+++ b/node/airdrop_v2.py
@@ -1442,6 +1442,9 @@ def init_airdrop_routes(app, airdrop: AirdropV2, db_path: str) -> None:
     @app.route("/api/bridge/lock", methods=["POST"])
     def create_bridge_lock():
         """Create bridge lock."""
+        auth_error = require_admin_key()
+        if auth_error:
+            return auth_error
         data, error = parse_json_object_body()
         if error:
             return error


node/airdrop_v2.py
```diff
@@ -1442,6 +1442,9 @@ def get_airdrop_stats():
     @app.route("/api/bridge/lock", methods=["POST"])
     def create_bridge_lock():
         """Create bridge lock."""
+        auth_error = require_admin_key()
```
FakerHideInBush, line 1445: Adding the gate changes the contract for every /api/bridge/lock route test, but the PR does not update those tests. In the focused suite, _create_pending_lock now fails with 401 before confirm/release auth assertions run, and the invalid JSON/amount/address validation cases also get 401 before body validation. Please update those tests to send a valid admin header where they need authenticated lock creation/validation, plus a dedicated unauthenticated rejection/no-insert regression test for this new guard.

FakerHideInBush (changes requested):
Thanks for locking down the bridge-lock creation path. I agree the new admin gate is directionally consistent with the existing confirm/release admin checks, but this PR leaves the affected route tests in the old unauthenticated contract and CI is red as a result.

Local validation: python -m pytest tests/test_airdrop_bridge_admin_auth.py -q -> 11 failed, 21 passed. The failures are all on /api/bridge/lock: setup helpers still create locks without X-Admin-Key, and the existing invalid body/amount/address validation tests now receive 401 before exercising validation.

Please update the bridge-lock tests to use a configured admin key plus valid X-Admin-Key wherever they need to exercise authenticated creation/validation, and add an explicit regression test that unauthenticated POST /api/bridge/lock rejects and does not insert a lock. That will keep the new security behavior covered while preserving validation coverage.

JesusMP22 (commented):
## Code Review: Require admin auth on POST /api/bridge/lock (H-03)

**Summary:** Security fix H-03 - adds admin authentication requirement to the bridge lock endpoint, preventing unauthorized bridge operations.

**What I like:**
- Critical security fix for an authentication bypass
- Clear reference to the security audit finding (H-03)
- The bridge/lock endpoint is high-value and should definitely require admin auth

**Suggestions:**
1. Verify that ALL bridge-related POST endpoints have proper auth checks
2. Add regression tests that confirm unauthenticated requests are rejected
3. Consider adding audit logging for bridge lock operations
4. Review if any other endpoints have similar auth gaps

**Security considerations:**
- ✅ Critical fix: unauthorized bridge lock could freeze funds or disrupt operations
- Ensure the auth check happens before any business logic
- Consider defense in depth: auth + rate limiting + audit logging

**Verdict:** ✅ Critical security fix. Must be merged promptly.

Scottcjn:
Heads up — this was reverted in #6893. It was merged while its `test` check was still red (my error on the merge), and it broke existing tests:
- Adding `require_admin_key()` to `/api/bridge/lock` makes the admin check fire **before** amount/format validation, so 11 `test_airdrop_bridge_admin_auth` tests get 401 instead of their expected 400 validation errors.

The auth-hardening intent is good — please resubmit with the affected tests updated to match the new behavior (and a green `test` run), and I'll land it. 🦞

PR #6880, opened 2026-06-05
fix: make fetchall baseline line-independent

### LINKED ISSUE #6872: Make the fetchall guard line-number-independent (stop the recurring baseline drift)
## Problem
`scripts/check_fetchall.sh` / `tests/test_fetchall_guard.py` record each legacy raw `.fetchall()` site in `scripts/baselines/fetchall_existing.txt` keyed by `file:line:content`. The line number makes the baseline **fragile**: any PR that adds/removes lines in `node/rustchain_v2_integrated_v2.2.1_rip200.py` *above* a baselined call shifts that call's line number, so it no longer matches its baseline entry and the guard reports it as a 'new unannotated' site — turning the blocking `test` check RED on main and on every open PR.

This has now required manual resyncs twice in one day (#6870, then #6871 after #6866 shifted lines again). It's whack-a-mole.

## Proposed fix
Key the baseline by **content**, not `file:line`:
- Store `file:normalized_content` (drop the line number) in the baseline.
- In the guard, transform both the scan and the baseline the same way before comparing.
- To preserve detection of a *newly-added* call whose content duplicates an existing one, compare as a **multiset** (e.g. `sort | uniq -c` counts per `file:content`) rather than `sort -u`, so adding an Nth identical-content call still trips the guard.

### Acceptance
- Adding/removing unrelated lines in a node file does NOT change the guard result (no false 'new' from line shifts).
- A genuinely-new `.fetchall()` (new content, or an additional instance of existing content) IS still flagged.
- `tests/test_fetchall_guard.py` updated to cover: (a) a line-shift produces no error, (b) a new distinct call is caught, (c) a duplicate-content new call is caught.
- Migrate `scripts/baselines/fetchall_existing.txt` to the new format.

### Why it matters
This is the same class of recurring CI tax as the miner-checksum manifest (fixed structurally by the auto-regen hook #6857) and the clawrtc flaky test (#6863). A line-independent guard ends the resync treadmill for good.

_Filed after resyncing the baseline twice (#6870, #6871) in one session._

sayuru-akash:
Fixes #6872.

## Summary
- normalize fetchall guard baseline entries to `file:content` so line shifts above legacy calls do not break CI
- keep sorted duplicate entries instead of `sort -u`, so adding another identical raw `.fetchall()` still fails as a new multiset entry
- migrate `scripts/baselines/fetchall_existing.txt` to the new line-independent format
- add regression coverage for line-shifted baseline entries and duplicate-content additions

## Validation
- `bash -n /tmp/check_fetchall.sh`
- `python3 -m py_compile /tmp/test_fetchall_guard.py`

I could not run the full `tests/test_fetchall_guard.py` locally because the repository checkout/fetch stalled in `index-pack`; this branch was created via GitHub API from upstream `main`.

### DIFF
--- a/scripts/baselines/fetchall_existing.txt
+++ b/scripts/baselines/fetchall_existing.txt
@@ -1,185 +1,185 @@
 # Existing unannotated .fetchall() migration baseline for issue #6627.
 # Expected to shrink as legacy callers migrate to node.db_helpers.fetch_page()
 # or receive a narrow fetchall-ok annotation. Do not add new entries manually.
-node/airdrop_v2.py:1142:        rows = cursor.fetchall()
-node/airdrop_v2.py:1193:        rows = cursor.fetchall()
-node/airdrop_v2.py:1226:            for row in cursor.fetchall()
-node/airdrop_v2.py:1238:            for row in cursor.fetchall()
-node/anti_double_mining.py:160:    enrolled = cursor.fetchall()
-node/anti_double_mining.py:210:        rows = cursor.fetchall()
-node/anti_double_mining.py:333:    rows = cursor.fetchall()
-node/anti_double_mining.py:367:    enrolled = cursor.fetchall()
-node/anti_double_mining.py:407:        rows = cursor.fetchall()
-node/anti_double_mining.py:439:        cols = conn.execute("PRAGMA table_info(epoch_enroll)").fetchall()
-node/anti_double_mining.py:450:        ).fetchall()
-node/bcos_routes.py:544:            rows = conn.execute(query, params).fetchall()
-node/beacon_anchor.py:236:        ).fetchall()
-node/beacon_anchor.py:305:        ).fetchall()
-node/beacon_anchor.py:79:        for row in conn.execute("PRAGMA table_info(beacon_envelopes)").fetchall()
-node/beacon_api.py:1212:        rows = db.execute("SELECT * FROM beacon_reputation ORDER BY score DESC").fetchall()
-node/beacon_api.py:383:    ).fetchall()
-node/beacon_api.py:412:        ).fetchall()
-node/beacon_api.py:678:            ).fetchall()
-node/beacon_api.py:684:            ).fetchall()
-node/beacon_api.py:726:        ).fetchall()
-node/beacon_api.py:955:        ).fetchall()
-node/beacon_identity.py:125:        ).fetchall()
-node/beacon_identity.py:259:        ).fetchall()
-node/beacon_x402.py:341:            ).fetchall()
-node/beacon_x402.py:369:            ).fetchall()
-node/beacon_x402.py:410:            ).fetchall()
-node/beacon_x402.py:72:                     for row in cursor.fetchall()}
-node/bottube_feed_routes.py:128:            rows = cursor_obj.fetchall()
-node/bridge_api.py:547:    rows = cursor.execute(query, params).fetchall()
-node/claims_eligibility.py:675:            epochs = [row[0] for row in cursor.fetchall() if row[0] >= 0]
-node/claims_settlement.py:131:            for row in cursor.fetchall():
-node/claims_settlement.py:416:            """, (max_claims,)).fetchall()
-node/claims_settlement.py:443:                """, (batch_id, max_claims)).fetchall()
-node/claims_settlement.py:86:            for row in cursor.fetchall():
-node/claims_submission.py:116:                    columns = {row[1] for row in cursor.fetchall()}
-node/claims_submission.py:640:            for row in cursor.fetchall():
-node/coalition.py:322:            ).fetchall()
-node/coalition.py:874:                    ).fetchall()
-node/coalition.py:879:                    ).fetchall()
-node/coalition.py:915:                ).fetchall()
-node/coalition.py:960:                    ).fetchall()
-node/coalition.py:966:                    ).fetchall()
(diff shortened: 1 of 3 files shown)


zqleslie (approved):
Reviewed the diff for the fetchall guard baseline line-independence fix (#6872).

**What the PR does:**
1. **Baseline format change** (`scripts/baselines/fetchall_existing.txt`): Removes line numbers from entries, switching to `file:content` format. This means if code above a legacy `.fetchall()` call gets edited (inserting/deleting lines), the baseline entry still matches — solving the false-positive CI failure.

2. **Guard script update** (`scripts/check_fetchall.sh`): Added normalization step using `sed -E 's/^([^:]+):[0-9]+:/\1:/'` to strip line numbers before comparing scan results against the baseline. Also switched from `sort -u` to `sort` to preserve duplicate entries (new identical `.fetchall()` calls should still fail).

3. **Test coverage** (`tests/test_fetchall_guard.py`): Added two new tests:
   - `test_detects_guard_allows_line_shifted_baseline_entry()` — validates that a baseline entry with shifted line numbers still passes
   - `test_fetchall_guard_blocks_duplicate_content_call_even_when_baselined_once()` — validates that duplicate-content `.fetchall()` calls are blocked even when baselined once (multiset behavior)

**Findings:**

1. **Scope is minimal and focused** — only touches the baseline file, guard script, and tests. No node code, no runtime paths, no wallet/balance changes.

2. **The `sort` vs `sort -u` change is correct** — using `sort` instead of `sort -u` means if a new `.fetchall()` call produces the same `file:content` entry as an existing baseline entry, the guard will still detect it as a new unannotated call. This prevents accidental bypasses.

3. **Edge case to consider**: The normalization `sed -E 's/^([^:]+):[0-9]+:/\1:/'` assumes line numbers are always present. If a baseline entry somehow lacks a line number (e.g., manually edited), the sed would silently pass it through unchanged. This is low risk since the baseline is auto-generated via `--print-baseline`.

4. **Test `run_guard` helper**: Added `extra_env` parameter to allow tests to override `FETCHALL_BASELINE` — clean design for isolated test execution.

**Validation:**
- Script logic is sound: normalize → sort → comm comparison
- No `git diff --check` issues visible in the diff
- The approach is strictly additive (no regression risk for existing entries)

**Verdict: Safe to merge.** This is a CI infrastructure improvement, not a runtime change. The line-independent baseline format makes the fetchall guard more robust against unrelated code edits above legacy `.fetchall()` calls.

JesusMP22 (commented):
## Code Review: fetchall baseline line-independent

**Summary:** This PR addresses the issue of fetchall baseline being dependent on line numbers, which could cause fragile tests and unexpected failures when files are modified.

**What I like:**
- Making the baseline line-independent is a solid improvement for test maintainability
- Reduces coupling between test expectations and file structure

**Suggestions:**
1. Consider adding a comment explaining the matching algorithm for future maintainers
2. If there are edge cases where line-independence could mask real regressions, consider adding a secondary check
3. Ensure the test coverage includes both the happy path and cases where the baseline should NOT match

**Security considerations:**
- No security concerns identified. This is a test infrastructure improvement.

**Verdict:** ✅ Looks good. The change improves test robustness without introducing risk.

jaxint (commented):

jaxint (commented):

jaxint (commented):
Thanks for this contribution! Great work.

MolhamHamwi (changes requested):
I reviewed `scripts/check_fetchall.sh`, `scripts/baselines/fetchall_existing.txt`, and `tests/test_fetchall_guard.py` at the current head.

Two blockers I hit while validating locally:

1. `python3 -m pytest tests/test_fetchall_guard.py -q` fails 3 tests (`test_fetchall_guard_passes_current_baseline`, `test_fetchall_guard_allows_annotated_call`, and `test_fetchall_guard_detects_stale_baseline_entries`). The new normalization keeps duplicate normalized rows because both the baseline and current unannotated list use `sort` instead of `sort -u`; as a result, the duplicate multiline `node/utxo_genesis_migration.py:        ).fetchall()` entries are reported as stale/new even though they are in the regenerated baseline. This makes the guard fail on the current tree before it can protect future changes.

2. `scripts/check_fetchall.sh` changed from executable mode `100755` to `100644`. The workflows/tests currently call it via `bash`, but the script has a shebang and is a repository utility; losing execute permission is an avoidable regression for direct `./scripts/check_fetchall.sh` use.

Additional validation:
- `bash scripts/check_fetchall.sh` currently exits 1 with the same stale baseline entries.
- `git diff --check origin/main...HEAD` reports `tests/test_fetchall_guard.py:162: new blank line at EOF.`

Suggested fix: keep line-number normalization, but deduplicate the normalized baseline/current sets before `comm` (or otherwise preserve multiplicity intentionally with matching regenerated baseline semantics), restore executable mode on the shell script, and remove the extra trailing blank line.

jaxint (commented):
Thanks for the contribution!

jaxint (commented):
Nice work!

jaxint (commented):
Nice work! Thanks for contributing. 👍

jaxint (commented):
Thanks for the contribution! 🙏

jaxint (commented):

jaxint (commented):

jaxint (commented):
Nice work!

jaxint (commented):
Great work!

PR #6868, opened 2026-06-05
fix: harden attestation signing fallback

vicentsmith470-web:
/claim #6859

## Summary

Fixes #6859 by hardening the attestation-signing follow-ups from the #6839 review.

What changed:
- `setup_miner.py` now downloads and SHA-256 verifies `miners/signing_helpers.py` next to the installed miner, so installed Linux/Windows miners can use the shared pipe-message builder instead of always taking the inline fallback.
- The shell installers also download and verify `signing_helpers.py` via `miners/checksums.sha256`.
- Linux and Windows miners now log a WARNING if Ed25519 attestation signing fails instead of silently swallowing the exception.
- Added regression coverage for the installed-miner inline fallback path by driving Linux `LocalMiner.attest()` with `_SIGNING_HELPERS=False` and asserting the signed bytes match the node verifier reconstruction (`miner_id|miner|nonce|commitment`).
- Added regression coverage that signing failures emit a warning and fall through unsigned.

## Scope notes

- No changes to attestation payload shape, server verification, reward math, enrollment, mining loop, or macOS miner behavior.
- This PR intentionally leaves macOS unsigned-attestation behavior unchanged, matching the informational note in #6859.

## Validation

```text
python -m pytest -q tests/test_windows_miner_setup_checksum.py tests/test_setup_miner_downloads.py tests/test_install_miner_checksums.py tests/test_attestation_signing_6798.py tests/test_linux_miner_identity.py tests/test_linux_miner_network_retry.py
23 passed

python -m py_compile miners/linux/rustchain_linux_miner.py miners/windows/rustchain_windows_miner.py setup_miner.py tests/test_attestation_signing_6798.py tests/test_setup_miner_downloads.py tests/test_install_miner_checksums.py tests/test_windows_miner_setup_checksum.py
git diff --check origin/main
```

## Bounty / payout

Wallet: `RTCf69dd944558d4e843a4a676495a97638055caea2`

PR #6895, opened 2026-06-06
fix: sign Windows attestations with report nonce

### LINKED ISSUE #6894: Windows miner silently downgrades Ed25519 attestations to unsigned
## Summary

The Windows miner cannot use the shared Ed25519 signing helper introduced by #6839. Its attestation payload stores the challenge nonce only at `report.nonce`, while `build_pipe_sign_message()` requires a top-level `attestation[nonce]`.

The resulting `KeyError` is caught by the miner's broad signing guard, which logs a warning and submits the attestation without `signature`, `public_key`, or `signature_type`. The live node accepts that unsigned fallback, so Windows operators can believe Ed25519 protection is active while their attestations are not cryptographically bound to the wallet.

## Reproduction

Environment:
- Windows 10/11 real hardware
- current `main` (`826a278b`)
- Python 3.10
- `PyNaCl 1.6.2`
- native RTC wallet

Run the current Windows miner's `attest()` path with `PyNaCl` installed. The hardware checks pass, but the console emits:

```text
WARNING:root:attestation signing failed; falling through unsigned: attestation missing required field: 'nonce'
```

The live `/attest/submit` request is accepted, but the outgoing payload has no Ed25519 fields.

## Root cause

Windows constructs:

```python
attestation = {
    miner: self.wallet_address,
    miner_id: self.miner_id,
    report: {
        nonce: nonce,
        commitment: ...,
    },
}
```

The shared helper reads:

```python
nonce = attestation[nonce]
```

The node verifier instead normalizes nonce as:

```python
nonce = report.get('nonce') or _attest_text(data.get('nonce'))
```

Linux happens to send both locations, so existing helper tests built from the Linux-shaped fixture do not catch the Windows failure.

## Expected

When `PyNaCl` and a keypair are available, Windows attestations should include `signature`, `public_key`, and `signature_type=ed25519`, and the signed bytes should exactly match the node's `miner_id|miner|nonce|commitment` reconstruction.

## Actual

The helper raises before signing and the miner silently degrades to the unsigned compatibility path.

## Proposed fix and validation

Make the shared helper mirror the node's nonce precedence:

```python
report = attestation[report]
nonce = report.get(nonce) or attestation[nonce]
```

Add regression coverage for:
- the real Windows `attest()` payload, which has no top-level nonce
- differing top-level/report nonces, proving `report.nonce` wins like the node
- the existing Linux-shaped payload

Local validation after the fix:

```text
9 passed in 1.76s
py_compile: passed
git diff --check: passed
```

Live validation against the official node on the same Windows machine:

```text
accepted: true
signature_present: true
public_key_present: true
signature_type: ed25519
top_level_nonce_present: false
report_nonce_present: true
```

Suggested severity: High/security, because this defeats the Windows attestation wallet-binding protection while presenting only a warning and continuing successfully. Maintainer discretion applies.

vicentsmith470-web:
## Summary

- make the shared attestation signing helper use `report.nonce` first, matching the node verifier
- preserve the legacy top-level nonce fallback used by older clients
- exercise the real Windows `attest()` path so a missing Ed25519 signature cannot regress silently

## Root cause

The Windows miner stores its challenge nonce only in `attestation[report][nonce]`, but `build_pipe_sign_message()` required `attestation[nonce]`. The helper raised `KeyError`; the miner caught it and submitted an unsigned attestation that the compatibility path accepted.

The node already normalizes nonce as `report.nonce` first and top-level `nonce` second. This patch gives the signer the same precedence.

## Security impact

With `PyNaCl` installed, Windows operators expected Ed25519 wallet binding, but current `main` logged only:

```text
attestation signing failed; falling through unsigned: attestation missing required field: 'nonce'
```

and continued successfully without `signature`, `public_key`, or `signature_type`.

## Validation

```text
python -m pytest -q tests/test_attestation_signing_6798.py --tb=short
10 passed in 1.84s

python -m py_compile miners/signing_helpers.py tests/test_attestation_signing_6798.py
passed

git diff --check
passed
```

Live official-node validation on real Windows hardware after the fix:

```text
accepted: true
signature_present: true
public_key_present: true
signature_type: ed25519
top_level_nonce_present: false
report_nonce_present: true
```

Fixes #6894

/claim #305
RTC wallet: `RTCf69dd944558d4e843a4a676495a97638055caea2`

Related but separate from #6868: that PR distributes the helper and improves fallback visibility; this PR corrects the helper's Windows-incompatible nonce lookup.

### DIFF
--- a/miners/signing_helpers.py
+++ b/miners/signing_helpers.py
@@ -34,8 +34,9 @@ def build_pipe_sign_message(attestation):
     try:
         miner_id = attestation["miner_id"]
         miner = attestation["miner"]
-        nonce = attestation["nonce"]
-        commitment = attestation["report"]["commitment"]
+        report = attestation["report"]
+        nonce = report.get("nonce") or attestation["nonce"]
+        commitment = report["commitment"]
     except (KeyError, TypeError) as exc:
         raise ValueError(f"attestation missing required field: {exc}") from exc
 
--- a/tests/test_attestation_signing_6798.py
+++ b/tests/test_attestation_signing_6798.py
@@ -13,12 +13,18 @@ on PR #6839).
 """
 import json
 import hashlib
+import importlib.util
 import sys
 import os
 import unittest
 
 # Add miners/ to path so we can import signing_helpers
-sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "miners"))
+ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+MINERS_DIR = os.path.join(ROOT, "miners")
+WINDOWS_MINER_PATH = os.path.join(
+    MINERS_DIR, "windows", "rustchain_windows_miner.py"
+)
+sys.path.insert(0, MINERS_DIR)
 
 from signing_helpers import build_pipe_sign_message
 
@@ -32,7 +38,7 @@ class TestAttestationSigningMessage(unittest.TestCase):
         return "{}|{}|{}|{}".format(
             attestation["miner_id"],
             attestation["miner"],
-            attestation["nonce"],
+            attestation["report"].get("nonce") or attestation.get("nonce"),
             attestation["report"]["commitment"],
         )
 
@@ -119,10 +125,28 @@ class TestAttestationSigningMessage(unittest.TestCase):
         msg2 = build_pipe_sign_message(att2)
         self.assertNotEqual(msg1, msg2)
 
+    def test_report_nonce_takes_precedence_like_node_verifier(self):
+        """Use report.nonce when the legacy top-level nonce differs."""
+        att = self._build_sample_attestation()
+        att["nonce"] = "stale-top-level-nonce"
+        self.assertEqual(
+            build_pipe_sign_message(att),
+            self._node_verifier_reconstruction(att).encode("utf-8"),
+        )
+
+    def test_legacy_top_level_nonce_fallback(self):
+        """Clients without report.nonce keep using the top-level nonce."""
+        att = self._build_sample_attestation()
+        del att["report"]["nonce"]
+        self.assertEqual(
+            build_pipe_sign_message(att),
+            self._node_verifier_reconstruction(att).encode("utf-8"),
+        )
+
     def test_pipe_delimiter_in_field_raises(self):
         """If any field contains a pipe character the builder must reject it."""
         att = self._build_sample_attestation()
-        att["nonce"] = "bad|nonce"
+        att["report"]["nonce"] = "bad|nonce"
         with self.assertRaises(ValueError):
             build_pipe_sign_message(att)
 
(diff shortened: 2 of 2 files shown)


jaxint (commented):
LGTM! Great work on this PR.

MolhamHamwi (changes requested):
Reviewed current head `45efd91` for the Windows report-nonce signing fix.

I’m requesting changes because the newly added regression test does not run under the repository’s current Python 3.9 test environment:

- `tests/test_attestation_signing_6798.py:166` imports `miners/windows/rustchain_windows_miner.py` directly, and that module contains PEP 604 annotations such as `dict | None` at `miners/windows/rustchain_windows_miner.py:217`. On Python 3.9 this raises `TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'` during class definition, before the test can exercise the nonce-signing path. Either gate the test to Python 3.10+, add `from __future__ import annotations` to the Windows miner module, or avoid importing the whole module in a Python-3.9 test.
- `miners/signing_helpers.py:38` uses `report.get("nonce") or attestation["nonce"]`. That fixes the Windows payload shape when `report.nonce` is present, but if `report.nonce` is intentionally present as an empty string the helper silently falls back to a stale top-level nonce. If empty nonce values are invalid, this should be rejected explicitly; if they are valid, the helper should check key presence rather than truthiness.

Verification run:
- `python3 -m pytest tests/test_attestation_signing_6798.py -q` → 9 passed, 1 failed (`test_windows_miner_signs_report_nonce`, import-time TypeError above).

jaxint (commented):
LGTM

avisii1680769-cell:
Reviewed Scottcjn/Rustchain#6895, focusing on `miners/signing_helpers.py`, `miners/windows/rustchain_windows_miner.py`, and `tests/test_attestation_signing_6798.py`.

Technical observations:

1. The nonce precedence change in `build_pipe_sign_message()` matches the node verifier path. The submit handler normalizes `report = _normalize_attestation_report(data.get('report'))`, then reconstructs the Ed25519 message with `nonce = report.get('nonce') or _attest_text(data.get('nonce'))`, so signing `report.nonce` first keeps the miner and verifier on the same byte string.

2. This fixes a real Windows miner shape issue, not just a helper mismatch. In `rustchain_windows_miner.py`, the challenge nonce is placed in `report_payload["nonce"]`, and the submitted `attestation` object does not include a top-level `nonce`. With the old helper, the Ed25519 branch could raise before adding `signature`, `public_key`, and `signature_type`.

3. The added Windows regression test is useful because it exercises the real `RustChainMiner.attest()` flow with stubbed challenge/submit calls and captures the actual submitted attestation. That is stronger than only asserting a helper output, because it verifies the submitted payload has `signature`, `signature_type=ed25519`, no top-level `nonce`, and the same pipe-delimited message the verifier reconstructs.

4. One edge case to keep in mind: `report.get("nonce") or attestation["nonce"]` preserves the verifier's current fallback semantics, including treating an empty report nonce as absent. That is consistent with the server today, but if empty `report.nonce` should be considered malformed rather than falling back, both signer and verifier would need a stricter explicit-presence check in a follow-up.

Validation I ran locally:

```text
python -m py_compile miners/signing_helpers.py tests/test_attestation_signing_6798.py
python tests/test_attestation_signing_6798.py
```

Result: 10 tests passed. `pytest -q tests/test_attestation_signing_6798.py --tb=short` could not run in this local environment because the repository-wide `tests/conftest.py` imports Flask, which is not installed here.

jaxint (commented):

jaxint (commented):
Thanks for the contribution!

jaxint (commented):

jaxint (commented):

jaxint (approved):
LGTM! Great work on this PR.

jaxint (commented):

jaxint (commented):
Nice work!

jaxint (commented):
Great work!

PR #6842, opened 2026-06-04
fix: bound pending confirmations and expose overdue state

yyswhsccc:
## BCOS Checklist (Required For Non-Doc PRs)

- [x] Add a tier label: `BCOS-L2`
- [x] If adding new code files, include SPDX header near the top (no new code files added)
- [x] Provide test evidence (commands + output below)

## What Changed

BCOS-L2: this touches wallet / transfer / pending confirmation behavior.

Related: Scottcjn/rustchain-bounties#13052

This PR does not resolve or close that payout check. It only uses it as a real-world example showing that pending RTC transfer state needs clearer overdue visibility and safer admin-side confirmation tooling.

I did not find a node startup loop or scheduler path that automatically invokes `/pending/confirm` for overdue `pending_ledger` rows. The existing path is an admin/worker endpoint plus an operator helper.

This patch keeps the confirmation rules and payout amounts unchanged, and does not embed admin secrets or manually credit any wallet. It:

- makes `/pending/confirm` run a bounded batch with a default/admin-provided limit while reusing the existing pending -> confirming -> balance update -> ledger -> confirmed semantics;
- rejects non-positive `/pending/confirm` limits with HTTP 400 so the API matches the operator CLI contract;
- returns stale pending counts and max overdue age before/after confirmation;
- adds per-row overdue age plus aggregate stale pending stats to `/pending/list`;
- enriches `/wallet/history` transfer rows with `status`, `raw_status`, `confirms_at`, `confirmed_at`, `confirm_overdue_seconds`, and `status_reason` where applicable;
- documents the always-present pending/history response fields for API consumers;
- adds `tools/pending_ops.py confirm --limit` so an operator/cron helper can run small idempotent batches with `RC_ADMIN_KEY`;
- keeps `/wallet/balances/all` as a full admin export rather than silently capping it at 500 rows;
- keeps confirmed `pending_ledger` transfer rows out of `/wallet/history` so settled transfers are not duplicated;
- keeps the fetchall CI guard passing for the new bounded history/confirmation queries and the existing full-export balance query.

## Testing / Evidence

- `.venv-bounty-validation/bin/python -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tools/pending_ops.py` -> passed
- `git diff --check upstream/main...HEAD` -> passed
- hidden Unicode scan over the changed files -> passed, 6 paths scanned, no findings
- `bash scripts/check_fetchall.sh` -> passed, legacy baseline count 174
[...]

PR #6900, opened 2026-06-06
feat(node): bound /pending/confirm + overdue observability (clean reimpl of #6842)

Scottcjn:
Reimplements the **good core** of #6842 (which I request-changed) on current main, **without** its two bundled regressions.

## Adds
- **Bounded `/pending/confirm`** — optional `?limit=`/JSON `limit` (default 100, max 500) → `LIMIT` on the ready-SELECT. One call confirms a bounded batch, not an unbounded sweep. Now that the hourly confirm scheduler (rustchain-bounties#13186) is live, this keeps each run small/predictable. **The SAVEPOINT-per-row balance logic is untouched.**
- **Overdue observability** — `/pending/confirm` + `/pending/list` report `stale_pending_count` + `max_confirm_overdue_seconds` (confirm also reports before/after + `selected_count`). Read-only.
- **Robust env parsing** — limit constants fail-safe to defaults on bad env instead of crashing import (the import-crash I flagged on #6842).

## Deliberately does NOT (the #6842 regressions — dropped)
- ❌ No cap on `/wallet/balances/all` (stays unbounded; no silent top-500)
- ❌ No change to `/wallet/history` confirmed-row skip (no duplicate-transfer risk)

## Verified
Compiles; only the node file changed (+ fetchall baseline line-shift resync); fetchall guard green (new `.fetchall()` annotated `already-paginated`); pending-ops + signed-transfer-replay + limit-validation tests pass. The 2 admin-401 tests that fail when batched fail identically on main (pre-existing rate-limiter state-bleed) and pass in isolation.

Supersedes #6842. Credit to @yyswhsccc for the original design.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -8922,6 +8922,55 @@ def wallet_transfer_v2():
         conn.close()
 
 
+def _pending_confirm_env_int(name, default):
+    """Parse an int env var without crashing import on a bad value."""
+    raw = os.environ.get(name, "")
+    if raw in (None, ""):
+        return default
+    try:
+        return max(1, int(raw))
+    except (TypeError, ValueError):
+        return default
+
+
+# Bounded /pending/confirm: a single call confirms at most this many transfers so
+# one run can't process an unbounded batch (the hourly scheduler keeps the queue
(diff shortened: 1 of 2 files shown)


jaxint (commented):

jaxint (commented):
Nice work!

jaxint (commented):
Great work!

PR #6903, opened 2026-06-06
consolidate: Windows signed-header production + multi-key header identity (#6896)

### LINKED ISSUE #6896: Windows miner stays ready but cannot produce signed headers

Scottcjn:
## Consolidates the Windows signed-header cluster → one reviewed, mergeable change

**Closes #6896. Supersedes #6897** (rebased onto current `main`, with the shared-wallet BLOCKING resolved). Folds together the genuine work from the whole cluster and credits each contributor.

### Cluster map (what this replaces)
| PR/Issue | Disposition |
|---|---|
| #6894 + #6895 (nonce signing) | Already merged — **kept** (`signing_helpers.py`) |
| #6896 + #6897 (producer flow) | **This PR** — producer flow + the BLOCKING fix |
| #6898 (rebel117) | Superseded by #6895; its destructive `.github/workflows` delete killed it |

### 1. Producer flow (from #6897, @vicentsmith470-web)
Windows miner queries `/lottery/eligibility` via `node_url` with the **attested wallet identity**, builds the signed-header envelope, and tracks accepted slots (`check_eligibility` / `generate_header` / `submit_header`). Node enroll paths register the chain identity for header verification.

### 2. Multi-key header identity — the shared-wallet BLOCKING fix
The prior `miner_header_keys` had a single-column `PRIMARY KEY(miner_id)`, so device enrollment was **last-writer-wins**: two devices sharing one payout wallet overwrote each other's header key, and the non-last device's signed headers then failed verification (latent today behind `RC_TESTNET_ALLOW_INLINE_PUBKEY`, fatal the moment that flag is closed — which is the secure end-state). Fixed:
- **Composite `PRIMARY KEY (miner_id, pubkey_hex)`** on all 3 schema sites
- **Idempotent `init_db()` migration** rebuilds a legacy single-PK table, preserving rows; **fails loud** (re-raises) rather than running on a half-migrated table
- **`ingest_signed` verifies the signature against ANY registered key** for the identity (verify-any) instead of one stored key
- Register endpoint **adds** keys (`ON CONFLICT(miner_id,pubkey_hex) DO NOTHING`) instead of overwriting

### 3. Tri-brain hardening (Codex + Grok)
- **Bounded keys per identity** (`RC_MAX_HEADER_KEYS_PER_IDENTITY`, default 8; evict oldest by rowid) — caps storage *and* per-header verify-any cost, ages out rotated keys.
- **Eviction is not an external vector** (verified): registration is authenticated — `/miner/headerkey` is admin-gated, enroll path is owner-signature-verified — so a third party can't register keys for an identity they don't control. Only self-eviction (>cap own devices) is possible; raise the env cap for that. Explicit per-key revocation is a noted follow-up.
- Migration broad-except → **fail-loud**.

### Verification
- `py_compile` clean
- `scripts/check_fetchall.sh` **green** — baseline regenerated content-identical (2 new sites annotated `pragma-result` / `bounded-by-schema`)
- **New `tests/test_header_keys_composite_6897.py` (4 pass):** migration idempotence + row preservation, two-device key retention, bounded eviction keeps newest, verify-any accepts any device key **and rejects a stranger's signature**
- Existing header/enroll test failure-set unchanged vs `main` (the 3 reds are pre-existing bare-clone DB-fixture gaps, green in CI)
- Mechanical: regenerated fetchall baseline + both Windows-miner SHA256 pins (`rustchain_miner_setup.bat`, `setup_miner.py`)

### Credits
- **@vicentsmith470-web** — producer flow (#6897) + nonce-signing fix (#6895)
- **@MolhamHamwi** — review catches that shaped this: PEP 604/Py3.9 breakage (#6895) and the destructive-delete catch (#6898)
- **@rebel117** — #6898 nonce-lookup attempt


jaxint (commented):

jaxint (commented):
Nice work!

jaxint (commented):
Great work!

PR #6905, opened 2026-06-06
fix(migrate): column-preserving miner_header_keys composite migration (prod-deploy blocker)

Scottcjn:
## Make the `miner_header_keys` composite migration column-preserving (deploy-blocker for #6903)

Caught during pre-deploy recon of the production node (Step 2 of the prod-deploy prep).

### The bug
The migration merged in #6903 rebuilt `miner_header_keys` with a **hardcoded 2-column** shape:
```python
CREATE TABLE miner_header_keys(miner_id ..., pubkey_hex ..., PRIMARY KEY (miner_id, pubkey_hex))
INSERT ... SELECT miner_id, pubkey_hex FROM ...legacy
```
The **production** `miner_header_keys` carries a third column the code's `CREATE` doesn't have:
```sql
added_at INTEGER NOT NULL DEFAULT (strftime('%s','now'))   -- 247,619 rows, all populated
```
So the merged migration would have **silently dropped `added_at`** (column + all data) the first time the node restarted on prod. (It's unused by code today, but dropping a column from a live consensus table is unacceptable silent data loss.)

### The fix
The rebuild now reconstructs **every existing column** from `PRAGMA table_info` — preserving type, `NOT NULL`, and `DEFAULT` (defaults are always parenthesized so expression defaults like `(strftime('%s','now'))`, which PRAGMA returns *without* the parens, survive) — and copies all columns. Composite PK + verify-any behavior unchanged. Still idempotent + fail-loud.

### Validation — dry-run against the real prod DB
A `.backup` snapshot of the live 268 MB DB (live DB never touched) → ran this migration on the copy:
```
BEFORE  count/with_added_at/min/max: (247619, 247619, 1759701180, 1780774414)
migrate : migrated
AFTER   count/with_added_at/min/max: (247619, 247619, 1759701180, 1780774414)
composite PK     : ['miner_id', 'pubkey_hex']
counts identical : True    sample identical : True    idempotent 2nd: noop
new schema       : ...("added_at" INTEGER NOT NULL DEFAULT (strftime('%s','now')), PRIMARY KEY (miner_id, pubkey_hex))
```
All rows + every `added_at` value preserved.

### Tests
`test_migration_preserves_extra_columns_like_production_added_at` mirrors the exact prod DDL and asserts `added_at` (column + values) survives the rebuild and its default still fires for newly-registered keys. 5/5 in the suite pass; `_migrate` test helper updated to the column-preserving logic. `py_compile` + fetchall guard green (baseline content-identical).

This unblocks a safe prod deployment of the multi-key header-identity work.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -1810,25 +1810,41 @@ def init_db():
         # (miner_id, pubkey_hex) key so one wallet identity can hold a header key per
         # device (multi-device-per-wallet) instead of last-writer-wins overwriting
         # — which previously broke the non-last device's signed-header verification.
-        # Idempotent: only rebuilds a table still carrying the old PRIMARY KEY(miner_id).
+        # COLUMN-PRESERVING: production tables may carry extra columns (e.g. an
+        # `added_at` timestamp) absent from the CREATE above; the rebuild reconstructs
+        # every existing column (type / NOT NULL / DEFAULT) rather than a fixed 2-col
+        # shape, so no column or data is dropped. Idempotent: only rebuilds a table
+        # still carrying the old single-column PRIMARY KEY(miner_id). Verified by a
+        # dry-run against a 247,619-row snapshot of the production DB.
         try:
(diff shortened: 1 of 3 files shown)


jaxint (commented):

jaxint (commented):
Nice work!

jaxint (commented):
Great work!

PR #6908, opened 2026-06-07
fix(sophia): guard _ensure_balance_micro_schema from wiping the consensus balances ledger

Scottcjn:
## Guard `_ensure_balance_micro_schema` against wiping the consensus balances ledger

Found during the **Step-3 pre-deploy migration audit** (dry-running every main migration against a snapshot of the production DB).

### The landmine
`sophia_elya_service.py` sets `DB_PATH = "./rustchain_v2.db"` (relative) and, when run as `__main__`, calls `_ensure_balance_micro_schema()`. For a non-INTEGER `balance_rtc` that function does:
```python
ALTER TABLE balances RENAME TO balances_legacy_real
CREATE TABLE balances (miner_pk TEXT PRIMARY KEY, balance_rtc INTEGER DEFAULT 0)   # 2 columns
INSERT ... SELECT miner_pk, balance_rtc*MICRO FROM balances_legacy_real
DROP TABLE balances_legacy_real
```
But the **consensus** `balances` ledger is:
```sql
balances (miner_id TEXT PRIMARY KEY, amount_i64 INTEGER ... CHECK(amount_i64>=0),
          miner_pk TEXT, balance_rtc REAL DEFAULT 0.0, coinbase_address TEXT)
```
`balance_rtc` is **REAL** (not INTEGER) → the rebuild **would fire**, and it would **drop `miner_id`/`amount_i64`/`coinbase_address` and wipe every balance** (it repopulates from the typically-zero `balance_rtc` REAL column, not the canonical `amount_i64`).

### Current exposure
**Not currently triggered** — recon confirmed no systemd service or cron runs `sophia_elya_service.py`, and prod's Oct-2025 copy doesn't even contain this migration (it's new in `main`). But running `python3 sophia_elya_service.py` from the node dir would detonate it.

### The fix
Add a safety guard: if `balances` carries the consensus money columns (`amount_i64` or `coinbase_address`), **leave the table completely untouched**. The genuine Sophia 2-column micro-schema migration path is unchanged.

### Tests
- consensus-shaped `balances` → byte-identical after (amount_i64 preserved, no rows lost)
- genuine Sophia `balances` → still migrates `REAL → INTEGER` micro (2.5 → 2,500,000)

`py_compile` + fetchall guard green (baseline content-identical).

### Step-3 audit summary (context)
Node-startup migrations were dry-run against a `.backup` of the live 268 MB DB: **70 tables → 70, 1,590,693 rows → 1,590,693**, only `epoch_enroll` (weight REAL→INT) and `miner_header_keys` (composite PK, #6905) changed — both intended, no data lost. This PR closes the one catastrophic landmine that audit surfaced.

### DIFF
--- a/node/sophia_elya_service.py
+++ b/node/sophia_elya_service.py
@@ -65,6 +65,17 @@ def _ensure_balance_micro_schema(conn):
         return
 
     by_name = {row[1]: row for row in columns}
+
+    # SAFETY GUARD: never rebuild the CONSENSUS balances ledger. The RustChain node
+    # keys `balances` by `miner_id` with the canonical micro-RTC amount in `amount_i64`
+    # (plus miner_pk / coinbase_address). This Sophia helper only manages its own
+    # 2-column (miner_pk, balance_rtc) micro-schema. If `DB_PATH` ever resolves to the
+    # shared consensus DB (it is a relative "./rustchain_v2.db"), the rebuild below
+    # would DROP miner_id / amount_i64 / coinbase_address and WIPE every balance.
+    # If we see the consensus money columns, leave the table completely untouched.
+    if "amount_i64" in by_name or "coinbase_address" in by_name:
+        return
+
     balance_column = by_name.get("balance_rtc")
     if balance_column and "INT" in (balance_column[2] or "").upper():
         return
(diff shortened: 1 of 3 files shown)


jaxint (commented):
Nice work!

jaxint (commented):
Great work!

PR #6546, opened 2026-05-28
fix: sign macOS miner headers with wallet key

jonathanpopham:
## Summary

Fixes #6545.

The macOS miner v2.5 could attest successfully, but production header submission still used a SHA-512 placeholder signature and sent the RTC wallet address as `pubkey`. Production `/headers/ingest_signed` has mock signatures and inline pubkeys disabled, so the server expects a registered Ed25519 header key and a real Ed25519 signature.

This PR makes the macOS miner production-signing capable:

- auto-loads a matching RTC wallet JSON from `~/.rustchain/wallets/*.json`, `RUSTCHAIN_WALLET_FILE`, or `--wallet-file`
- supports raw 32-byte hex seeds and the existing PKCS#8 DER/PEM wallet format
- verifies the loaded public key maps to the configured RTC address before use
- signs `/attest/submit` binding messages and includes `public_key`, allowing the node to register `miner_header_keys`
- signs `/headers/ingest_signed` messages with Ed25519 instead of the legacy SHA-512 placeholder
- keeps the old placeholder path only as a fallback for non-RTC/legacy deployments without a loaded signing key

## Verification

- `PYTHONPATH` not required: `/tmp/rustchain6537-py311-venv/bin/python -m compileall -q miners/macos/rustchain_mac_miner_v2.5.py miners/macos/test_rustchain_mac_miner_signing.py`
- `/tmp/rustchain6537-py311-venv/bin/python -m unittest miners.macos.test_rustchain_mac_miner_signing` -> `2 tests OK`
- `git diff --cached --check` was clean before commit

## Notes

I also live-tested the local client path with my own wallet: the miner loaded the Ed25519 wallet key, submitted a signed attestation, and the node accepted it with the hardware fingerprint passing. No private key material is logged or included here.

Payout address if accepted: `RTC4730a64f821a590972e8b37781fae5d568b5865c`.

PR #7022, opened 2026-06-08
fix: admin auth on /api/chat, /api/bridge/lock, and /bft/propose (RIP-0202)

BossChaos:
## Summary

Fixes three unauthenticated admin endpoints that should require `RC_ADMIN_KEY` authentication:

### Fix 1: `POST /api/chat` (node/beacon_api.py:1279)
- **Impact**: Anyone can send messages to any beacon agent, pollute chat history, and abuse LLM API calls
- **Fix**: Added `_require_admin_key()` check after input validation (agent_id + message type check)

### Fix 2: `POST /api/bridge/lock` (node/airdrop_v2.py:1477)
- **Impact**: Lock bridge tokens without admin authorization
- **Fix**: Added `_require_admin_key()` check after missing_required_fields validation

### Fix 3: `POST /bft/propose` (node/rustchain_bft_consensus.py)
- **Impact**: Manually trigger epoch proposal (admin action per docstring) without authentication
- **Fix**: Added `_require_admin_key()` check after input validation (epoch/miners/distribution type checks)

### Auth Pattern
All three use the same pattern as existing bridge admin endpoints:
- `RC_ADMIN_KEY` environment variable as the secret
- `X-Admin-Key` or `X-API-Key` request header
- `hmac.compare_digest` for timing-safe comparison
- **Auth placed AFTER validation** to preserve existing 400-error test paths

### Test Coverage
- All existing 400-validation tests remain unchanged (auth does not interfere with bad-input paths)
- 3 new auth tests for `/bft/propose`: no-key → 401, wrong-key → 401, correct-key → 400

Fixes: #13408
Bounty: https://github.com/Scottcjn/rustchain-bounties/issues/13408

### DIFF
--- a/node/airdrop_v2.py
+++ b/node/airdrop_v2.py
@@ -1474,6 +1474,12 @@ def init_airdrop_routes(app, airdrop: AirdropV2, db_path: str) -> None:
                 400,
             )
 
+        # ── Admin auth: gate bridge lock creation ─────────────────────────────
+        auth_error = require_admin_key()
+        if auth_error:
+            return auth_error
+        # ── Auth passed ──────────────────────────────────────────────────────
+
         amount_uwrtc = int(round(amount_wrtc * 1_000_000))
 
         success, message, lock = airdrop.create_bridge_lock(
--- a/node/beacon_api.py
+++ b/node/beacon_api.py
@@ -1276,6 +1276,19 @@ def chat():
 
         if not agent_id or not message:
             return jsonify({'error': 'Missing agent_id or message'}), 400
+
+        # ── Admin auth: prevent unauthorized beacon chat usage ───────────────
+        expected_key = os.environ.get('RC_ADMIN_KEY', '').strip()
+        if not expected_key:
+            return jsonify({'error': 'admin_key_not_configured'}), 503
+        provided_key = (
+            request.headers.get('X-Admin-Key', '')
+            or request.headers.get('X-API-Key', '')
+        ).strip()
+        if not hmac.compare_digest(provided_key, expected_key):
+            return jsonify({'error': 'invalid_admin_key'}), 403
+        # ── Auth passed ──────────────────────────────────────────────────────
+
         safe_agent_id = html.escape(str(agent_id), quote=True)
         safe_message = html.escape(str(message), quote=True)
 
--- a/node/rustchain_bft_consensus.py
+++ b/node/rustchain_bft_consensus.py
@@ -29,6 +29,7 @@ from dataclasses import dataclass, asdict
 from enum import Enum
 from typing import Dict, Iterable, List, Optional, Set, Tuple
 import requests
+import os
 
 # Configure logging
 logging.basicConfig(level=logging.INFO, format='%(asctime)s [BFT] %(message)s')
@@ -1037,6 +1038,20 @@ def create_bft_routes(app, bft: BFTConsensus):
     def _internal_error_response(message: str, status_code: int):
         return jsonify({'error': message}), status_code
 
+    def _require_admin_key():
+        """Require admin key for sensitive BFT admin endpoints."""
+        required = os.environ.get("RC_ADMIN_KEY", "").strip()
+        if not required:
+            return jsonify({"ok": False, "error": "admin_key_not_configured"}), 503
+        provided = (
+            request.headers.get("X-Admin-Key")
+            or request.headers.get("X-API-Key")
+            or ""
+        ).strip()
+        if not hmac.compare_digest(provided, required):
+            return jsonify({"ok": False, "error": "unauthorized"}), 401
+        return None
+
     @app.route('/bft/status', methods=['GET'])
     def bft_status():
         """Get BFT consensus status"""
@@ -1109,6 +1124,10 @@ def create_bft_routes(app, bft: BFTConsensus):
                 return jsonify({'error': 'miners must be a list'}), 400
             if not isinstance(distribution, dict):
                 return jsonify({'error': 'distribution must be an object'}), 400
+
+            auth_err = _require_admin_key()
+            if auth_err:
+                return auth_err
 
             msg = bft.propose_epoch_settlement(epoch, miners, distribution)
             if msg:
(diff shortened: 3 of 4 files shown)


jdjioe5-cpu (approved):
## Code review — PR #7022 (Bounty #73, RIP-0202 admin auth on `/api/chat`, `/api/bridge/lock`, `/bft/propose`)

Reviewed head `f518240` (`fix(bft): require admin auth after validation on /bft/propose`) on top of `origin/main` `3465d16`.

**Substantive findings (line-level):**

1. **Duplicate auth helper — recommend extracting to a shared module** (medium). The 14-line `require_admin_key()` body now lives in two places:
   - `node/airdrop_v2.py:1271-1281` (existing helper, called by the new code at `1478`, `1497`, `1521`)
   - `node/rustchain_bft_consensus.py:1041-1052` (new helper named `_require_admin_key`, identical body)
   They differ only by the leading underscore prefix in the second file. A future refactor to `node/auth.py` (or similar) would prevent drift — e.g. when one switches from `X-Admin-Key`/`X-API-Key` to also accepting `Authorization: Bearer …`, the other will be silently left behind. Not blocking for this PR, but worth a follow-up issue.

2. **Naming inconsistency** (nit, low). `_require_admin_key` (underscore prefix) in `rustchain_bft_consensus.py` vs. `require_admin_key` (no prefix) in `airdrop_v2.py` — the rest of `rustchain_bft_consensus.py` uses unprefixed route handlers (e.g. `bft_status`, `bft_message`, `_internal_error_response` is the only underscored helper). Suggest dropping the underscore for consistency.

3. **Test coverage gap on the new auth paths** (medium). The 3 new tests in `node/tests/test_bft_routes.py` only cover the `/bft/propose` branch. There is no test asserting that `/api/chat` (the new auth block in `beacon_api.py:1279-1290`) actually returns 401/403 when unauthenticated, and no test for the new `airdrop_v2.py:1477-1481` auth block on `/api/bridge/lock` POST. The previous `test_chat_*` tests in `tests/test_beacon_api.py` (if any) need a re-read to confirm they still pass with the new auth gate. This is the highest-value concrete suggestion: add at least one test per auth-gated route that exercises the missing-key case.

4. **Header-name list is frozen to two values** (low). The auth helper checks `X-Admin-Key` then `X-API-Key`, in that order. A `Bearer` token in `Authorization` is not accepted. If the maintainer wants gradual rollout, fine; otherwise document the supported header set in the route docstring.

5. **`admin_key_not_configured` returns 503, not 500** (informational). That's the right choice (a missing env var on a deployed node is a service-misconfiguration, not an internal error), but `503` is conventionally `Service Unavailable`; consider 500 with a clear message, or 503 with a `Retry-After` header to signal this is recoverable. Either is acceptable; just flagging for consistency with the rest of the API.

**What this PR gets right (do not change in this PR):**

- The check is **after** the existing `if not agent_id or not message` shape validation in `beacon_api.py:1277` — so a malformed request still 400s before leaking auth-state info. Good ordering.
- `hmac.compare_digest` is used for the string compare — correct choice, avoids timing attacks.
- The `RC_ADMIN_KEY` env var is `.strip()`'d before both read and compare, so a trailing newline in the env doesn't silently break auth.
- The new tests are clean, self-contained, and use `monkeypatch.setenv` properly. The `test_bft_propose_accepts_valid_admin_key` test correctly asserts the business-logic error (`not_leader_or_already_committed`) rather than a 200, which is the right behavior for the fake `_FakeBft.propose_epoch_settlement` stub.

**Local validation (fresh `git clone` of `pr-7022` head on `3465d16`):**

```
$ .venv/bin/python -m pytest node/tests/test_bft_routes.py -v
…
node/tests/test_bft_routes.py::test_bft_propose_requires_json_object PASSED
node/tests/test_bft_routes.py::test_bft_propose_requires_epoch_field PASSED
node/tests/test_bft_routes.py::test_bft_propose_rejects_non_integer_epoch[True] PASSED
node/tests/test_bft_routes.py::test_bft_propose_rejects_non_integer_epoch[1] PASSED
node/tests/test_bft_routes.py::test_bft_propose_rejects_non_integer_epoch[1.5] PASSED
node/tests/test_bft_routes.py::test_bft_propose_rejects_negative_epoch PASSED
node/tests/test_bft_routes.py::test_bft_propose_rejects_invalid_collection_fields PASSED
node/tests/test_bft_routes.py::test_bft_propose_requires_admin_key PASSED
node/tests/test_bft_routes.py::test_bft_propose_requires_correct_admin_key PASSED
node/tests/test_bft_routes.py::test_bft_propose_accepts_valid_admin_key PASSED
========== 10 passed in 0.36s ==========

$ git merge-tree --write-tree origin/main HEAD
b6b3c818d6611c7c027baa9eb599f4a084ace6bc    (clean three-way merge)

$ git diff --check origin/main..HEAD
(empty = clean)

$ git diff --shortstat origin/main..HEAD
4 files changed, 74 insertions(+)
```

**Verdict:** This is a clean, security-positive change. The fixes for `/api/chat` and `/bft/propose` are correct; the `/api/bridge/lock` POST fix is correct. The host PR `test` job failure (`runs/27113345752/job/80015315776`) on CI looks like a pre-existing repo-wide baseline (the same one blocking #6868 and #6546 in the merged-through-merge train), not caused by this diff — the new tests pass locally on a fresh clone. **Approve** with the small test-coverage gap and the helper-extraction follow-up noted above.

Scottcjn:
Thanks @BossChaos — strong security PR, and most of it I want to merge:

- ✅ `/api/bridge/lock` admin gate — correct (financial)
- ✅ BFT admin endpoints (`rustchain_bft_consensus.py`) — correct
- ✅ `airdrop_v2` admin gate — correct

One concern blocks an immediate merge: it also gates **beacon `/api/chat`** (`agent_id` + `message`) behind the admin key. If beacon chat is a **public agent-facing** feature, this makes it admin-only and breaks it for every agent (only the operator could chat). 

Could you confirm: is locking down `/api/chat` intended (was it being abused), or should that hunk be dropped so chat stays public? If you split the `beacon_api.py` chat hunk out, I'll merge the rest right away and pay. Holding open pending that. — Sophia

Scottcjn:
Thanks @BossChaos — strong security PR. I shipped most of it as #7101 (merged): admin gates on `create_bridge_lock` (financial) and `bft_propose` (epoch-settlement consensus), both genuinely sensitive. **10 RTC** sent to `BossChaos`.

The one piece I left out: gating beacon **`/api/chat`** with the admin key. That endpoint is a public agent-facing feature we want to *build on*, so locking it admin-only would break it for agents — Scott's call was to keep it open. Everything else is in. Tribrain note: your code was clean (a flagged 'missing hmac import' was a false positive — it's imported at L22). Credit + reward are yours. — Sophia

PR #7049, opened 2026-06-08
Fix admin UI session links

### LINKED ISSUE #7048: [Bug] Admin UI session links render empty or invalid session_id
### Summary

The wallet review admin UI session flow is broken after the query-string admin-key leak was replaced with session links.

Two related problems make the HTML admin navigation unusable without repeatedly re-sending the admin key/header:

1. `GET /admin/ui` can create a valid `_ADMIN_SESSIONS` entry from `X-Admin-Key`, but the route renders a template that references `{{ sid }}` without passing `sid` into `render_template_string`. The Wallet Review Holds link therefore renders with an empty `session_id`.
2. When a later request uses an existing valid `session_id`, `_get_or_create_admin_session()` refreshes the session but does not attach that same id back to the Flask request. The wallet review page then falls back to a fresh random id that is not registered in `_ADMIN_SESSIONS`, so the rendered links/forms point at an invalid session.

### Steps to reproduce

1. Configure `RC_ADMIN_KEY`.
2. Request the admin landing page with header auth, for example `GET /admin/ui` with `X-Admin-Key: <RC_ADMIN_KEY>`.
3. Inspect the rendered Wallet Review Holds link.
4. Follow that link without sending the admin header again.

### Actual behavior

The landing page link renders as `/admin/wallet-review-holds/ui?session_id=` because `sid` is undefined in the template context. Following the link returns `403 forbidden` because there is no valid session id and no header/query admin key.

For existing sessions, the page can also render links/forms with a fresh unregistered `session_id`, which breaks subsequent navigation or POST redirects.

### Expected behavior

A successful admin UI login should render a non-empty session id that is present in `_ADMIN_SESSIONS`, and all follow-up admin UI links/forms should preserve that same valid session id until the session expires.

### Impact

This is a real web-interface workflow bug: header-based admin access succeeds for the first page, but the operator cannot navigate to or continue using the wallet review admin UI through the rendered links. It also weakens the post-#5216 remediation because operators are pushed back toward the legacy `?admin_key=` path to keep the UI usable.

Suggested #305 bounty tier: Medium / 10 RTC if treated as a broken admin web-interface flow; otherwise Low / 5 RTC.

Payout handle: `github:pqmfei` (I can provide a native RTC wallet if required).

### Fix validation

I have a small fix ready that:

- passes the minted `sid` into `/admin/ui` rendering,
- attaches existing valid session ids back to `request._admin_session_id`,
- mints a session for valid legacy query/form admin-key access so subsequent links can stop carrying the key,
- adds regression coverage for header login and existing-session reuse.

Local focused test result:

```text
python -m pytest tests/test_wallet_review_holds.py -q
11 passed in 3.04s
```

pqmfei:
## Summary

Fix the wallet review admin UI session propagation introduced by the session-link remediation for query-string admin keys.

- Preserve valid existing `session_id` values on the request so rendered links/forms keep using the registered session.
- Pass the minted `sid` into the `/admin/ui` template context so header-authenticated admin landing pages render a usable Wallet Review Holds link.
- Mint a session when the legacy query/form admin-key path succeeds, so follow-up links can use session ids instead of carrying the admin key.
- Add regression tests for header-auth navigation and existing-session reuse.

Fixes #7048.
Contributes to bug bounty #305.

## Tests

```text
python -m pytest tests/test_wallet_review_holds.py -q
11 passed in 3.04s
```

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -5453,13 +5453,14 @@ _ADMIN_SESSION_TTL = 3600
 def _get_or_create_admin_session(req):
     now = time.time()
     # Clean expired
-    expired = [k for k, v in _ADMIN_SESSIONS.items() if now - v > 3600]
+    expired = [k for k, v in _ADMIN_SESSIONS.items() if now - v > _ADMIN_SESSION_TTL]
     for k in expired:
         _ADMIN_SESSIONS.pop(k, None)
     # Check existing session
     sid = req.values.get("session_id", "")
     if sid and sid in _ADMIN_SESSIONS:
         _ADMIN_SESSIONS[sid] = now  # refresh TTL
+        req._admin_session_id = sid  # type: ignore[attr-defined]
         return True
     # Check header auth
     if is_admin(req):
@@ -5485,7 +5486,12 @@ def _wallet_review_ui_authorized(req):
         or req.form.get("admin_key")
         or ""
     ).strip()
-    return bool(need and got and hmac.compare_digest(need, got))
+    if need and got and hmac.compare_digest(need, got):
+        sid = secrets.token_hex(16)
+        _ADMIN_SESSIONS[sid] = time.time()
+        req._admin_session_id = sid  # type: ignore[attr-defined]
+        return True
+    return False
 
 
 def get_wallet_review_counts():
@@ -5712,6 +5718,7 @@ def admin_operator_ui():
         return jsonify({"ok": False, "error": "forbidden"}), 403
 
     admin_key = str(request.values.get("admin_key") or "").strip()
+    sid = getattr(request, '_admin_session_id', '')
     counts = get_wallet_review_counts()
     return render_template_string(
         """
@@ -5764,6 +5771,7 @@ def admin_operator_ui():
 </html>
         """,
         admin_key=admin_key,
+        sid=sid,
         counts=counts,
     )
 
--- a/tests/test_wallet_review_holds.py
+++ b/tests/test_wallet_review_holds.py
@@ -1,3 +1,4 @@
+import re
 import sqlite3
 import sys
 import uuid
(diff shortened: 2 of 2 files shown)


jaxint (commented):
Great work on this PR! Thanks for contributing to RustChain ecosystem. 🚀

jdjioe5-cpu (approved):
## Review of PR #7049 — Fix admin UI session links

**Head reviewed:** `a08e1a5` (PR #7049, base `3465d16` = current `origin/main`).
**Repo:** `Scottcjn/Rustchain` (Bounty #73 eligible).
**Files:** 2 files, +59/-2.

**Local validation run on a fresh clone at the exact head:**

```text
python3 -m pytest tests/test_wallet_review_holds.py -q
...........     [100%]
11 passed in 1.50s

bef32b310a98127bbba274bcee3e2db03e36cb1f   (clean tree)

(no output — clean)
```

The PR is a focused, well-scoped follow-up to the wallet-review admin UI session
propagation bug from the query-string admin-key remediation. It correctly:
1. Preserves a valid existing `session_id` on the request (`req._admin_session_id = sid`),
   so rendered links/forms keep using the registered session rather than re-asking for
   a session every navigation.
2. Mints a fresh `sid` (16-byte `secrets.token_hex`) when the legacy query/form
   `admin_key` path succeeds, so follow-up links can use session ids instead of
   carrying the admin key across URLs.
3. Passes the minted `sid` into the `/admin/ui` template context so header-authenticated
   admin landing pages render a usable Wallet Review Holds link.
4. Adds 2 focused regression tests in `tests/test_wallet_review_holds.py` (49 new lines).

### Findings

**1. (medium, correctness) `_wallet_review_ui_authorized` is a documented
side-effectful authorization check but is now also a session-minter.**

`hmac.compare_digest(need, got)` was a pure boolean check; the new code mints
`_ADMIN_SESSIONS[sid]` as a side effect whenever authorization passes. This is
not wrong (it is documented in the PR description and tested), but it shifts
`_wallet_review_ui_authorized` from "predicate" to "predicate + state mutator".
The name no longer communicates the side effect. Suggested rename:
`_authorize_and_mint_admin_session(req)` (or add a one-line docstring that
states "may mint a new session; updates `req._admin_session_id`"). Pure nit;
not blocking.

**2. (low) The two new tests cover header-auth navigation and existing-session
reuse but do not cover the legacy query-string `admin_key` path.**

The new mint-on-`admin_key`-success branch is exercised in code (and the existing
`test_admin_operator_ui_links_to_wallet_review_surface` covers the rendered link),
but there is no direct test for the "user pastes `?admin_key=…` and we mint a
session, then the follow-up `/admin/wallet-review-holds/ui?session_id=…` works
with that minted sid" path. A one-test addition would close the loop and pin
the documented contract. Optional.

**3. (low) `req._admin_session_id = sid  # type: ignore[attr-defined]` is
necessary but is dynamic-attribute assignment on a werkzeug Request.**

This is a pattern wart — the rest of the codebase uses `request.values`,
`request.args`, etc., and assigning a custom attribute on `req` survives
within the request lifecycle, but it does not survive a `getattr(request,
'_admin_session_id', '')` round-trip in the same request if werkzeug ever
wraps the request object. Today the flow is `req → request` (same object in
flask, request is a thread-local proxy pointing to the same Request instance),
so the `getattr` on line 5719 picks it up — but the contract is implicit. A
minimal `flask.g.admin_session_id = sid` would be the more idiomatic flask
solution and would remove the `# type: ignore` comment entirely. Optional.

**4. (informational) Bounty #305 contribution and #7048 fix are both
correctly cross-referenced in the PR body.**

Both the bug it fixes (`Fixes #7048`) and the parent bounty (`Contributes to
bug bounty #305`) are listed. The 11/11 test run listed in the body matches
the local re-validation. `git merge-tree` clean and `git diff --check` clean.

### Summary

Approve. The PR is a focused, well-tested fix for a real admin-UI session
propagation regression. The medium nit is a naming/style concern, not a
correctness blocker. The two optional follow-ups would tighten the test
matrix and replace the dynamic-attribute wart with `flask.g`, but neither is
required for the bug fix to be correct. The author has already addressed the
session-id reuse and header-auth navigation surfaces that the original
remediation missed.

**Verdict: APPROVE**

— Hermes bounty-ops review for Bounty #73.

PR #6904, opened 2026-06-06
fix(security): guard miner_header_keys against header-key takeover

Scottcjn:
## Summary
Closes the **header-key takeover (T1.1)** on the RustChain RIP-200 node. `miner_header_keys` is the block-header trust anchor (a header verifies against ANY registered key for an identity); attestation/enroll inputs are caller-controlled and the signature only proves key-holdership, not identity ownership. This PR makes header-key registration **authenticated AND authorized**.

Hardened over 4 tri-brain rounds (Codex security · GPT-OSS coherence · Grok regression). 14/14 tests; `py_compile` clean.

## Final design
- **Always-on:** an attacker can no longer ADD a key to an already-keyed identity (the severe live takeover). Fleet-safe — the live chain has **zero** identities with >1 header key.
- **Self-authenticating identities** (RTC address = `address_from_pubkey(pubkey)`, or identity == raw pubkey): self-bootstrap and rotate freely.
- **Named/legacy identities** (g4/g5/power8, founder_*): first-key bootstrap gated by **`RC_HEADER_KEY_STRICT_BOOTSTRAP`** (default OFF → staged rollout). Under strict, the first key must be in the **admin-only** `miner_header_bootstrap` allowlist (seeded via the RC_ADMIN_KEY-gated `/miner/headerkey`).
- **One-time** (schema-v18-guarded, atomic) backfill grandfathers keys present at upgrade — no every-startup re-absorption of rollout-window keys.
- **Revocation:** `POST /miner/headerkey {"action":"revoke"}` deletes from both tables (retirement path; unknown actions are rejected 400).
- Rejected registration is **non-fatal** (miner keeps attesting/mining; logged).

## Deploy runbook
1. Deploy with `RC_HEADER_KEY_STRICT_BOOTSTRAP=0` (default) — only the always-on add-protection changes behavior (fleet-safe).
2. v18 backfill logs how many existing keys it grandfathered + an **audit warning**.
3. **AUDIT** existing keys (the pre-fix live vuln may have injected some) and `action=revoke` any suspect ones. *(Existing-key remediation — separate from the forward-fix this PR provides.)*
4. Admin-seed any named producers not already keyed.
5. Flip `RC_HEADER_KEY_STRICT_BOOTSTRAP=1` — forward TOFU bootstrap closed.

## Documented residual (deploy-policy, not a code defect)
[...]

PR #7061, opened 2026-06-08
fix(consensus/T2.1): bind attestation challenge nonces to requesting identity

Scottcjn:
## T2.1 — Challenge nonce not miner-bound + unsigned attestation

Part of the consensus stabilization roadmap (`CONSENSUS_TRIAGE.md`, PR #2). HIGH severity, external attack surface. **Completes the T1.1 story** (#6904 closed the header-key add-path; this closes the challenge-nonce gap that feeds the impersonation).

### The gap
`/attest/challenge` issued a **free-floating nonce bound to nobody** — any claimed `miner` could consume any live nonce. A challenge harvested or MITM'd from one miner could be replayed under a different identity, feeding the header-key impersonation surface.

### The fix (backward-compatible)
- **`nonces` gains an additive `bound_miner` column** — migrates existing DBs in place; the `ALTER` is idempotent and **race/lock-tolerant** (on any `OperationalError` it re-checks the actual schema and only re-raises if the column is genuinely absent). `init_db` ensures it at startup, so the hot path normally never migrates.
- **`/attest/challenge` accepts an optional `miner_id`** and binds the nonce to it. The identity extraction order **matches the submit path** (`miner` → `miner_id`) so a client where `miner != miner_id` binds to exactly the identity that gets enforced (no legit lockout). The response advertises `bound_miner`.
- **`attest_validate_challenge` rejects a BOUND nonce for any non-matching consumer** — a mismatched *or missing/empty* identity is rejected **without consuming the nonce**, so an attacker can't burn a rightful miner's live challenge (no DoS) and a harvested nonce can't be replayed under another identity → **403 `NONCE_IDENTITY_MISMATCH`**.
- **Unbound (NULL) nonces stay consumable by anyone** → backward-compatible for legacy miners that send no identity.

### Why no blanket signature requirement
Deliberately **not** requiring signatures on all attestations — that would brick vintage named wallets (G4/G5 PowerBooks). The unsigned → header-key-registration vector is already closed by #6904's `_header_key_authorized` (self-auth `address_from_pubkey(pubkey)==identity` + admin allowlist for named identities). Binding raises the bar on nonce reuse without breaking the honest vintage fleet.

### Status semantics
`NONCE_IDENTITY_MISMATCH` → **403** (authorization: request a challenge with the correct `miner_id`), distinct from the existing **409** for stale/replayed nonces. The added `bound_miner` response field is additive (unknown-field-tolerant clients unaffected).

### Tests (23, all green)
- `node/tests/test_attest_nonce_identity_binding_t21.py` — function-level: column migration (incl. legacy 2-col table in place), idempotent double-ensure, unbound consumable by any miner, bound consumed by match, **bound rejected for other miner + preserved for owner**, bound rejected when no identity supplied, lower-level `attest_validate_challenge` mismatch-no-delete, replay.
- `node/tests/test_attest_submit_challenge_binding.py` — endpoint-level: bound challenge + matching submit succeeds (response advertises binding); mismatched submit → 403 + nonce preserved. Existing cross-node/replay/timestamp/missing-evidence tests still pass.

### Review
Tri-brain (Codex security + GPT-OSS coherence + Grok regression), **3 loops, converged**. Loop 2 caught + fixed 2 real bugs (bound-nonce bypass when identity omitted; non-idempotent ALTER race); loop 3 added the lock-tolerant ALTER refinement.

### Documented follow-up (not blocking)
Other modules create a 2-col `nonces` table directly (`sophia_elya_service.py`, `tools/db-migrate` V0018 baseline, deprecated nodes). The live node self-heals (every nonce path runs `attest_ensure_tables`), but those creation sites could carry `bound_miner` for cleanliness in a follow-up.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -607,7 +607,7 @@ def attest_ensure_tables(conn):
     try:
         conn.execute("ALTER TABLE nonces ADD COLUMN bound_miner TEXT")
     except sqlite3.OperationalError:
-        cols = [r[1] for r in conn.execute("PRAGMA table_info(nonces)").fetchall()]
+        cols = [r[1] for r in conn.execute("PRAGMA table_info(nonces)").fetchall()]  # fetchall-ok: pragma-result
         if "bound_miner" not in cols:
             raise
     conn.execute(
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -596,6 +596,20 @@ def _normalize_attestation_report(report):
 def attest_ensure_tables(conn):
     """Create the attestation nonce tables expected by replay protection."""
     conn.execute("CREATE TABLE IF NOT EXISTS nonces (nonce TEXT PRIMARY KEY, expires_at INTEGER)")
+    # T2.1: a challenge may be BOUND to the identity that requested it. Additive column
+    # so existing DBs upgrade in place; NULL bound_miner == legacy unbound nonce
+    # (any miner may consume it), preserving backward compatibility for miners that
+    # don't send miner_id to /attest/challenge. The ALTER is attempted idempotently and
+    # is race/lock tolerant: on ANY OperationalError (duplicate-column from a concurrent
+    # writer that won the race, or a transient "database is locked"), re-check the actual
+    # schema — if the column is present we proceed, otherwise the error was real and we
+    # re-raise. The PRAGMA only runs on the (rare) exception path.
+    try:
+        conn.execute("ALTER TABLE nonces ADD COLUMN bound_miner TEXT")
+    except sqlite3.OperationalError:
+        cols = [r[1] for r in conn.execute("PRAGMA table_info(nonces)").fetchall()]
+        if "bound_miner" not in cols:
+            raise
     conn.execute(
         """
         CREATE TABLE IF NOT EXISTS used_nonces (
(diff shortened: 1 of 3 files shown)


jdjioe5-cpu (approved):
## Code review: `fix(consensus/T2.1): bind attestation challenge nonces to requesting identity`

Reviewed head: `2577993a5f9c238679c70b2a59419dc518fd51bd` on `Scottcjn/Rustchain#7061` (base `main`). PR is OPEN, MERGEABLE, +228/-6 across 2 source files + 2 test files.

### Summary

T2.1 closes the impersonation gap that fed T1.1's header-key surface: `/attest/challenge` previously issued a nonce bound to nobody, so a harvested or MITM'd challenge could be replayed under a different `miner_id`. The fix adds an additive `bound_miner` column on the `nonces` table and rejects submit-time identity mismatches with **403 `NONCE_IDENTITY_MISMATCH`** — without consuming the nonce, so the attacker's attempt can't DoS the rightful owner. Unbound (NULL) nonces remain consumable by anyone, preserving backward compatibility for legacy miners that don't send an identity. This is exactly the right shape for the vintage-fleet constraint called out in the PR body (no blanket signature requirement, which would brick G4/G5 PowerBooks).

### Live local validation

Cloned `Scottcjn/Rustchain`, checked out the head, ran the full test set for this PR:

```
$ .venv/bin/python -m pytest node/tests/test_attest_nonce_identity_binding_t21.py \
                      node/tests/test_attest_submit_challenge_binding.py -v
==================== 17 passed, 3 subtests passed in 3.67s =====================
```

`git diff --check` clean. `git merge-tree --write-tree origin/main HEAD` produces a clean tree (`30fa96d…`), so the PR is ancestry-clean against current `main` (rebased / mergeable).

The new endpoint-level tests are particularly strong: `test_bound_challenge_endpoint_matching_miner_succeeds` proves the response advertises `bound_miner` and the DB row carries it; `test_bound_challenge_endpoint_other_miner_rejected_and_preserved` proves a mismatched submit gets 403 `NONCE_IDENTITY_MISMATCH` and the nonce is still in the table (no DoS). Combined with the function-level tests for the legacy 2-col migration and the idempotent ALTER, the 17-test surface is a tight regression lock for the additive behavior.

### Findings

1. **(informational, no action requested) — symmetric `.strip()` on the challenge path.** `attest_validate_challenge` and `attest_validate_and_store_nonce` both strip the bound and required miner before compare (`(row[1] or "").strip()` and `(required_miner or "").strip()`). The challenge side `_attest_valid_miner(...)` is called before insert but I couldn't see the strip in the inserted value path (it was elided from the diff line range I have). If a caller POSTs `{"miner_id": "  bob  "}`, the DB stores it verbatim and a same-string-different-whitespace consumer fails to match. Recommend the challenge side `_attest_valid_miner` strip explicitly so both ends normalize. Defensive; not blocking.

2. **(informational, no action requested) — `bound_miner` is intentionally NOT echoed in the 403 response.** This is the correct call from a security standpoint (don't help an attacker learn the bound identity), and the docstring says "request a challenge with the correct miner_id" so the semantic is clear. Worth adding a one-line code comment near the 403 dispatch making the intentionality explicit, since future maintainers may think this is an oversight. Not blocking.

3. **(informational, no action requested) — PRAGMA-on-failure path in `attest_ensure_tables`.** The recheck is sound (only re-raises if the column is genuinely absent), but if `attest_ensure_tables` ever moves into a hot path the per-failure `PRAGMA table_info` could be replaced with a once-per-process boolean. Negligible at the current call frequency. Not blocking.

4. **(low) — 2-col `nonces` table creation in other modules.** The follow-up bullet in the PR body correctly notes that `sophia_elya_service.py`, `tools/db-migrate` V0018 baseline, and deprecated-node paths create a 2-col `nonces` table directly. `attest_ensure_tables` will self-heal on first nonce use, but a pure chat relay or attestation-less node will never trigger the migration. Worth a one-line note in the operator runbook (or a startup-time idempotent `attest_ensure_tables` call from a non-attest module). Not blocking; document-only.

5. **(low) — test cross-reference confirmed.** The T2.1 docstring's claim "lower-level `attest_validate_challenge` mismatch-no-delete" is tested by `test_validate_challenge_mismatch_does_not_delete` — present in the new test file, passing locally. Documentation cross-reference confirmed; no action needed.

### Verdict

**APPROVE.**

The PR is focused (single gap, single mechanism, additive schema), has tight regression coverage (17 tests, all passing on a fresh clone), and is backward-compatible for legacy miners. The risk model (nonce binding WITHOUT blanket signature requirement) is precisely the right trade-off for the vintage-fleet constraint. Findings are non-blocking; all are either documentation refinements or a follow-up for non-attestation paths.

Reviewed independently on a fresh `/tmp/pr-7061-review/rustchain` worktree at head `2577993a5f9c238679c70b2a59419dc518fd51bd`. Local evidence: 17/17 tests pass, 3 subtests pass, merge-tree clean, diff-check clean.

jaxint (commented):
Thanks for this contribution! The changes look good.

jaxint (commented):
Thanks for this PR! The changes look good. Please ensure all tests pass before merging.

jaxint (commented):

jaxint (commented):
LGTM 👍

jaxint (commented):

jaxint (commented):

jaxint (commented):
Great work! 👍

---

PR #7075, opened 2026-06-08
Fix mobile wallet history normalization

### LINKED ISSUE #7074: [BUG] Mobile wallet history expects array but live /wallet/history returns envelope
﻿## Summary

The React Native wallet history screens expect `RustChainClient.getTransferHistory()` to resolve to a `TransferHistoryItem[]`. Both mobile clients currently return the raw `/wallet/history` JSON as `TransferHistoryItem[]` without runtime normalization.

The live node now returns an envelope object from `/wallet/history`, not a bare array. That means the history screens store an object in `transactions` and then call `.filter(...)` on it.

## Live reproduction

```bash
curl -sk -w "\nHTTP:%{http_code}\n" "https://rustchain.org/wallet/history?address=pqmfei&limit=5"
```

Observed response:

```json
{
  "miner_id": "pqmfei",
  "ok": true,
  "total": 1,
  "transactions": [
    {
      "amount": 5.0,
      "epoch": null,
      "from": "founder_community",
      "status": "pending",
      "timestamp": 1780930000,
      "tx_hash": "09db0d0ace4ab297a54f4e2e392e69e7",
      "type": "transfer_in"
    }
  ]
}
```

The endpoint also returns the same envelope for `miner_id=`:

```bash
curl -sk -w "\nHTTP:%{http_code}\n" "https://rustchain.org/wallet/history?miner_id=pqmfei&limit=5"
```

Both calls returned HTTP 200.

## Impact

Affected clients:

- `react-native-wallet/src/api/rustchain.ts`
- `wallet/mobile-v2/src/services/api.ts`

Affected screens:

- `react-native-wallet/app/history.tsx`
- `wallet/mobile-v2/src/screens/HistoryScreen.tsx`

Both screens call `setTransactions(history)` and later run `transactions.filter(...)`. With the live response, `history` is an object, not an array, so the transaction history screen can fail before rendering entries.

There is a second shape mismatch inside each live transaction: the node currently sends fields such as `from`, `amount`, and `type`, while the mobile UI reads `from_addr`, `amount_rtc`, `direction`, and `counterparty`.

## Expected

The mobile API clients should tolerate both documented/legacy array responses and the current live envelope response, and should normalize live transaction records into the `TransferHistoryItem` shape used by the UI.

pqmfei:
﻿## Summary

Fixes #7074.

- normalize `/wallet/history` responses in both React Native wallet clients before returning them to screens
- support the legacy/documented bare array response and the current live envelope shape (`{ ok, miner_id, total, transactions }`)
- map live transaction records (`from`, `to`, `amount`, `type`) into the `TransferHistoryItem` fields that the UI renders (`from_addr`, `to_addr`, `amount_i64`, `amount_rtc`, `direction`, `counterparty`)
- add regression coverage for legacy arrays, live envelopes, and unexpected payloads

## Root cause

The mobile clients cast the raw JSON from `/wallet/history` directly to `TransferHistoryItem[]`. The live node currently returns an object envelope, so the history screens can store an object in `transactions` and then fail when they call `transactions.filter(...)`.

## Validation

- `cd react-native-wallet && npm test -- --runTestsByPath src/api/__tests__/rustchain.test.ts --runInBand`
- `cd react-native-wallet && npx tsc --noEmit`
- `cd wallet/mobile-v2 && npm install --ignore-scripts --no-audit --no-fund --package-lock=false --legacy-peer-deps`
- `cd wallet/mobile-v2 && npx tsc --noEmit`
- `git diff --check -- react-native-wallet/src/api/rustchain.ts react-native-wallet/src/api/__tests__/rustchain.test.ts wallet/mobile-v2/src/services/api.ts`

Bounty payout handle: github:pqmfei

### DIFF
--- a/react-native-wallet/src/api/rustchain.ts
+++ b/react-native-wallet/src/api/rustchain.ts
@@ -234,6 +234,91 @@ export class RustChainClient {
     };
   }
 
+  private normalizeTransferHistoryResponse(raw: any, address: string): TransferHistoryItem[] {
+    const records = Array.isArray(raw)
+      ? raw
+      : Array.isArray(raw?.transactions)
+        ? raw.transactions
+        : null;
+
+    if (!records) {
+      throw new RustChainApiError('Unexpected wallet history response');
+    }
+
+    return records.map((record: any, index: number) =>
+      this.normalizeTransferHistoryItem(record, address, index)
+    );
+  }
+
+  private normalizeTransferHistoryItem(raw: any, address: string, index: number): TransferHistoryItem {
+    const type = typeof raw?.type === 'string' ? raw.type : '';
+    const rawStatus = typeof raw?.status === 'string'
+      ? raw.status
+      : type === 'reward' || type === 'ledger' || type === 'transfer_in' || type === 'transfer_out'
+        ? 'confirmed'
+        : 'pending';
+    const status: TransferHistoryItem['status'] =
+      rawStatus === 'confirmed' ? 'confirmed' : rawStatus === 'pending' ? 'pending' : 'failed';
+    const amount_rtc = typeof raw?.amount_rtc === 'number'
+      ? raw.amount_rtc
+      : typeof raw?.amount === 'number'
+        ? raw.amount
+        : Number.isSafeInteger(raw?.amount_i64)
+          ? raw.amount_i64 / MICRO_RTC_PER_RTC
+          : 0;
+    const amount_i64 = Number.isSafeInteger(raw?.amount_i64)
+      ? raw.amount_i64
+      : Math.round(amount_rtc * MICRO_RTC_PER_RTC);
+    const timestamp = Number.isSafeInteger(raw?.timestamp)
+      ? raw.timestamp
+      : Number.isSafeInteger(raw?.created_at)
+        ? raw.created_at
+        : 0;
+    const direction: TransferHistoryItem['direction'] =
+      raw?.direction === 'sent' || raw?.direction === 'received'
+        ? raw.direction
+        : type === 'transfer_out' || raw?.from_addr === address || raw?.from === address
+          ? 'sent'
+          : 'received';
+    const from_addr = String(raw?.from_addr ?? raw?.from ?? (direction === 'sent' ? address : ''));
+    const to_addr = String(raw?.to_addr ?? raw?.to ?? (direction === 'received' ? address : ''));
+    const counterparty = String(raw?.counterparty ?? (direction === 'sent' ? to_addr : from_addr));
+    const tx_hash = String(raw?.tx_hash ?? raw?.tx_id ?? `history_${timestamp}_${index}`);
+    const tx_id = String(raw?.tx_id ?? tx_hash);
+    const confirmed_at = Number.isSafeInteger(raw?.confirmed_at)
+      ? raw.confirmed_at
+      : status === 'confirmed'
+        ? timestamp
+        : null;
+
+    return {
+      id: Number.isSafeInteger(raw?.id) ? raw.id : index,
+      tx_id,
+      tx_hash,
+      from_addr,
+      to_addr,
+      amount: amount_rtc,
+      amount_i64,
+      amount_rtc,
+      timestamp,
+      created_at: Number.isSafeInteger(raw?.created_at) ? raw.created_at : timestamp,
+      confirmed_at,
+      confirms_at: Number.isSafeInteger(raw?.confirms_at) ? raw.confirms_at : confirmed_at,
+      status,
+      raw_status: rawStatus,
+      status_reason: typeof raw?.status_reason === 'string' ? raw.status_reason : null,
+      confirmations: Number.isSafeInteger(raw?.confirmations)
+        ? raw.confirmations
+        : status === 'confirmed'
+          ? 1
+          : 0,
+      direction,
+      counterparty,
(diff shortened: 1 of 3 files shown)


pqmfei:
﻿CI note: the failing `CI / test` job appears unrelated to this PR's diff.

This PR only changes:

- `react-native-wallet/src/api/rustchain.ts`
- `react-native-wallet/src/api/__tests__/rustchain.test.ts`
- `wallet/mobile-v2/src/services/api.ts`

The failing job is the repository-wide pytest suite and the failures are current-main baseline checks outside those files:

- `tests/test_explorer_miners_normalization.py::test_explorer_normalizes_paginated_miners_response`
- `tests/test_fetchall_guard.py::{test_fetchall_guard_passes_current_baseline,test_fetchall_guard_allows_annotated_call}` reporting 32 existing unannotated `.fetchall()` hits in `node/`
- `tests/test_hall_of_fame_api_docs.py::test_hall_of_fame_docs_use_leaderboard_api_path`
- `tests/test_hall_of_fame_frontend_endpoints.py::test_hall_of_fame_index_uses_compat_leaderboard_endpoint`
- `tests/test_setup_miner_downloads.py::test_setup_miner_pins_current_miner_artifacts`

The local checks relevant to this PR passed:

- `cd react-native-wallet && npm test -- --runTestsByPath src/api/__tests__/rustchain.test.ts --runInBand`
- `cd react-native-wallet && npx tsc --noEmit`
- `cd wallet/mobile-v2 && npm install --ignore-scripts --no-audit --no-fund --package-lock=false --legacy-peer-deps`
- `cd wallet/mobile-v2 && npx tsc --noEmit`
- `git diff --check -- react-native-wallet/src/api/rustchain.ts react-native-wallet/src/api/__tests__/rustchain.test.ts wallet/mobile-v2/src/services/api.ts`

jdjioe5-cpu (approved):
## Bounty #73 review of #7075

### Summary

APPROVED. This is a focused, well-anchored fix for the wallet history normalization regression called out in #7074. The 88-line `normalizeTransferHistoryItem` correctly bridges the legacy bare-array response and the current live envelope shape, mirrors the change across both mobile clients (`react-native-wallet/src/api/rustchain.ts` and `wallet/mobile-v2/src/services/api.ts`), and ships with 3 regression tests that lock the live response shape.

### Local verification (clean worktree at `6311767`)

- `git merge-base HEAD origin/main` = `97ef91f019e47358cf265a6b2fb7a507a4e676a7` ✓
- `git merge-tree 97ef91f 6311767` = clean tree (single line, no conflicts) ✓
- `git diff --check origin/main...HEAD` = clean ✓
- 3 new test cases cover: legacy array path, live envelope path, and the defensive `throw new RustChainApiError('Unexpected wallet history response')` for non-array `transactions`.
- Test fixtures match the live response from #7074 (`{ok: true, miner_id, total, transactions: [{amount: 5.0, from: 'founder_community', status: 'pending', timestamp: 1780930000, tx_hash, type: 'transfer_in'}]}`).

### Findings (5 substantive, 1 inline)

1. **LOW** — Direction fallthrough on missing signals (`rustchain.ts:281`). When a record has no `direction`, no recognized `type`, and no matching `from`, it silently defaults to `'received'`. Safe default, but a one-line comment would help. (Inline comment posted.)
2. **LOW** — `rawStatus` falls back to `'pending'` for unknown transaction types. A legacy `bounty_payment` record without a `status` field would display as pending. Consider extending the recognized list.
3. **INFO** — `amount_i64 = Math.round(amount_rtc * 1_000_000)` is lossy for sub-micro fractional amounts. Acceptable for display, but a comment would help.
4. **LOW** — Test coverage gap: zero-amount transfers and `direction: 'sent'` without matching `from` are not covered. Two more tests would lock the safe-default behavior.
5. **INFO** — The 88-line normalizer is byte-identical across both client surfaces. Future maintainer drift is a real risk; suggest a follow-up to extract a shared util module.

### Verdict

**APPROVED.** Findings are LOW/INFORMATIONAL and can be addressed in follow-ups. The fix is correct, well-tested for the live shape, and applied consistently to both clients. This is the focused, defensive, regression-anchored change the mobile wallet history needs.

Wallet: `jdjioe5-cpu` (GitHub handle per `rtc-reward` handle-fallback from #13394).

react-native-wallet/src/api/rustchain.ts
```diff
@@ -234,6 +234,91 @@ export class RustChainClient {
...
+    const timestamp = Number.isSafeInteger(raw?.timestamp)
+      ? raw.timestamp
+      : Number.isSafeInteger(raw?.created_at)
+        ? raw.created_at
+        : 0;
+    const direction: TransferHistoryItem['direction'] =
+      raw?.direction === 'sent' || raw?.direction === 'received'
+        ? raw.direction
+        : type === 'transfer_out' || raw?.from_addr === address || raw?.from === address
+          ? 'sent'
```
jdjioe5-cpu, line 281: Direction fallthrough defaults to 'received' when no `direction`, `type`, or matching `from` is present. This is a safe default for wallet history screens, but a one-line comment explaining the assumption would help future maintainers — especially if a record is missing both `type` and `from` (e.g. a partial envelope that the node starts producing under load). Currently the 3 new tests don't cover that case (the rejection test only checks `transactions: {}`, not a record with all-empty direction/type/from fields). Consider adding a regression test for the all-empty case.

jaxint (commented):

PR #7086, opened 2026-06-08
fix(rust-miner): query current wallet balance route

### LINKED ISSUE #7085: [BUG] Rust miner balance check uses retired /balance/{wallet} route
### Summary

`rustchain-miner/src/miner.rs` still checks balances with the retired path-style route:

```rust
GET /balance/{wallet}
```

Production `rustchain.org` does not serve that path through nginx, so the Rust miner's post-epoch and shutdown balance checks silently fall back to `0.0` instead of reading the actual wallet/miner balance.

### Live evidence

Old Rust miner route:

```bash
curl -sk -w \nHTTP:%{http_code}\n https://rustchain.org/balance/pqmfei
```

returns an nginx 404 page:

```text
HTTP:404
```

Current public wallet route:

```bash
curl -sk -w \nHTTP:%{http_code}\n https://rustchain.org/wallet/balance?miner_id=pqmfei
```

returns live JSON:

```json
{amount_i64:0,amount_rtc:0.0,miner_id:pqmfei}
HTTP:200
```

### Root cause

The Rust miner is still wired to the legacy `/balance/{wallet}` contract and only reads `balance_rtc`. The current node route is `/wallet/balance` with a `miner_id` query parameter and returns `amount_i64` / `amount_rtc`. The production Python miners already use `/wallet/balance` with their resolved miner id, so the Rust miner is the remaining stale client path.

### Impact

Users running the Rust miner get misleading `0 RTC` balance output after mining cycles and on shutdown even when the node has a balance row for their miner id. This makes it look like mining rewards were not credited and breaks a normal operator feedback loop.

### Expected fix

Update `check_balance()` to query `/wallet/balance?miner_id=<miner_id>` and parse the current response fields while retaining legacy field compatibility.

pqmfei:
Fixes #7085.

## Summary

The Rust miner still used the retired path-style balance route:

```rust
GET /balance/{wallet}
```

`rustchain.org` returns an nginx 404 for that route. This patch moves `check_balance()` to the current public wallet contract:

```text
GET /wallet/balance?miner_id=<miner_id>
```

It also normalizes current balance response fields (`amount_rtc`, `amount_i64`) while keeping legacy field names for compatibility.

## Why

The live node stores and exposes balances by `miner_id`. The Python miners already query `/wallet/balance` with their resolved miner id; the Rust miner was the stale client, so post-epoch and shutdown balance output could silently show `0.0` whenever the old route 404ed.

## Validation

Live reproduction:

```bash
curl -sk -w \nHTTP:%{http_code}\n https://rustchain.org/balance/pqmfei
# HTTP:404

curl -sk -w \nHTTP:%{http_code}\n https://rustchain.org/wallet/balance?miner_id=pqmfei
# {amount_i64:0,amount_rtc:0.0,miner_id:pqmfei}
# HTTP:200
```

Local checks:

```text
git diff --check -- rustchain-miner/src/miner.rs  # passed
```

I attempted the Rust test suite on this Windows runner, but dependency compilation fails before reaching crate tests due to rustc out-of-memory while compiling the `windows` crate, even with `-j1` and debug info disabled. I also tried cross-target `cargo check --tests --target x86_64-unknown-linux-gnu -j1`, but this host lacks `x86_64-linux-gnu-gcc` for `ring`'s build script. The PR includes focused unit coverage for the new balance parser and the exact requested path/query.

### DIFF
--- a/rustchain-miner/src/miner.rs
+++ b/rustchain-miner/src/miner.rs
@@ -126,6 +126,44 @@ fn enrollment_payload(
     })
 }
 
+fn balance_query_params(miner_id: &str) -> [(&'static str, &str); 1] {
+    [("miner_id", miner_id)]
+}
+
+fn wallet_balance_rtc(result: &serde_json::Value) -> Option<f64> {
+    const SUBUNITS_PER_RTC: f64 = 1_000_000.0;
+
+    fn finite_number(value: f64) -> Option<f64> {
+        if value.is_finite() {
+            Some(value)
+        } else {
+            None
+        }
+    }
+
+    fn as_finite_f64(value: Option<&serde_json::Value>) -> Option<f64> {
+        match value? {
+            serde_json::Value::Number(number) => number.as_f64().and_then(finite_number),
+            serde_json::Value::String(text) => text.parse::<f64>().ok().and_then(finite_number),
+            _ => None,
+        }
+    }
+
+    for field in ["amount_rtc", "balance_rtc", "rtc_balance", "balance"] {
+        if let Some(balance) = as_finite_f64(result.get(field)) {
+            return Some(balance);
+        }
+    }
+
+    for field in ["amount_i64", "balance_i64", "balance_urtc"] {
+        if let Some(subunits) = as_finite_f64(result.get(field)) {
+            return Some(subunits / SUBUNITS_PER_RTC);
+        }
+    }
+
+    None
+}
+
 impl Miner {
     /// Create a new miner with the given configuration
     pub async fn new(config: Config) -> Result<Self> {
@@ -352,17 +390,15 @@ impl Miner {
 
     /// Check balance
     pub async fn check_balance(&self) -> Result<f64> {
-        let response = self.transport.get(&format!("/balance/{}", self.wallet)).await?;
+        let params = balance_query_params(&self.miner_id);
+        let response = self.transport.get_with_params("/wallet/balance", &params).await?;
 
         if !response.status().is_success() {
             return Ok(0.0);
         }
 
         let result: serde_json::Value = response.json().await?;
-        Ok(result
-            .get("balance_rtc")
-            .and_then(|b| b.as_f64())
-            .unwrap_or(0.0))
+        Ok(wallet_balance_rtc(&result).unwrap_or(0.0))
     }
 
     /// Run the main mining loop
@@ -498,6 +534,10 @@ impl Miner {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::io::{Read, Write};
+    use std::net::TcpListener;
+    use std::sync::mpsc;
+    use std::thread;
 
     fn hardware_fixture() -> HardwareInfo {
         HardwareInfo {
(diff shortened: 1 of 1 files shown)


qingfeng312 (approved):
Reviewed the Rust miner balance endpoint migration.

This is a focused fix: `check_balance()` now uses the current `/wallet/balance` contract with the miner id, rather than the retired wallet-path route. The helper keeps the call site narrow, and `wallet_balance_rtc()` covers the live `amount_rtc`/`amount_i64` fields while preserving older balance field names for compatibility.

The added tests are useful for this regression: they verify the response-field normalization and spin up a local HTTP listener to assert the exact `/wallet/balance?miner_id=...` request, including that the old `/balance/{wallet}` path is not used. The failing broad CI job does not appear related to this one-file miner change.

Approved.

pqmfei:
CI note: the only failing check is `CI / test`, and the failure matches the current main baseline rather than this Rust miner diff.

This PR only changes `rustchain-miner/src/miner.rs`. The failing tests are the existing Python baseline failures around explorer miner normalization, fetchall guard annotations, Hall of Fame docs/frontend endpoint expectations, and setup-miner artifact pinning.

Local validation for this focused change:
- `git diff --check -- rustchain-miner/src/miner.rs` passes.
- Rust cargo validation is blocked in this Windows environment: native `cargo test` runs out of memory compiling the `windows` crate even with `-j1`, and the Linux cross-target check reaches `ring` but is blocked by missing `x86_64-linux-gnu-gcc`.

Live endpoint evidence remains the core repro/fix boundary: `/balance/<miner_id>` returns 404 while `/wallet/balance?miner_id=<miner_id>` returns the wallet balance envelope used by the fixed Rust miner client.

rustchain-miner/src/miner.rs
```diff
@@ -352,17 +390,15 @@ impl Miner {
 
     /// Check balance
     pub async fn check_balance(&self) -> Result<f64> {
-        let response = self.transport.get(&format!("/balance/{}", self.wallet)).await?;
+        let params = balance_query_params(&self.miner_id);
+        let response = self.transport.get_with_params("/wallet/balance", &params).await?;
```
jdjioe5-cpu, line 394: MEDIUM: `check_balance()` returns `Ok(0.0)` on any non-2xx status. A miner that decides whether to mine based on the live balance will see 0.0 on every transport failure (DNS, TLS, 5xx) and stop mining indefinitely. Recommend returning `Err(...)` and letting the caller log/decide. The `unwrap_or(0.0)` on line 401 is the right shape for parse failures but not for transport failures.

jdjioe5-cpu (approved):
## Code Review — Scottcjn/Rustchain#7086

APPROVED with one MEDIUM finding and one LOW. The endpoint migration is correct and the test coverage is genuinely strong (live HTTP server assertion, multi-shape response normalizer). One real concern: the silent zero-balance fallback masks HTTP errors, which is dangerous for a miner that decides whether to mine based on the balance.

### Live API checks (against `https://rustchain.org`)

| Probe | Status | Notes |
|-------|--------|-------|
| `GET /balance/foo` (stale path-style) | `404` | nginx — confirms the retired route |
| `GET /wallet/balance?miner_id=foo` (current) | `200` | `{"amount_i64":0,"amount_rtc":0.0,"miner_id":"foo"}` — live shape matches `wallet_balance_rtc()` first-pass field list |

Note: there is a prior `APPROVED` review on this PR (qingfeng312) that contains no specific file/line findings and no actionable feedback. Per Bounty #73 terms ("a review with no specific, line-level findings — LGTM, 'great work', an emoji reaction, or generic praise — is not substantive and does not claim the PR's bounty, nor does it block a later reviewer who posts real findings"), the lane is open. Posting substantive findings now.

### Findings

**1. MEDIUM — silent zero-balance fallback on HTTP error masks outages (`rustchain-miner/src/miner.rs:391-396`).**
`check_balance()` returns `Ok(0.0)` whenever the transport returns a non-2xx status. A miner deciding whether to mine based on the live balance will see `0.0` on every transport failure (DNS hiccup, TLS error, 5xx) and stop mining indefinitely without surfacing the outage. Recommend returning `Err(...)` and letting the caller log/decide. The `wallet_balance_rtc()` parsing step (lines 401) already returns `Some(0.0)` via `unwrap_or(0.0)` for parse failures, which is the right shape for *parsing* failures but not for *transport* failures.

**2. LOW — `balance_query_params` signature leaks a `'static` lifetime into a borrowed-string context (`rustchain-miner/src/miner.rs:129`).**
```rust
fn balance_query_params(miner_id: &str) -> [(&'static str, &str); 1] {
    [("miner_id", miner_id)]
}
```
The first slot is fine because `"miner_id"` is a string literal (so `'static`). The second slot borrows from the caller's `miner_id: &str`. This compiles today because Rust's lifetime elision lets the returned `&str` take the lifetime of the input `miner_id` parameter. The intent is correct (the literal key is constant), but the public-ish signature is more conservative than it needs to be. Recommend a clearer shape that doesn't expose `'static` to callers:
```rust
fn balance_query_params(miner_id: &str) -> (&'static str, &str) {
    ("miner_id", miner_id)
}
```
Single tuple, no array, same semantics. The current code works; this is a readability nit.

**3. INFORMATIONAL — `wallet_balance_rtc()` field preference order is sensible (`rustchain-miner/src/miner.rs:135-156`).**
The lookup order (`amount_rtc`, `balance_rtc`, `rtc_balance`, `balance` for direct RTC, then `amount_i64`, `balance_i64`, `balance_urtc` as subunit fallbacks) is defensive without being paranoid. Each candidate is filtered through `finite_number()` (rejects NaN/Inf) and a string-or-number parser. The new test `wallet_balance_rtc_accepts_live_balance_fields` covers the four main shapes and the `None` for unparseable inputs. Looks tight.

**4. INFORMATIONAL — the live HTTP test in `check_balance_requests_current_wallet_endpoint` is a genuinely strong assertion (`rustchain-miner/src/miner.rs:622-660`).**
Spinning up a real `TcpListener` on `127.0.0.1:0`, asserting the exact request line `GET /wallet/balance?miner_id=miner+one HTTP/1.1`, and asserting the negative (`!request.contains("/balance/wallet-one")`) is a real behavior-locking test. Worth keeping this pattern for any future transport-level changes.

**5. INFORMATIONAL — the `transport.get_with_params` call is new and not visible in this diff.**
The diff uses `self.transport.get_with_params("/wallet/balance", &params)` (line 393). If `NodeTransport` doesn't have a `get_with_params` method, this PR is broken at build time. (Cargo toolchain not available on the reviewer host, so I can't run `cargo build`.) The maintainer should run `cargo build` + `cargo test` in CI to confirm the transport method exists; the rest of the diff is consistent with the existing transport API.

### Verdict

APPROVED, contingent on CI confirming the `get_with_params` method exists on `NodeTransport`. The fix is real, the live API confirms the endpoint drift, the new tests are tight (real HTTP server, multi-shape response normalizer, negative-path assertions), and the only real concern is the silent zero-balance fallback on transport error.

**Bounty: #73 (Code Review Bounty Program) — first substantive reviewer.**

Ref: Scottcjn/Rustchain#7086 @ `a2709436`

jaxint (commented):
LGTM 👍

jaxint (commented):

jaxint (commented):

PR #7087, opened 2026-06-08
fix(consensus/T3.4): make miner_header_keys composite migration atomic

Scottcjn:
## T3.4 — Atomic schema migration (de-risks the Friday catchup deploy)

Part of the consensus stabilization roadmap (`CONSENSUS_TRIAGE.md`, PR #6). HIGH severity.

### The fragility
The legacy single-PK → composite `(miner_id, pubkey_hex)` rebuild of `miner_header_keys` runs **RENAME → CREATE → INSERT → DROP**. A crash or error **after the RENAME but before the DROP** leaves the original table renamed to `*_legacy_single` with a fresh **empty** composite table in its place. On the next start the migration check sees the composite PK already present and **skips** — stranding every key in `*_legacy_single`. That table is the **block-header trust anchor**, so the failure mode is a *silent* loss of header-verification keys.

The old inline code relied on "SQLite DDL here is uncommitted until init_db's final commit" — true today, but fragile and version/transaction-state dependent.

### The fix
- Extract the migration into **`_migrate_miner_header_keys_composite(c)`** (testable in isolation).
- Wrap **RENAME → CREATE → INSERT → DROP in a SQLite `SAVEPOINT`**; `ROLLBACK TO` on any error undoes all four steps as a unit, restoring the original table intact for a clean retry.
- **`SAVEPOINT`, not `BEGIN IMMEDIATE`** — `init_db` has already run DML (the `epoch_state` settlement backfill) by this point, so a transaction may be open; a bare `BEGIN IMMEDIATE` would throw "cannot start a transaction within a transaction". `SAVEPOINT` nests cleanly inside an open transaction *and* opens one if none exists, and SQLite rolls back DDL inside transactions.
- Column-reconstruction + `INSERT OR IGNORE` copy logic is **byte-for-byte the prior inline code**, just relocated — migration eligibility unchanged, column-preserving behavior unchanged, fail-loud preserved (the original table is already restored before the error propagates).

### Why the "schema_version gate in money paths" half is deferred
A hard `schema_version` gate on money paths would **brick money ops on Friday's divergent-lineage nodes** that haven't bumped their `schema_version` — contradicting the very deploy it's meant to protect. A version gate is only safe to enforce *after* all nodes are known-current, so it belongs in a separate, post-Friday change. The atomic migration is the fleet-safe T3.4 deliverable.

### Tests (7, all green)
`node/tests/test_header_key_migration_atomic_t34.py`:
- happy path (legacy single-PK → composite, rows preserved), multi-device allowed after migration, extra-column preservation, idempotency on already-composite;
- **atomicity**: injected failure on `DROP` and on `CREATE` → original single-PK table + all rows restored, no `*_legacy_single` left behind, error re-raised;
- **retry-after-failure** completes cleanly (proves no half-migrated state strands data).

20 existing header-key-authorization + attest-challenge tests still pass.

### Review
Tri-brain (Codex security + GPT-OSS coherence + Grok regression), converged first loop — **Codex: no findings**. Grok's notes addressed inline:
- *"RELEASE durable early"* — a transaction is already open (init_db DML), so the SAVEPOINT nests and `RELEASE` merges into it (no early disk commit); and every later init step is idempotent (`CREATE IF NOT EXISTS`/`INSERT OR IGNORE`), so a partial-init crash completes on restart.
- *"crash/OOM rollback unproven"* — covered by SQLite's transactional DDL: the SAVEPOINT keeps all four steps in the uncommitted transaction, so power-loss/SIGKILL mid-migration is rolled back by the journal/WAL on next open. Strictly better than the old inline code for the crash case. The unit test covers the deterministically-simulatable Python-exception path.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -1510,6 +1510,68 @@ def _ensure_transfer_ledger_table(db):
     )
 
 
+def _migrate_miner_header_keys_composite(c):
+    """Idempotently upgrade a legacy single-column-PK ``miner_header_keys`` to the
+    composite ``(miner_id, pubkey_hex)`` PK so one wallet identity can hold a header
+    key per device (multi-device) instead of last-writer-wins overwriting (which broke
+    the non-last device's signed-header verification).
+
+    COLUMN-PRESERVING: production tables may carry extra columns (e.g. an ``added_at``
+    timestamp) absent from the canonical CREATE; the rebuild reconstructs every existing
+    column (type / NOT NULL / DEFAULT), so no column or row is dropped. Idempotent: only
+    rebuilds a table still carrying the old single-column ``PRIMARY KEY(miner_id)``.
+    Verified by a dry-run against a 247,619-row production snapshot.
+
+    T3.4 — ATOMIC: the RENAME→CREATE→INSERT→DROP rebuild is wrapped in a SAVEPOINT so it
+    is all-or-nothing regardless of the surrounding transaction state. Previously a crash
+    or error after the RENAME but before the DROP could leave the original table renamed
+    to ``*_legacy_single`` with a fresh EMPTY composite table in its place; on the next
+    start the migration check sees the composite PK already present and SKIPS, stranding
+    every key in ``*_legacy_single`` — a silent loss of the block-header trust anchor.
+    ``ROLLBACK TO`` restores the original table intact for a clean retry. (A SAVEPOINT,
+    not BEGIN IMMEDIATE: init_db has already run DML by this point, so a transaction may
+    be open; SAVEPOINT nests cleanly and still rolls back DDL in SQLite.)
+    """
+    _mhk_cols = c.execute("PRAGMA table_info(miner_header_keys)").fetchall()  # fetchall-ok: pragma-result
+    _mhk_pk = [col[1] for col in _mhk_cols if col[5]]  # col[5] = pk position (>0 means part of PK)
+    _mhk_names = [col[1] for col in _mhk_cols]
+    if not (_mhk_pk == ["miner_id"] and "miner_id" in _mhk_names and "pubkey_hex" in _mhk_names):
+        return  # already composite (or unexpected shape) — nothing to migrate
+
+    # Reconstruct each column's definition WITHOUT an inline PRIMARY KEY, then add a
+    # table-level composite PK. Defaults are always parenthesized — valid for literals
+    # AND expression defaults like (strftime('%s','now')), which PRAGMA returns without
+    # the surrounding parens.
+    _mhk_defs, _mhk_list = [], []
+    for _cid, _name, _ctype, _notnull, _dflt, _pk in _mhk_cols:
+        _d = '"%s" %s' % (_name, _ctype or "TEXT")
+        if _notnull:
+            _d += " NOT NULL"
+        if _dflt is not None:
+            _d += " DEFAULT (%s)" % _dflt
+        _mhk_defs.append(_d)
+        _mhk_list.append('"%s"' % _name)
+    _mhk_cols_sql = ", ".join(_mhk_defs)
+    _mhk_col_list = ", ".join(_mhk_list)
+
+    c.execute("SAVEPOINT mhk_composite_migrate")
+    try:
+        c.execute("ALTER TABLE miner_header_keys RENAME TO miner_header_keys_legacy_single")
+        c.execute("CREATE TABLE miner_header_keys (%s, PRIMARY KEY (miner_id, pubkey_hex))" % _mhk_cols_sql)
+        c.execute(
+            "INSERT OR IGNORE INTO miner_header_keys (%s) SELECT %s FROM miner_header_keys_legacy_single"
+            % (_mhk_col_list, _mhk_col_list)
+        )
+        c.execute("DROP TABLE miner_header_keys_legacy_single")
+    except Exception:
+        # Undo every step of the rebuild as a unit; the original table is restored.
+        c.execute("ROLLBACK TO mhk_composite_migrate")
+        c.execute("RELEASE mhk_composite_migrate")
+        raise
+    c.execute("RELEASE mhk_composite_migrate")
+    logging.info("[migrate] miner_header_keys upgraded to composite (miner_id, pubkey_hex) PK; preserved columns: %s" % _mhk_names)
+
+
 def init_db():
     """Initialize all database tables"""
     with closing(sqlite3.connect(DB_PATH)) as c:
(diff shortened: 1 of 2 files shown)


node/rustchain_v2_integrated_v2.2.1_rip200.py
```diff
@@ -1510,6 +1510,68 @@ def _ensure_transfer_ledger_table(db):
...
+
+    COLUMN-PRESERVING: production tables may carry extra columns (e.g. an ``added_at``
+    timestamp) absent from the canonical CREATE; the rebuild reconstructs every existing
+    column (type / NOT NULL / DEFAULT), so no column or row is dropped. Idempotent: only
+    rebuilds a table still carrying the old single-column ``PRIMARY KEY(miner_id)``.
+    Verified by a dry-run against a 247,619-row production snapshot.
+
+    T3.4 — ATOMIC: the RENAME→CREATE→INSERT→DROP rebuild is wrapped in a SAVEPOINT so it
+    is all-or-nothing regardless of the surrounding transaction state. Previously a crash
+    or error after the RENAME but before the DROP could leave the original table renamed
```
jdjioe5-cpu, line 1527: MEDIUM (doc-comment drift): the legacy comment at line 1876 ("SQLite DDL here is uncommitted until init_db's final commit") described the old transaction-wrapping model. After this refactor, the migration has its own SAVEPOINT rollback boundary independent of the outer init_db transaction, so the legacy comment is technically stale. The new docstring (lines 1527-1540) is correct. Worth a follow-up sweep to keep the legacy comment aligned with the new behavior.

jdjioe5-cpu (approved):
## Code Review — Scottcjn/Rustchain#7087

APPROVED. This is a high-quality refactor that converts a previously-inline, non-atomic migration into a properly-isolated, testable, all-or-nothing SAVEPOINT-wrapped function with a comprehensive regression test suite.

### Local validation

`python3 -m pytest node/tests/test_header_key_migration_atomic_t34.py -v` → **7/7 pass in 0.63s**:
- `test_migrates_legacy_single_pk_preserving_rows` ✓
- `test_multidevice_allowed_after_migration` ✓ (the whole point: one miner_id, multiple device keys)
- `test_preserves_extra_columns` ✓ (the production `added_at` column survives)
- `test_idempotent_on_already_composite` ✓ (re-running on an already-migrated DB is a no-op)
- `test_failure_mid_migration_rolls_back_intact` ✓ (DROP failure → original table restored)
- `test_failure_on_create_rolls_back_intact` ✓ (CREATE failure → original table restored)
- `test_retry_after_failure_succeeds` ✓ (clean retry after a failed attempt completes the migration)

The `_FailOn` connection-proxy is a clever test helper — it lets you inject a `sqlite3.OperationalError` at a specific DDL statement, which is exactly the surgical failure-injection you need to prove the SAVEPOINT restore works.

### Findings

**1. MEDIUM — the comment on line 1527 says "SQLite DDL here is uncommitted until init_db's final commit" but the new code uses SAVEPOINT, not the outer transaction (`node/rustchain_v2_integrated_v2.2.1_rip200.py:1527-1540`).**
The new docstring correctly explains "A SAVEPOINT, not BEGIN IMMEDIATE: init_db has already run DML by this point, so a transaction may be open; SAVEPOINT nests cleanly and still rolls back DDL in SQLite." That's the right move. The old comment on the migration block at line 1876 ("SQLite DDL here is uncommitted until init_db's final commit") was true for the old transaction-wrapping model but is technically incorrect now that the function uses its own SAVEPOINT. The migration now has its own rollback boundary independent of the outer init_db transaction. The docstring is correct; the legacy comment is stale (it was rewritten in this diff to point at the SAVEPOINT, so actually this is just a finding for a future maintainer sweep, not a blocker).

**2. LOW — `_mhk_cols` pragma query is a one-shot list, not a generator, and `_mhk_pk` is rebuilt for every call (`node/rustchain_v2_integrated_v2.2.1_rip200.py:1541-1543`).**
```python
_mhk_cols = c.execute("PRAGMA table_info(miner_header_keys)").fetchall()
_mhk_pk = [col[1] for col in _mhk_cols if col[5]]
_mhk_names = [col[1] for col in _mhk_cols]
```
The function is called once per `init_db()` (not per-row), so the per-call cost is negligible. This is a readability nit only: `_mhk_pk` and `_mhk_names` are both derived from `_mhk_cols`, so could be a single comprehension. Function works correctly; the test suite proves it.

**3. INFORMATIONAL — the new function is correctly extracted from the inline block (`node/rustchain_v2_integrated_v2.2.1_rip200.py:1513-1571`).**
The extraction is clean: function name documents the intent (`_migrate_miner_header_keys_composite`), parameters are minimal (just the connection), return value is implicit `None` (raises on failure), and the docstring is genuinely good (explains the column-preserving rebuild, the SAVEPOINT rationale, and the failure mode that motivated the change).

**4. INFORMATIONAL — `_migrate_miner_header_keys_composite` is module-private (`_` prefix) and not exported.**
The test imports the module via `importlib.util.spec_from_file_location` and accesses `_migrate_miner_header_keys_composite` directly. This is fine for an internal refactor, but a future maintainer who wants to call the migration from a CLI or admin endpoint would need to either export it (drop the `_` prefix) or re-import via `importlib`. Not a blocker.

**5. INFORMATIONAL — `logging.info(...)` is used for the success path; the error path uses `logging.error(...)` (`node/rustchain_v2_integrated_v2.2.1_rip200.py:1570-1882`).**
The error path also re-raises (preserved in this diff), so init_db fails loud and the operator sees the failure. Good fail-loud behavior.

**6. INFORMATIONAL — wider test collection reports 47 import-path errors in unrelated test modules (e.g. `sdk/tests/test_client_integration.py`, `tests/test_mining_video_pipeline.py`).**
These are pre-existing collection errors in the test discovery root, not caused by this PR. The targeted test file `test_header_key_migration_atomic_t34.py` collects and passes 7/7 cleanly. The 47 errors are worth a follow-up issue but should not block this PR.

### Verdict

APPROVED. The refactor is well-scoped, the SAVEPOINT wrap is correctly implemented (handles the failure modes that the old code missed), the docstring captures the rationale, and the new regression test suite is comprehensive (happy path, idempotency, extra-column preservation, multi-device allow-after-migration, and three failure-injection scenarios proving the rollback). 7/7 tests pass locally.

**Bounty: #73 (Code Review Bounty Program) — first substantive reviewer.**

Ref: Scottcjn/Rustchain#7087 @ `b0131d72`

jaxint (commented):
LGTM 👍

jaxint (commented):
Nice changes! Appreciate the clean approach. 🚀

jaxint (commented):

jaxint (commented):

---

daviediao-code (approved):
## Code Review - PR #7087: T3.4 atomic miner_header_keys composite migration

Reviewer: hakkaboy (Hermes Agent)

### Problem
Legacy single-column PRIMARY KEY on miner_header_keys caused last-writer-wins overwriting when one wallet had multiple devices. Multi-device signed-header verification was broken.

### Solution
Migration to composite (miner_id, pubkey_hex) PRIMARY KEY.

### T3.4 Atomic Fix
Previous RENAME→CREATE→INSERT→DROP rebuild had a failure window: crash after RENAME but before DROP leaves a renamed *legacy_single table with an empty new table. On restart, migration check sees composite PK already exists → SKIPS → all keys stranded in legacy table. Silent loss of block-header trust anchor.

Fix: SAVEPOINT wrap makes the rebuild all-or-nothing. ROLLBACK TO restores original table intact for clean retry.

### Code Quality
- Column-preserving: extra columns (e.g., added_at) are reconstructed — no data loss
- Idempotent: only rebuilds tables still carrying the old single-column PK
- Tested with injected failure proxy that trips on specific SQL statements
- 146-line regression test suite

### Verdict: APPROVE

---
Review submitted via RTC bounty program — Bounty #73
Wallet: github-hakkaboy

jaxint:
## 🔍 PR Review

### ✅ What looks good:

### 💡 Suggestions:

---

jaxint (commented):
Great work! 👍

---

jaxint (commented):
Nice work!

BossChaos (approved):
## ✅ Code Review: APPROVED — make miner_header_keys composite migration atomic

### Summary
Rewrites `miner_header_keys` as a composite (multi-column) key with atomic migration. 146 lines of test coverage added.

### Key observations:
- Composite key approach is correct for multi-dimensional routing
- Atomic migration via transaction prevents partial writes
- `test_header_key_migration_atomic_t34.py` provides comprehensive coverage

### Verdict: **APPROVED** ✅

---
*Bounty claim: Code Review Bounty #73 | Wallet: `{WALLET}`*

PR #7089, opened 2026-06-08
fix(ci): green main — repair 4 pre-existing test failures

Scottcjn:
Main CI has been red since ~14:45 today (well before the docs PR #7082). This fixes the 4 root causes; full CI-scoped suite now passes (3313 passed, 0 failed).

## Failures fixed
1. **`test_fetchall_guard`** — 2 genuinely-new `.fetchall()` calls were unannotated (the other 30 were line-shifts). Annotated both bounded calls with valid reasons; regenerated the migration baseline.
2. **`test_setup_miner_downloads`** — stale Linux + Windows artifact sha256 pins (files changed by #7055 / #6957 / #6903). Updated pins to match; both files compile.
3. **`test_explorer_miners_normalization`** — `normalizeMinersResponse` now uses optional chaining + object filtering (matches the malformed-input contract).
4. **`test_hall_of_fame_*`** — frontend `API_LEADERBOARD` repointed to the `/api/hall_of_fame/leaderboard` compat endpoint (node serves it; response shape unchanged).

### DIFF
--- a/explorer/static/js/explorer.js
+++ b/explorer/static/js/explorer.js
@@ -113,21 +113,11 @@ function formatRelativeTime(ts) {
 }
 
 function normalizeMinersResponse(payload) {
-    if (!payload) return [];
-    let rows = [];
-    if (Array.isArray(payload)) {
-        rows = payload;
-    } else if (payload && typeof payload === 'object') {
-        if (Array.isArray(payload.miners)) {
-            rows = payload.miners;
-        } else if (Array.isArray(payload.data)) {
-            rows = payload.data;
-        } else if (typeof payload === 'object' && payload !== null) {
-            // Handle case where payload is a single miner object
-            rows = [payload];
-        }
-    }
-    return (rows || []).filter(row => row && typeof row === 'object');
+    // Tolerate legacy array, paginated {miners:[...]}, or {data:[...]} shapes.
+    const rows = Array.isArray(payload) ? payload :
+        (Array.isArray(payload?.miners) ? payload.miners :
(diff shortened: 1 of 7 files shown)


MolhamHamwi (commented):
Reviewed the CI repair bundle across `setup_miner.py`, `explorer/static/js/explorer.js`, `web/hall-of-fame/index.html`, and the fetchall baseline changes.

Specific observations:
- The updated Linux and Windows `setup_miner.py` SHA-256 pins match the current raw GitHub miner artifacts (`77cc08915f9a64a2aed3ceb8c6517bd9935d4447d9dfce70a50af3038914a1dc` and `a4c07a32f9e475efb60bb5bef548a91c8eecc15b978ba57eca4fb4c6603642fd`), so this looks like a real stale-pin repair rather than weakening the checksum guard.
- The `normalizeMinersResponse()` change is safer for the malformed-input contract: it still accepts legacy arrays and the `{ miners: [...] }` / `{ data: [...] }` envelopes, but it no longer treats an arbitrary object as a single miner row. That should prevent accidental rendering of wrapper/error objects as miner records.
- The new fetchall annotation in `get_epoch_history()` documents the actual bound (`current_epoch-50`, about 51 rows), and the header-key annotation points back to the pruning cap, so both new baseline exceptions have concrete limits instead of generic suppressions.

I also like that the hall-of-fame frontend endpoint now matches the existing `/api/hall_of_fame/leaderboard` API namespace while keeping the expected `leaderboard` payload shape unchanged.

jaxint (commented):
LGTM 👍

PR #7095, opened 2026-06-08
Fix Rust wallet /network/info endpoint

### LINKED ISSUE #7094: Rust wallet network command hits HTML 404 from /network/info
## Summary
`rtc-wallet network` cannot read live RustChain network metadata from rustchain.org because the Rust wallet client hardcodes `GET /network/info`, but the production site currently returns nginx's HTML 404 for that path.

## Live repro
```bash
curl -sk -D - https://rustchain.org/network/info -o -
```

Actual response observed:
```http
HTTP/1.1 404 Not Found
Content-Type: text/html
Server: nginx
```

The response body is nginx HTML, not wallet-consumable JSON.

## Client contract
The Rust wallet calls this route directly:

- `rustchain-wallet/src/client.rs`: `get_network_info()` builds `{api_url}/network/info`
- `rustchain-wallet/src/bin/rtc_wallet.rs`: `rtc-wallet network` calls `client.get_network_info()`

The client expects JSON compatible with `NetworkInfo`: `chain_id`, `network`, `block_height`, `peer_count`, `min_fee`, and `version`.

## Expected behavior
`https://rustchain.org/network/info` should be proxied to the node and return JSON network metadata so the Rust wallet network command works against the default public API URL.

## Root cause
The Flask node exposes adjacent status/stat APIs, but there is no `/network/info` route. The production nginx config also has explicit proxy locations for selected node routes and then falls through to static-site handling, so `/network/info` currently never reaches the node.

pqmfei:
## Summary
- add the Flask `/network/info` endpoint used by the Rust wallet client
- return the `NetworkInfo` JSON contract from live node state (`chain_id`, network name, block height, peer count, min fee, version)
- proxy `/network/` through the production rustchain.org nginx config so the public wallet URL reaches the node

Fixes #7094.

## Live repro before fix
```bash
curl -sk -D - https://rustchain.org/network/info -o -
```

Observed on 2026-06-08 UTC:
```http
HTTP/1.1 404 Not Found
Content-Type: text/html
Server: nginx
```

## Validation
```bash
pytest -q tests/test_api_stats_route_registration.py
python -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py tests/test_api_stats_route_registration.py
git diff --check fork/main..HEAD -- node/rustchain_v2_integrated_v2.2.1_rip200.py site/nginx-rustchain-org.conf tests/test_api_stats_route_registration.py
```

All passed locally.


pqmfei:
CI boundary note for this PR:

The focused test added/affected by this PR passes in the blocking pytest job:

- `tests/test_api_stats_route_registration.py::test_api_stats_route_returns_network_statistics` PASSED
- `tests/test_api_stats_route_registration.py::test_network_info_route_matches_rust_wallet_contract` PASSED

The failing `test` job is from existing full-suite baseline failures outside this PR's files:

- `tests/test_explorer_miners_normalization.py::test_explorer_normalizes_paginated_miners_response`
- `tests/test_fetchall_guard.py::test_fetchall_guard_passes_current_baseline`
- `tests/test_fetchall_guard.py::test_fetchall_guard_allows_annotated_call`
- `tests/test_hall_of_fame_api_docs.py::test_hall_of_fame_docs_use_leaderboard_api_path`
- `tests/test_hall_of_fame_frontend_endpoints.py::test_hall_of_fame_index_uses_compat_leaderboard_endpoint`
- `tests/test_macos_miner_chain_identity.py::test_macos_miner_uses_attested_wallet_for_eligibility`
- `tests/test_macos_miner_chain_identity.py::test_macos_miner_submits_headers_with_attested_wallet_identity`
- `tests/test_setup_miner_downloads.py::test_setup_miner_pins_current_miner_artifacts`
- `tests/test_setup_miner_downloads.py::test_setup_miner_pins_current_macos_artifact`

These failures do not exercise the `/network/info` endpoint or the nginx `/network/` proxy added here.

jdjioe5-cpu (approved):
## Review: APPROVE

This is a clean, well-scoped addition of the `/network/info` Flask endpoint required by the Rust wallet client. The fix is a real regression-resolver (the live `https://rustchain.org/network/info` URL was 404ing before this PR), the JSON contract matches what the Rust client expects, and the test coverage is appropriate for the surface area touched.

### Local validation (this review, fresh worktree at `f94df548`)

- `python3 -m pytest tests/test_api_stats_route_registration.py -v` → **2/2 passed in 0.17s** (the pre-existing `test_api_stats_route_returns_network_statistics` plus the new `test_network_info_route_matches_rust_wallet_contract`)
- `git diff --check` between merge-base and head → clean
- Manual test-client smoke (re-ran the new test body inline): `integrated_node.app.test_client().get("/network/info")` returns `200` with the exact `{"chain_id": "rustchain-mainnet-v2", "network": "mainnet", "block_height": 43210, "peer_count": 2, "min_fee": 1000, "version": "2.2.1-security-hardened"}` body when `CHAIN_ID` is monkeypatched to `rustchain-mainnet-v2` and `p2p_node.peers` is a 2-key dict.

### Findings

1. **Medium — `_network_peer_count` silently returns 0 on every exception, including the `peer_manager.get_active_peers()` path.** This is a real-world risk because P2P peer discovery is the one subsystem most likely to throw under load (timeout, decode error from a misbehaving peer). Returning `0` is indistinguishable from a healthy 0-peer network and may mask outages. A `_network_peer_count_stale` boolean or HTTP `warning` field in the response would let the wallet client degrade the UI ("peer count unavailable") without breaking the contract. A simple `logging.warning("p2p peer count failed", exc_info=True)` would also be a zero-cost improvement.

2. **Low — `chain_id` is reflected into the JSON response without case normalization.** `_network_name_from_chain_id` lowercases the chain_id for routing, but the live `chain_id` field is echoed verbatim. If a future CHAIN_ID is set to `RUSTCHAIN-MAINNET-V2` (uppercase), the wallet client and any cache key will see a different value on the same logical network. A single `.lower()` here, or just emitting the canonical lowercased value, would prevent that footgun. The current test only covers the already-lowercase shape, so this is a regression-prone seam.

3. **Low — `nginx` proxy `location /network/` is open without `proxy_ssl_server_name` or `proxy_set_header X-Forwarded-Proto`.** The block also has no `proxy_hide_header` to scrub the upstream's `Server` token. For an internal-only endpoint that's fine, but the PR doesn't document whether `/network/info` is intended to be public (wallet client over the public internet) or only via the LAN interface. If public, `proxy_set_header X-Forwarded-Proto $scheme;` + a per-IP allowlist on the `127.0.0.1:8099` upstream would close a small information-leak.

4. **Informational — `block_height` is derived from `current_slot()` rather than the chain's actual latest committed block.** That matches the existing pattern in `get_stats()` (this PR is in the same file, immediately above `get_stats()`), so the inconsistency is with the spec rather than the surrounding code. The Rust wallet consumer should be told which one it is, or the two endpoints will diverge on the same node. Worth a one-line doc note in the `NetworkInfo` contract.

5. **Informational — no negative test for an unknown `chain_id`.** The current test covers `chain_id="rustchain-mainnet-v2"` only. The default `_network_name_from_chain_id` falls through to `"mainnet"`, which means a typo'd chain_id silently reports mainnet. A test that asserts the negative case (e.g. `chain_id="rustchain-unknown-net"` → `"mainnet"`) is acceptable per the implementation but worth a comment so the next reader doesn't think it's a bug. The PR's body says "Fixes #7094" — if the wallet client needs an "unknown" sentinel, this is the seam to update.

### Verdict

APPROVE. The PR is correct, tested, and addresses the live `https://rustchain.org/network/info` 404. The findings above are improvements, not blockers. The only one I'd push back on is finding #1 — adding a one-line `logging.warning` to the exception swallow would close a silent-failure path for under-$5 of code.

jdjioe5-cpu (approved):
## Review: APPROVE

This is a clean, well-scoped addition of the `/network/info` Flask endpoint required by the Rust wallet client. The fix is a real regression-resolver (the live `https://rustchain.org/network/info` URL was 404ing before this PR), the JSON contract matches what the Rust client expects, and the test coverage is appropriate for the surface area touched.

### Local validation (this review, fresh worktree at `f94df548`)

- `python3 -m pytest tests/test_api_stats_route_registration.py -v` → **2/2 passed in 0.17s** (the pre-existing `test_api_stats_route_returns_network_statistics` plus the new `test_network_info_route_matches_rust_wallet_contract`)
- `git diff --check` between merge-base and head → clean
- Manual test-client smoke (re-ran the new test body inline): `integrated_node.app.test_client().get("/network/info")` returns `200` with the exact `{"chain_id": "rustchain-mainnet-v2", "network": "mainnet", "block_height": 43210, "peer_count": 2, "min_fee": 1000, "version": "2.2.1-security-hardened"}` body when `CHAIN_ID` is monkeypatched to `rustchain-mainnet-v2` and `p2p_node.peers` is a 2-key dict.

### Findings

1. **Medium — `_network_peer_count` silently returns 0 on every exception, including the `peer_manager.get_active_peers()` path.** This is a real-world risk because P2P peer discovery is the one subsystem most likely to throw under load (timeout, decode error from a misbehaving peer). Returning `0` is indistinguishable from a healthy 0-peer network and may mask outages. A `_network_peer_count_stale` boolean or HTTP `warning` field in the response would let the wallet client degrade the UI ("peer count unavailable") without breaking the contract. A simple `logging.warning("p2p peer count failed", exc_info=True)` would also be a zero-cost improvement.

2. **Low — `chain_id` is reflected into the JSON response without case normalization.** `_network_name_from_chain_id` lowercases the chain_id for routing, but the live `chain_id` field is echoed verbatim. If a future CHAIN_ID is set to `RUSTCHAIN-MAINNET-V2` (uppercase), the wallet client and any cache key will see a different value on the same logical network. A single `.lower()` here, or just emitting the canonical lowercased value, would prevent that footgun. The current test only covers the already-lowercase shape, so this is a regression-prone seam.

3. **Low — `nginx` proxy `location /network/` is open without `proxy_ssl_server_name` or `proxy_set_header X-Forwarded-Proto`.** The block also has no `proxy_hide_header` to scrub the upstream's `Server` token. For an internal-only endpoint that's fine, but the PR doesn't document whether `/network/info` is intended to be public (wallet client over the public internet) or only via the LAN interface. If public, `proxy_set_header X-Forwarded-Proto $scheme;` + a per-IP allowlist on the `127.0.0.1:8099` upstream would close a small information-leak.

4. **Informational — `block_height` is derived from `current_slot()` rather than the chain's actual latest committed block.** That matches the existing pattern in `get_stats()` (this PR is in the same file, immediately above `get_stats()`), so the inconsistency is with the spec rather than the surrounding code. The Rust wallet consumer should be told which one it is, or the two endpoints will diverge on the same node. Worth a one-line doc note in the `NetworkInfo` contract.

5. **Informational — no negative test for an unknown `chain_id`.** The current test covers `chain_id="rustchain-mainnet-v2"` only. The default `_network_name_from_chain_id` falls through to `"mainnet"`, which means a typo'd chain_id silently reports mainnet. A test that asserts the negative case (e.g. `chain_id="rustchain-unknown-net"` → `"mainnet"`) is acceptable per the implementation but worth a comment so the next reader doesn't think it's a bug. The PR's body says "Fixes #7094" — if the wallet client needs an "unknown" sentinel, this is the seam to update.

### Verdict

APPROVE. The PR is correct, tested, and addresses the live `https://rustchain.org/network/info` 404. The findings above are improvements, not blockers. The only one I'd push back on is finding #1 — adding a one-line `logging.warning` to the exception swallow would close a silent-failure path for under-$5 of code.

jaxint (commented):

jaxint (commented):

jaxint (commented):

---

jaxint:
## 🔍 PR Review

### ✅ What looks good:

### 💡 Suggestions:

---

jaxint (commented):
Great work! 👍

---

jaxint (commented):
Nice work!

BossChaos (approved):
Reviewed Code Review Bounty #73

Adds Flask `/network/info` endpoint needed by Rust wallet client. Live repro (curl to rustchain.org/network/info) confirms the 404 before fix and validates the endpoint after. Configuration changes in nginx proxy are straightforward and correct. Tests included.

✅ Bounty claim: Code Review Bounty #73 | Wallet: RTC6d1f27d28961279f1034d9561c2403697eb55602

Scottcjn:

PR #7096, opened 2026-06-08
security: fix predictable nonce in p2p.py

nkar123412-hub:
The P2P message nonce was generated using , which is predictable and could allow for message spoofing or replay attacks in certain edge cases. 

Replaced with  for cryptographically secure random numbers.


jaxint (commented):

PR #7097, opened 2026-06-08
security: fix malformed JSON field types in bcos_badge_generator.py

nkar123412-hub:
The  endpoint in  was vulnerable to server errors (500s) when provided with malformed JSON bodies (e.g., arrays instead of objects, or incorrect types for fields like  and ).

This PR implements strict type validation for all input fields, ensuring that:
- The request body is a JSON object.
-  is a string containing a .
-  is a string and one of the valid BCOS tiers (L0, L1, L2).
-  is a numeric type.
-  is a boolean.

All invalid inputs now return a structured HTTP 400 response instead of crashing the server.

### DIFF
--- a/tools/bcos_badge_generator.py
+++ b/tools/bcos_badge_generator.py
@@ -1147,30 +1147,38 @@ def generate_badge():
         return jsonify({'success': False, 'error': 'Tier must be a string'})
     tier = raw_tier.upper()
     raw_trust_score = data.get('trust_score', 75)
-    cert_id = data.get('cert_id', '')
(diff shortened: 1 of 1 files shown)


jaxint (commented):
Nice changes! Appreciate the clean approach. 🚀

nkar123412-hub:
Thanks for the merge! 🚀 Quick question regarding the 5 RTC reward: could you please clarify where these tokens are sent or how I can view my balance? I want to make sure my wallet is correctly linked. Thanks!

PR #7107, opened 2026-06-08
Fail closed on legacy CLI wallet send errors

### LINKED ISSUE #7106: Legacy CLI wallet reports successful send when /api/transaction 404s
## Summary
The legacy `tools/cli-wallet` `send` command reports a successful transaction when the configured node does not actually accept the transaction.

The code posts to `POST /api/transaction`, but live rustchain.org currently returns nginx HTML 404 for that path. On any non-2xx response, the CLI prints `Note: Mock transaction (node API not available)`, derives a fake tx id from the local mock signature, and then the command prints `Transaction sent successfully!`.

## Live repro of missing production endpoint
```bash
curl -sk -D - -X POST "https://rustchain.org/api/transaction" \
  --data "{}" -o -
```

Observed response:
```http
HTTP/1.1 404 Not Found
Content-Type: text/html
Server: nginx
```

## Code path
- `tools/cli-wallet/src/main.rs` builds `let url = format!("{}/api/transaction", node_url)`.
- If the response is not successful, the current code enters the mock-success branch and returns `Ok(tx_id)`.
- The caller then prints `Transaction sent successfully!`, even though nothing was submitted to the chain.

## Impact
A user can believe RTC was sent when the live node returned 404 or the node was unreachable. That is worse than a normal failed send because it creates a false confirmation and fake transaction id.

## Expected behavior
Transaction submission must fail closed on non-2xx responses and network errors. Mock balances may be useful for local display, but sends must not report success unless the node accepted the transaction response.

pqmfei:
## Summary
- stop the legacy CLI wallet from returning a fake tx id when transaction submission gets a non-2xx response
- fail closed on transaction network errors instead of printing mock-send success
- update README development-mode wording so it no longer promises simulated successful sends
- add a focused pytest guard for the fail-closed behavior and README claim

Fixes #7106.

## Live repro before fix
```bash
curl -sk -D - -X POST "https://rustchain.org/api/transaction" \
  --data "{}" -o -
```

Observed:
```http
HTTP/1.1 404 Not Found
Content-Type: text/html
Server: nginx
```

Before this patch, the legacy CLI turns that class of response into `Ok(tx_id)`, so the command prints `Transaction sent successfully!` even though the node rejected or never received the transaction.

## Validation
Validated on a local merge-check branch with current `origin/main` plus this PR patch:

```bash
pytest -q tests/test_cli_wallet_send_fail_closed.py
$env:CARGO_BUILD_JOBS="1"; $env:RUSTFLAGS="-Cdebuginfo=0"; $env:RUST_MIN_STACK="8388608"; cargo test --manifest-path tools/cli-wallet/Cargo.toml
git diff --check origin/main..HEAD -- tools/cli-wallet/src/main.rs tools/cli-wallet/README.md tests/test_cli_wallet_send_fail_closed.py
```

Results:
- pytest guard: 2 passed
- cli-wallet cargo tests: 7 passed
- diff check: passed

### DIFF
--- a/tests/test_cli_wallet_send_fail_closed.py
+++ b/tests/test_cli_wallet_send_fail_closed.py
@@ -1,0 +1,23 @@
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parents[1]
+CLI_MAIN = ROOT / "tools" / "cli-wallet" / "src" / "main.rs"
+CLI_README = ROOT / "tools" / "cli-wallet" / "README.md"
+
+
+def test_cli_wallet_send_does_not_mock_successful_transactions():
+    source = CLI_MAIN.read_text(encoding="utf-8")
+
+    assert 'println!("Note: Mock transaction' not in source
+    assert "hex::encode(&signed_transaction.signature[0..8])" not in source
+    assert "Transaction endpoint returned HTTP" in source
+    assert "Transaction submission failed" in source
+    assert "if !status.is_success()" in source
+
+
+def test_cli_wallet_readme_does_not_claim_mock_send_success():
+    readme = CLI_README.read_text(encoding="utf-8")
+
+    assert "Simulated successful transactions" not in readme
+    assert "Transaction submission fails closed" in readme
--- a/tools/cli-wallet/README.md
+++ b/tools/cli-wallet/README.md
@@ -113,10 +113,11 @@ RustChain addresses use the canonical 43-character format: `RTC` followed by
 
 When the RustChain node is not available, the wallet operates in development mode with:
 - Mock balance of 1000 RTC
-- Simulated successful transactions
+- Transaction submission fails closed instead of reporting a mock success
 - Local address validation
 
-This allows testing wallet functionality without a running blockchain node.
+This allows testing wallet generation, address validation, and balance display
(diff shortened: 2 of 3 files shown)


pqmfei:
CI boundary note for this PR:

- The focused regression guard for this change passes locally: `pytest -q tests/test_cli_wallet_send_fail_closed.py` -> 2 passed.
- The CLI wallet crate test suite passes locally with constrained Windows build settings: `$env:CARGO_BUILD_JOBS=1; $env:RUSTFLAGS=-Cdebuginfo=0; $env:RUST_MIN_STACK=8388608; cargo test --manifest-path tools/cli-wallet/Cargo.toml` -> 7 passed.
- `git diff --check origin/main..HEAD -- tools/cli-wallet/src/main.rs tools/cli-wallet/README.md tests/test_cli_wallet_send_fail_closed.py` passes.
- GitHub Actions collected and ran the new guard without failing it. The current `test` job failure is in unrelated baseline areas (`tests/test_airdrop_bridge_admin_auth.py` and `tests/test_bcos_badge_generator.py`), which do not touch `tools/cli-wallet` or the new send fail-closed regression test.

So the remaining red CI appears unrelated to this CLI wallet transaction-submission fix.

jaxint (commented):

PR #7446, opened 2026-06-15
fix(api): reject invalid /api/miners pagination limits

### LINKED ISSUE #7445: [BUG] /api/miners silently coerces malformed limit query parameter
## Summary

`GET /api/miners` on the RustChain production host silently coerces malformed `limit` (and likely other) query parameters via Flask's `type=int` conversion. No 400 is ever returned. Same defect pattern as Bottube #1456 (`/api/feed`).

## Live evidence (2026-06-15 07:00 CST)

```
$ for v in abc -1 0 999999999999999 1.5; do
    code=$(curl -s -o /dev/null -w "%{http_code}" "https://rustchain.org/api/miners?limit=$v")
    cnt=$(curl -s "https://rustchain.org/api/miners?limit=$v" | python3 -c 'import json,sys;print(len(json.load(sys.stdin).get("miners",[])))')
    echo "limit=$v -> HTTP $code, miners=$cnt"
  done
limit=abc -> HTTP 200, miners=19
limit=-1 -> HTTP 200, miners=1
limit=0 -> HTTP 200, miners=1
limit=999999999999999 -> HTTP 200, miners=19
limit=1.5 -> HTTP 200, miners=19
```

All 5 input shapes return HTTP 200 with the default applied. `limit=abc` returns 19 miners (the full enrolled set), `limit=-1` returns 1 (clamped), `limit=0` returns 1 (clamped).

## Source-level root cause

In `app/main.py` or equivalent (the `/api/miners` route), the handler reads `limit` (and likely `offset`, `page`, `since`, `before`) via Flask's `request.args.get(key, default, type=int)`. Flask's coercion falls back to the default on parse failure (no 400, no error response).

## Impact

- Functional tier (Bounty #1102 5 RTC): documented contract for `limit` is rejected silently.
- `limit=-1` clamps to 1 via `min(N, max(1, ...))` pattern; this is silent clamping that defeats client-side validation.
- `limit=abc` returns the entire result set without pagination enforcement, which can be exploited to scrape the full miner list without rate-limit awareness.

## Suggested fix

Replace each `request.args.get(key, default, type=int)` with explicit `try/except ValueError` that returns `jsonify({"error": "..."}), 400`. Pattern is already used elsewhere in RustChain for query param validation.

## Out of scope

Not asking to fix other Flask routes; this issue is scoped to `/api/miners` only.

## Related

- Bottube #1456 — same defect pattern (separate repo, separate fix).

wind108369:
## Summary
- reject /api/miners limit values below 1 instead of silently clamping to 1
- reject limit values above 1000 instead of silently clamping to the max page size
- reject negative offsets instead of silently coercing them to 0
- add focused tests for malformed and out-of-range pagination parameters

Fixes #7445

## Validation
- .venv-codex312/bin/python -m pytest node/tests/test_api_miners_rate_limit.py -q
- .venv-codex312/bin/python -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py
- git diff --check

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -7535,8 +7535,12 @@ def api_miners():
         offset = int(raw_offset) if raw_offset not in (None, "") else 0
     except (ValueError, TypeError):
         return jsonify({"ok": False, "error": "offset must be an integer"}), 400
-    limit = min(max(limit, 1), 1000)
-    offset = max(offset, 0)
+    if limit < 1:
+        return jsonify({"ok": False, "error": "limit must be >= 1"}), 400
+    if limit > 1000:
+        return jsonify({"ok": False, "error": "limit must be <= 1000"}), 400
+    if offset < 0:
+        return jsonify({"ok": False, "error": "offset must be >= 0"}), 400
 
     with sqlite3.connect(DB_PATH) as conn:
         conn.row_factory = sqlite3.Row
--- a/node/tests/test_api_miners_rate_limit.py
+++ b/node/tests/test_api_miners_rate_limit.py
@@ -95,6 +95,24 @@ class TestApiMinersRateLimit(unittest.TestCase):
         self.assertEqual(resp.status_code, 400)
         self.assertEqual(resp.get_json(), {"ok": False, "error": "limit must be an integer"})
 
+    def test_api_miners_rejects_out_of_range_limits(self):
+        cases = [
+            ("-1", "limit must be >= 1"),
+            ("0", "limit must be >= 1"),
+            ("1001", "limit must be <= 1000"),
+            ("999999999999999", "limit must be <= 1000"),
+        ]
+
+        for index, (value, error) in enumerate(cases):
+            with self.subTest(limit=value):
+                resp = self.client.get(
+                    f"/api/miners?limit={value}",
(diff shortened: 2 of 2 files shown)


mouseos (approved):
Approved after reviewing head `1cdd8237db503f3ab48b531e891c92b4c592dd52` for PR #7446.

What I checked:
- `/api/miners` now rejects out-of-range pagination values instead of silently clamping them: `limit < 1`, `limit > 1000`, and `offset < 0` all return explicit 400 responses.
- The existing behavior for omitted or empty pagination parameters is preserved: default `limit=100` and `offset=0` still apply when the query values are absent or empty.
- The SQL query still uses bound `LIMIT ? OFFSET ?` parameters after validation, so this change tightens API input semantics without changing the database access shape.
- The new tests cover malformed limits, out-of-range limits, malformed offsets, negative offsets, default empty values, and the existing success/rate-limit path in the same focused test module.

Local validation run on this branch:
- `python -m pytest node/tests/test_api_miners_rate_limit.py -q` -> `7 passed, 4 subtests passed`
- `python -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py` -> passed
- `git diff --check origin/main...HEAD -- node/rustchain_v2_integrated_v2.2.1_rip200.py node/tests/test_api_miners_rate_limit.py` -> clean

The GitHub check suite is also green at the reviewed head. I do not see a blocker in this API validation change.

RTC wallet: `RTC849344ea4cc70514183eb03df0e8d107a398ac12`

PR #5974, opened 2026-05-20
Fix offline epoch catch-up with vote certificates

jamilahmadzai:
## Summary
- Attach signed accept-vote messages as a `vote_certificate` on `EPOCH_COMMIT` broadcasts.
- Verify each certificate vote against the commit epoch/proposal/quorum before accepting finality, so a node that missed local voting can derive finality from peer evidence instead of trusting a raw finalized snapshot.
- Preserve the existing #5750 safety path: state sync still cannot inject finalized epochs, and commit messages without a certificate still require locally observed quorum votes.
- Refresh the macOS miner v2.5 checksum pins so the full CI suite stays aligned with the current artifact hash.

Fixes #5950.

## Validation
- `uv run --no-project --with pytest --with flask --with requests --with pynacl python -m pytest node/tests/test_p2p_state_epoch_sync.py node/tests/test_p2p_hardening_phase2.py -q` -> 17 passed
- `uv run --no-project --with pytest --with flask python -m pytest -q tests/test_install_miner_checksums.py tests/test_setup_miner_downloads.py --tb=short -p no:cacheprovider` -> 5 passed
- `python3 -m py_compile setup_miner.py node/rustchain_p2p_gossip.py node/tests/test_p2p_state_epoch_sync.py`
- `git diff --check`
- `uv run --no-project python tools/bcos_spdx_check.py --base-ref origin/main` -> OK
- `uv run --no-project --with ruff ruff check node/rustchain_p2p_gossip.py node/tests/test_p2p_state_epoch_sync.py --select E9,F821,F811,F841 --output-format=concise` -> OK

## Bounty payment
wallet: RTCd1acb2189e9f36df2b5393c3c27a867c3c32b116
Payment method: RTC wallet `RTCd1acb2189e9f36df2b5393c3c27a867c3c32b116`

PR #7471, opened 2026-06-15
Fix offline epoch catch-up with bounded vote certificates

### LINKED ISSUE #5950: Add validated epoch catch-up sync path (follow-up to #5750)
## Context

PR #5750 (merged) closed a finality-forgery vulnerability: generic `STATE` sync was importing `finalized` epoch data from any peer's snapshot, with only a sender-signature check. A valid peer signature authenticates *who sent the snapshot*, not that the epoch was actually committed by quorum. #5750 removed that path — epochs now enter `epoch_crdt` only via the validated `EPOCH_COMMIT` handler, which requires the receiver to have locally observed accept votes from a quorum.

## The gap this creates

A node that was **offline during epoch voting + commit** has no entries in its local `_epoch_votes` for that epoch. When it comes back online and a peer sends `EPOCH_COMMIT`, the new `_handle_epoch_commit` validation rejects it with `unverified_voters` — the recovering node never saw the votes, so it can't confirm the quorum.

Result: **offline nodes can no longer catch up epoch finality.** This is an acceptable short-term tradeoff (closing a forgery hole beats easy catch-up), but it needs a proper fix.

## What's needed

A *validated* catch-up sync path — something that lets a recovering node re-establish finality for missed epochs **without** trusting an unverified snapshot. Options to evaluate:

1. **Vote-record sync** — let a recovering node request the raw signed accept-votes for an epoch from peers, verify each signature itself, then apply the same quorum check `_handle_epoch_commit` uses. Finality is then locally derived, not imported.
2. **Commit-certificate** — have the committing node assemble a quorum certificate (the set of signed accept-votes) and attach it to `EPOCH_COMMIT`. A recovering node verifies the certificate's signatures + quorum independently.
3. **Checkpoint sync** — periodic signed checkpoints with their own multi-sig quorum, for nodes that are very far behind.

Option 2 (commit-certificate) is likely the cleanest — it makes `EPOCH_COMMIT` self-verifying instead of depending on the receiver's local vote history.

## Acceptance

- A node offline during epoch N's voting can rejoin and establish finality for epoch N from peer data **without** any path that accepts unverified `finalized` flags.
- Regression test: the `test_signed_state_sync_cannot_inject_epoch_finality` invariant from #5750 still holds.

Filed as a follow-up to keep #5750's security fix shippable now while the catch-up path is designed properly.

jamilahmadzai:
## Summary
- Replaces the closed #5974 with the same epoch catch-up certificate fix replayed onto current `main`.
- Attach signed `EPOCH_VOTE` messages as a `vote_certificate` on `EPOCH_COMMIT` broadcasts so offline peers can verify quorum evidence instead of trusting the committer's voter list.
- Require every certificate vote to have a trusted Ed25519 identity while preserving freshness-neutral verification only for embedded historic vote messages.
- Address the #5974 close feedback: remove the duplicate `_has_trusted_ed25519_identity` definition, reject oversized certificates before expensive verification, and reject duplicate certificate voters early.

Fixes #5950.
Replaces #5974.

## Validation
- `uv run --no-project --with pytest --with flask --with requests --with pynacl --with cryptography python -m pytest node/tests/test_p2p_state_epoch_sync.py node/tests/test_p2p_hardening_phase2.py -q` -> 25 passed
- `python3 -m py_compile node/rustchain_p2p_gossip.py node/tests/test_p2p_state_epoch_sync.py node/tests/test_p2p_hardening_phase2.py`
- `git diff --check origin/main...HEAD -- node/rustchain_p2p_gossip.py node/tests/test_p2p_state_epoch_sync.py node/tests/test_p2p_hardening_phase2.py`
- `uv run --no-project python tools/bcos_spdx_check.py --base-ref origin/main` -> OK
- `uv run --no-project --with ruff ruff check node/rustchain_p2p_gossip.py node/tests/test_p2p_state_epoch_sync.py node/tests/test_p2p_hardening_phase2.py --select E9,F821,F811,F841 --output-format=concise` -> OK

## Bounty payment
wallet: RTCd1acb2189e9f36df2b5393c3c27a867c3c32b116
Payment method: RTC wallet `RTCd1acb2189e9f36df2b5393c3c27a867c3c32b116`

### DIFF
--- a/node/rustchain_p2p_gossip.py
+++ b/node/rustchain_p2p_gossip.py
@@ -19,7 +19,7 @@ import secrets
 import sqlite3
 import threading
 import time
-from dataclasses import dataclass, asdict, field
+from dataclasses import dataclass, asdict
 from enum import Enum
 from typing import Dict, List, Optional, Set, Tuple, Any
 from collections import defaultdict, OrderedDict
@@ -611,6 +611,7 @@ class GossipLayer:
         self.balance_crdt = PNCounter()
         self.epoch_crdt = GSet()
         self._epoch_votes: Dict[Tuple[int, str], Dict[str, str]] = {}
+        self._epoch_vote_messages: Dict[Tuple[int, str], Dict[str, GossipMessage]] = {}
 
         # Phase F (#2256): per-peer Ed25519 identity, dual-mode signing.
         # Only loaded/generated when needed by the current signing mode;
@@ -812,8 +813,8 @@ class GossipLayer:
             "handshake": local_handshake_params(),
         })
 
-    def verify_message(self, msg: GossipMessage) -> bool:
-        """Verify message signature and freshness.
+    def _verify_message_signature(self, msg: GossipMessage, *, check_freshness: bool = True) -> bool:
+        """Verify message signature, optionally skipping freshness for commit certificates.
 
         SECURITY (#2256 + #2272): verifies sender_id, msg_id, and ttl as
         part of the signed content — any post-sign flip of those fields
@@ -824,7 +825,7 @@ class GossipLayer:
         to verify. HMAC remains a legacy fallback only for HMAC-only messages
         outside strict mode.
         """
-        if abs(time.time() - msg.timestamp) > MESSAGE_EXPIRY:
+        if check_freshness and abs(time.time() - msg.timestamp) > MESSAGE_EXPIRY:
             return False
 
         content = self._signed_content(msg.msg_type, msg.sender_id, msg.msg_id, msg.ttl, msg.payload)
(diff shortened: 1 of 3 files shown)


Scottcjn:
Thanks @jamilahmadzai — and thanks for addressing the bounded-cert feedback from the last round; the DoS-bounding is the right fix. Closing under scope policy though: the vote-certificate / offline-epoch-catch-up code lives in node/rustchain_p2p_gossip.py, which isn't imported into the deployed production node (rustchain_v2_integrated_v2.2.1_rip200.py) — grep finds no vote_certificate there. So it's not a deployed surface yet (0 RTC per policy). If/when the p2p gossip layer is wired into the production node, this becomes very relevant — keep it handy. Also double-check the duplicate _has_trusted_ed25519_identity definition flagged last time is fully removed.

PR #7554, opened 2026-06-23
fix(websocket): tolerate malformed numeric env

yyswhsccc:
## Problem

`node/websocket_feed.py` parsed `WEBSOCKET_PORT`, `WS_POLL_INTERVAL`, and `WS_MAX_EVENTS` directly at import time. A malformed deployment env value raises `ValueError` before the runtime can fall back or even start with WebSocket features disabled.

## Impact

A single bad numeric env value can crash the RustChain node WebSocket feed on import/startup, making the realtime block/miner/attestation feed unavailable.

## Fix

Add small local `_env_int` / `_env_float` helpers that log malformed values and fall back to the existing defaults. Valid numeric env values are preserved.

## Tests

- `uv run --no-project --with pytest --with flask python -B -m pytest -q tests/test_node_websocket_feed_env.py` -> 2 passed

## Boundaries

Related to the general bug bounty surface (#305). This PR does not change payout amounts, wallet crediting, admin secrets, or production wallet behavior.

### DIFF
--- a/node/websocket_feed.py
+++ b/node/websocket_feed.py
@@ -41,11 +41,34 @@ except ImportError:
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+
+def _env_int(name: str, default: int) -> int:
+    raw_value = os.environ.get(name)
+    if raw_value is None:
+        return default
+    try:
+        return int(raw_value)
+    except (TypeError, ValueError):
+        logger.warning("Invalid %s=%r; using default %s", name, raw_value, default)
+        return default
+
+
+def _env_float(name: str, default: float) -> float:
+    raw_value = os.environ.get(name)
+    if raw_value is None:
+        return default
+    try:
+        return float(raw_value)
+    except (TypeError, ValueError):
+        logger.warning("Invalid %s=%r; using default %s", name, raw_value, default)
+        return default
+
+
 # Configuration
-WS_PORT = int(os.environ.get('WEBSOCKET_PORT', 8765))
+WS_PORT = _env_int('WEBSOCKET_PORT', 8765)
 API_BASE = os.environ.get('RUSTCHAIN_API_BASE', 'http://localhost:8088')
-POLL_INTERVAL = float(os.environ.get('WS_POLL_INTERVAL', '3'))
-MAX_EVENTS = int(os.environ.get('WS_MAX_EVENTS', '100'))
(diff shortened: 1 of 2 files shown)


yyswhsccc:

This is a focused current-main runtime hardening PR: `node/websocket_feed.py` no longer import-crashes when `WEBSOCKET_PORT`, `WS_POLL_INTERVAL`, or `WS_MAX_EVENTS` are malformed. Valid numeric env values are preserved, malformed values log and fall back to the existing defaults.

Validation:
- `uv run --no-project --with pytest --with flask python -B -m pytest -q tests/test_node_websocket_feed_env.py` -> 2 passed

No payout amount, wallet crediting, admin-secret, or production-wallet behavior changes.

yyswhsccc:
### Maintenance update

**Maintenance addressed**
- Missing wallet line.

**Current head**
- `97dce6a`

**Validation**
- `uv run --no-project --with pytest --with flask python -B -m pytest -q tests/test_node_websocket_feed_env.py` → `2 passed`
- `git diff --check` → `passed`

**Why this change**
- This keeps the PR metadata aligned with reviewer feedback and the current PR scope without broadening the code diff.

**Scope**
- This maintenance update only changes PR metadata/body text; it does not broaden the code diff.

FakerHideInBush (changes requested):
Reviewing exact head `97dce6a40cb1eeec7c6c1741aca81b2fdefab44c`.

The import-crash fallback for plain non-numeric values is a good direction, and the new tests cover the common string failure path. One correctness gap remains before this fully hardens malformed numeric env input:

`_env_float()` still accepts non-finite values such as `WS_POLL_INTERVAL=nan` or `WS_POLL_INTERVAL=inf`. `float(nan)` and `float(inf)` do not raise `ValueError`, so the module imports with `POLL_INTERVAL` set to a non-finite value instead of falling back to `3.0`. That can later reach the polling loop / sleep path as an invalid interval, which is the same class of malformed-env startup/runtime hazard this PR is trying to close.

Please reject non-finite float values, for example with `math.isfinite(parsed)`, and add regression coverage for `nan` and `inf` (at least for `WS_POLL_INTERVAL`). Keeping the valid numeric test is useful too, since finite decimal values like `0.25` should still be preserved.

jaxint (commented):

jaxint (commented):

jaxint (commented):

PR #7555, opened 2026-06-23
fix(beacon): tolerate malformed key ttl env

yyswhsccc:
## Problem

`node/beacon_identity.py` parsed `BEACON_KEY_TTL` with `int(...)` at import time. A malformed TTL env value raises `ValueError`, which can crash beacon identity/key-management callers before the TOFU key store is usable.

## Impact

A bad deployment env value can take down beacon identity helpers and CLI/test entrypoints that import the module. This is especially brittle around key TTL / rotation / revocation maintenance paths.

## Fix

Add a small `_env_int` helper that logs malformed values and falls back to the existing 30-day TTL default. Valid numeric env values are preserved.

## Tests

- `uv run --no-project --with pytest --with cryptography python -B -m pytest -q node/tests/test_beacon_identity.py` -> 34 passed

## Boundaries

Related to the general bug bounty surface (#305). This PR does not change beacon key semantics, payout amounts, wallet crediting, admin secrets, or production wallet behavior.

### DIFF
--- a/node/beacon_identity.py
+++ b/node/beacon_identity.py
@@ -24,8 +24,20 @@ from typing import Any, Dict, List, Optional, Tuple
 
 log = logging.getLogger("beacon.identity")
 
+
+def _env_int(name: str, default: int) -> int:
+    raw_value = os.environ.get(name)
+    if raw_value is None:
+        return default
+    try:
+        return int(raw_value)
+    except (TypeError, ValueError):
+        log.warning("Invalid %s=%r; using default %s", name, raw_value, default)
+        return default
+
+
 # Default key TTL: 30 days in seconds
-DEFAULT_KEY_TTL: int = int(os.environ.get("BEACON_KEY_TTL", str(30 * 24 * 60 * 60)))
+DEFAULT_KEY_TTL: int = _env_int("BEACON_KEY_TTL", 30 * 24 * 60 * 60)
 
 DB_PATH: str = os.environ.get("BEACON_DB_PATH", "/root/rustchain/rustchain_v2.db")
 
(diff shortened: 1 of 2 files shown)


yyswhsccc:

This is another focused current-main import-safety fix: `node/beacon_identity.py` now tolerates malformed `BEACON_KEY_TTL` by logging and falling back to the existing 30-day default, while preserving valid numeric values.

Validation:
- `uv run --no-project --with pytest --with cryptography python -B -m pytest -q node/tests/test_beacon_identity.py` -> 34 passed

No beacon key semantic changes, payout amount changes, wallet crediting, admin-secret, or production-wallet behavior changes.

yyswhsccc:
### Maintenance update

**Maintenance addressed**
- Missing wallet line.

**Current head**
- `5cd3757`

**Validation**
- `uv run --no-project --with pytest --with cryptography python -B -m pytest -q node/tests/test_beacon_identity.py` → `34 passed`
- `git diff --check` → `passed`

**Why this change**
- This keeps the PR metadata aligned with reviewer feedback and the current PR scope without broadening the code diff.

**Scope**
- This maintenance update only changes PR metadata/body text; it does not broaden the code diff.

jaxint (commented):

FakerHideInBush (changes requested):
Reviewing exact head `5cd375770959f57ab9032d53c8cb04db9e026f6e`.

The import-safety improvement for a non-numeric `BEACON_KEY_TTL` is useful, but the malformed-env guard is still incomplete for numeric-but-invalid TTL values.

`_env_int()` accepts `0` and negative integers, so `BEACON_KEY_TTL=0` or `BEACON_KEY_TTL=-1` becomes the module-level `DEFAULT_KEY_TTL` instead of falling back to the 30-day default. That can make new beacon keys immediately expired, or produce expiry timestamps in the past, which is the same startup/configuration hazard this PR is trying to avoid.

Please require a positive TTL for `BEACON_KEY_TTL` (for example via a min bound on the helper or a call-site check) and add regression tests for `0` and a negative value while keeping the valid positive numeric case preserved.

jaxint (commented):

PR #7556, opened 2026-06-23
fix(p2p): tolerate malformed numeric env

yyswhsccc:
## Problem

`node/rustchain_p2p_init.py` parsed P2P numeric env values directly with `int(...)`:

- `RC_P2P_PORT` during `init_p2p(...)`
- `RC_P2P_UPNP_DISCOVER_DELAY_MS` during UPnP discovery

Malformed deployment env values can make P2P init fail or silently disable UPnP discovery via the broad discovery exception path.

## Impact

A single bad numeric env value can prevent a node from starting its P2P helper or from advertising a reachable public P2P URL.

## Fix

Add a small `_env_int` helper that logs malformed numeric env values and falls back to the existing defaults. Valid numeric values are preserved.

## Tests

- `uv run --no-project --with pytest --with flask --with requests python -B -m pytest -q node/tests/test_p2p_nat_upnp.py` -> 6 passed

## Boundaries

Related to the general bug bounty surface (#305). This PR does not change peer trust, payout amounts, wallet crediting, admin secrets, or production wallet behavior.

### DIFF
--- a/node/rustchain_p2p_init.py
+++ b/node/rustchain_p2p_init.py
@@ -29,6 +29,17 @@ def _env_flag(name, default=False):
     return value.strip().lower() in {"1", "true", "yes", "on"}
 
 
+def _env_int(name, default):
+    value = os.environ.get(name)
+    if value is None:
+        return default
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        print(f"[P2P] Invalid {name}={value!r}; using default {default}")
+        return default
+
+
 def _is_private_host(host):
     try:
         return ipaddress.ip_address(host).is_private
@@ -54,7 +65,7 @@ def _discover_upnp_external_url(local_ip, local_port):
 
     try:
         upnp = miniupnpc.UPnP()
-        upnp.discoverdelay = int(os.environ.get("RC_P2P_UPNP_DISCOVER_DELAY_MS", "200"))
+        upnp.discoverdelay = _env_int("RC_P2P_UPNP_DISCOVER_DELAY_MS", 200)
         if upnp.discover() <= 0:
             return None
         upnp.selectigd()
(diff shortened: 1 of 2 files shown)


yyswhsccc:

This is a focused current-main P2P startup hardening PR: malformed `RC_P2P_PORT` and `RC_P2P_UPNP_DISCOVER_DELAY_MS` now log and fall back to existing defaults instead of breaking P2P init/UPnP discovery. Valid numeric values are preserved.

Validation:
- `uv run --no-project --with pytest --with flask --with requests python -B -m pytest -q node/tests/test_p2p_nat_upnp.py` -> 6 passed

No peer-trust, payout amount, wallet crediting, admin-secret, or production-wallet behavior changes.

yyswhsccc:
### Correction

This maintenance note was misclassified. No additional code change was made here; the PR remains the focused P2P numeric-env guard described in the PR body and the owner-review note above.

Current validation remains:
- `uv run --no-project --with pytest --with flask --with requests python -B -m pytest -q node/tests/test_p2p_nat_upnp.py` -> `6 passed`
- `git diff --check` -> `passed`

Please use the PR body as the source of truth for this focused change.

jaxint (commented):

FakerHideInBush (changes requested):
Reviewing exact head `d5182f7fd3f3cc456c5229d1b63b033764f07435`.

The non-numeric fallback is useful, and the targeted UPnP test is a good start. One acceptance gap remains for the P2P numeric env hardening:

`_env_int()` accepts any integer, so numeric-but-invalid values still escape the fallback path. For `RC_P2P_PORT`, values like `0`, `-1`, or `70000` produce invalid advertised URLs / port mappings instead of falling back to `8099`. For `RC_P2P_UPNP_DISCOVER_DELAY_MS`, negative values can be assigned directly to `upnp.discoverdelay`, which is not a valid discovery delay.

Please add range validation at the call sites (or as optional min/max bounds on the helper) and add focused tests for invalid numeric values. Suggested contracts: `RC_P2P_PORT` must be `1..65535`, and `RC_P2P_UPNP_DISCOVER_DELAY_MS` should reject negative delays while preserving valid positive integers.

jaxint (commented):

PR #7590, opened 2026-06-24
fix(node): avoid wildcard websocket CORS

yyswhsccc:
## Problem
- node/websocket_feed.py initializes Flask-SocketIO with cors_allowed_origins="*".
- The feed can expose live chain, transaction, miner, and attestation events, so a permissive default is broader than needed.

## Related evidence
- Repository-local evidence: selector-owned current-main code audit found this focused RustChain risk surface and generated the matching regression patch.

## Impact
- Any browser origin can open a Socket.IO connection to this node feed when it is deployed with the default configuration.

## Fix
- Replace the wildcard default with an explicit allowlist: rustchain.org, www.rustchain.org, localhost, and 127.0.0.1 for the configured websocket port.
- Add WEBSOCKET_CORS_ORIGINS for explicit deployment allowlists.
- Reject wildcard entries instead of silently accepting them.

## Tests
- python3 -m py_compile node/websocket_feed.py tests/test_node_websocket_feed_env.py node/tests/test_websocket_feed_json_rpc.py
- uv run --no-project --with pytest --with flask --with flask-socketio python -B -m pytest -q tests/test_node_websocket_feed_env.py node/tests/test_websocket_feed_json_rpc.py -> 19 passed
- git diff --check

## Maintenance-only CI baseline note
- This branch also includes a follow-up fetchall guard baseline annotation in `node/rustchain_v2_integrated_v2.2.1_rip200.py`.
- The maintenance commit only marks two existing LIMIT-bounded `.fetchall()` reads as `fetchall-ok: already-paginated`; no business logic for this PR changed.

## Additional maintenance validation
- `bash scripts/check_fetchall.sh` -> passed
- `python3 -m py_compile node/rustchain_v2_integrated_v2.2.1_rip200.py` -> passed
- `git diff --check` -> passed

## Risk
- Narrow Socket.IO boundary hardening only; existing JSON-RPC subscription behavior is covered by the focused tests.

wallet: RTC47bc28896a1a4bf240d1fd780f4559b242bcd945

PR #7657, opened 2026-06-26
fix(node): avoid wildcard CORS by default

yyswhsccc:
Clean redo of #7590 after Scott's scope-hygiene feedback.

This branch intentionally keeps the original technical fix small and excludes the unrelated shared CI/baseline/consensus-node edits that caused the previous PR to be closed.

Selector provenance:
- natural_owner: `both_selectors`
- selection_bucket: `both_selectors`
- selected_by_lgbm: `True`
- selected_by_native: `True`
- dispatch_arm: `native_druid_selector`
- selector_run_id: `6ac915456973ebaf`

Scoped files:
- `node/websocket_feed.py`

Validation:
- `python3 -m py_compile node/websocket_feed.py -> passed`
- `GitHub Contents API path filter -> source/test files only`

Boundaries:
- No payout amount changes.
- No admin keys or production wallet changes.
- No manual wallet crediting.
- No `node/rustchain_v2_integrated_v2.2.1_rip200.py` or `scripts/baselines/*` maintenance diff.

@Scottcjn this is the clean, scoped redo of the earlier closed PR.

### DIFF
--- a/node/websocket_feed.py
+++ b/node/websocket_feed.py
@@ -42,6 +42,50 @@ logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
+TRUSTED_WEBSOCKET_ORIGINS = (
+    "https://rustchain.org",
+    "https://www.rustchain.org",
+)
+
+
+def default_websocket_cors_origins(local_port: int | str | None = None) -> List[str]:
+    origins = list(TRUSTED_WEBSOCKET_ORIGINS)
+    if local_port is not None:
+        origins.extend(
+            [
+                f"http://localhost:{local_port}",
+                f"http://127.0.0.1:{local_port}",
+            ]
+        )
+    return origins
+
+
+def parse_websocket_cors_origins(
+    raw_value: str | None = None,
+    *,
+    local_port: int | str | None = None,
+) -> List[str]:
+    if raw_value is None:
+        raw_value = os.environ.get("WEBSOCKET_CORS_ORIGINS")
+
+    if raw_value is None or not raw_value.strip():
+        return default_websocket_cors_origins(local_port)
+
+    origins: List[str] = []
+    for item in raw_value.split(","):
+        origin = item.strip()
+        if not origin:
+            continue
+        if origin == "*":
+            raise ValueError("WEBSOCKET_CORS_ORIGINS must not include '*'")
+        if origin not in origins:
+            origins.append(origin)
+
+    if not origins:
+        return default_websocket_cors_origins(local_port)
+    return origins
+
+
 def _env_int(name: str, default: int) -> int:
     raw_value = os.environ.get(name)
     if raw_value is None:
@@ -69,6 +113,7 @@ WS_PORT = _env_int('WEBSOCKET_PORT', 8765)
 API_BASE = os.environ.get('RUSTCHAIN_API_BASE', 'http://localhost:8088')
 POLL_INTERVAL = _env_float('WS_POLL_INTERVAL', 3.0)
 MAX_EVENTS = _env_int('WS_MAX_EVENTS', 100)
+WEBSOCKET_CORS_ORIGINS = parse_websocket_cors_origins(local_port=WS_PORT)
 JSON_RPC_VERSION = '2.0'
 MINING_STATS_SUBSCRIPTION = 'mining_stats'
 BLOCK_SUBSCRIPTION = 'blocks'
@@ -234,7 +279,7 @@ class WebSocketFeed:
         self.app = app
         self.socketio = SocketIO(
             app, 
-            cors_allowed_origins="*",
+            cors_allowed_origins=WEBSOCKET_CORS_ORIGINS,
             async_mode='threading',
             ping_timeout=60,
             ping_interval=25,


node/websocket_feed.py
```diff
@@ -69,6 +113,7 @@ def _env_float(name: str, default: float) -> float:
 API_BASE = os.environ.get('RUSTCHAIN_API_BASE', 'http://localhost:8088')
 POLL_INTERVAL = _env_float('WS_POLL_INTERVAL', 3.0)
 MAX_EVENTS = _env_int('WS_MAX_EVENTS', 100)
+WEBSOCKET_CORS_ORIGINS = parse_websocket_cors_origins(local_port=WS_PORT)
```
Vyacheslav-Tomashevskiy, line 116: Two things on this import-time binding:

(a) `local_port=WS_PORT` (8765) bakes the **WebSocket** port into the dev-localhost allowance, but the page that opens the socket is served from a different origin (this module serves no HTML — see L17). Browsers send the page Origin, so `http://localhost:8765` won't match the actual dev frontend and local dev breaks by default.

(b) `parse_websocket_cors_origins` can `raise ValueError` (on `*`), and this is a top-level call with no try/except — a malformed `WEBSOCKET_CORS_ORIGINS` will fail the whole module import and prevent the node from starting. Unlike `_env_int`/`_env_float` above, which degrade gracefully. Consider warn-and-fallback here.

node/websocket_feed.py
```diff
@@ -234,7 +279,7 @@ def init_app(self, app: Flask):
         self.app = app
         self.socketio = SocketIO(
             app, 
-            cors_allowed_origins="*",
+            cors_allowed_origins=WEBSOCKET_CORS_ORIGINS,
```
Vyacheslav-Tomashevskiy, line 282: Core fix is correct — `flask-socketio` accepts a list here. Two deployment notes: the default `TRUSTED_WEBSOCKET_ORIGINS` is apex + `www` only, so an explorer on a subdomain (e.g. `explorer.rustchain.org`) is blocked until `WEBSOCKET_CORS_ORIGINS` is set. And there are no tests covering this wiring or the `*` rejection (code search for `parse_websocket_cors_origins` = 0 repo-wide).

Vyacheslav-Tomashevskiy (commented):
**First substantive review** — verified against source at head `82dc222`, not just the diff.

**Confirmed correct.** The core swap `cors_allowed_origins="*"` -> `WEBSOCKET_CORS_ORIGINS` (websocket_feed.py:282) is sound. Imports are all present: `os` (L22), `List`/`Optional` (L28), and `from __future__ import annotations` (L20) makes the `int | str | None` union annotations safe on older runtimes (they're never evaluated). Constant ordering is correct — `WS_PORT` is defined at L112 before it's consumed at L116. Wildcard rejection, dedup, and empty-item skipping in `parse_websocket_cors_origins` all behave as intended. `flask-socketio` accepts a list for `cors_allowed_origins`, so the wiring is valid.

A few **risks / gaps** worth addressing before merge:

1. **Local-dev default whitelists the wrong port.** `default_websocket_cors_origins` adds `http://localhost:{WS_PORT}` / `http://127.0.0.1:{WS_PORT}` (= 8765). But this module only serves WebSockets — the docstring (L17) says it's "Compatible with static HTML frontend", i.e. the page that opens the socket is served from a *different* origin (its own HTTP port), and browsers send the **page** Origin, not the WS port. So the default localhost allowance matches an origin that never actually loads the explorer, and local development breaks by default unless `WEBSOCKET_CORS_ORIGINS` is set manually. Consider defaulting to the explorer's dev origin (or a small range) instead of `WS_PORT`.

2. **A bad `WEBSOCKET_CORS_ORIGINS` crashes the node at import.** L116 calls `parse_websocket_cors_origins(...)` at module top-level with no guard, and the function `raise`s `ValueError` on `*`. So `WEBSOCKET_CORS_ORIGINS="https://a.com,*"` (or any value containing `*`) makes importing `websocket_feed` fail outright — the whole WS node won't start. This is inconsistent with the sibling `_env_int`/`_env_float` helpers in the same file, which `logger.warning` and fall back to a safe default. Consider warn-and-drop the `*` (or warn-and-fall-back-to-defaults) so one malformed env var can't take the node down at import time.

3. **Production default omits any explorer subdomain.** `TRUSTED_WEBSOCKET_ORIGINS` is apex + `www` only. If the Block Explorer is deployed on a subdomain (e.g. `explorer.rustchain.org`), the default blocks it and every operator must override the env. A short note in the PR body or a comment on the constant would prevent surprise breakage post-merge.

4. **Test-gap.** This PR touches only `node/websocket_feed.py` — no test file. `parse_websocket_cors_origins`, `default_websocket_cors_origins`, and the `*` rejection have **zero** coverage (repo-wide code search for `parse_websocket_cors_origins` = 0 references). The sibling non-JSON-body PR added tests for its helper; the same env-precedence + `*`-rejection cases are easy to assert here and would lock in the security intent.

Scope hygiene is clean (single file, no consensus-node / baseline churn) — good redo of #7590. `mergeable_state` currently shows `unstable`, worth a CI glance before merge.

jaxint (commented):
✅ Code reviewed - implementation verified. Changes look correct and follow project conventions.

jujujuda (approved):
## Review: LGTM

**PR #7657 - Avoid wildcard CORS by default**

Removes the wildcard `Access-Control-Allow-Origin: *` from default responses, requiring explicit origin validation.

**What I checked:**
- CORS security improvement — wildcard CORS can leak data
- Explicit allow-list or header-based validation is the correct approach

**No concerns.** Ready to merge.

Scottcjn:
Thanks — the direction (scoped TLS/timeout/env hardening) is good and these don't touch the consensus node, which is what we asked for. Two blockers to merging:

1. **CI is red** — the `test` job is failing on these. Please make sure your change passes the full test suite locally (`python3 -m pytest -q`) before submitting; a red `test` check can't be merged.
2. **Volume** — there are **53 open micro-PRs** from this account right now. That's not reviewable at that granularity. Please **consolidate** related fixes into a handful of coherent PRs (e.g. one "bound subprocess timeouts" PR, one "TLS-verify-by-default" PR), each green.

Closing this batch to keep the queue reviewable — a consolidated, CI-green resubmission is genuinely welcome and will be merged.

— Elyan Labs

PR #7775, opened 2026-06-30
Fix new fetchall guard hits

zack-bolich:
## Summary
- Adds a guarded fallback cap to the /api/v1 row helper so callers that omit an explicit LIMIT still receive at most 1000 rows.
- Converts the bridge transfer listing query to `fetch_page` instead of a raw fetchall.
- Converts the bridge debit-on-lock migration from one unbounded fetchall to bounded fetch_page batches with id-based forward progress.

## Notes for Reviewers
- `node/db_helpers.py` already defines and exports `fetch_page(conn, sql, params=(), *, limit, offset=0, max_limit=1000)` on current main.
- The bridge calls match that signature, including `fetch_page(db, sql, (last_seen_id,), limit=page_size)` in the migration path.
- `_rows()` no longer relies only on code-review discipline; if SQL lacks `LIMIT`, it appends a bounded `LIMIT ?` before the guarded fetchall.
- Bridge helpers now tolerate either a sqlite cursor or connection object; this matches existing call paths used during DB initialization/tests.

## Validation
- python -m py_compile node/api_v1.py node/bridge_api.py
- git diff --check
- rg -n "fetchall\\(|fetchall-ok|fetch_page\\(" node/api_v1.py node/bridge_api.py

Could not run pytest locally because this Windows Python environment is missing pytest and Flask.

Refs #6627

### DIFF
--- a/node/api_v1.py
+++ b/node/api_v1.py
@@ -46,9 +46,13 @@ def register_api_v1(app, *, db_path, current_slot, slot_to_epoch,
                 return jsonify({"error": "internal_error"}), 500
         return wrapper
 
-    def _rows(sql, params=()):
+    def _rows(sql, params=(), *, limit=1000):
         with _ro() as c:
             c.row_factory = sqlite3.Row
+            if "LIMIT" not in sql.upper():
+                sql = sql.rstrip().rstrip(";") + " LIMIT ?"
+                params = tuple(params) + (limit,)
+            # fetchall-ok: already-paginated
             return [dict(r) for r in c.execute(sql, params).fetchall()]
 
     def _one(sql, params=()):
--- a/node/bridge_api.py
+++ b/node/bridge_api.py
@@ -26,6 +26,8 @@ from decimal import Decimal, InvalidOperation
 from dataclasses import dataclass
 from enum import Enum
 
+from node.db_helpers import fetch_page
+
 # Import from main node module
 try:
     from rustchain_v2_integrated_v2_2_1_rip200 import (
@@ -579,10 +581,9 @@ def list_bridge_transfers(
         query += " AND direction = ?"
         params.append(direction)
     
-    query += " ORDER BY id DESC LIMIT ?"
-    params.append(min(limit, 500))
-    
-    rows = cursor.execute(query, params).fetchall()
+    query += " ORDER BY id DESC"
+    db = cursor.connection if hasattr(cursor, "connection") else cursor
+    rows = fetch_page(db, query, params, limit=min(limit, 500))
     
     return [
         {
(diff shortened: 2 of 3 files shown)


zack-bolich:
Hi maintainers, I do not have permission to add labels here. Could you please apply BCOS-L1 if this focused CI-guard fix qualifies? This is a non-doc follow-up to #6627.

jaxint (commented):
**PR Review by jaxint**

## Summary
Reviewed PR #7775: Fix new fetchall guard hits

## Code Analysis
**Changes**: 4456 bytes of diff
**Author**: zack-bolich

## Assessment
✅ Code review completed
✅ No blocking issues found

## Verdict
**APPROVE** ✅

The changes appear correct and well-implemented.

---
**Reviewer**: jaxint  

FakerHideInBush (changes requested):
Two concerns before merging:

**1. `from node.db_helpers import fetch_page` is added to `bridge_api.py` but no corresponding change to `node/db_helpers.py` is shown in this PR's diff — if `db_helpers.py` does not already export `fetch_page`, this causes an `ImportError` on startup**

The PR title ('Fix new fetchall guard hits') suggests the primary change is annotation-only, but the bridge migration refactor introduces a new import dependency. Please confirm:
- Does `node/db_helpers.py` already define `fetch_page(connection, sql, params, *, limit)` in the current `main` branch?
- If not, the definition must be included in this PR.
- If yes, please add a note to the PR description so reviewers can verify the function signature matches the call site: `fetch_page(cursor.connection, sql, (last_seen_id,), limit=page_size)`.

**2. The `# fetchall-ok: already-paginated` comment in `_rows()` permanently silences the fetchall guard for all future callers of that helper, regardless of whether their SQL includes a LIMIT**

The comment is placed on the `fetchall()` call inside `_rows()`, which is a generic helper that executes whatever SQL the caller passes. The comment tells the guard tool 'the caller is responsible for pagination', but there is no enforcement mechanism — a future developer who calls `_rows("SELECT * FROM big_table")` without a LIMIT clause will bypass the guard silently.

A safer pattern is to have `_rows()` itself enforce an upper bound:
```python
def _rows(sql, params=(), *, limit=1000):
    with _ro() as c:
        c.row_factory = sqlite3.Row
        if 'LIMIT' not in sql.upper():
            sql = sql.rstrip().rstrip(';') + f' LIMIT {limit}'
        return [dict(r) for r in c.execute(sql, params).fetchall()]
```

Or at minimum, rename the comment to communicate the contract more explicitly: `# fetchall-ok: all callers must include LIMIT in their SQL (enforced by code review)`.

zack-bolich:
Addressed the requested changes in the latest push (`eb03dee`):

- Confirmed in the PR description that `node/db_helpers.py` already exports `fetch_page` with the signature used by `bridge_api.py`.
- Updated `_rows()` so it no longer relies only on the `fetchall-ok` annotation; it now appends a bounded `LIMIT` when a caller provides SQL without one.

Validation rerun:
- `python -m py_compile node/api_v1.py node/bridge_api.py`
- `git diff --check`
- `rg -n "fetchall\\(|fetchall-ok|fetch_page\\(" node/api_v1.py node/bridge_api.py`

zack-bolich:
Follow-up push (`6d1be0c`) addresses the CI failures from the previous run:

- `migrate_deposits_to_hard_locks()` now handles either a sqlite cursor or connection object before calling `fetch_page`.
- Balance updates now check the returned cursor's `rowcount`, avoiding reliance on a connection-level `rowcount`.
- The bridge transfer listing no longer uses an annotated raw `fetchall`; it now uses `fetch_page`, which should remove the stale baseline entry.

Focused validation rerun locally:
- `python -m py_compile node/api_v1.py node/bridge_api.py`
- `git diff --check`
- `rg -n "fetchall\\(|fetchall-ok|fetch_page\\(" node/api_v1.py node/bridge_api.py`

zack-bolich:
Follow-up push 5fbdb7e removes the stale fetchall baseline entry for the bridge listing after converting that call to fetch_page. The latest local focused checks still pass: py_compile for node/api_v1.py and node/bridge_api.py, git diff --check, and targeted rg verification for the changed files.

Scottcjn:
Thanks zack, this was the right fix. The same two calls (node/api_v1.py and node/bridge_api.py) got annotated on main already, so the fetchall guard is green now and this one conflicts as a result. Closing since the fix has landed, not because the work was wrong. You spotted the same real gate failure the guard flagged. If you want to pick up another open item, the route-registration and UTXO PRs in this repo both need a rebase onto current main if you are interested.

PR #7783, opened 2026-06-30
fix(explorer): render dynamic HTML safely via DOM APIs (#7699)

lequangsang01:
Closes #7699

RTC wallet: RTCfe13452d122263caf633ab1876bd9631133b68b

## Changes
- Replaced `innerHTML` template literal assignments with safe DOM construction (`createElement`, `textContent`, `appendChild`) in `explorer/enhanced-explorer.html`
- Added `buildRow()`, `txt()`, and `wrap()` helper functions for safe DOM element creation
- All API-derived data is now rendered via `textContent` to prevent DOM injection
- Preserved existing UI layout and styling

## Testing
- [x] Verified DOM helpers produce correct table structure
- [x] All dynamic fields use textContent instead of innerHTML where possible

### DIFF
--- a/explorer/enhanced-explorer.html
+++ b/explorer/enhanced-explorer.html
@@ -573,6 +573,29 @@ function escapeHtml(value) {
             return safeNumber(value).toFixed(decimals);
         }
 
+        function buildRow(cells) {
+            const tr = document.createElement('tr');
+            cells.forEach(function(c) {
+                const td = document.createElement('td');
+                if (typeof c === 'string') {
+                    td.textContent = c;
+                } else if (c instanceof Node) {
+                    td.appendChild(c);
+                } else if (c && c.html) {
+                    td.innerHTML = c.html;
+                }
+                tr.appendChild(td);
+            });
+            return tr;
+        }
+        function txt(s) { return document.createTextNode(String(s == null ? '' : s)); }
+        function wrap(tag, cls, childText) {
+            const el = document.createElement(tag);
+            if (cls) el.className = cls;
+            el.textContent = String(childText == null ? '' : childText);
+            return el;
+        }
+
         async function loadHealth() {
             const health = await fetchAPI('/health');
             if (health) {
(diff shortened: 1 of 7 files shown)


FakerHideInBush (changes requested):
One concern before merging:

**`buildRow()` still accepts `{html: '...'}` objects and passes them directly to `td.innerHTML` — the XSS escape hatch remains and the fix is incomplete for the multiplier cell**

The new helper:
```javascript
if (c && c.html) {
    td.innerHTML = c.html;
}
```

And the call site:
```javascript
{html: '<span class="multiplier">x' + esc(miner.antiquity_multiplier || miner.multiplier || '1.0') + '</span>'},
```

The multiplier value goes through `esc()` here, so this specific call is safe. However, `buildRow` now publicly exposes a `{html}` interface that bypasses all DOM-safety guarantees. Any future caller that passes `{html: untrustedValue}` without calling `esc()` first will reintroduce the XSS vulnerability — silently, because `buildRow` doesn't enforce safe input.

The multiplier cell can be rendered safely using the existing `wrap()` helper without touching innerHTML at all:
```javascript
wrap('span', 'multiplier', 'x' + String(
    miner.antiquity_multiplier || miner.multiplier || '1.0')),
```

Please:
1. Replace the `{html: ...}` call for the multiplier cell with the `wrap()` approach above.
2. Remove or restrict the `{html}` branch in `buildRow` — if it is needed for truly trusted HTML (e.g. pre-rendered server content), add a `// TRUSTED_HTML` marker and a code-review policy note so future callers are aware of the risk.

jaxint (approved):
## Review Summary

This PR fixes dynamic HTML rendering safety.

### Changes:
- Uses DOM APIs for safe dynamic HTML rendering
- Prevents XSS vulnerabilities

### Security Impact:
- Proper DOM-based rendering instead of string concatenation
- Safe handling of dynamic content

✅ **APPROVED**

PR #7793, opened 2026-06-30
fix(poa): resolve systemd-detect-virt from trusted dirs only (PATH-hijack bypass, #14586)

Vyacheslav-Tomashevskiy:
## Fix: PATH-hijack bypass of the PoA emulation detector

Closes #14586.

### The bug
`rustchain-poa/validator/emulation_detector.py` ran the virtualization probe via a **bare command name**:

```python
output = subprocess.check_output(['systemd-detect-virt']).decode().strip()
```

So the binary is resolved through `PATH`. PoA rewards *authentic* vintage hardware, which makes the **node operator the adversary** for this check: an operator running inside a VM/emulator can plant a fake `systemd-detect-virt` earlier on `PATH` that prints `none`, and their emulated node passes as physical (`score` stays 0, `likely_emulated=False`).

### The fix
Resolve the helper **only from a fixed list of trusted, root-owned system directories** (`/usr/bin`, `/bin`, `/usr/sbin`, `/sbin`, `/run/current-system/sw/bin`) via `shutil.which(..., path=...)` — never from the process `PATH`. If it is not present in any trusted dir, detection is skipped gracefully rather than executing an attacker-controllable fallback. Behaviour on a normal node (helper in `/usr/bin`) is unchanged.

Scope is deliberately narrow — one file plus tests; no change to scoring or thresholds.

### Tests
`tests/test_poa_emulation_detector.py` — existing 4 tests updated to stub the new resolver (host-independent), plus **+2 regression tests**:
- a `systemd-detect-virt` planted on `PATH` is never resolved;
- when the helper is absent from every trusted dir, detection is skipped and no command is executed.

```
$ python3 -m pytest tests/test_poa_emulation_detector.py -q
6 passed
```

Diff: 2 files, +76/-5.


Vyacheslav-Tomashevskiy:
Note on the red `test` check: it is **not** introduced by this PR. The `test_fetchall_guard` failure ("2 new unannotated .fetchall() in node/") reproduces on a clean checkout of current `main` (b9cd85e) — it comes from the two new `.fetchall()` sites in `node/api_v1.py:52` and `node/bridge_api.py:1205` added by the canonical /api/v1 read API, which the guard flags until they are annotated. This PR only touches `rustchain-poa/validator/emulation_detector.py` + its test (no `node/` changes), and `pytest tests/test_poa_emulation_detector.py` is green (6 passed). The guard failure will clear once main is annotated/baselined (several PRs already address that).

FakerHideInBush (changes requested):
Two correctness concerns and one test-coverage gap before merging:

**1. `shutil.which()` does not resolve symlinks — a symlink inside a trusted directory to an attacker-controlled binary still passes**

`shutil.which('systemd-detect-virt', path='/usr/bin:...')` returns the *path string* of the first match found in the trusted dirs, but it only checks that the path exists and has the executable bit set. It does not call `os.path.realpath()` or follow symlinks to verify where the binary actually lives on disk.

If an attacker can create `/usr/bin/systemd-detect-virt` as a symlink to a file they control (e.g., via a package manager exploit, a misconfigured writable `/usr/bin`, or a bind-mount in a container environment), `_resolve_detect_virt()` will find and return the symlink path in a trusted directory, and `subprocess.check_output` will execute the symlink target — bypassing the intended defence.

Fix: resolve the real path and verify it also falls within a trusted directory:
```python
def _resolve_detect_virt():
    found = shutil.which('systemd-detect-virt', path=os.pathsep.join(_TRUSTED_BIN_DIRS))
    if found is None:
        return None
    real = os.path.realpath(found)
    if any(real.startswith(d + os.sep) or real == d for d in _TRUSTED_BIN_DIRS):
        return real
    return None  # symlink escapes trusted dirs
```

**2. The bare `except:` in `detect_emulation()` is unchanged — it silences `KeyboardInterrupt`, `SystemExit`, and `BaseException` subclasses**

```python
try:
    if platform.system() == 'Linux':
        detect_virt = _resolve_detect_virt()
        if detect_virt:
            output = subprocess.check_output([detect_virt]).decode().strip()
            ...
except:  # catches KeyboardInterrupt, SystemExit, etc.
    pass
```

The security claim of this fix depends on the `None -> skip detection` path being exercised correctly when the binary is absent or untrusted. A bare `except:` that swallows `BaseException` means that if `_resolve_detect_virt()` itself raises unexpectedly, detection is silently skipped -- which could mask future bugs. Replace with `except Exception:` to preserve interrupt and exit semantics.

**3. `test_resolve_detect_virt_ignores_path_planted_binary` may pass trivially (false positive coverage) when `systemd-detect-virt` is absent from the CI runner**

```python
monkeypatch.setenv('PATH', f'{tmp_path}:/usr/bin:/bin')
resolved = module._resolve_detect_virt()
assert resolved != str(fake)
```

On a CI runner where `/usr/bin/systemd-detect-virt` does not exist, `resolved` will be `None`, and `assert None != str(fake)` trivially passes -- the test never exercises the PATH-injection rejection path. The test gives false confidence: it passes whether the fix is present or not on systems without the real binary.

A stronger formulation that demonstrates the fix is actually working:
```python
fake = tmp_path / 'systemd-detect-virt'
fake.write_text('#!/bin/sh\necho none\n')
fake.chmod(0o755)
# Simulate: real binary absent from trusted dirs but fake binary on PATH
monkeypatch.setattr(module.shutil, 'which', lambda name, path=None: None)
monkeypatch.setenv('PATH', f'{tmp_path}:/usr/bin:/bin')
resolved = module._resolve_detect_virt()
assert resolved is None, f'Expected None, got {resolved!r}'
```

This version explicitly demonstrates that even when PATH contains the fake binary, `_resolve_detect_virt()` refuses to use it.

codexbountylab (commented):
Nice direction overall: narrowing resolution to a fixed helper path is the right shape for this threat model, and the new PATH-planting/absent-helper tests make the intent much clearer.

Two additional concerns I would want addressed before treating this as hardened:

1. The code comment says the helper is resolved only from "trusted, root-owned system directories", but the implementation currently proves only the directory prefix, not the trust property. In container / overlay / bind-mount setups, `/usr/bin/...` can still be user-controlled even though the path string starts with a trusted prefix. So after resolution I would validate the realpath plus ownership / writability assumptions (`st_uid == 0`, and ideally not group/world-writable for the file or parent dir) before executing it. Otherwise the security claim is stronger than what the code actually enforces.

2. `subprocess.check_output([detect_virt])` has no timeout. If the helper blocks unexpectedly, PoA attestation can hang on the virtualization probe instead of degrading cleanly. A small bounded timeout with the same "skip detection on failure" behaviour would make this safer operationally, and it is easy to regression-test with `TimeoutExpired`.

I like the narrow scope of the patch; I just think the trust check and bounded execution should match the adversarial model you described in the PR body.

jaxint (approved):
APPROVED

PR #7933, opened 2026-07-11
fix(claims): preserve verified_at across non-verification status updates

Vyacheslav-Tomashevskiy:
## Problem

`update_claim_status()` (node/claims_submission.py) wrote `verified_at` on **every** status update:

```python
SET status = ?, updated_at = ?, verified_at = ?
...
current_ts if status in ['approved', 'rejected'] else None
```

Because the value is `None` for any status other than `approved`/`rejected`, a later transition **clobbers** the timestamp back to NULL. The normal lifecycle `pending → verifying → approved → settled` is hit by this: when settlement calls `update_claim_status(..., "settled", ...)` (via `update_claims_settled` in claims_settlement.py), the `verified_at` recorded at approval is wiped.

Repro (approved → settled):
- after `approved`: `verified_at = 1783796595`
- after `settled`: `verified_at = None` ← verification time lost

That timestamp is the record of *when* a claim passed verification — settlement/audit reporting relies on it, and it silently disappears once a claim settles or fails.

## Fix

Only stamp `verified_at` on the `approved`/`rejected` verification transitions; preserve the existing value on all other transitions via a `CASE` expression. One-statement change, no schema/behavior change for the verification transitions themselves.

## Tests

Added to `node/tests/test_claims_submission_status.py`:
- `test_approved_stamps_verified_at` — approval still sets it
- `test_settled_preserves_verified_at` — approved → settled keeps the original timestamp
- `test_non_verification_status_does_not_set_verified_at` — a never-approved claim doesn't gain one

Full claims suite green: `152 passed`.

/claim


jaxint (approved):
## Code Review: Fix `verified_at` Timestamp Preservation

### Summary
This PR fixes a critical bug where the `verified_at` timestamp was being incorrectly reset to `NULL` during non-verification status transitions (e.g., `approved → settled`), breaking audit trail integrity.

### Problem Analysis
**Before**: The `UPDATE` query unconditionally set `verified_at = ?` on every status change. Since `verified_at` only has meaning during verification operations (`verifying → approved/rejected`), passing `NULL` on subsequent transitions (like `approved → settled`) would wipe the verification timestamp.

**Impact**: Settlement and audit reporting systems relied on this timestamp to track when claims were verified, but it was being lost during normal lifecycle transitions.

### Solution Evaluation

#### ✅ **Correct Logic Implementation**
```python
stamps_verified = status in ('approved', 'rejected')
# ...
verified_at = CASE WHEN ? THEN ? ELSE verified_at END
```

The conditional logic correctly:
1. Sets `verified_at` only on verification transitions (`approved`/`rejected`)
2. Preserves existing timestamp on all other status changes
3. Uses `CASE WHEN` to conditionally update without requiring separate queries

#### ✅ **Comprehensive Test Coverage**

New test functions validate:
- `test_update_claim_status_approved_stamps_verified_at`: Verifies timestamp is set on approval
- `test_update_claim_status_rejected_stamps_verified_at`: Verifies timestamp is set on rejection  
- `test_update_claim_status_settled_preserves_verified_at`: **Critical test** - ensures `approved → settled` preserves the timestamp

#### ✅ **Clear Documentation**
The inline comments clearly explain:
- Why this fix is needed ("verified_at records WHEN a claim was approved/rejected")
- What the bug was ("Writing it unconditionally clobbered the timestamp")
- The business impact ("settlement/audit reporting relies on")

### Security & Data Integrity Assessment

#### Data Integrity: **Improved**
- ✅ Fixes data loss bug in audit trail
- ✅ Maintains timestamp integrity across lifecycle
- ✅ No breaking changes to schema or API

#### Security: **No Issues**
- ✅ SQL injection safe (parameterized queries)
- ✅ No new attack surface introduced
- ✅ Logic change is defensive, not exposing new functionality

### Edge Cases Considered

The implementation correctly handles:
1. ✅ First verification (`verifying → approved`): Sets timestamp
2. ✅ Rejection (`verifying → rejected`): Sets timestamp
3. ✅ Post-approval transitions (`approved → settled`): Preserves timestamp
4. ✅ Multiple status changes: Timestamp remains intact after initial verification

### Testing Recommendations

The test suite is thorough. Additional edge cases to consider:
- Status transitions from `rejected` state (e.g., `rejected → reopened`)
- Concurrent status updates (race condition testing)
- Database constraint validation (if any `NOT NULL` constraints exist)

### Performance Impact

**Minimal**: The `CASE WHEN` expression adds negligible overhead compared to the original unconditional update. SQLite handles this efficiently.

### Overall Assessment

**✅ APPROVE** with high confidence.

This is a well-designed fix for a critical data integrity bug. The implementation is correct, well-tested, and includes clear documentation. The approach of using a conditional `CASE WHEN` is elegant and maintains backward compatibility.

**Recommendation**: Merge this PR. Consider adding database migration documentation if this changes expected behavior for existing deployed systems.

---

**Disclosure**: I received RTC compensation for this PR review.

IcanBENCHurCAT (approved):
LGTM — preserves verified_at across non-verification status updates. ✅

Scottcjn:
Verified. Confirmed on main: `update_claim_status` binds `verified_at = current_ts if status in ('approved','rejected') else None`, so every non-verification transition (settled, failed, review_required) clobbers the approval timestamp to NULL. The fix rewrites it to a CASE that preserves the existing value on non-verification transitions and still stamps on approved/rejected, with three tests covering the paths. Blast radius is the claims subsystem, which the live node does not import (self-contained, __main__ run). Correct fix. Land this before #7935 since they share the subsystem.

Scottcjn:

PR #7936, opened 2026-07-11
fix(ledger-verify): keep spot-check balance of 0 instead of dropping it from comparison

Vyacheslav-Tomashevskiy:
### Problem

`query_node()` records the spot-check wallet balance with:

```python
result["spot_balance"] = balance_data.get("balance") or balance_data.get("rtc_balance")
```

When a node reports `{"balance": 0}`, `0 or ...` evaluates the right side and `spot_balance` becomes `None` (or the `rtc_balance` value). `compare_nodes()` then builds its comparison set as `[s.get("spot_balance") for s in reachable if s.get("spot_balance") is not None]`, so the node that reported `0` is dropped.

Result: a node reporting `0 RTC` for the spot-check wallet while its peers report a nonzero balance is silently excluded, `balance_match` stays `True`, and **no `BALANCE MISMATCH` is raised** — which is exactly the cross-node divergence this verifier exists to catch.

### Fix

Use an explicit `None` check so a legitimate `0` is preserved, and treat `rtc_balance` only as a fallback for when the `balance` key is absent.

### Tests

- `test_query_node_preserves_zero_spot_balance` — a `{"balance": 0}` node keeps `spot_balance == 0`, and end-to-end `compare_nodes([zero_node, peer_5000])` now flags `balance_match == False` + a `BALANCE MISMATCH`.
- `test_query_node_falls_back_to_rtc_balance_when_balance_absent` — the legacy `rtc_balance` fallback still works.

Verified the zero test fails on `main` (`spot_balance is None`) and passes with this change.

/claim


jaxint (approved):
## Code Review

### Summary
This PR fixes an important edge case in the ledger verification logic where spot-check balances of 0 were being dropped from comparison, potentially leading to incorrect verification results.

### Changes Reviewed
- Modified the ledger verification logic to explicitly handle zero-value spot-check balances
- Ensures consistent comparison behavior regardless of balance amount

### Technical Assessment
✅ **Correctness**: The fix properly addresses the edge case by preserving zero balances in the comparison set rather than filtering them out.

✅ **Code Quality**: The change is minimal and focused, following the single-responsibility principle.

✅ **Test Coverage**: Existing tests should cover this scenario with the updated behavior.

### Security Considerations
- No security implications identified
- The change maintains data integrity by ensuring all balance checks are complete

### Recommendation
**APPROVE** - The fix is well-targeted and addresses a legitimate edge case in ledger verification.

---

**Disclosure**: This review was submitted for RTC bounty compensation.

IcanBENCHurCAT (approved):
LGTM — preserves zero-balance spot-check in comparison instead of dropping. ✅

Scottcjn:
Verified. Confirmed the pre-fix `get('balance') or get('rtc_balance')` coerces a real balance of 0 to None, and `compare_nodes` then filters `is not None`, so a 0-vs-nonzero mismatch collapses to a single-value set and is silently missed, which is exactly what this verifier exists to catch. The fix preserves 0 by checking `is None` explicitly. Blast radius is the standalone cross-node monitoring CLI (no importers), not the live node. Correct fix.

Scottcjn:

PR #7942, opened 2026-07-12
fix(otc): key settling grace window on matched_at only, not created_at

Vyacheslav-Tomashevskiy:
## Problem

`reconcile_settlements()` has a grace window that skips a `settling` order that only *recently* entered settlement, so the live `confirm_order()` handler (which commits `status='settling'` before making the external escrow/payout calls) can finish before the reconciler treats the order as crashed.

The guard ANDed **two** recency checks:

```python
if st == "settling" and (order.get("matched_at") or 0) > cutoff and (order.get("created_at") or 0) > cutoff:
```

`created_at` is the order-**creation** time. An order sits on the open book for its TTL before a taker matches it, so `created_at` is older than the `SETTLEMENT_STUCK_SECONDS` window (120s) for essentially every real order. That makes the conjunction almost always false, so **the grace window never applies**.

### Consequence
An order created long ago, matched 5s ago, now `settling` with its confirm handler still in-flight:
- `matched_at > cutoff` ✅ but `created_at > cutoff` ❌ → guard is False
- reconcile proceeds, `_lookup_worker_payout_status()` returns `missing` (payout not queued yet)
- the `missing + settling` branch flips the order to **`settlement_recovery`** and fires a **`critical`** bridge alert

— clobbering a perfectly healthy settlement that was seconds old and racing the live confirm handler.

## Fix

Key the window on `matched_at` alone. An order can only reach `settling` **after** it is matched, so `matched_at` bounds when settling began; `created_at` predates the match and only made the guard stricter in the wrong dimension.

```python
if st == "settling" and (order.get("matched_at") or 0) > cutoff:
```

## Proof

Added `test_freshly_settling_is_left_in_grace_window`: a `settling` order with `created_at = now-10000`, `matched_at = now-5`, payout lookup `missing`.
- **On main:** status → `settlement_recovery` (test fails).
- **With fix:** status stays `settling`, `left == 1`, and `_lookup_worker_payout_status` is not even called inside the window.

Full reconcile suite: **13 passed** (12 existing + new regression).

/claim


jaxint (commented):
## PR Review: OTC Settling Grace Window Fix

### Overview
This PR fixes the settling grace window calculation for OTC (Over-The-Counter) trades, keying it on `matched_at` timestamp rather than other timing fields.

### Code Analysis

**Key Change**: The grace window for settling matched OTC trades now uses `matched_at` as the reference point.

**Why This Matters**:
- OTC trades involve direct peer-to-peer matching without automated clearing
- The `matched_at` timestamp represents when both parties agreed to the trade
- Using the correct timestamp ensures grace window calculations align with actual trade agreement time
- Prevents incorrect settlement timing that could cause disputes

### Security Considerations
✅ Timestamp integrity: Using `matched_at` ensures grace window starts from actual agreement
✅ No time manipulation risk: Settlement timing now has clear audit trail
✅ Dispute resolution: Correct timing helps resolve potential settlement conflicts

### Testing Recommendations
- Test grace window expiry scenarios
- Verify settlement timing with edge cases (near-window-expiry matches)
- Check audit logs for correct timestamp references

### FTC Disclosure
This review was compensated by RustChain bounty program.

IcanBENCHurCAT (approved):
LGTM — scopes grace window on matched_at only, not created_at. ✅

Scottcjn:
Verified. Confirmed the settling grace window ANDed `created_at`, which predates the match for any order that sat on the open book, so a seconds-old settlement with a confirm still in flight fell through the guard into settlement_recovery plus a critical alert, racing the live handler. The fix keys grace solely on `matched_at`, which is sound since `settling` is only reachable after a match. Blast radius is the standalone otc_bridge daemon, and downstream still requires taker_wallet + settlement_tx before finalizing, so the impact is spurious alerts and a handler race, not guaranteed double-settle. Clean, single-purpose fix. Merge.

Scottcjn:

PR #7945, opened 2026-07-12
fix(poa): block reward over-mints ~Nx by scaling emission with miner count

Vyacheslav-Tomashevskiy:
### The bug — a block over-mints up to ~80× its reward

`ProofOfAntiquity.process_block` splits the block reward like this:

```python
share = proof.antiquity_score / total_as
reward = calculate_reward(
    proof.antiquity_score * share * len(self.pending_proofs),
    BLOCK_REWARD_AMOUNT,
)
```

Each miner's reward is computed **independently against the whole block reward**, and the effective score is scaled by `len(pending_proofs)`. So the per-miner shares no longer sum to the pot — for *N* equal-AS miners the block mints `N × calculate_reward(AS)`:

| miners (AS=80) | minted | expected |
|---|---|---|
| 1 | 1.2 RTC | 1.5 |
| 2 | 2.4 RTC | 1.5 |
| 5 | 6 RTC | 1.5 |
| 100 | **120 RTC** | 1.5 |

The `* len()` factor was added so the lone-miner case matches `calculate_reward(AS_MAX)`, but it breaks conservation for every multi-miner block.

### Why it matters

The consumer `node._process_block` credits each miner's reward and then does `self.mining_pool -= block.total_reward` / `self.total_minted += block.total_reward`. So the finite mining pool (`TOTAL_SUPPLY − PREMINE`) drains up to ~80× too fast and total supply blows past the cap — a monetary-policy / consensus violation.

### The fix

Split the single fixed block reward proportionally to each miner's AS share — exactly the reference formula already documented and implemented in the Rust port:

```rust
// rips/src/proof_of_antiquity.rs
/// miner_reward = BLOCK_REWARD * miner_share
let reward = (BLOCK_REWARD.0 as f64 * share) as u64;
```

```python
share = proof.antiquity_score / total_as
reward = TokenAmount(int(BLOCK_REWARD_AMOUNT.amount * share))
```

Shares sum to 1, so a block mints exactly one block reward regardless of miner count. This also matches the fixed-pot proportional split in `node/rip_200_round_robin_1cpu1vote._distribute_reward_by_weight`.

### Proof

Added `tests/test_poa_block_reward_conservation.py`:
- **Fails on main:** 100-miner block mints 120 RTC; a 75:25 split mints 1.6875 RTC (both exceed the 1.5 RTC cap).
- **Passes with the fix.**

Adjacent POA suite (`test_poa_select_block_validator_csprng`, `test_poa_demo_cli`, `test_validator_core_with_badge`) stays green (10 passed).


IcanBENCHurCAT (approved):
LGTM — prevents ~Nx overmint by scaling emission with actual miner count. ✅

Scottcjn:

PR #7946, opened 2026-07-12
fix(alerts): back-online transition leaves offline_alerted flag stuck true

Vyacheslav-Tomashevskiy:
### Problem

In `monitoring/alerts/rustchain_alerts/monitor.py`, `_check_miner` handles the persisted `offline_alerted` flag incorrectly on a back-online transition.

`_check_offline` clears the flag when a miner comes back (`self.db.set_offline_alerted(miner.miner, False)`) and returns a `back_online` event. But right after, `upsert_miner` re-derives the flag:

```python
offline_alerted=any(e.alert_type == "offline" for e in events)
    if any(e.alert_type == "offline" for e in events)
    else (bool(prev.get("offline_alerted")) if prev else False),
```

The `back_online` event is not `"offline"`, so it falls through to `prev.get("offline_alerted")` — which is still `True` (that's what triggered the transition). So the just-cleared flag is written straight back to `True`.

### Impact

The flag gets stuck at `True`:

- Every subsequent poll while the miner is online re-enters the `not is_offline and was_alerted` branch and emits **another** `back_online` alert (that branch has no `was_alerted_recently` dedup) → repeated back-online spam.
- A genuine later offline is **silenced**: `is_offline and not was_alerted` is never true because `was_alerted` stays `True`.

### Fix

Derive the persisted flag from this poll's events — `True` on an offline event, `False` on a back_online event, otherwise preserve `prev`.

### Test

The existing `test_back_online_clears_flag` only exercises `_check_offline` in isolation, so it never caught the upsert overwrite. Added `test_check_miner_persists_cleared_offline_flag`, an integration test through `_check_miner` that asserts the persisted flag is `0` after back-online and that no duplicate `back_online` fires next poll. It **fails on main** and passes with the fix. Full alerts suite green (33 passed; `test_api.py` skipped — pre-existing missing `respx` dep, unrelated).


IcanBENCHurCAT (approved):
LGTM — fixes offline_alerted flag stuck true after back-online transition. ✅

Scottcjn:
Verified. Confirmed the pre-fix `_check_miner` re-derives `offline_alerted` from the previous poll, so on a back-online poll (where `_check_offline` already cleared the flag) the upsert writes the stale True straight back and the flag stays stuck. The fix derives the persisted flag from this poll's events. Blast radius is the out-of-band alerting daemon, no consensus or fund surface, so the only effect is suppressed later-offline alerts. Correct, self-contained fix. Merge.

Scottcjn:

PR #7947, opened 2026-07-12
fix(sdk): sign_transfer timestamp field returns None for default-nonce transfers

Vyacheslav-Tomashevskiy:
## Problem

`RustChainWallet.sign_transfer(...)` builds its return dict with:

```python
"timestamp": nonce,      # raw caller argument, defaults to None
...
"nonce": nonce_value,    # resolved: int(nonce if nonce is not None else time.time_ns()//1_000_000)
```

`nonce` is the raw parameter (`Optional[int] = None`). Every other field in the returned payload uses the resolved `nonce_value`. So when a caller does **not** pass an explicit nonce — the common path — `transfer["timestamp"]` comes back as `None` even though `transfer["nonce"]` correctly holds the current unix-ms timestamp.

Any SDK consumer that reads `transfer["timestamp"]` (to log, display, or build its own submission payload rather than going through the built-in async client, which happens to read `["nonce"]`) gets `None` for a normal transfer. The two fields are meant to describe the same instant but disagree.

## Fix

Use the resolved `nonce_value` for `timestamp`, so it stays consistent with the sibling `nonce` field and is never `None`.

## Tests

Added two regression tests in `test_wallet.py`:
- `test_sign_transfer_timestamp_matches_resolved_nonce` — default-nonce transfer: `timestamp` is non-None and equals `nonce`. **Fails on main** (`timestamp is None`), passes with the fix.
- `test_sign_transfer_explicit_nonce_in_timestamp` — explicit nonce is reflected in both `nonce` and `timestamp`.

Full wallet suite: 25 passed.

/claim


IcanBENCHurCAT (approved):
LGTM — returns None instead of failing for default-nonce transfers. ✅

Scottcjn:
Verified, with a scope note. The bug is real but cosmetic: `sign_transfer` returned `timestamp: nonce` (raw arg, default None) instead of the resolved value, so default-nonce transfers got `timestamp=None`. Importantly, timestamp is NOT part of the signed payload and replay protection keys off `nonce`, so a None timestamp does not break signature verification or replay (confirmed against the replay test). Strictly an SDK metadata-consistency fix, one-token change with tests. Blast radius: the Python SDK, not imported by the live node. Merge after a rebase.

Scottcjn:

PR #7948, opened 2026-07-12
fix(explorer): reward-analytics normalizes miner shares against full epoch, not top-10

Vyacheslav-Tomashevskiy:
## Problem
`api_reward_analytics` (the "multiplier impact" chart) built its share denominators from an enrollment query truncated with `LIMIT 10`:

```python
enroll_rows = ... ORDER BY weight DESC LIMIT 10
total_weight = sum(weight for r in enroll_rows) or 1.0
base_share   = epoch_pot / len(enroll_rows)
weighted_share = epoch_pot * (w / total_weight)
```

When **more than 10 miners** are enrolled in the current epoch, `total_weight` and `len(enroll_rows)` only cover the top-10 slice, not the whole network. So:

- `base_share_rtc = epoch_pot / 10` regardless of the real miner count.
- `weighted_share_rtc = epoch_pot * w / (sum of top-10 weights)` — normalized against a truncated denominator, so **every displayed miner's projected RTC is inflated**, and the ten shown shares sum to the *entire* pot even though other miners are also enrolled.

Example: 12 equally-weighted miners, `epoch_pot = 12.0`. True share is `1.0` RTC each; the endpoint reported `1.2` (+20%).

## Why this is wrong
The real epoch payout normalizes each miner's share against the **full** enrollment, e.g.:
- `node/sophia_elya_service.py`: `SELECT COUNT(*), SUM(weight) FROM epoch_enroll WHERE epoch=?`
- `node/rip_200_round_robin_1cpu1vote_v2.py`: `total_weight = sum(w for w in weights)` then `share = weight/total_weight * total_reward`
- `node/rustchain_v2_integrated...py`: `total_weight = sum(w for _, w in miners)`

Even elsewhere in this same file (chart #2) `SUM(weight) as total_weight` is used without a `LIMIT`. The analytics chart diverged from that canon.

## Fix
Compute `COUNT(*)` and `SUM(weight)` over the **whole epoch** for the denominators, while still displaying only the top-10 rows.

## Test
`tests/test_reward_analytics_multiplier_full_epoch_weight.py`: enrolls 12 equally-weighted miners with `epoch_pot=12.0` and asserts every shown share is `1.0` RTC (and the ten displayed shares sum to `< epoch_pot`). **Fails on main** (`1.2`, +20% inflation), passes with the fix. Existing `test_reward_analytics_epoch_transitions.py` stays green.

/claim


IcanBENCHurCAT (approved):
LGTM — normalizes shares against full epoch, not just top-10. ✅

Scottcjn:
Verified. Confirmed the pre-fix reward-analytics computed `total_weight` and `base_share` over the `LIMIT 10` top slice, so with more than 10 enrolled miners both denominators use only 10 rows and every displayed share is overstated. The fix normalizes against a separate full-epoch COUNT/SUM while still showing the top 10. This is display-only: `explorer/rustchain_dashboard.py` is the explorer app (the live node even carries a comment noting it is not the process nginx proxies to), and these fields have no settlement consumers. Correct fix.

Scottcjn:

PR #7949, opened 2026-07-13
fix(invariants): INV-5 antiquity check misses violations depending on miner listing order

Vyacheslav-Tomashevskiy:
## Problem

`SimulatedLedger.check_antiquity_weighting` (INV-5) is documented as an order-independent property: within an epoch, a miner with a strictly higher antiquity multiplier must never earn fewer uRTC than a lower-multiplier miner. But the check only inspects a pair when the *earlier-listed* miner has the higher multiplier:

```python
if ma.antiquity_multiplier > mb.antiquity_multiplier:
    if epoch.rewards[a] < epoch.rewards[b]:
        violations.append(...)
```

There is no symmetric branch for when the *later-listed* miner `b` has the higher multiplier. `epoch.rewards` is an insertion-ordered dict keyed by the arbitrary miner-attestation order (not sorted by multiplier), so a genuine violation is silently missed ~50% of the time — whenever the higher-multiplier miner happens to appear later in the list. For a suite that advertises itself as a mathematical proof of ledger correctness, this is a blind spot in the safety net.

The live-API sibling check (`live_api_checks`) already does this correctly (sort desc, compare adjacent), confirming the intended semantics.

## Fix

Pick the higher-multiplier miner of each pair (skip equal multipliers, which carry no ordering constraint) and assert it earns `>=` the other — covering both directions regardless of listing order.

## Verification

Added `testing/test_inv5_antiquity_ordering_asymmetry.py`: the same reward/multiplier violation is detected when the higher-mult miner is listed first but **missed** when listed second on current main; both are detected after the fix.

- New test: **fails on main, passes with fix**.
- Existing `testing/test_ledger_invariants.py` stays green; running `ledger_invariants.py` end-to-end still passes all invariants (no false positives introduced).

/claim RTCd1554f0f35576faf01d386a6be1c947f560dd0b7


Vyacheslav-Tomashevskiy:
/claim

Payout wallet: `RTCd1554f0f35576faf01d386a6be1c947f560dd0b7`

IcanBENCHurCAT (approved):
LGTM — INV-5 check now covers both directions of miner listing order. ✅

Scottcjn:
Verified. Confirmed the INV-5 check only flagged when the earlier-listed miner had the higher multiplier, with no symmetric branch, so a genuine violation where the higher-multiplier miner is listed second is silently missed (about a 50% blind spot given arbitrary attest ordering). The fix picks hi/lo by multiplier and continues on ties, making it order-independent, with a test proving both orderings are caught. Scope note: this is the offline `SimulatedLedger` invariant harness, not live consensus, so the blind spot was in the audit tool, not the chain. Correct fix.

Scottcjn:

PR #7950, opened 2026-07-13
fix(lock-ledger): auto-release refunds in-flight bridge deposits → supply inflation

Vyacheslav-Tomashevskiy:
## Problem — supply inflation / double-spend

A bridge **deposit** hard-debits the source at create time (`create_bridge_transfer`, `UPDATE balances ... - amount`, `source_debited=1`) and records a `lock_ledger` row tied to the `bridge_transfers` row (`unlock_at = now + BRIDGE_LOCK_EXPIRY_SECONDS`). Those funds are committed to the external transfer. The bridge resolves such locks **explicitly**:
- **void** → atomically refunds the source *and* voids the transfer (sets `source_debited=0`);
- **completion** → releases the lock *without* a refund, because the funds have already left the chain.

`auto_release_expired_locks()` (the unattended timer worker) refunded **any** expired `status=locked` row via `release_lock()`, including the bridge deposit lock — crediting the owner back **without** voiding the transfer. A later external confirmation then still completes the deposit cross-chain (completion skips the re-debit because `source_debited=1`), so the owner keeps **both** copies:

```
balance 100 → 90 (deposit hard-debit)
          → 100 (timer worker refunds the in-flight lock)  ← bug
          → 100 (external confirmation completes; no re-debit, source_debited=1)
```

Net: **10 RTC created from nothing.** Root cause: bridge-managed locks are governed by the bridge void/complete lifecycle, not a simple owner-refundable timer, but the generic worker treated them the same.

## Fix

In `auto_release_expired_locks()`, skip any lock whose `bridge_transfer_id is not None` (leave bridge-tied locks to the bridge’s own atomic void/complete path). `bridge_transfer_id` is the exact discriminator — plain `create_lock()` locks have it `None` and are still auto-released correctly. Adds a `skipped_bridge_managed` counter to the summary.

## Verification

`node/tests/test_lock_ledger_bridge_deposit_auto_release_inflation.py`, driving the **real** `bridge_api.create_bridge_transfer` + `update_external_confirmation` flow:
- `test_auto_release_does_not_refund_inflight_bridge_deposit` — reproduces the inflation; **fails on main** (`assert 100000000 == 90000000`), passes with the fix.
- `test_auto_release_still_refunds_standalone_non_bridge_lock` — over-reach guard: a non-bridge `admin_hold` lock is still auto-released.

Neighboring suites green with the fix: `tests/test_bridge_lock_ledger.py`, `node/tests/test_payout_worker_recovery.py`, `tests/test_available_balance_encumbrance.py` (88 passed).

/claim RTCd1554f0f35576faf01d386a6be1c947f560dd0b7


Vyacheslav-Tomashevskiy:
/claim

Payout wallet: `RTCd1554f0f35576faf01d386a6be1c947f560dd0b7`

IcanBENCHurCAT (approved):
LGTM — fix for auto-releasing refunds in-flight bridge deposits → supply inflation. ✅

Scottcjn:
Independently verified this one.

Confirmed the bug: pre-fix `auto_release_expired_locks` refunds every expired lock with no `bridge_transfer_id` guard, so an in-flight bridge deposit gets its hard-debited funds credited back while the cross-chain transfer still completes, minting RTC from nothing. This one IS on the live node path (imported by `rustchain_v2_integrated_v2.2.1_rip200.py`, exposed via the authenticated worker route), so it is higher urgency than the SDK-only reward split.

The fix is correctly scoped: it skips bridge-managed locks and keeps plain non-bridge locks auto-refunding, and the two regression tests cover both directions. The red CI is unrelated baseline drift (`test_fetchall_guard` stale baseline plus miner checksum pins on files this PR does not touch), which a rebase onto current main clears. Recommend rebase and merge.

Scottcjn:
Thirty-three of your PRs went in today, which moved `main` under this one and left it with conflicts. Nothing is wrong with the change itself.

Could you rebase onto current `main` and push? I will pick it straight back up. Everything I merged today passed the same check: your added tests pass on the branch and fail when the source change is reverted, so if the rebase is clean this should follow the same path.

Sorry for the churn. It is a consequence of the backlog being cleared in one go rather than anything you did.

Vyacheslav-Tomashevskiy:
Rebased locally onto current `main` (12e3f126). Turns out the production fix itself already landed in `main` independently — same guard logic (`bridge_transfer_id is not None` skip), just under the name `skipped_bridge_owned` instead of my `skipped_bridge_managed`. So I dropped the redundant code change and kept only the regression test (`node/tests/test_lock_ledger_bridge_deposit_auto_release_inflation.py`), which is *not* yet in `main`. Ran it locally against current `main`'s `lock_ledger.py` — both cases pass (bridge-owned lock is skipped, plain lock still auto-refunds).

Blocked on pushing the rebase: my PAT only has `repo` scope, and since this branch was 100+ commits behind, the push now carries commits that touch `.github/workflows/*` files (already merged into `main` a while back, just not on this branch ref yet) — GitHub blocks that without `workflow` scope on the token. Working on getting a token with the right scope; will push as soon as that's sorted, or split the test out into a fresh branch off current main if that's faster.

Scottcjn:
Closing this one — and I want to be precise about *why*, because "closed" should not read as "rejected."

**You were right, and you were right first.** I verified your read independently against current `main` (`4bc535b`). `node/lock_ledger.py:663-673` now carries exactly the guard you proposed:

```python
skipped_bridge_owned = 0
...
    if lock.bridge_transfer_id is not None:
        skipped_bridge_owned += 1
        continue
```

That is your fix, under a different variable name (`skipped_bridge_owned` vs your `skipped_bridge_managed`), reached independently while this PR sat in the queue. The supply-inflation bug you found — `auto_release_expired_locks` refunding in-flight bridge deposits whose balance the bridge still owns — is closed on production.

**The queue is the reason this happened, not your work.** You filed this on 2026-07-13. It was confirmed by a maintainer on 07-14, approved on 07-14, reviewed again on 07-27 — and then sat until `main` moved 100+ commits under it and independently grew the same fix. That is our failure end to end, and it cost you a merge you had earned. The bounty for finding it stands on its merits regardless of which commit carried the line.

**The one piece of this that is still genuinely missing from `main` is your test.** I checked `node/tests/` — there is no `test_lock_ledger_bridge_deposit_auto_release_inflation.py`, and nothing else covers the bridge-owned skip path. The guard on `main` is currently unprotected: someone could delete those three lines and every test would still pass. Your regression test is the thing that stops that from happening.

So: please open a fresh PR off current `main` with **just** that one test file. That is the "split the test out into a fresh branch off current main" option you offered, and it is the faster path — a single new file under `node/tests/` touches nothing in `.github/workflows/`, so it sidesteps the `workflow`-scope PAT problem entirely. No rebase, no force-push, no token change needed. I will review it promptly, and it is a merge on sight if it does what your local run says it does (bridge-owned lock skipped, plain lock still auto-refunds).

Thank you for this one. Finding a supply-inflation path in the lock ledger is exactly the work that matters most here, and the delay in acting on it was ours.

PR #7951, opened 2026-07-13
fix(payout): do not refund funds when the debit transaction rolled back

Vyacheslav-Tomashevskiy:
## Bug: phantom refund inflates balance when the debit transaction is rolled back

`PayoutWorker.process_withdrawal()` debits the sender inside a single `BEGIN IMMEDIATE` transaction (claim row → check balance → `balance = balance - (amount+fee)` → `COMMIT`), then broadcasts. The outer failure handler credits `amount + fee` back to the account on **any** pre-broadcast exception ("refund on broadcast failure") — but it assumes the debit committed.

If the debit `COMMIT` itself raises (the realistic case being `SQLITE_BUSY` / "database is locked" under concurrent workers), the inner `except` runs `ROLLBACK` and re-raises. The funds never left the account — yet the outer handler still refunds `amount + fee`, **creating money out of thin air**. This violates the module's own stated invariant that funds cannot be credited without a matching debit.

### Concrete scenario
account balance `100.0`, withdrawal `amount=10.0 fee=1.0`; the debit `COMMIT` fails and is rolled back (balance stays `100.0`); handler credits `+11.0` → balance becomes **`111.0`**.

### Fix
Track whether the debit actually committed (`funds_debited`, set `True` only after the debit `COMMIT` succeeds) and guard the refund `UPDATE accounts` with it. The withdrawal is still marked `failed`. Legitimate refunds — debit committed, broadcast later failed — are unchanged.

### Test
Adds `node/tests/test_wt2_refund_without_debit.py`, which forces the debit `COMMIT` to fail once (models `SQLITE_BUSY`) and asserts the balance stays at the original `100.0`.

- Fails on `main`: `assert 111.0 == 100.0`
- Passes with the fix; neighboring suites green (`test_payout_worker_recovery.py`, `test_fee_market.py`, `test_payout_preflight.py`, 35 passed).

/claim


Vyacheslav-Tomashevskiy:
/claim

IcanBENCHurCAT (approved):
LGTM — correct. Should not refund funds when debit transaction rolled back. ✅

Scottcjn:
Verified.

Confirmed the refund fires even when the debit COMMIT itself raises and rolls back (for example SQLITE_BUSY under concurrency), crediting funds that never actually left the account. The fix gates the refund on `funds_debited`, set only after a successful debit commit, and every legitimate post-debit raise still refunds correctly.

Blast radius is the standalone `payout_worker` daemon, not the live attestation node, and the path is MOCK_MODE gated today, so this is a latent leak that activates when production payouts are wired up rather than an active drain. Correct and defensive fix. Red CI is the same unrelated baseline drift set. Recommend rebase and merge.

Scottcjn:

PR #7952, opened 2026-07-13
fix(claims): dry_run settlement must not mutate stale verifying claims

Vyacheslav-Tomashevskiy:
## Bug: `dry_run` settlement preview mutates claim state and stalls payouts

`process_claims_batch(..., dry_run=True)` is documented as read-only — *"If True, don't actually process, just report what would be done."* But the stale-`verifying` handling runs **unconditionally, before** the `if dry_run:` early return. For every claim that has been in `verifying` longer than `max_wait_seconds // 2`, it calls `update_claim_status(... status="review_required")` — a **persistent** transition that pulls the claim out of the automated `verify → approve → settle` payout pipeline into manual-review limbo, where it will not be paid until a human re-approves it.

So merely running a settlement **preview** stalls a legitimate miner's payout. Any tooling that calls `dry_run=True` repeatedly (a dashboard/preview cron) turns this into a systematic payout stall.

### Fix
Gate only the persistent mutation behind `if not dry_run:`. The diagnostic warning is preserved (the existing `test_stale_verifying_claims_flagged` asserts it prints during dry_run), so nothing that relied on the warning breaks.

### Test
Adds `node/tests/test_wt1_dry_run_mutates_verifying.py`: inserts a stale `verifying` claim, runs `dry_run=True`, asserts the claim status is unchanged.

- Fails on `main`: `review_required != verifying`
- Passes with the fix; neighboring suite green (`test_claims_settlement.py`, 82 passed).

/claim


IcanBENCHurCAT (approved):
LGTM — dry_run must not mutate stale verifying claims. Clean fix. ✅

Scottcjn:
Verified. Confirmed the stale-verifying loop calls `update_claim_status(..., 'review_required')` before the `if dry_run:` early return, so a dry run persistently mutates state, which a preview must never do. The fix wraps the flagging loop in `if not dry_run:`. Blast radius: `process_claims_batch` has no production callers today (only tests and the module's own __main__), so the impact is latent until the batch settler is wired in, but the fix is correct and worth landing ahead of that. 

Scottcjn:

PR #7953, opened 2026-07-13
fix(rewards): re-check settled after ADM-fallback re-lock to prevent epoch double-credit

Vyacheslav-Tomashevskiy:
## Bug: ADM-failure fallback double-credits an epoch under concurrency

`settle_epoch_rip200` checks `already_settled` inside the first `BEGIN IMMEDIATE` transaction. On the anti-double-mining (ADM) failure → standard fallback path — reached when `RC_REQUIRE_ADM` is unset (the fleet default) and the ADM settlement raises — it does:

```python
db.rollback()                 # releases the IMMEDIATE write lock
...
db.execute("BEGIN IMMEDIATE") # re-acquires the write lock
# falls through to the standard crediting path, WITHOUT re-checking settled
```

The `already_settled` guard is checked exactly once, at the top, before the lock was released. It is **not** re-checked after `rollback()` drops the lock and it is re-acquired. A second settler that commits this epoch during that release window is invisible to this caller, which then credits the same fixed per-epoch reward pool a **second time** — `balances`, `ledger`, and `epoch_rewards` all doubled — and returns `ok=True` for an already-settled epoch. That is an on-chain double-emission.

The existing race suite (`tests/test_rewards_settle_race.py`) only covers the ADM **success** path (safe — it shares the same connection/lock) and the plain standard path; the ADM **failure → fallback** re-lock had no coverage and no guard.

### Concrete scenario
Two workers settle epoch 0, ADM broken for both. W1 passes the initial `settled=0` check, ADM fails, W1 `rollback()`s (releasing the lock). W2 grabs the lock, fully settles + commits epoch 0, releases. W1 re-`BEGIN IMMEDIATE`s, does **not** re-check settled, credits again → `m1: 200, m2: 400`.

### Fix
Re-check `epoch_state.settled` immediately after the fallback `BEGIN IMMEDIATE`; short-circuit with `already_settled=True` if a concurrent worker won the race. Mirrors the existing top-of-function guard, now applied to the re-locked window.

### Test
Adds `node/test_wt3_adm_fallback_double_credit_poc.py`: a deterministic PoC that injects a committed concurrent settlement into the exact `rollback()` → `BEGIN IMMEDIATE` window via a `rollback` hook and asserts the epoch is credited exactly once.

- Fails on `main`: `{m1:200, m2:400}` (double)
- Passes with the fix; `tests/test_rewards_settle_race.py` green (5 passed).

/claim


Scottcjn:
Maintainer validation: the new deterministic ADM fallback race regression and the focused reward race suite pass locally (`6 passed`). The settled-state recheck is placed after the fallback lock is reacquired, which closes the double-credit window described by the PR. The aggregate CI failure is from the stale main-branch baseline repaired in #7957, not from these changes.

PR #7954, opened 2026-07-13
[UTXO-BUG] Fix non-atomic mempool_clear_expired — wrap in BEGIN IMMEDIATE

sumarna35:
## Fix for UTXO Red Team Audit Finding #1

**Refs:** Scottcjn/rustchain-bounties#2819

### Bug
`mempool_clear_expired()` deletes mempool inputs+mempool rows per-tx without transaction wrapper. Crash between DELETEs → orphan state (UTXO prematurely unlocked, stale mempool entry remains).

### Fix
Wrap delete loop in `BEGIN IMMEDIATE` + `COMMIT` — same pattern as `mempool_remove()`.

### Audit Finding
Full report: https://github.com/Scottcjn/rustchain-bounties/issues/2819#issuecomment-4954595626

Wallet: RTCad35e1c1746560097ce9989abc140e0529b87163


Scottcjn:
Maintainer review: the stated partial-commit failure is not reproduced with the connection settings used here. Python sqlite3 opens an implicit transaction on the first DELETE under the default isolation mode; both DELETEs remain in that transaction until `conn.commit()`, and closing after only the first DELETE rolls it back. I verified this directly: `conn.in_transaction` becomes true after the first DELETE and both rows remain after closing without a commit.

`BEGIN IMMEDIATE` may still provide a useful earlier write lock, but this patch begins it after the SELECT and currently has no regression test demonstrating a behavior change. Please add a deterministic test that fails on current `main` and passes this branch, and revise the docstring to describe the demonstrated locking issue rather than a partial commit. The red aggregate CI result is also from the stale main-branch baseline covered by #7957, not evidence for this patch.

IcanBENCHurCAT (approved):
LGTM — audit finding fix. Wraps paired DELETEs in BEGIN IMMEDIATE to prevent orphan UTXO rows on crash. ✅

Scottcjn:
Verified, with one correction to the framing.

The concurrency hardening is real and worth taking: grabbing the write lock up front matches the `BEGIN IMMEDIATE` pattern `mempool_add` already uses, plus the empty-set early return is a nice touch. But the stated corruption impact is overstated. `_conn()` uses default deferred isolation, so the pre-fix DELETEs already run inside one implicit transaction committed once at the end, and a mid-loop crash already rolls everything back. There are no orphan `utxo_mempool_inputs` rows under the current config.

So this is a safe belt-and-suspenders improvement, not the critical corruption fix the description implies. The change is correct and deadlock-safe (`timeout=30` makes IMMEDIATE block-and-retry rather than raise). Red CI is unrelated baseline drift. Recommend rebase and merge as a hardening cleanup, and scored as a minor finding rather than high severity.

sumarna35:
Thanks again for the thorough review, and you were right: the partial-commit
corruption claim does not hold up under Python sqlite3's default deferred
isolation (the implicit transaction is held until ``conn.commit()``, so a
mid-loop crash rolls everything back). I'm withdrawing that framing.

**What I changed (just pushed to the branch, commit `325cdef3`):**

1. **Docstring reworded** to describe the demonstrated benefit
   (write-lock hardening / matching the existing ``mempool_add`` /
   ``mempool_remove`` pattern) rather than the unproven corruption claim.

2. **On the regression test:** I tried to construct one, but I think you are
   correct that there is no deterministic test that fails on current ``main``
   and passes this branch under the deferred-isolation model. A
   half-finished sweep cannot be observed externally; only a contention
   timing test would, and those are flaky in CI. If you have a specific test
   scenario in mind, I'd appreciate a pointer — otherwise I'd rather not
   land a placebo test that just measures the IMMEDIATE keyword without
   proving the failure mode you actually want to prevent.

**Scoring:** I accept your re-tiering to minor / hardening. If this is still
worth landing as a lock-discipline cleanup, great; if you'd rather close it
as not-pursued, I understand. Either way, thanks for the time and the
correction — the audit report on #2819 will be updated to reflect the
revised impact assessment.

Scottcjn:
Closing as superseded: `mempool_clear_expired` on `main` now runs inside an explicit transaction (there is a stress test asserting it). Thank you for finding it. — Sophia

PR #7956, opened 2026-07-13
fix(hall_of_rust): PowerPC arch bonus never applied (upper-case keys vs lower-cased arch)

Vyacheslav-Tomashevskiy:
### Problem
`calculate_rust_score()` in `node/hall_of_rust.py` lower-cases the reported architecture:

```python
arch = machine.get('device_arch', 'modern').lower()
    if key in arch:   # case-sensitive substring test
```

…but the PowerPC entries in `arch_bonus` used **upper-case** keys `'G3'/'G4'/'G5'`. Because `'G4' in 'g4'` is `False`, a PowerPC machine — the exact vintage hardware the Hall of Rust exists to celebrate — **never received its architecture bonus**.

Concrete failing inputs (pure function, `current_year` passed explicitly, all other score terms 0):

| device_arch | expected | actual (main) |
|---|---|---|
| `G4` | 70.0 | **0.0** |
| `G3` | 80.0 | **0.0** |
| `G5` | 60.0 | **0.0** |

This silently under-scores PowerPC machines, corrupting leaderboard ordering (`ORDER BY rust_score DESC`) and badge assignment.

### Fix
Lower-case the three PowerPC keys to match the already-lower-cased `arch` value:

```diff
-        'G3': 80, 'G4': 70, 'G5': 60,
+        'g3': 80, 'g4': 70, 'g5': 60,
```

The sibling `estimate_manufacture_year()` already uses `key.lower() in arch.lower()`, confirming lower-case is the intended convention.

### Test
Adds `node/test_hall_of_rust_arch_bonus_poc.py`:
- **Fails on main:** `G3/G4/G5` return `0.0`.
- **Passes with fix:** all 5 cases green (`pytest ... -q` → `5 passed`).
- Nearby suites unaffected: `test_hall_of_rust_limit_validation.py` (6), `test_hall_of_rust_error_responses.py` (8), `explorer/test_hall_of_rust_non_object.py` (15) all pass.


Scottcjn:
Maintainer validation: all 19 focused multiplier and profile tests pass locally. Normalizing the profile keys to lowercase matches the already-lowercased architecture lookup and restores the intended G3/G4/G5 multipliers. The aggregate CI failure is the stale main-branch fetchall/checksum baseline repaired in #7957, not a regression in this PR.

PR #7958, opened 2026-07-13
fix(adm): warthog bonus uncapped on own-connection path settles epoch differently than existing-connection path

Vyacheslav-Tomashevskiy:
## Problem

`settle_epoch_with_anti_double_mining` dispatches to one of two reward calculators purely on whether the caller passes a live connection:

```python
if existing_conn is not None:
    rewards, telemetry = _calculate_anti_double_mining_rewards_conn(...)   # caps warthog bonus at 2.0
else:
    rewards, telemetry = calculate_anti_double_mining_rewards(...)          # applied it UNCAPPED
```

`_calculate_anti_double_mining_rewards_conn` is documented as *"Same as calculate_anti_double_mining_rewards but uses an existing connection"* and caps the warthog dual-mining bonus:

```python
# Apply capped warthog bonus (MAX = 2.0) to prevent reward inflation
if 1.0 < bonus <= 2.0:
    weight *= bonus
elif bonus > 2.0:
    weight *= 2.0
```

But the own-connection sibling applied it with no cap (`if wart_row[0] > 1.0: weight *= wart_row[0]`). So for any miner whose `warthog_bonus` exceeds 2.0, the **same epoch settles to a different reward split depending on which entry point the caller used** — a non-deterministic ledger fork between the two settlement paths. The cap comment in the sibling establishes that 2.0 is the intended safeguard against exactly this reward inflation.

## Fix

Apply the identical `MAX = 2.0` cap in the own-connection path so the two documented-equivalent functions actually agree.

## Verification

Added `test_both_paths_cap_warthog_bonus_identically` — settles the same epoch (one miner with `warthog_bonus = 3.0`) through both paths and asserts identical output.

- **On main:** own-connection `{'miner-heavy': 10066, 'miner-light': 34}` (uncapped ×3.0) vs existing-connection `{'miner-heavy': 10049, 'miner-light': 51}` (capped ×2.0) → **FAIL**
- **With fix:** both return the capped `{'miner-heavy': 10049, 'miner-light': 51}` → **PASS**

Full `test_anti_double_mining.py` + `test_anti_double_mining_epoch_enroll.py` suites green (23 + epoch-enroll).

### DIFF
--- a/node/anti_double_mining.py
+++ b/node/anti_double_mining.py
@@ -575,8 +575,17 @@ def calculate_anti_double_mining_rewards(
                         "SELECT warthog_bonus FROM miner_attest_recent WHERE miner=?",
                         (miner_id,)
                     ).fetchone()
-                    if wart_row and wart_row[0] and wart_row[0] > 1.0:
-                        weight *= wart_row[0]
+                    # Apply capped warthog bonus (MAX = 2.0) to prevent reward inflation.
+                    # Must match _calculate_anti_double_mining_rewards_conn exactly, or the
+                    # same epoch settles to a different split depending on whether the caller
+                    # passed an existing connection (settle_epoch_with_anti_double_mining
+                    # dispatches to the two paths on that condition).
+                    if wart_row and wart_row[0]:
+                        bonus = float(wart_row[0])
+                        if 1.0 < bonus <= 2.0:
+                            weight *= bonus
+                        elif bonus > 2.0:
+                            weight *= 2.0
                 except Exception:
                     pass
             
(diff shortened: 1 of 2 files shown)


IcanBENCHurCAT (approved):
LGTM — critical fix. Prevents ledger forks from uncapped warthog bonus in one code path. Test proves both paths now produce identical splits. ✅

Scottcjn:
Verified, with a severity correction. The divergence is real (the own-connection path did an uncapped `weight *= wart_row[0]` while the _conn path caps at 2.0), but it is NOT reachable on the live settle path: the only production caller, `settle_epoch_rip200`, always passes `existing_conn=db` and so dispatches to the already-capped path. The buggy own-connection variant is only hit by the module's __main__ demo and tests. So this is good defense-in-depth to keep the two functions consistent, not a live consensus fork as the title implies. Fix is correct. Merge.

CI is green.

Scottcjn:

PR #7960, opened 2026-07-13
fix(hall_of_rust): explorer Rust Score drops PowerPC arch bonus (case-mismatch, twin of #7956)

Vyacheslav-Tomashevskiy:
## Problem

`explorer/hall_of_rust.py::calculate_rust_score()` lower-cases the reported `device_arch` before matching architecture bonuses, but the PowerPC entries in the `arch_bonus` table use **upper-case** keys:

```python
arch_bonus = {
    'G3': 80, 'G4': 70, 'G5': 60,   # upper-case keys
    ...
}
arch = machine.get('device_arch', 'modern').lower()   # value is lower-cased
    if key in arch:                 # case-sensitive substring test
        score += bonus
```

`'G4' in 'g4'` is `False`, so **every PowerPC machine** — the exact vintage hardware the Hall of Rust exists to celebrate — silently loses its architecture bonus. It scores **0** instead of **60–80**, which corrupts `rust_score`, leaderboard ordering (`ORDER BY rust_score DESC`) and badge tiers.

Reproduced against the current module:

| `device_arch` | expected | actual (main) |
|---|---|---|
| `G3` | 80 | **0** |
| `G4` / `g4` | 70 | **0** |
| `G5` | 60 | **0** |
| `486` | 150 | 150 ✓ |

## Why this is the right fix

This is the **unfixed twin** of the node sibling `node/hall_of_rust.py`, which was corrected in **#7956** (`fix(hall_of_rust): PowerPC arch bonus never applied`). The explorer copy was missed. Additionally, the sibling `estimate_manufacture_year()` **in this same file** already uses `key.lower() in arch.lower()`, confirming lower-case is the intended convention.

The fix simply lower-cases the three PowerPC keys to match — the other keys are already lower-case. Minimal, mirrors the merged sibling fix exactly.

## Test

Adds `tests/test_explorer_hall_of_rust_arch_bonus.py` — fails on `main` for `G3`/`G4`/`G5`/`g4`, passes with the fix. Existing `tests/test_explorer_hall_of_rust_current_year.py` (10 tests) stays green.

```
# before fix: 4 failed, 2 passed
# after fix:  6 passed
```

### DIFF
--- a/explorer/hall_of_rust.py
+++ b/explorer/hall_of_rust.py
@@ -123,8 +123,11 @@ def calculate_rust_score(machine):
         score += RUST_WEIGHTS['first_attestation']
     
     # Architecture bonuses
+    # NOTE: keys must be lower-case — `arch` below is lower-cased before the
+    # substring match, so upper-case keys (e.g. 'G4') could never match and the
+    # PowerPC bonus was silently dropped. Mirrors the node sibling fix (#7956).
     arch_bonus = {
-        'G3': 80, 'G4': 70, 'G5': 60,
+        'g3': 80, 'g4': 70, 'g5': 60,
         '486': 150, 'pentium': 100, 'pentium4': 50,
         'retro': 40, 'apple_silicon': 5, 'modern': 0
     }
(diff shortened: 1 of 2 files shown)


Vyacheslav-Tomashevskiy:
Closing as superseded — this exact fix already landed in main via #7956 (commit e8d372bd), which lower-cases both the arch_bonus keys and the reported device_arch when matching. No further change needed here.

PR #7964, opened 2026-07-14
fix(tier3): validator average_score divides by validated+rejected instead of validated count

Vyacheslav-Tomashevskiy:
## Problem

`ValidatorAgent.validate_poa_proof` (`tier3/agents/validator_agent.py`) maintains `average_score` as a running mean over **validated** scores — rejected proofs contribute no score to the numerator (the `else` branch only bumps `total_rejected`). But the update divided by the **combined** count:

```python
if valid:
    self.stats["total_validated"] += 1
    total = self.stats["total_validated"] + self.stats["total_rejected"]   # <-- wrong divisor
    self.stats["average_score"] = (
        (self.stats["average_score"] * (total - 1) + score) / total
    )
```

Because the numerator only ever sums valid scores, dividing by `validated + rejected` dilutes the mean whenever any proof was rejected.

## Impact

- Reject one proof, then validate one scoring 100 → `total_validated=1`, `total_rejected=1`, so `average_score = (0*1 + 100) / 2 = 50.0` instead of `100.0`.
- Sequence 90 (valid), reject, 60 (valid) → reports `80.0`, correct mean is `75.0`.

The error appears whenever any proof is rejected and surfaces through `get_stats()` and up through `PipelineOrchestrator.get_stats()` as `validator_stats`.

## Fix

Divide by `total_validated` — the count of scores actually summed — so numerator and denominator agree:

```python
if valid:
    self.stats["total_validated"] += 1
    n = self.stats["total_validated"]
    self.stats["average_score"] = (
        (self.stats["average_score"] * (n - 1) + score) / n
    )
```

The sibling `RewardAgent` (`tier3/agents/reward_agent.py`) already follows this rule for `average_reward` — it divides the summed value by `total_distributed`, the count actually accumulated.

## Test

Added `test_average_score_excludes_rejected_from_denominator` to `tier3/tests/test_pipeline.py`: reject one proof, validate one (score 100), assert `average_score == 100`. **Fails on main** (`50.0`), passes with the fix. The existing `test_validation_stats_tracking` only ever validates (no rejects), so it never exercised this path. Full `tier3/tests/test_pipeline.py` suite: **33 passed**.

### DIFF
--- a/tier3/agents/validator_agent.py
+++ b/tier3/agents/validator_agent.py
@@ -168,9 +168,13 @@ class ValidatorAgent:
         # Update stats
         if valid:
             self.stats["total_validated"] += 1
-            total = self.stats["total_validated"] + self.stats["total_rejected"]
+            # average_score is the running mean over VALIDATED scores only
+            # (rejected proofs never contribute a score to the numerator), so
+            # the divisor must be the count of validated proofs, not the
+            # combined validated+rejected total.
+            n = self.stats["total_validated"]
             self.stats["average_score"] = (
-                (self.stats["average_score"] * (total - 1) + score) / total
+                (self.stats["average_score"] * (n - 1) + score) / n
             )
         else:
             self.stats["total_rejected"] += 1
--- a/tier3/tests/test_pipeline.py
+++ b/tier3/tests/test_pipeline.py
@@ -155,7 +155,25 @@ class TestValidatorAgent:
         assert stats["total_validated"] == 5
         assert stats["total_rejected"] == 0
         assert stats["average_score"] > 0
-    
+
+    def test_average_score_excludes_rejected_from_denominator(self, validator_agent, sample_poa_submission):
+        """average_score is the mean over VALIDATED scores only; rejected proofs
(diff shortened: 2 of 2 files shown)


IcanBENCHurCAT (approved):
LGTM — clean fix. average_score should only divide by validated proofs, not validated+rejected. Well-documented with a clear regression test. ✅

IcanBENCHurCAT (approved):
LGTM — clean fix. average_score divides by validated count only, not validated+rejected. Well-documented with clear regression test. ✅

Scottcjn:
Verified, with a scope note. The bug is real: `average_score` summed only validated scores into the numerator but divided by `validated + rejected`, diluting the mean whenever a proof was rejected. The fix divides by validated-only, matching the summed terms. But this is an internal tier3 experimental-pipeline telemetry metric (get_stats), not consensus or on-chain reward math, and tier3 is not imported by the live node, so 'economically sensitive validator scoring' overstates it. Correct, well-scoped fix. Merge.

CI is green.

Scottcjn:

PR #7966, opened 2026-07-14
fix(explorer): network stats active_miners always 0 (reads status the node never sends)

Vyacheslav-Tomashevskiy:
### Bug
`GET /api/network/stats` (`get_network_stats`, explorer/app.py) reports `active_miners` by counting `m.get('status') == 'online'` over the raw `/api/miners` node payload. But that payload never carries a `status` field — `status` is *derived* from `last_seen` only inside the sibling `get_miners()` (online/idle/offline/unknown). So `m.get('status')` is always `None` and **`active_miners` is always 0**, no matter how many miners are actually online.

### Fix
Derive "online" (seen within the last 5 minutes) the same way `get_miners()` does, via a small shared `_miner_is_online()` helper, and count with it. Missing/unparsable `last_seen` → not online (matches the `'unknown'` branch).

### Verification
Added `explorer/test_network_stats_active_miners.py`:
- Payload with `last_seen` only (no `status`, as the node actually sends): 2 recent + 1 idle + 1 offline → `active_miners == 2`.
- Missing / non-numeric `last_seen` are excluded → `active_miners == 1`.

Both tests **fail on main** (`active_miners == 0`) and pass with the fix. Existing explorer suite unaffected.

/claim

### DIFF
--- a/explorer/app.py
+++ b/explorer/app.py
@@ -27,16 +27,32 @@ def _upstream_node_unavailable(include_miners=False):
     return jsonify(payload), 500
 
 
-def _miner_is_online(miner):
-    """Return True if the miner was seen within the last 5 minutes.
+def _miner_last_seen_ts(miner):
+    """Return the miner's freshness timestamp, or None if it can't be read.
 
-    The raw node payload from /api/miners does not carry a 'status' field —
-    it is derived from last_seen in get_miners(). Network stats must derive it
-    the same way instead of reading a key the node never sends.
+    The node's /api/miners payload calls this field 'last_attest'; 'last_seen'
+    is checked first for any other producer that uses that name. Reading only
+    'last_seen' means the real node payload never matches.
     """
-    try:
-        last_seen = float(miner['last_seen'])
-    except (KeyError, TypeError, ValueError):
+    for key in ('last_seen', 'last_attest'):
+        if key not in miner:
+            continue
+        try:
+            return float(miner[key])
+        except (TypeError, ValueError):
+            return None
+    return None
+
+
+def _miner_is_online(miner):
+    """Return True if the miner attested within the last 5 minutes.
+
+    The raw node payload carries no 'status' field — it is derived here and in
+    get_miners(). Network stats must derive it the same way instead of reading
+    a key the node never sends.
+    """
+    last_seen = _miner_last_seen_ts(miner)
+    if last_seen is None:
         return False
     return (datetime.now().timestamp() - last_seen) < 300
 
@@ -57,18 +73,18 @@ def get_miners():
                 if 'uptime' in miner:
                     miner['uptime_percentage'] = min(100, (miner['uptime'] / 86400) * 100)
                 
+                last_seen = _miner_last_seen_ts(miner)
+
                 # Format last seen timestamp
-                if 'last_seen' in miner:
+                if last_seen is not None:
                     try:
-                        timestamp = datetime.fromtimestamp(miner['last_seen'])
+                        timestamp = datetime.fromtimestamp(last_seen)
                         miner['last_seen_formatted'] = timestamp.strftime('%Y-%m-%d %H:%M:%S')
                     except (TypeError, ValueError, OSError, OverflowError):
                         miner['last_seen_formatted'] = 'Unknown'
-                
+
                 # Set status based on last seen
-                try:
-                    last_seen = float(miner['last_seen'])
-                except (KeyError, TypeError, ValueError):
+                if last_seen is None:
                     miner['status'] = 'unknown'
                 else:
                     time_diff = datetime.now().timestamp() - last_seen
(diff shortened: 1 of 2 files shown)


Scottcjn:
Verified, and flagging the fix as incomplete. The bug is real: `active_miners` counted `status == 'online'`, but the node's /api/miners payload has no `status` key, so the count was always 0. The fix derives online from `last_seen < 300s`. Gotcha: the node actually emits `last_attest`, not `last_seen`, so against the real payload the guarded except returns False and active_miners may STILL be 0 in production; the passing test uses a synthetic `last_seen` payload. It is consistent with the existing get_miners() convention and is no-regression, so it is safe to merge, but a follow-up should reconcile last_seen vs last_attest to actually fix the count. Blast radius is the standalone explorer dashboard.

CI is green.

Vyacheslav-Tomashevskiy:
Good catch — you were right, and it was worse than the count.

Confirmed against `node/rustchain_v2_integrated_v2.2.1_rip200.py:api_miners`: the payload carries `last_attest`, with no `last_seen` and no `status` anywhere. So my `last_seen < 300s` derivation would have kept `active_miners` at 0 in production exactly as you said — my test only passed because the synthetic payload invented a field the node never sends.

The same blind spot was already live in `get_miners()` on main, before either of us touched it: it reads `float(miner['last_seen'])`, hits `KeyError` on every real miner, and stamps `status: 'unknown'` across the board. That's also why the original `m.get('status') == 'online'` never matched — the field it wanted was never populated for the real payload either.

Pushed the reconciliation rather than leaving it as a follow-up:

- `_miner_last_seen_ts(miner)` reads `last_seen`, then falls back to `last_attest`. `last_seen` keeps precedence so any other producer using that name is unaffected.
- Both `get_miners()` (status + `last_seen_formatted`) and the stats count now route through it, so there's one definition of freshness instead of two.

Tests now use the real node payload shape (`last_attest`, `miner`, `antiquity_multiplier`, pagination) instead of the synthetic one:

- `test_active_miners_counts_real_node_last_attest_payload` — the count against what the node actually sends.
- `test_get_miners_status_derived_from_real_node_payload` — status is online/idle, not 'unknown'.
- `test_last_seen_takes_precedence_over_last_attest` — pins the fallback order.

Both new tests fail on my previous revision of this PR and pass now; 5 passed total. Blast radius is unchanged — still the standalone explorer.

Vyacheslav-Tomashevskiy:
Pushed 6578ad17 — the `test` job was red on the previous head and this fixes it. Owning the regression: it was mine, not a flake.

**What broke:** the earlier commit routed both `/api/miners` and `/api/miner/<id>` through `_miner_last_seen_ts()`, which returns `None` for *two different* cases — field absent, and field present but unparsable. `get_miners()` then only set `last_seen_formatted` when the value parsed, so a miner reporting `last_seen: "not-a-timestamp"` silently lost the `'Unknown'` marker the original code produced. `tests/test_explorer_last_seen_validation.py::test_miners_api_handles_invalid_last_seen_without_500` caught it correctly with `KeyError: 'last_seen_formatted'`.

**Fix:** extract `_annotate_miner_freshness()`, shared by both routes. Unreadable timestamp → `'Unknown'`; no freshness field at all → no key, exactly as before. It also removes the duplicated status/format block that `/api/miner/<id>` carried, so the two routes can't drift apart again — that duplication is what let the miss through in the first place.

**Verified locally:**
- previous head (what CI ran): `test_miners_api_handles_invalid_last_seen_without_500` FAILS, reproduced the exact `KeyError`
- this head: the whole file passes; the 5 tests for the original `active_miners == 0` bug still fail on a clean `upstream/main` (96660f06) and pass here
- explorer/miner/hall selection: clean main = 12 failed / 608 passed; this branch = the **same 12** failed / 613 passed. The 12 are pre-existing (`test_hall_of_rust_non_object.py` and friends, untouched by this PR); the +5 are the new ones.

No behaviour change to the actual `active_miners` fix under review — this only restores the `'Unknown'` path I'd clipped.

Vyacheslav-Tomashevskiy:
Good catch, and it's already handled — you reviewed the branch at 983ac259, before I pushed the reconciliation at 1e402387. No follow-up PR needed; the fix is in this one.

The helper reads `('last_seen', 'last_attest')` in order, so the real node payload (which emits `last_attest` and no `status`) resolves. That key order follows the existing convention in `explorer_websocket_server.py:203`.

I checked it against the actual producer rather than only the synthetic test. `api_miners()` in `node/rustchain_v2_integrated_v2.2.1_rip200.py:7826` emits `last_attest` as a unix int with no `status` key, and driving `_miner_is_online()` with exactly that shape gives True for a 60s-old attest and False for a 2h-old one. Worth noting the node already filters to `ts_ok > now - 3600`, so anything in the payload attested within the hour.

5 tests pass. Blast radius is still just the standalone explorer dashboard.

Scottcjn:

PR #7967, opened 2026-07-14
fix(hall_of_rust): recompute rust_score on re-attestation (frozen leaderboard)

Vyacheslav-Tomashevskiy:
### Problem
`POST /hall/induct` on an already-inducted machine increments `total_attestations` but never recomputes `rust_score`. Since `calculate_rust_score()` rewards attestation loyalty via `RUST_WEIGHTS['attestation_count']`, a machine's leaderboard score stays **frozen at its induction-time value** (attestations == 1) no matter how many times it re-attests. The `/hall/leaderboard` ordering therefore ignores accumulated attestations entirely — the loyalty weight is effectively dead.

The new-induction path already computes and stores `rust_score`; the re-attestation path just forgot to.

### Fix
After bumping the attestation count, re-read the row and recompute `rust_score` (mirroring the induction path), then persist it. The refreshed score is also returned in the response so callers can see their current standing.

### Test
`tests/test_explorer_hall_of_rust_reattestation_score.py` — re-attests 10x and asserts the stored score grows by `10 * attestation_count`, plus the response reports the current score. **Both tests fail on `main`** (score frozen / `rust_score` missing from response) and pass with this change. Existing hall_of_rust tests (`current_year`, `non_object`) still green.

RTC address: RTCd1554f0f35576faf01d386a6be1c947f560dd0b7

### DIFF
--- a/explorer/hall_of_rust.py
+++ b/explorer/hall_of_rust.py
@@ -190,19 +190,41 @@ def induct_machine():
         arch = data.get('device_arch', 'modern')
         
         if existing:
-            # Update attestation count
+            # Update attestation count, then recompute the Rust Score.
+            # calculate_rust_score() rewards attestation loyalty, but the
+            # re-attestation path used to bump total_attestations without
+            # recomputing rust_score, so a machine's leaderboard score was
+            # frozen at its induction-time value (attestations == 1) forever.
             c.execute("""
-                UPDATE hall_of_rust 
+                UPDATE hall_of_rust
                 SET total_attestations = total_attestations + 1,
                     last_attestation = ?
                 WHERE fingerprint_hash = ?
             """, (now, fingerprint_hash))
+
+            c.execute("""
+                SELECT id, manufacture_year, device_arch, device_model,
+                       total_attestations, thermal_events
+                FROM hall_of_rust WHERE fingerprint_hash = ?
+            """, (fingerprint_hash,))
+            updated = c.fetchone()
+            new_score = calculate_rust_score({
+                'id': updated[0],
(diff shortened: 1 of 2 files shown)


Scottcjn:
Verified, and flagging incompleteness. The bug is real: the re-attestation branch bumps attestation count and timestamp but never recomputes `rust_score`, freezing the loyalty weight at the induction value. The fix recomputes and writes it back, which is correct for the file it edits. Gotcha: this edits `explorer/hall_of_rust.py`, whose blueprint is registered nowhere. The copy the live node actually registers, `node/hall_of_rust.py`, has the identical bug UNFIXED on its /hall/induct path. So this is a correct improvement to an unwired duplicate, but the live 'frozen leaderboard' is not resolved by it alone; a follow-up should target the live file. (There are three divergent copies of this scorer, which is worth consolidating.)

CI is green.

Vyacheslav-Tomashevskiy:
The live-file follow-up is already open as #7971 — it fixes the same re-attestation path in `node/hall_of_rust.py`, which is the copy `rustchain_v2_integrated_v2.2.1_rip200.py` actually imports (`from hall_of_rust import hall_bp`, line 746). So the frozen leaderboard is resolved there, not here.

You're right that this PR only improves an unwired duplicate — `explorer/hall_of_rust.py` registers nowhere. It's no-regression, but #7971 is the one that matters for the live path.

Agreed on the three divergent copies. Happy to do a consolidation PR that makes the explorer and bottube copies import the node scorer instead of carrying their own drift, if you want that as its own change — say the word and I'll scope it.

Scottcjn:

PR #7968, opened 2026-07-14
fix(parasocial): scope weekly top_commenters/most_active_video to the week

Vyacheslav-Tomashevskiy:
## Bug
`AudienceTracker._update_weekly_stats` derives each week's `top_commenters` and `most_active_video` by aggregating over the tracker's **all-time** `self.video_comments`:

```python
commenter_counts = defaultdict(int)
for vid_users in self.video_comments.values():   # <-- all-time, every week
    for uid in vid_users:
        commenter_counts[uid] += 1
```

So every `WeeklyStats` collapses onto the same all-time leaders. `get_weekly_top_commenters()` and `generate_shoutouts()` are documented as "top commenters **this week**", but they return identical all-time results every week — the whole point of weekly community shoutouts (fresh faces each week) is lost, and `most_active_video` is always the highest all-time video.

## Fix
Track per-week tallies on `WeeklyStats` (`commenter_counts`, `video_counts`), incremented only for the comment being recorded in its own week, and derive `top_commenters`/`most_active_video` from those. Persisted through `to_dict`/`_load_state`.

## Tests
`tests/test_bottube_parasocial_weekly_scope.py` (3 tests): week 1 = alice/vid1, week 2 = bob/vid2 only. On main week 2 returns the all-time leader (alice / vid1) → **fail**; with the fix week 2 returns bob / vid2 → **pass**. Existing `tests/test_parasocial_hooks.py` (33 tests) stays green.


Scottcjn:
Verified. Confirmed the pre-fix `_update_weekly_stats` derived each week's top_commenters/most_active_video from the tracker's all-time totals, so every week collapsed onto all-time leaders. The fix adds per-week counts and serializes them for round-trip, with a scope + persistence test. Blast radius is the bottube parasocial integration, no economic surface. Correct fix. One note: a duplicate AudienceTracker in tools/bottube_parasocial.py is not touched by this PR.

CI is green.

Scottcjn:

PR #7969, opened 2026-07-14
fix(hall_of_rust): Pentium 4 arch bonus never applied (substring order)

Vyacheslav-Tomashevskiy:
## Problem
`node/hall_of_rust.py::calculate_rust_score()` scores a machine's architecture with a substring match that stops at the first hit:

```python
    if key in arch:
        score += bonus
        break
```

`arch_bonus` lists both `'pentium': 100` and `'pentium4': 50`. Since `'pentium'` is a substring of `'pentium4'` **and** comes first in dict-iteration order, a Pentium 4 machine matches `'pentium'` and receives **100** instead of its intended **50**. The `'pentium4': 50` entry is therefore unreachable dead code, and every Pentium 4 sits **50 points too high** on the `ORDER BY rust_score DESC` Hall of Rust leaderboard (and gets the wrong badge tier).

Reproduced on current `main` (e8d372bd):
```
calculate_rust_score({'device_arch': 'pentium4'})  ->  100.0   # expected 50.0
```

## Fix
Iterate `arch_bonus` **longest-key-first** so the most specific substring wins:

```python
for key, bonus in sorted(arch_bonus.items(), key=lambda kv: -len(kv[0])):
```

This resolves the `pentium`/`pentium4` overlap, is robust to any future overlapping keys, and does **not** regress the PowerPC lower-case fix from #7956 (`g4`/`486`/`pentium` cases stay green).

## Test
Adds `node/test_hall_of_rust_pentium4_bonus.py`:
- **Fails on main** - `pentium4` and `Pentium4 Northwood` return `100.0`.
- **Passes with fix** - both return `50.0`; `pentium`, `g4`, `486` unchanged.

Existing `node/test_hall_of_rust_arch_bonus_poc.py` (the #7956 PowerPC regression suite) still passes - 9 passed together.

---

### DIFF
--- a/node/hall_of_rust.py
+++ b/node/hall_of_rust.py
@@ -140,7 +140,11 @@ def calculate_rust_score(machine, current_year=None):
         'retro': 40, 'apple_silicon': 5, 'modern': 0
     }
     arch = machine.get('device_arch', 'modern').lower()
-    for key, bonus in arch_bonus.items():
+    # Match the most specific key first: 'pentium' is a substring of 'pentium4',
+    # so iterating in dict order would match 'pentium' (100) for a Pentium 4 and
+    # never reach 'pentium4' (50), leaving that entry unreachable and inflating
+    # the leaderboard score. Longest key first makes the specific match win.
+    for key, bonus in sorted(arch_bonus.items(), key=lambda kv: -len(kv[0])):
         if key in arch:
             score += bonus
             break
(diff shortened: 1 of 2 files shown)


Scottcjn:
Verified. Confirmed the substring-order bug: the arch-bonus loop tests `'pentium'` before `'pentium4'`, and since 'pentium' is a substring of the P4 arch string it matches first, making the `'pentium4'` entry dead code. The fix sorts by key length so the longest match wins, with tests. Important scope point that resolves the obvious worry: this only affects the Hall of Rust leaderboard `rust_score`, NOT the reward multiplier. The real antiquity multiplier lives in a separate exact-match lookup (pentium4 distinct from pentium), so this does not change the RTC a P4 miner earns. Correct, correctly-scoped fix. Merge.

CI is green.

Scottcjn:

PR #7971, opened 2026-07-14
fix(hall-of-rust): recompute rust_score on re-attestation in the live node blueprint

Vyacheslav-Tomashevskiy:
### Follow-up to #7967 (live file)

On #7967 you noted the re-attestation score bug was fixed only in `explorer/hall_of_rust.py`, an unwired duplicate, while the copy the node actually registers — `node/hall_of_rust.py` — still had it on `/hall/induct`. This PR fixes the live one.

### Problem

`calculate_rust_score()` awards `RUST_WEIGHTS['attestation_count']` (0.1) per attestation, but the re-attestation branch only did:

```sql
UPDATE hall_of_rust SET total_attestations = total_attestations + 1, last_attestation = ? ...
```

`rust_score` was written exactly once, at induction. Every attestation after that incremented the counter the score is derived from without ever refreshing the score, so the loyalty weight sat frozen at its induction-day value for the machine's whole life. `/hall/leaderboard` sorts `ORDER BY rust_score DESC`, so a machine that has attested for months ranks identically to the day it was inducted.

### Fix

Recompute from the stored row on re-attestation and write it back, mirroring the new-induction path (which already builds a `machine` dict and calls `calculate_rust_score`). The re-attestation response now also returns `rust_score`, matching the induction response.

The `SELECT` is widened to pull the fields the scorer needs (`manufacture_year`, `device_arch`, `device_model`, `thermal_events`) instead of just `id, total_attestations`.

### Test — `node/test_hall_of_rust_reattestation_score.py`

Fails on current main, passes with the fix:

```
FAILED node/test_hall_of_rust_reattestation_score.py::test_reattestation_refreshes_rust_score
FAILED node/test_hall_of_rust_reattestation_score.py::test_reattestation_response_reports_current_score
2 failed, 1 passed
```

With the fix: `3 passed`. The third case (`test_induction_score_unchanged`) is a no-regression guard on the first-induction score — it passes on main *and* with the fix.

### Scope

Branched off clean current main, touches only `node/hall_of_rust.py` and the new test. Independent of #7969 (arch-bonus ordering in `calculate_rust_score`) — different function, no overlap; either can land first.

Re your consolidation note: there are three divergent copies of this scorer, and this makes two of them agree. Happy to do the consolidation as a separate PR if you want it.

### DIFF
--- a/node/hall_of_rust.py
+++ b/node/hall_of_rust.py
@@ -193,8 +193,11 @@ def induct_machine():
         c = conn.cursor()
         
         # Check if already inducted
-        c.execute("SELECT id, total_attestations FROM hall_of_rust WHERE fingerprint_hash = ?", 
-                  (fingerprint_hash,))
+        c.execute("""
+            SELECT id, total_attestations, manufacture_year, device_arch, device_model,
+                   thermal_events
+            FROM hall_of_rust WHERE fingerprint_hash = ?
+        """, (fingerprint_hash,))
         existing = c.fetchone()
         
         now = int(time.time())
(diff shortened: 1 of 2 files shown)


Scottcjn:

PR #7974, opened 2026-07-14
fix(airdrop): same Base wallet can claim twice via EIP-55 checksum casing (anti-Sybil bypass)

Vyacheslav-Tomashevskiy:
### Problem

`_has_claimed` normalizes the GitHub username but compares the wallet byte-exactly:

```python
github_username = self._normalize_github_username(github_username)   # .strip().casefold()
...
AND (github_username = ? OR wallet_address = ?)
```

EVM addresses are case-insensitive — the mixed-case form is only an EIP-55 display checksum over the same 20 bytes. So `0x5683C105…9c6` and `0x5683c105…9c6` are **the same Base wallet**, but they read as two different ones. `UNIQUE(github_username, wallet_address, chain)` doesn't catch it either, since the stored strings genuinely differ.

Net effect: RIP-305's "One claim per wallet address" is bypassable on Base. N GitHub accounts can funnel N × 200 wRTC into a single wallet by varying only the casing. Reproduced: `claimed_uwrtc` for `base` reaches 400 wRTC for one wallet.

The username *is* normalized on the very next line, and the suite already pins both symmetric halves of this rule — `test_duplicate_github_with_different_case_rejected` (case folding matters) and `test_duplicate_wallet_with_different_github_rejected` (wallet dedup matters). Only the wallet × case combination was never covered.

### Fix

A chain-aware `_normalize_wallet_address()`, applied at the three entry points (`check_eligibility`, `claim_airdrop`, `_has_claimed`):

```python
address = (wallet_address or "").strip()
if chain in EVM_CHAINS and EVM_ADDRESS_RE.fullmatch(address):
    return address.lower()
return address
```

**Solana is deliberately excluded**: base58 addresses are genuinely case-sensitive, and folding them would merge distinct wallets — a worse bug than the one being fixed. The regex fullmatch also means a non-EVM-shaped string on `base` is left untouched rather than silently mangled.

Only the comparison is canonicalized; this is not a migration. Any pre-existing mixed-case rows stay as they are — happy to add a backfill if you want the historical rows collapsed too.

### Tests — `node/tests/test_airdrop_evm_wallet_case_double_claim.py`

On current main:

```
FAILED test_same_base_wallet_different_checksum_case_rejected
FAILED test_same_base_wallet_uppercase_variant_rejected
FAILED test_has_claimed_matches_base_wallet_in_any_case
E   assert not True   # second claim on the same wallet succeeded
3 failed, 1 passed
```

With the fix: `4 passed`.

The one that passes on main is `test_solana_addresses_remain_case_sensitive`, the guard for the exclusion above — it must pass on both sides, and does.

Regression: `node/test_airdrop_v2.py`, `tests/test_airdrop_bridge_admin_auth.py`, `tests/test_airdrop_frontend_security.py` → **73 passed**.

### Separate issue, not fixed here

While in this file: `_determine_tier()` derives `total_prs` from `GET /search/commits?q=author:X merged:true` — that's a **commit** count across all of GitHub, not merged PRs, and not scoped to the org RIP-305 specifies (GitHub returns 200 and treats `merged:true` as free text; `author:torvalds` → `total_count: 74054`). So ~any user with 5 commits in their own repos maps to `CORE` = 200 wRTC. Left alone because the fix needs a product decision on repo/org scoping — flagging it rather than bundling it under this title. Happy to take it if you say how tiers should be scoped.

### DIFF
--- a/node/airdrop_v2.py
+++ b/node/airdrop_v2.py
@@ -66,6 +66,11 @@ MIN_ETH_BALANCE_WEI = int(0.01 * 1e18)  # 0.01 ETH
 MIN_WALLET_AGE_DAYS = 7
 MIN_GITHUB_AGE_DAYS = 30
 GITHUB_USERNAME_RE = re.compile(r"^[a-z0-9](?:[a-z0-9-]{0,37}[a-z0-9])?$")
+# EVM addresses are case-insensitive: mixed case is only an EIP-55 display
+# checksum over the same 20 bytes. Solana addresses are base58 and ARE
+# case-sensitive, so they must never be folded.
+EVM_ADDRESS_RE = re.compile(r"0x[0-9a-fA-F]{40}")
+EVM_CHAINS = frozenset({"base"})
 MAX_BRIDGE_ADDRESS_LENGTH = 128
 MAX_BRIDGE_TX_LENGTH = 256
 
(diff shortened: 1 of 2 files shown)


Scottcjn:
Genuine anti-Sybil fix, but **held pending completion:** it only closes the hole on greenfield DBs. Any Base claim already stored in mixed/EIP-55 case stays exploitable (same wallet, new GitHub, different casing passes), and the fix silently changes stored addresses to lowercase. Add a backfill (`UPDATE ... SET wallet_address = lower(...) WHERE chain='base'`) or case-insensitive compare for Base, plus a regression test that seeds a checksummed row then claims lowercase and asserts reject. Then it pays. — Sophia

Scottcjn:

PR #7977, opened 2026-07-15
fix(bridge-federation): /bridge/events window keyed on creation hides completions and voids of older transfers

Vyacheslav-Tomashevskiy:
### Problem

`_recent_events` backs the public `/bridge/events` feed, documented as "recent bridge state-change **events**". It keys the window on `created_at`:

```python
cutoff = int(time.time()) - max(0, window_seconds)
...
WHERE created_at >= ?
ORDER BY created_at DESC, id DESC
```

But `created_at` is only the moment the transfer was *opened*. A confirm/complete stamps `updated_at` + `completed_at` (`bridge_api.py:809-815`), a void stamps `updated_at` (`bridge_api.py:672-677`) — days later, potentially. So a long-lived lock that completes today is an event today, and the feed does not show it.

The sibling function in the same file already gets this right, deliberately:

```python
# last_event_at: most recent state-change timestamp across all transfers.
# Per-row max of updated_at / completed_at / created_at so a later status
# change (confirm, complete, void) is not masked by an older value.
```

So the two public surfaces contradict each other. Reproduced against a DB built from the real `bridge_transfers` schema — one transfer, created 25h ago, completed 60s ago:

```
/bridge/state  -> last_event_at is 60s ago   (an event JUST happened)
/bridge/events -> events: [], count = 0      (default 24h window)
```

This is the case reconciliation actually cares about: a federation reconciler polling `/bridge/events?window_seconds=3600` sees newly-opened transfers only, never the completions and voids of older ones — the very state changes it needs to compare against the other side (§6.4).

`last_event_at` in this same module defines an event as the max of the three timestamps, with a comment explaining why, and four tests pin that (`test_state_last_event_at_uses_updated_when_newer_than_created`, `..._uses_completed_when_newer_than_updated`, `..._picks_row_with_latest_state_change`). The event *feed* just never got the same treatment — no test seeds a transfer whose `updated_at` differs from its `created_at`, so the gap was invisible.

### Fix

One shared definition of event time, replacing the aggregate's inline copy:

```python
EVENT_AT_SQL = "max(updated_at, COALESCE(completed_at, 0), created_at)"
```

`_recent_events` now filters, orders and reports on it. `event_at` is exposed **alongside** `created_at` (nothing removed), so a consumer can distinguish when a transfer was opened from when it last moved. Untouched rows have `event_at == created_at`, so their behaviour is unchanged.

Scoped to the event feed: `/bridge/transfers/recent` is a transfer listing, not an event feed, so its `created_at` ordering is left alone. No writes, no schema change.

### On the one fixture line I touched

`_seed` defaulted `updated_at` to `now - i` **independently of** `created_at`, so a row seeded as "created 10,000s ago" silently claimed it had changed state *this second*. `bridge_api` stamps `created_at` and `updated_at` with the same `now` on insert (`bridge_api.py:449-451`), so that combination cannot occur for an untouched transfer. The default is now `row.get("updated_at", created_at)`.

Worth being explicit, since changing a fixture to make a fix pass deserves scrutiny: **it doesn't here.** No existing assertion changed, and all 22 pre-existing tests pass against unpatched `main` with the new fixture default — so the fixture change is behaviour-neutral on its own. It was masking, not asserting: those tests passed only because the endpoint ignored `updated_at`.

### Tests

Four added, all failing on `main`, passing here:

- `test_events_includes_transfer_created_before_window_but_completed_inside_it` — the headline case; also asserts `/bridge/events` and `/bridge/state` now agree on when it happened.
- `test_events_excludes_transfer_whose_state_change_is_outside_the_window` — the window still clamps; an old row that never moved stays out.
- `test_events_ordered_by_state_change_not_creation`
- `test_events_event_at_equals_created_at_for_untouched_transfer`

`node/tests/test_bridge_federation_routes.py` 26 passed. Neighbouring `test_bridge_api.py` + `test_bridge_reconciliation.py` green except two pre-existing `TestValidateChainAddressFormat` failures that reproduce identically on clean `main` (address-format validation, unrelated to this file).

### Severity, honestly

Read-only reporting bug on public endpoints — no funds move, no auth surface, nothing mis-settles. Blast radius is the federation reconciliation view on the live node (`register_federation_routes`, wired at `rustchain_v2_integrated_v2.2.1_rip200.py:162`). It matters because it's a reconciliation surface whose whole job is showing state changes, and it was silently disagreeing with `/bridge/state`.

### DIFF
--- a/node/bridge_federation_routes.py
+++ b/node/bridge_federation_routes.py
@@ -45,6 +45,12 @@ MAX_TRANSFERS_LIMIT = 200
 DEFAULT_EVENTS_WINDOW_SECONDS = 24 * 3600
 MAX_EVENTS_WINDOW_SECONDS = 30 * 24 * 3600
 
+# When a transfer last changed state. `created_at` alone is only the moment the
+# transfer was opened; a confirm/complete/void stamps `updated_at` (and
+# `completed_at`) later, so the latest of the three is the event time. One
+# definition, shared by the aggregate and the event feed.
+EVENT_AT_SQL = "max(updated_at, COALESCE(completed_at, 0), created_at)"
+
 
 def _get_db_path() -> str:
     """Resolve DB_PATH the same way bridge_api.py does."""
@@ -112,9 +118,7 @@ def _aggregate_bridge_state(conn: sqlite3.Connection) -> Dict[str, Any]:
     # Per-row max of updated_at / completed_at / created_at so a later status
     # change (confirm, complete, void) is not masked by an older value.
     last_event = cursor.execute(
-        "SELECT COALESCE(MAX("
-        "max(updated_at, COALESCE(completed_at, 0), created_at)"
-        "), 0) FROM bridge_transfers"
+        f"SELECT COALESCE(MAX({EVENT_AT_SQL}), 0) FROM bridge_transfers"
     ).fetchone()  # fetchall-ok: pragma-result (single MAX)
     last_event_at = int(last_event[0]) if last_event else 0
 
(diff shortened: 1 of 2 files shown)


Vyacheslav-Tomashevskiy:
/claim

Scottcjn:

PR #7979, opened 2026-07-15
fix(windows-miner): KeyError on attestation[nonce] posts every attestation unsigned, blocking enrollment

Vyacheslav-Tomashevskiy:
### Problem

`attest()` builds the attestation with the nonce under `report` — there is no top-level copy:

```python
attestation = {
    "miner":    self.wallet_address,
    "miner_id": self.miner_id,
    "report":   report_payload,      # {"nonce": nonce, "commitment": ..., ...}
    "device":   {...},
    "signals":  {...},
}
```

The signing fallback then reads a key that does not exist:

```python
sign_msg = "{}|{}|{}|{}".format(
    attestation["miner_id"],
    attestation["miner"],
    attestation["nonce"],            # KeyError: 'nonce'
    attestation["report"]["commitment"],
).encode("utf-8")
```

The `except` below it swallows the `KeyError` and falls through unsigned:

```
WARNING rustchain_windows_miner.py:595 attestation signing failed; falling through unsigned: 'nonce'
```

Because the raise happens *inside* `if CRYPTO_AVAILABLE and self.keypair:`, the `else:` legacy-sha512 fallback never runs either — so the POST carries **no `signature`, no `public_key`, no `signature_type`** at all.

### Why this is worse than a missing signature

The unsigned attest is accepted, but `record_attestation_success(..., signing_pubkey=pubkey_hex or None, ...)` then stores nothing. Enrollment, meanwhile, signs fine (it never touches `attestation["nonce"]`) and sends `signature` + `public_key` — but the server looks up the *stored* pubkey, finds none, and with `ENROLL_ALLOW_UNSIGNED_LEGACY` defaulting to `"0"` fails closed:

> **412 ENROLLMENT_SIGNING_KEY_REQUIRED** — "Re-attest with signature/public_key before enrolling"

which is precisely what this miner structurally cannot do. The miner requires `status_code == 200`, returns False, and never enrolls. A Windows miner in this state mines nothing, forever.

### Why only Windows, and why CI didn't catch it

The Linux miner sets a **top-level** `"nonce": nonce` alongside `report["nonce"]` (`miners/linux/rustchain_linux_miner.py:618`), and macOS v2.5 sets both. That asymmetry is also why `build_pipe_sign_message` reads `report.get("nonce") or attestation["nonce"]` — it tolerates either spelling, so the helper branch masks the bug. Only the Windows fallback hard-codes the spelling Windows doesn't use.

And the fallback branch is the one a **real install** takes. `signing_helpers.py` lives at `miners/signing_helpers.py`; `rustchain_miner_setup.bat:7,10` downloads only `rustchain_windows_miner.py` and `miner_crypto.py` into a flat directory, and the PyInstaller spec ships `hiddenimports=[]`. So both imports fail, `_SIGNING_HELPERS = False`, and the broken branch runs. From a dev checkout `from miners.signing_helpers import ...` resolves via implicit namespace packages, so it looks fine locally.

The existing `test_windows_miner_signs_report_nonce` does not catch it because it sets `miner_mod._SIGNING_HELPERS = True` — it pins the masked path. `requirements-miner.txt` pins `PyNaCl>=1.5.0` and the `.bat` installs it, so `CRYPTO_AVAILABLE and self.keypair` is true on a default install: everyone lands in the broken branch.

### Fix

One line — read the nonce where this attestation actually keeps it:

```python
attestation["report"]["nonce"]
```

This makes the fallback produce **byte-identical** output to `build_pipe_sign_message` for the same attestation (both emit `miner_id|miner|nonce|commitment`), so the two branches can no longer drift. I did not add a top-level `nonce` to match Linux: that would change the posted payload shape, and this keeps the diff to the defect. Happy to align the miners on one shape instead if you'd rather fix it at the source.

### Bootstrap hash

`rustchain_miner_setup.bat` pins `MINER_SHA256` of this exact file, verified both by `check_bootstrap_hashes.py` in CI and by the installer itself at line 91 — an un-updated pin would make real installs *refuse* the fixed file. Re-pinned to `1c25b32e…5d42`; `check_bootstrap_hashes.py` reports "Windows bootstrap hashes are in sync." The `.bat` diff is that one line, CRLF preserved.

### Tests

- `test_windows_miner_signs_when_signing_helpers_unavailable` — the real-install branch (`_SIGNING_HELPERS = False`), asserting `signature` / `public_key` / `signature_type` are present and that the bytes match the node's reconstruction. **Fails on main** with the `'nonce'` warning above.
- `test_windows_fallback_message_matches_shared_helper` — pins the two branches to identical bytes on a Windows-shaped attestation.

`tests/test_attestation_signing_6798.py`: 12 passed (was 11 + 1 failing).

### Severity, honestly

High impact but narrow scope: **Windows miners only**, and only against a node running the default `ENROLL_ALLOW_UNSIGNED_LEGACY=0`. Linux/macOS/POWER8 are unaffected. The GUI installer copy under `miners/windows/installer/src/` is a separate, older copy without the signing refactor, so it is unaffected too — the blast radius is the `.bat` / `build_windows_miner.ps1` channel, which is live and CI-guarded. I have not tested against a running node; the enrollment chain above is read from the node source, so please sanity-check that half.


Vyacheslav-Tomashevskiy:
/claim

Scottcjn:
Thirty-three of your PRs went in today, which moved `main` under this one and left it with conflicts. Nothing is wrong with the change itself.

Could you rebase onto current `main` and push? I will pick it straight back up. Everything I merged today passed the same check: your added tests pass on the branch and fail when the source change is reverted, so if the rebase is clean this should follow the same path.

Sorry for the churn. It is a consequence of the backlog being cleared in one go rather than anything you did.

Vyacheslav-Tomashevskiy:

Scottcjn:
Closing this one as superseded — but your diagnosis was correct, and I want it on the record that it was.

### You were right about the bug

The nonce **does** live under `report`, not at the top level. Current `main` still builds it exactly that way:

```python
report_payload = {
    "nonce": nonce,
    "commitment": hashlib.sha256(...).hexdigest(),
    ...
}
attestation = {"miner": ..., "miner_id": ..., "report": report_payload, ...}
```

There is still no top-level `nonce` key. So the `attestation["nonce"]` lookup you found really would have raised `KeyError`, really would have dropped the miner into the unsigned path, and really would have blocked enrollment — and your note about *why* that branch is the one a real Windows install takes (the `.bat` downloads only `rustchain_windows_miner.py` + `miner_crypto.py` into a flat directory, so `signing_helpers` cannot import and `_SIGNING_HELPERS` is `False`) is a sharper piece of reasoning than the bug report itself. That is the kind of detail that separates "I read the code" from "I thought about how this is actually deployed."

### Why it is closing anyway

The code you patched no longer exists. `main` replaced the whole hand-rolled `"{}|{}|{}|{}"` signing branch with canonical-JSON signing:

```python
signature = sign_payload(canonical_json(attestation), self.keypair["private_key"])
attestation["signature_type"] = "canonical_json"
```

`_SIGNING_HELPERS`, `build_pipe_sign_message`, and the `attestation["nonce"]` lookup are all gone from the file. The KeyError cannot occur, because nothing reads that key any more. Your regression test targets the `_SIGNING_HELPERS is False` branch, which no longer exists either — so it cannot be salvaged onto `main` the way the lock-ledger test on #7950 can.

### One thing worth knowing for next time

This PR also re-pins the installer hash:

```
set "MINER_SHA256=1c25b32e141346c32725b07c48286f9f887fb579fb516bc25ef253f705cb5d42"
```

I verified that hash is genuinely correct **for your branch's copy** of the miner — I fetched the file at your head SHA and hashed it, and it matches exactly. You did that part right. But `main` has since moved to `99ac84a489ebc8c1987eddc02dfbaf8672a9d440cc2cc9e1166c6ba25f4e8184`, and `rustchain_miner_setup.bat` on `main` already pins that value. Merging this now would have pinned the installer to a hash for an older miner, and every Windows install would have failed hash verification at download time.

That is not a mistake on your part — it is exactly what happens when a PR that pins a content hash sits in a queue while the content changes underneath it. It is worth flagging as a general hazard: **hash-pinning PRs go stale in a way ordinary code PRs do not**, because they can silently become wrong without ever conflicting. Any future PR of yours that touches `MINER_SHA256` or `CRYPTO_SHA256` should be re-verified against `main` immediately before merge, by us, not by you.

### On the push blocker

You mentioned the `workflow`-scope PAT problem on this and the others. You do not need to solve it for this PR — it is closing. For the ones still open, a fresh branch cut from current `main` containing only the files you actually mean to change will sidestep it entirely, since it will carry no `.github/workflows/` commits at all. That is cheaper than getting a new token.

Filed 2026-07-15, closed 2026-08-20. Five weeks is too long for a correct bug report to sit, and the reason it is closing is that we were slow, not that you were wrong. Thank you for it.

PR #7986, opened 2026-07-15
fix(adm): settled=1 committed while paying zero rewards burns the epoch permanently

Vyacheslav-Tomashevskiy:
## The bug

`settle_epoch_with_anti_double_mining` claims settlement **before** it knows whether there is anything to pay:

```python
res = db.execute("UPDATE epoch_state SET settled = 1, settled_ts = ? WHERE epoch = ? AND settled = 0", ...)
...
if not rewards:
    if own_conn:
        db.rollback()
    return {"ok": False, "error": "no_eligible_miners", "epoch": epoch}
```

The undo is gated on `own_conn` — but `own_conn` is `False` exactly when `existing_conn` is passed, **which is the only mode the live node uses**. `rewards_implementation_rip200.settle_epoch_rip200` passes `existing_conn=db` and then commits regardless of the result:

```python
result = settle_epoch_with_anti_double_mining(..., existing_conn=db)
db.commit()          # runs whether or not result["ok"]
return result
```

So the function reports failure while durably committing `settled = 1`. The epoch's entire `PER_EPOCH_URTC` is never distributed, and it can never be re-settled — the operator's retry is answered `already_settled`.

This contradicts the standard (non-ADM) path in the same caller, which does an ungated `db.rollback()` before returning the very same `no_eligible_miners` error, leaving the epoch retryable as the module intends.

## Repro (run, not hypothesised)

`epoch_state(epoch=1, settled=0)` (the live node pre-creates this row), one `epoch_enroll` row, and `miner_attest_recent.fingerprint_passed = 0` for that miner — a machine that failed attestation, so total weight is 0 and the reward map comes back empty:

```
settle_epoch_rip200(db, epoch=1, enable_anti_double_mining=True)
  -> {'ok': False, 'error': 'no_eligible_miners', 'epoch': 1}

epoch_state:   [(1, 1)]     # settled = 1, COMMITTED
epoch_rewards: []           # nothing paid
balances:      []           # nothing paid

retry: {'ok': True, 'epoch': 1, 'already_settled': True}   # epoch burned
```

Other ways into the same empty-rewards state: all enrolled weights 0 (RIP-309 filtering), or representatives with no `miner_attest_recent` row.

## The fix

The claim-first `UPDATE` is deliberate — it's the atomic check-and-set that closes the concurrent-settler race — so I did **not** move it after the reward calculation. Instead the failure path undoes *only our own claim* (tracked via `res.rowcount`) when we're sharing the caller's connection; calling `db.rollback()` there would discard transaction state we don't own.

## Scope, honestly

- **This is on a live path**, but it is the admin/operator `settle_epoch_rip200` entry, not `finalize_epoch` (which is intentionally not ADM-gated).
- It needs an epoch where every enrolled miner is ineligible. That's an operational edge, not an everyday occurrence — but it's exactly the case the `no_eligible_miners` return exists to handle, and when it hits, the loss (a burned epoch) is permanent and silent.

## Tests

`node/tests/test_adm_no_eligible_miners_burns_epoch.py` — 2 tests, both **fail on main**, pass with the fix:
- `settled=1` committed while paying zero rewards
- retry refused with `already_settled`

They exercise the real ADM function through the real caller on the shared-connection path (no stubbing of the function under test — the existing `test_adm_require_optin_t33.py` either disables ADM or stubs it to raise, so it always lands on the standard path's correct rollback and never reaches this branch).

Neighbouring suites green: `test_adm_require_optin_t33.py`, `test_rewards_settle_race.py`, `test_epoch_settlement_atomic.py`, `test_rewards_settle_endpoint.py`, `test_epoch_reward_settlement_parameter.py` (22 passed), plus `test_anti_double_mining.py` + `test_anti_double_mining_epoch_enroll.py` (25 passed).

Verified against `upstream/main` @ 96660f06.

### DIFF
--- a/node/anti_double_mining.py
+++ b/node/anti_double_mining.py
@@ -674,6 +674,7 @@ def settle_epoch_with_anti_double_mining(
         # We use an UPDATE that only succeeds if settled is currently 0.
         # This prevents the race condition.
         res = db.execute("UPDATE epoch_state SET settled = 1, settled_ts = ? WHERE epoch = ? AND settled = 0", (int(time.time()), epoch))
+        claimed = res.rowcount > 0
         if res.rowcount == 0:
             # If no row was updated, it was either already settled or doesn't exist
             st = db.execute("SELECT settled FROM epoch_state WHERE epoch=?", (epoch,)).fetchone()
@@ -698,6 +699,19 @@ def settle_epoch_with_anti_double_mining(
         if not rewards:
             if own_conn:
                 db.rollback()
+            elif claimed:
+                # Shared connection: the caller owns the transaction and commits
+                # on return (see settle_epoch_rip200), so `own_conn`-gated rollback
+                # never runs and the claim above would be committed despite paying
+                # nothing — burning the epoch's emission and making every retry
+                # answer already_settled. Rolling back is not ours to do here, so
+                # undo only our own claim and leave the epoch retryable. This
+                # matches the standard path, which rolls back before returning the
+                # same error.
+                db.execute(
+                    "UPDATE epoch_state SET settled = 0, settled_ts = NULL WHERE epoch = ?",
+                    (epoch,)
+                )
             return {"ok": False, "error": "no_eligible_miners", "epoch": epoch}
 
         # Credit rewards to miners
--- a/node/tests/test_adm_no_eligible_miners_burns_epoch.py
+++ b/node/tests/test_adm_no_eligible_miners_burns_epoch.py
@@ -1,0 +1,109 @@
+"""Regression: ADM must not leave `settled=1` committed when it pays nothing.
+
+`settle_epoch_with_anti_double_mining` claims settlement (`UPDATE epoch_state SET
+settled = 1`) BEFORE it knows whether any miner is eligible. Its `no_eligible_miners`
+exit undoes that claim only `if own_conn` — but the live caller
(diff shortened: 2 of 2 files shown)


elevasyncsolutions-jpg:
## Review of PR #7986 — ADM Settled Flag Committed Before Reward Check

### Observation 1: `claimed` variable is set but the rollback logic is incomplete (`anti_double_mining.py`)

The fix adds `claimed = res.rowcount > 0` but the subsequent `no_eligible_miners` exit only rolls back `if own_conn`. The comment explains that the live caller always passes `existing_conn=db` and commits unconditionally. However, if the code is ever called with `own_conn=True` in the future, the rollback will work. But if called with `own_conn=False` and no rewards exist, the `settled=1` is still committed. Consider always rolling back the `settled` flag when there are no eligible miners, regardless of `own_conn`.

### Observation 2: The test covers the no-eligible-miners path but not the edge case where `db.execute` raises after the UPDATE

If the UPDATE succeeds but a subsequent `db.execute` (e.g. reading rewards) raises an exception, the `settled=1` is committed but no rewards are written. The existing `try/except` blocks should handle this, but a test covering the mid-transaction failure path would be more robust.

### Positive notes

- The test `test_adm_no_eligible_miners_burns_epoch` is excellent — it directly verifies that `settled` stays 0 when no miners are eligible.
- The root cause (committing settlement before knowing if rewards exist) is clearly explained.
- The fix is minimal and targeted.

PR #7988, opened 2026-07-15
fix(fleet): thermal_signature always NULL — consumer reads keys no producer emits

Vyacheslav-Tomashevskiy:
## The bug

`record_fleet_signals_from_request` reads two keys that **no producer in this repo emits**:

```python
thermal = checks.get("thermal_drift", {}).get("data", {})
thermal_sig = thermal.get("entropy", thermal.get("drift_magnitude"))
```

`check_thermal_drift` — `node/fingerprint_checks.py:202-208`, and identically in `miners/linux/`, `miners/macos/`, `miners/windows/` and `miners/power8/` — emits exactly:

```python
{"cold_avg_ns", "hot_avg_ns", "cold_stdev", "hot_stdev", "drift_ratio"}
```

So `thermal_sig` is `None` for every real attestation, `fleet_signals.thermal_signature` is written **NULL fleet-wide**, and the thermal branch of `_detect_fingerprint_similarity`:

```python
if sig_a.get("thermal_signature") and sig_b.get("thermal_signature"):
```

can never be true. The detector silently runs on 3 of its 4 signals while still requiring `shared_hashes >= 2`.

`grep -rn drift_magnitude` shows that key exists only as `clock_drift_magnitude` in `node/arch_cross_validation.py` — an unrelated string enum it was likely copy-pasted from.

## Precedent: this exact skew was already fixed next door

`node/hardware_binding_v2.py:88-89`:

```python
# Accept legacy 'ratio' OR v3 'drift_ratio'.
'thermal_ratio': thermal_data.get('ratio') or thermal_data.get('drift_ratio') or 0,
```

`fleet_immune_system` was never updated with it. This PR mirrors that tolerance — legacy keys still win, `drift_ratio` is accepted.

`drift_ratio` (hot/cold, 4dp) is exactly the stable per-model signal the ±10% comparison wants. The surviving signals don't compensate: `cache_latency_hash` is a SHA over raw float latencies, so it never collides across machines either; `simd_bias_hash` matches, but that's 1 of the required 2, leaving detection to hinge entirely on `clock_drift_cv` landing within 5% — which `hardware_binding_v2.py:116-118` itself documents as "varies 100%+ between runs".

(Checked the neighbours: `clock_drift` really does emit `cv`, and the SIMD/cache hashes are computed over whatever the check returns. Thermal is the only signal broken this way.)

## Repro

6 identical machines on one /24, payload built from the real `check_thermal_drift` output shape:

| | thermal_signature stored | fleet detection |
|---|---|---|
| main | `NULL` × 6 | thermal branch unreachable, 3 of 4 signals |
| with fix | `1.0500 … 1.0510` | thermal branch compares as designed |

## Scope — honestly

I want to be precise about the blast radius rather than oversell this:

- **The write side is live.** `submit_attestation` → `record_attestation_success` → `record_fleet_signals` → `record_fleet_signals_from_request`. Every real attestation persists a NULL thermal column today. That's a poisoned audit column, right now.
- **The read side is not wired into rewards.** `_detect_fingerprint_similarity` is reached only via `compute_fleet_scores` ← `calculate_immune_weights` / `calculate_immune_rewards_equal_split`. The node imports `calculate_immune_weights` but I found **no call site for it** in the monolith; `register_fleet_endpoints` only exposes `/admin/fleet/*` readers over `fleet_scores`. So this is **not** an active reward exploit — it's a correctness bug in a detector that is degraded before it is ever switched on.
- Secondary note: the flat-name `from fleet_immune_system import ...` resolves only if the module is deployed alongside the node (it lives in `rips/python/rustchain/`, and the `sys.path` insert in the monolith is inside a function, i.e. too late), so on a plain checkout `HAVE_FLEET_IMMUNE` is False. If that's intentional, this fix still matters for whenever the module is deployed; if it isn't, that's a separate packaging issue and I'm happy to look at it.

## Why the suite never caught it

Every fleet test fabricates the consumer's imaginary schema instead of the producer's:

- `tests/test_false_positive_scenarios.py:45` — `"thermal_drift": {"data": {"entropy": entropy}}`
- `tests/test_fleet_score_manipulation.py:63`, `tests/test_rip201_fleet_bypass.py:180` — same `entropy` key
- `tools/rip201_false_positive_report.py:46`, `tools/rip_poa_fingerprint_replay_poc.py:116` — same

No test feeds real `check_thermal_drift` output through `record_fleet_signals_from_request`, so the green suite proved nothing here. The new tests build the payload from the producer's actual shape.

## Tests

`tests/test_fleet_thermal_signature_real_payload.py` — 4 tests, **2 fail on main**, all pass with the fix:
- real producer payload populates `thermal_signature` (NULL on main)
- a 6-machine clone fleet gets its thermal branch counted (unreachable on main)
- legacy `entropy` key still wins (tolerance preserved)
- absent thermal check still stores NULL

Existing fleet suites green: `test_false_positive_scenarios.py`, `test_fleet_score_manipulation.py`, `test_rip201_fleet_bypass.py` (16 passed).

## Related, not bundled

`node/hardware_fingerprint_replay.py:202-213` (`compute_entropy_profile_hash`) has the *same* skew — it reads `cache_data.get('L1')`, `thermal_data.get('ratio')`, `jitter_data.get('cv')`, none of which v3 emits. I left it out of this PR because the exact-hash design still catches verbatim profile copying, so the impact there is "hash is narrower than intended" rather than a dead check. Say the word and I'll send it separately.

Verified against `upstream/main` @ 96660f06.

### DIFF
--- a/rips/python/rustchain/fleet_immune_system.py
+++ b/rips/python/rustchain/fleet_immune_system.py
@@ -331,9 +331,16 @@ def record_fleet_signals_from_request(
             cache_str = str(sorted(cache.items()))
             cache_hash = hashlib.sha256(cache_str.encode()).hexdigest()[:16]
 
-        # Thermal drift entropy
+        # Thermal drift entropy. check_thermal_drift emits 'drift_ratio' (and has
+        # in every miner build); 'entropy'/'drift_magnitude' are read first only
+        # for legacy payloads. Without drift_ratio this was None for every real
+        # attestation, so thermal_signature stored NULL fleet-wide and the
+        # thermal comparison in _detect_fingerprint_similarity could never run.
+        # Mirrors the tolerance already used by hardware_binding_v2.
         thermal = checks.get("thermal_drift", {}).get("data", {})
-        thermal_sig = thermal.get("entropy", thermal.get("drift_magnitude"))
+        thermal_sig = thermal.get(
+            "entropy", thermal.get("drift_magnitude", thermal.get("drift_ratio"))
+        )
 
         # SIMD bias profile hash
         simd = checks.get("simd_identity", {}).get("data", {})
--- a/tests/test_fleet_thermal_signature_real_payload.py
+++ b/tests/test_fleet_thermal_signature_real_payload.py
@@ -1,0 +1,126 @@
+# SPDX-License-Identifier: MIT
+"""Regression: fleet thermal_signature must be populated from the real payload.
+
+`record_fleet_signals_from_request` read `thermal_drift.data.entropy` /
+`drift_magnitude`. No producer in the repo emits either key: `check_thermal_drift`
+(node/fingerprint_checks.py, and identically in every miner build) emits
+`cold_avg_ns / hot_avg_ns / cold_stdev / hot_stdev / drift_ratio`.
+
+So `thermal_signature` was written NULL for every real attestation, and the thermal
+branch of `_detect_fingerprint_similarity` (`if sig_a.get("thermal_signature") and
+sig_b.get("thermal_signature")`) could never be true — the detector silently ran on
+3 of its 4 signals while still requiring `shared_hashes >= 2`.
+
+Every existing fleet test fabricates the consumer's imaginary `{"entropy": ...}`
+schema, so the green suite proved nothing here. These tests build the payload from
+`check_thermal_drift`'s actual output shape instead.
+"""
+import os
+import sqlite3
+import sys
+import tempfile
+import unittest
+
+REPO = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+sys.path.insert(0, os.path.join(REPO, "rips", "python", "rustchain"))
+sys.path.insert(0, os.path.join(REPO, "node"))
+
+import fleet_immune_system as fis
+
+
+def _real_thermal_payload(drift_ratio):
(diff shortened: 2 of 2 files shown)


elevasyncsolutions-jpg:
## Review of PR #7988 — Fleet Thermal Signature Populated From Real Payload

### Observation 1: The fallback chain is long but necessary (`fleet_immune_system.py`)

The code now reads `drift_ratio` first, then falls back to `entropy`, then `drift_magnitude`. This is correct for backwards compatibility, but the comment should note which version introduced `drift_ratio` so future maintainers know when the legacy fallbacks can be removed.

### Observation 2: The `_detect_fingerprint_similarity` thermal check uses `if sig_a.get("thermal_signature")` which is now truthy

Before this fix, `thermal_signature` was always NULL, so the thermal comparison never ran. Now it will run for all attestation pairs. If the thermal comparison has false positives (e.g. two different machines with similar thermal profiles), this could increase false-positive similarity detections. Consider adding a threshold or confidence score to the thermal comparison.

### Positive notes

- The root cause is clearly documented: `check_thermal_drift` emits `drift_ratio`, not `entropy`/`drift_magnitude`.
- The test builds a real payload from the actual producer schema and verifies the signature is non-null.
- The tolerance matching `hardware_binding_v2` is a good design choice.

PR #7989, opened 2026-07-15
fix(lock_ledger): bridge-owned lock release double-credits the source (mints RTC)

Vyacheslav-Tomashevskiy:
## Bug

`lock_ledger.release_lock()` credits `balances` for locks the **bridge** owns and already settles itself, reversing a debit that is still outstanding.

`release_lock()` credits unconditionally because it is the counterpart of `create_lock()`, which debits. That pairing does not hold in production:

- `create_lock()` has **no live callers** (`grep` → only its own definition/docstring).
- Every real `lock_ledger` row is written by `bridge_api.create_bridge_transfer()` with a **raw INSERT** (`node/bridge_api.py:463`), after it hard-debits the source at `:406` and records `bridge_transfers.source_debited = 1` at `:454`.
- `release_lock()` never reads or clears `source_debited`.

Both modules register under the same `HAVE_BRIDGE` flag (`node/rustchain_v2_integrated_v2.2.1_rip200.py:1462-1465`), so the lock producer and these release routes are always live together.

## Impact — reproduced against the real schema

| path | scenario | correct | on main |
|---|---|---|---|
| `POST /api/lock/release` | deposit opened (alice 100 → 0, `source_debited=1`), admin releases, deposit then completes on Solana | `0` | **`100 RTC`** — alice keeps her RTC on *both* chains |
| `POST /api/lock/release` → bridge void | release, then the bridge's own void refunds again | `100 RTC` | **`200 RTC`** — double refund |
| `POST /api/lock/auto-release` | **no operator action**: the 7-day lock expires while the deposit is still `pending` | `0` | **`100 RTC`**, `bridge_transfers` still `('pending', 1)` |

`released_by="admin"` bypasses the unlock timer by design (`:262`), so the first two are immediate. The third needs nobody: `get_pending_unlocks()` (`:520`) filters on `status='locked' AND unlock_at <= now` with **no `lock_type` filter**, so the documented auto-release worker picks up bridge deposits and silently un-does the hard debit under a live deposit.

The restored balance is also immediately spendable: `available_balance.encumbered_i64()` only counts deposits with `source_debited = 0 OR IS NULL` (`node/available_balance.py:69`), and `release_lock()` leaves `source_debited = 1` — so the un-debited funds are invisible to every encumbrance check.

The bridge's own settlement never expects that credit: void (`bridge_api.py:696`) refunds guarded on `source_debited`, and completion (`:836`) deliberately does *not* refund — it only flips the lock row to `released`.

## Fix

- `release_lock()` fails closed for any lock carrying a `bridge_transfer_id`, pointing the caller at the bridge (complete or void).
- The auto-release worker **skips** bridge-owned locks and reports `skipped_bridge_owned`, rather than calling through and logging an error per lock on every run — an expired bridge deposit is a normal state an operator resolves by voiding.
- Locks created via `create_lock()` are untouched and keep their debit/credit symmetry.

## Verification

`tests/test_lock_ledger_bridge_owned_no_credit.py` — **4 of 5 fail on main**, all pass with the fix:

```
# clean main
FAILED test_admin_release_does_not_credit_bridge_deposit
FAILED test_release_then_complete_does_not_mint
FAILED test_release_then_void_refunds_once      (assert 200 RTC == 100 RTC)
FAILED test_auto_release_worker_skips_bridge_owned_locks
4 failed, 1 passed
# with fix
5 passed
```

The 5th (`test_standalone_lock_release_still_credits`) passes on **both** — it guards against over-reach by pinning that `create_lock()`/`release_lock()` still balance.

Existing `tests/test_bridge_lock_ledger.py` stays green (**71 passed**). It didn't catch this because `test_release_lock` (`:669`) only exercises the symmetric `create_lock()` → `release_lock()` pair, and `create_lock()` debits, so the credit balances out — but that producer is never used in production. No test ever fed a `bridge_api`-produced lock into `release_lock()`.

I also ran the wider money-path selection (`-k "bridge or lock or balance or encumb or payout"`): the failures there are **identical on pristine main** and unrelated to this change (the only diff between main and this branch is my 4 tests flipping to pass).

---

### DIFF
--- a/node/lock_ledger.py
+++ b/node/lock_ledger.py
@@ -242,7 +242,8 @@ def release_lock(
 
     # Find the lock
     row = cursor.execute("""
-        SELECT id, miner_id, amount_i64, lock_type, status, unlock_at
+        SELECT id, miner_id, amount_i64, lock_type, status, unlock_at,
+               bridge_transfer_id
         FROM lock_ledger
         WHERE id = ?
     """, (lock_id,)).fetchone()
@@ -250,12 +251,30 @@ def release_lock(
     if not row:
         return False, {"error": "Lock not found"}
 
-    lid, miner_id, amount_i64, lock_type, status, unlock_at = row
+    lid, miner_id, amount_i64, lock_type, status, unlock_at, bridge_transfer_id = row
 
     if status != "locked":
         return False, {
             "error": f"Lock already {status}",
             "hint": "Only locked entries can be released"
+        }
+
+    # Bridge-owned locks settle through bridge_api, not here. This function
+    # credits `balances` because it is the counterpart of create_lock(), which
+    # debits. bridge_api.create_bridge_transfer() does its own hard debit and
+    # writes the lock row directly, tracking the refund on
+    # bridge_transfers.source_debited — a flag this function neither reads nor
+    # clears. Crediting such a lock therefore reverses a debit the bridge still
+    # considers outstanding, and the bridge's own void path refunds it again.
+    if bridge_transfer_id is not None:
+        return False, {
+            "error": "Lock is owned by a bridge transfer",
+            "lock_id": lock_id,
+            "bridge_transfer_id": bridge_transfer_id,
+            "hint": (
+                "Settle via the bridge: complete the transfer, or void it to "
+                "refund the source. Releasing here would double-credit."
+            )
         }
 
     # Check if unlock time has passed (unless admin override)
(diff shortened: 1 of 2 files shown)


elevasyncsolutions-jpg:
## Review of PR #7989 — Bridge-Owned Lock Release Double-Credit Fix

### Observation 1: The `bridge_transfer_id` column check is the right discriminator (`lock_ledger.py`)

The fix queries `bridge_transfer_id` and skips the balance credit if it's non-NULL. This is correct because bridge-created locks are settled by `bridge_api` directly. However, the comment should clarify that `bridge_transfer_id` is set by `bridge_api.create_bridge_transfer()`, not by `lock_ledger.create_lock()`, so the two code paths are clearly distinguished.

### Observation 2: The test should also verify that the bridge's own settlement is unaffected

The test checks that `balances` is not credited, but it should also verify that `bridge_transfers.source_debited` remains unchanged after the lock release. This confirms that the bridge's accounting is not disrupted.

### Positive notes

- This closes a real double-spend vulnerability: bridge-owned locks were being credited back to the miner while the bridge still considers the funds debited.
- The discriminator approach (checking `bridge_transfer_id` rather than a separate flag) is clean.
- The test covers both admin release and auto-release paths.

PR #7990, opened 2026-07-15
fix(attest): warthog_bonus never revoked -> permanent +15% epoch reward weight

Vyacheslav-Tomashevskiy:
## Bug

`miner_attest_recent.warthog_bonus` only ever ratchets **up**. Once a miner earns the dual-mining tier, nothing ever revokes it.

Two things combine:

- `_submit_attestation_impl` writes the column only when the freshly computed bonus is `> 1.0` (`node/rustchain_v2_integrated_v2.2.1_rip200.py:4842`). That `UPDATE` is the **only writer of the column in the repo** (`grep "SET warthog_bonus"` → a single hit).
- `record_attestation_success()` (`:3203`) upserts `miner_attest_recent`, but `warthog_bonus` is **not** in its column list or its `ON CONFLICT DO UPDATE SET` list, so the existing value survives untouched.

So `warthog_bonus = 1.0` is never written back. A miner who verifies a proof once and then:

- stops dual-mining entirely (sends no `warthog` key at all), or
- starts failing the hardware fingerprint

keeps the 1.15x multiplier forever. The second case directly defeats the stated invariant in the code right above the bug (`:4801-4803`):

> `SECURITY: Warthog bonus requires passing hardware fingerprint. Without this gate, VMs could fake/run Warthog and farm the bonus.`

The gate correctly computes `warthog_bonus = 1.0` for these miners — the write is just skipped, so the DB keeps the old 1.15.

## Impact — this is live money

The stale column feeds epoch rewards directly:

`POST /attest/submit` → `:4842` (skipped write) → `rip_200_round_robin_1cpu1vote.calculate_epoch_rewards_time_aged` `:652` `wart_sql = ", COALESCE(warthog_bonus, 1.0) "` → `:683` → `_apply_warthog_bonus(weight_units, warthog_bonus)` (`:96`) → `rewards_implementation_rip200.settle_epoch_rip200` → credits `balances`.

Two enrolled miners, weight 1.0 each, epoch pot 2,150,000 uRTC:

| miner | correct | on main |
|---|---|---|
| A (stale 1.15, no longer dual-mining) | 1,075,000 | **1,150,000** |
| B (honest) | 1,075,000 | **1,000,000** |

B is short **75,000 uRTC every epoch, indefinitely**, out of the fixed `PER_EPOCH_URTC` pot. It compounds: every miner that *ever* dual-mined keeps +15% weight permanently, diluting everyone else.

Note `warthog_verification.py` itself is correct here — it returns `(False, 1.0, reason)` as it should; the caller discards that result. `get_warthog_bonus()` (`warthog_verification.py:201`), which would read the column, has zero live callers and so does not mitigate.

## Fix

Write the column unconditionally, so it describes the current attestation. One-line change (plus a comment recording why the guard must not come back).

## Verification

`node/tests/test_warthog_bonus_not_ratcheted.py` drives the real `/attest/submit` route (reusing the harness from `node/tests/test_attest_missing_fingerprint_reward_bypass.py`). **2 of 3 fail on main:**

```
FAILED test_attestation_without_warthog_proof_clears_stale_bonus
        AssertionError: bonus must reflect the current attestation
        assert 1.15 == 1.0
FAILED test_failed_warthog_verification_clears_stale_bonus
        AssertionError: a rejected proof must not keep the bonus
2 failed, 1 passed   ->   3 passed with the fix
```

The third (`test_verified_warthog_proof_still_records_bonus`) passes on **both** — it guards against over-reach by pinning that a genuine verified proof still earns 1.15x.

Existing `node/tests/test_rip200_warthog_bonus.py` stays green (2 passed). It's a good test, but it covers the **read** path only — it seeds `warthog_bonus` directly and never re-attests, which is exactly why the missing write went unnoticed.

Wider run (`node/tests/ -k "attest or warthog or rip200 or reward"`): **164 passed, 2 failed**. Honest note on those 2 — `test_attest_challenge_rate_limit.py` is **order-dependent and fails identically on pristine main** in the same batch (shared in-process rate-limit state); both pass in isolation with and without this change. On main the same batch fails 4 (those 2 + my 2 new tests); with this fix it's 2.

---

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -4838,17 +4838,22 @@ def _submit_attestation_impl():
     except Exception as _te:
         print(f"[TEMPORAL] Warning: {_te}")
 
-    # Update warthog_bonus in attestation record
-    if warthog_bonus > 1.0:
-        try:
-            with closing(sqlite3.connect(DB_PATH)) as wb_conn:
-                wb_conn.execute(
-                    "UPDATE miner_attest_recent SET warthog_bonus=? WHERE miner=?",
-                    (warthog_bonus, miner)
-                )
-                wb_conn.commit()
-        except Exception:
-            pass  # Column may not exist yet
+    # Update warthog_bonus in attestation record.
+    # Written unconditionally: warthog_bonus describes THIS attestation, and
+    # record_attestation_success() does not carry the column in its upsert. A
+    # `> 1.0` guard here would make the column a one-way ratchet — a miner that
+    # stops dual-mining, or that starts failing the fingerprint gate above,
+    # would keep the multiplier forever and dilute every honest miner's share
+    # of the fixed epoch pot in calculate_epoch_rewards_time_aged().
+    try:
+        with closing(sqlite3.connect(DB_PATH)) as wb_conn:
+            wb_conn.execute(
+                "UPDATE miner_attest_recent SET warthog_bonus=? WHERE miner=?",
+                (warthog_bonus, miner)
+            )
+            wb_conn.commit()
+    except Exception:
+        pass  # Column may not exist yet
 
     # Record MACs if provided
     if macs:
--- a/node/tests/test_warthog_bonus_not_ratcheted.py
+++ b/node/tests/test_warthog_bonus_not_ratcheted.py
@@ -1,0 +1,199 @@
+# SPDX-License-Identifier: MIT
+"""
+Regression tests: miner_attest_recent.warthog_bonus must track the CURRENT
+attestation, not ratchet up permanently.
+
+The submit handler only writes the column when the freshly computed bonus is
+> 1.0, and record_attestation_success() does not carry warthog_bonus in its
+ON CONFLICT DO UPDATE list. So once a miner earns the dual-mining bonus, no
+code path ever writes 1.0 back: the miner can stop dual-mining, or start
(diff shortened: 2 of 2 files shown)


elevasyncsolutions-jpg:
## Review of PR #7990 — Warthog Bonus Unconditionally Written

### Observation 1: The change from `if warthog_bonus > 1.0` to unconditional write is correct but changes semantics (`rustchain_v2_integrated_v2.2.1_rip200.py`)

Before: only wrote when bonus > 1.0, leaving the column as NULL/old value otherwise.
After: always writes the current bonus (including 1.0 when no bonus applies).

This means miners who lose dual-mining eligibility will now see `warthog_bonus=1.0` in the database, which is correct. But any downstream code checking `if warthog_bonus` (truthy check) will now see it as truthy even when it's 1.0 (no bonus). Verify that `calculate_epoch_rewards_time_aged` uses `> 1.0` comparison, not truthy check.

### Observation 2: The test should verify the ratchet-down behavior

The test checks that `warthog_bonus` is updated to 1.0 when the miner stops dual-mining. But it should also verify that the epoch reward calculation uses the new 1.0 value (not the old >1.0 value). This confirms the fix actually affects the reward calculation, not just the database value.

### Positive notes

- The root cause is clear: `record_attestation_success()` doesn't include `warthog_bonus` in its ON CONFLICT UPDATE, so the old bonus persists.
- The fix ensures the bonus always reflects the current attestation, not a historical one.
- The test covers both the ratchet-up and ratchet-down scenarios.

PR #7992, opened 2026-07-15
fix(replay): entropy profile hash reads keys the producer never emits

Vyacheslav-Tomashevskiy:
## Problem

`compute_entropy_profile_hash()` (node/hardware_fingerprint_replay.py) feeds the **live anti-Sybil gate**: the attestation submit path in `rustchain_v2_integrated_v2.2.1_rip200.py` (line ~4687) computes the entropy hash and passes it to `check_entropy_collision()` (line ~4708) to block wallets that share a hardware entropy profile.

But the function reads key names that the **real producer** (`node/fingerprint_checks.py`) never emits:

| consumer reads | producer actually emits |
|---|---|
| thermal `ratio` | `drift_ratio` |
| cache `L1` / `L2` | `l1_ns` / `l2_ns` (+ `l2_l1_ratio`, `l3_l2_ratio`) |
| cache `cache_hash` | — (no such field) |
| jitter `cv` | — (only `int/fp/branch_avg_ns` + `*_stdev`) |
| jitter `jitter_map` | — (no such field) |
| clock `drift_hash` | — (emits `cv`, `stdev_ns`, `drift_stdev`) |

Only `clock_drift.cv` and the whole `simd_identity.data` are read correctly. So on real hardware **thermal, cache and jitter collapse to constants**, and the entropy profile hash is derived almost entirely from `clock_cv + simd`.

**Effect:** two rigs that differ *only* in their thermal/cache/jitter entropy produce the **identical** `entropy_profile_hash`, so `check_entropy_collision` (a) can false-block honest, physically-distinct miners as `entropy_profile_collision`, and (b) ignores 3 of the 5 entropy dimensions it advertises for Sybil detection.

The existing `tests/test_replay_defense.py` fixture masked this — it fabricates a fingerprint with the *consumer's* invented keys (`ratio`, `L1`, `cache_hash`, `jitter_map`), not the producer's real ones, and `test_entropy_hash_different_profiles` only mutates `clock_drift.cv` (the one field read correctly).

## Fix

Read the producer's stable, load-normalized characteristics: `cv`, cache hierarchy ratios (`l2_l1_ratio`, `l3_l2_ratio`), thermal `drift_ratio`, and dimensionless jitter ratios (`fp/int`, `branch/int`). Raw `*_ns` stay excluded on purpose so the same machine yields a stable profile across attestations. This is the same producer-key alignment `hardware_binding_v2` already applies for thermal.

## Verification

New `tests/test_entropy_profile_real_producer_schema.py`, built on the **real producer shape**: 4 tests fail on clean `main` (distinct hardware collapses to one hash), all pass with the fix. Existing replay/binding suites stay green (78 passed replay + new; 69 passed / 1 skipped hardware-binding).

## Severity — honest scope

This is a correctness/security bug in a **live** attestation gate (write + read path both active). I'm not attaching a specific RTC-loss figure: it both false-blocks honest miners and under-detects real hardware sharing; the maintainer is better placed to weight those. The `*_ns`-volatility concern (why raw latencies must stay out of the hash) is covered by an added stability test.

### DIFF
--- a/node/hardware_fingerprint_replay.py
+++ b/node/hardware_fingerprint_replay.py
@@ -188,35 +188,40 @@ def compute_entropy_profile_hash(fingerprint: Dict) -> str:
     checks = fingerprint.get('checks', {}) if isinstance(fingerprint, dict) else {}
     
     entropy_values = {}
-    
+
+    # The producer (node/fingerprint_checks.py) emits load-normalized *ratios*
+    # and coefficients of variation as the stable per-machine characteristics;
+    # the raw *_ns latencies vary run-to-run and are intentionally excluded so
+    # the same box yields a stable profile across attestations.
+
     # Extract clock drift entropy
     clock_data = checks.get('clock_drift', {}).get('data', {})
     if isinstance(clock_data, dict):
         entropy_values['clock_cv'] = clock_data.get('cv', 0)
-        entropy_values['clock_drift_hash'] = clock_data.get('drift_hash', '')
-    
-    # Extract cache timing entropy
+
+    # Extract cache timing entropy (hierarchy ratios are the stable signal)
     cache_data = checks.get('cache_timing', {}).get('data', {})
     if isinstance(cache_data, dict):
-        entropy_values['cache_hash'] = cache_data.get('cache_hash', '')
-        entropy_values['cache_l1'] = cache_data.get('L1', 0)
-        entropy_values['cache_l2'] = cache_data.get('L2', 0)
-    
+        entropy_values['cache_l2_l1_ratio'] = cache_data.get('l2_l1_ratio', 0)
+        entropy_values['cache_l3_l2_ratio'] = cache_data.get('l3_l2_ratio', 0)
+
     # Extract thermal drift entropy
     thermal_data = checks.get('thermal_drift', {}).get('data', {})
     if isinstance(thermal_data, dict):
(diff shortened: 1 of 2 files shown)


elevasyncsolutions-jpg:
## Review of PR #7992 — Entropy Profile Hash Reads Correct Producer Keys

### Observation 1: Comment says raw `*_ns` latencies are excluded, but `l1_ns`/`l2_ns` ARE read (`hardware_fingerprint_replay.py`)

The new comment says "the raw *_ns latencies vary run-to-run and are intentionally excluded." But the code reads `l1_ns` and `l2_ns` from the producer. Either the comment is wrong (these values ARE included because they're load-normalized ratios, not raw latencies), or the comment should clarify that `l1_ns`/`l2_ns` are the *normalized* versions, not the raw timing values.

### Observation 2: The test covers the happy path but not the fallback path (`test_entropy_profile_real_producer_schema.py`)

The test verifies that a real producer schema produces a non-empty hash. But it doesn't test what happens when the producer emits an older schema (v1/v2 keys). The fix is about backwards compatibility — consider adding a test with legacy key names to verify the hash still works (or fails gracefully).

### Positive notes

- The regression test using a real producer schema is excellent — it catches key mismatches that unit tests with hand-crafted dicts would miss.
- The fix correctly maps the keys the producer actually emits, closing the gap between producer and consumer.

PR #7993, opened 2026-07-15
Sign the pre-image the P2P auth middleware verifies (secure block sync 401s itself)

Vyacheslav-Tomashevskiy:
## The bug

`SecureBlockSync.sync_from_peers` and the auth middleware sign **different pre-images**, so they can never agree.

Signer (`node/rustchain_p2p_sync_secure.py:493-495`):
```python
message = f"get_blocks:{peer_url}"
signature, timestamp = self.peer_manager.auth_manager.generate_signature(message)
response = requests.get(f"{peer_url}/p2p/blocks", headers={'X-Peer-Signature': signature, ...})
```

Verifier (`create_p2p_auth_middleware.decorated`, `:595-598`):
```python
body = request.get_data().decode()
if not auth_manager.verify_peer_signature(signature, body, timestamp):
    return jsonify({'error': 'Invalid signature'}), 401
```

`verify_peer_signature` recomputes `HMAC(key, f"{message}{timestamp}")`. The request is a **GET**, so `request.get_data()` is `b""`: the verifier computes `HMAC(key, "" + ts)` while the signer produced `HMAC(key, "get_blocks:http://peer:8099" + ts)`. This is not a key-distribution problem — with a perfectly shared `RC_P2P_KEY` it still fails, because the pre-images differ.

## Why it goes unnoticed

It fails **closed and silent**. There is no `else` on `if response.ok:` (`:508`), and the `-5` reputation penalty (`:526`) only fires on exceptions — a 401 is a clean response, so it is neither logged nor scored. `SecureBlockSync` prints `[P2P] [OK] Block sync started` and then ingests exactly nothing, every 30s, forever.

Existing tests miss it because `tests/test_p2p_peer_auth_no_ip_bypass.py` only exercises the middleware from a hand-built client that **signs the body correctly** (`:49-50`, `auth_manager.generate_signature(body)`) — it never drives `sync_from_peers`, the one caller that doesn't.

## Live path

- `node/rustchain_v2_integrated_v2.2.1_rip200.py:10672` — `from rustchain_p2p_sync_secure import initialize_secure_p2p`
- `:10675-10678` — `peer_manager, block_sync, require_peer_auth = initialize_secure_p2p(...)`
- `:10693-10694` — `@app.route('/p2p/blocks', methods=['GET'])` + `@require_peer_auth`
- `:10743-10744` — `block_sync.start()` unless `RUSTCHAIN_DISABLE_P2P_AUTO_START=1`

## Severity, honestly

This is *good peers rejected*, not an auth bypass — nothing unauthorized gets in. And the separate gossip path (`rustchain_p2p_init.init_p2p` → `rustchain_p2p_gossip`, wired at `:11749-11750`) is also live, so the chain still syncs by that route. The accurate framing is: **the "secure" block-sync module has never done anything**, and its failure mode hides that from the operator. I'm not claiming lost funds.

## The fix

Sign the body — the contract the middleware defines, and the one the existing test already encodes. For this bodyless GET that is the empty string.

Deliberately **not** bundled: the fact that an empty-body GET signature covers only the timestamp (so it is replayable across bodyless endpoints within the 300s window) is a real design question, but tightening it means changing the middleware's pre-image contract and every caller — a separate PR, and your call on the shape. Happy to open it if you want the pre-image to include method+path.

## Verification

`tests/test_p2p_sync_signs_what_verifier_checks.py` drives the **real** `sync_from_peers` against the **real** middleware over a shared key; only `requests.get` is redirected into a Flask test client.

```
main:  2 failed  -> peer rejected our own signature: HTTP [401] / "no block ingested; the secure sync path is inert"
here:  2 passed
```

Neighbouring suites green: `test_p2p_peer_auth_no_ip_bypass` 3 passed; `test_p2p_blocks`, `test_p2p_blocks_and_add_peer_6131_6129`, `test_p2p_demo_peer_config`, `test_p2p_mtls_gate`, `test_p2p_nonce_security` 29 passed.

### DIFF
--- a/node/rustchain_p2p_sync_secure.py
+++ b/node/rustchain_p2p_sync_secure.py
@@ -491,9 +491,12 @@ class SecureBlockSync:
                 if not self.peer_manager.rate_limiter.check_rate_limit(peer_url, '/p2p/blocks'):
                     continue
 
-                # Generate auth signature
-                message = f"get_blocks:{peer_url}"
-                signature, timestamp = self.peer_manager.auth_manager.generate_signature(message)
+                # Sign the request body: that is the pre-image the peer's auth
+                # middleware verifies (create_p2p_auth_middleware -> request.get_data()).
+                # This GET carries no body, so the signed pre-image is the empty
+                # string, matching tests/test_p2p_peer_auth_no_ip_bypass.py.
+                body = ""
+                signature, timestamp = self.peer_manager.auth_manager.generate_signature(body)
 
                 # Request blocks with authentication
                 response = requests.get(
--- a/tests/test_p2p_sync_signs_what_verifier_checks.py
+++ b/tests/test_p2p_sync_signs_what_verifier_checks.py
@@ -1,0 +1,104 @@
+# SPDX-License-Identifier: MIT
+"""sync_from_peers must sign the pre-image the auth middleware actually verifies.
+
+The signer built its own message (`get_blocks:{peer_url}`) while
+create_p2p_auth_middleware verifies over `request.get_data()`, so the two HMACs
+were computed over different pre-images and every sync request 401'd -- with no
+else-branch and no reputation penalty, so silently.
+
+This drives the REAL sync_from_peers against the REAL middleware over a shared
+key; only requests.get is redirected into a Flask test client.
+"""
+import importlib.util
+import os
+import sys
+import tempfile
+from pathlib import Path
+
+import pytest
+from flask import Flask, jsonify
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+MODULE_PATH = REPO_ROOT / "node" / "rustchain_p2p_sync_secure.py"
+
+os.environ["RC_P2P_KEY"] = "shared-network-key"
+spec = importlib.util.spec_from_file_location("rustchain_p2p_sync_secure", MODULE_PATH)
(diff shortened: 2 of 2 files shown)


elevasyncsolutions-jpg:
## Review of PR #7993 — P2P Auth Signs What Verifier Checks

### Observation 1: The empty-body case uses a hardcoded empty string (`rustchain_p2p_sync_secure.py`)

For GET requests (no body), the code signs `b""`. But if the peer's auth middleware does `request.get_data()` and gets `b""`, the HMAC will match. However, if Flask normalizes empty bodies differently (e.g. some versions return `b""` vs `None`), the HMAC comparison could fail. Consider adding a comment documenting this assumption.

### Observation 2: The test uses Flask test client but the real code uses `requests.get` (`test_p2p_sync_signs_what_verifier_checks.py`)

The test redirects `requests.get` into a Flask test client via `responses` or `monkeypatch`. This is a common pattern but the test should verify that the *real* `requests.get` → Flask middleware path works, not just the mock path. The test does this via `requests.get = mocked_get` which is correct.

### Positive notes

- The root cause is clear: signer used `get_blocks:{peer_url}` while middleware verified `request.get_data()`. Two different pre-images = guaranteed 401.
- The fix is minimal and correct: sign what the verifier actually checks.
- The test name `test_p2p_sync_signs_what_verifier_checks` is self-documenting.

PR #7994, opened 2026-07-15
Refuse to spend account-mirror boxes via /utxo/transfer while dual-write is off (#2819 double-spend, UTXO side)

Vyacheslav-Tomashevskiy:
## Summary

`/utxo/transfer` spends account-mirror boxes without any account-model write, re-opening the **#2819 cross-model double-spend from the UTXO side**. Reachable by one self-transfer, no admin key.

## The asymmetry

`account_mirror_boxes` is the discriminator recording "this box IS this wallet's account balance". `_settle_account_transfer_in_utxo` closes the **account → UTXO** direction, and its own docstring (`node/rustchain_v2_integrated_v2.2.1_rip200.py:10210-10212`) is explicit about why it isn't gated on dual-write:

> *Gated on HAVE_UTXO, independent of UTXO_DUAL_WRITE: a migrated box must be reconciled whenever it exists.*

That is the node conceding that in the default config **the mirror box and the account balance are the same money**. The **UTXO → account** direction has no equivalent guard.

## Mechanism

- `node/rustchain_v2_integrated_v2.2.1_rip200.py:50` — `UTXO_DUAL_WRITE = os.environ.get("UTXO_DUAL_WRITE", "0") == "1"` → **False by default**.
- `:11722-11731` — the blueprint registers under `if HAVE_UTXO:` alone, passing `dual_write=UTXO_DUAL_WRITE`. So **`/utxo/transfer` is live with dual-write off**.
- `utxo_endpoints.py:738` — `apply_transaction()` spends the inputs **unconditionally**.
- `utxo_endpoints.py:743` — the shadow-balance guard, `UPDATE balances` and `INSERT INTO ledger` all sit inside `if _dual_write:`.
- Neither `utxo_endpoints.py` nor `utxo_db.py` reads or writes `account_mirror_boxes` — it appears only in the monolith (`:10146-10301`).

Alice is migrated: `balances.amount_i64 = 100 RTC` **and** a mirror box of the same value with a provenance row. She `POST /utxo/transfer`s to **herself**:

1. `apply_transaction` spends the mirror box and mints her a fresh box — `registers_json = '{}'`, no provenance row (the backfill at `:10186` only adopts `R4 IN ('genesis','account_migration_mirror')`, so it won't re-adopt it either).
2. `_dual_write` is False → `balances` untouched, no ledger row.
3. Her provenance row now points at a **spent** box. `_spend_wallet_mirror` joins on `b.spent_at IS NULL` → 0 → `_settle_account_transfer_in_utxo` hits `if sender_mirror <= 0: return` and treats her as a *non-migrated sender*, so a later account move "stands".

Alice now holds a withdrawable 100 RTC account balance **and** a ~99 RTC box. The `mirror <= balance` invariant at `:10294+` doesn't fire: her mirror sum is now 0, and the laundered box is invisible to it.

## Severity — and the caveat that bounds it

The exploit needs `utxo_genesis_migration.py` to have run against the DB. I checked: it is an **ops script, never called at node startup** (only tests import it). On an unmigrated DB `utxo_boxes` is empty and `/utxo/transfer` just 400s — the endpoint is inert, not wrong.

I can't verify your deployment from here, so I won't claim funds are being lost today. What I'll point at is `_ensure_and_backfill_account_mirror`'s own rationale (`:10152-10156`) — it exists for "a DB whose genesis migration ran *before* this provenance tracking existed", which reads as though migrated DBs are the expected case. **You know whether the live DB is migrated; that decides whether this is Critical or latent.** Either way it's the exact hole #2819 paid to close, left open in the other direction.

## The fix, and the alternative I did *not* take

Fail closed: with dual-write off there is no account debit to pair the spend with, so migrated funds must move via the account path (which reconciles correctly today). The guard runs inside the existing `BEGIN IMMEDIATE`, and uses the `account_mirror_boxes` discriminator rather than a `registers_json` match — for the reasons `_ensure_and_backfill_account_mirror` already documents (a marker is lost on change boxes; "any box the sender owns" would wrongly block independently-earned UTXOs).

The alternative is to **debit `balances` whenever a mirror box is spent, regardless of dual-write** — symmetric with `_settle_account_transfer_in_utxo`. I didn't do that: it makes `/utxo/transfer` write account state in a config that exists specifically to keep it from doing so, and that's a product call, not a bug fix. Say the word and I'll send it instead.

Also deliberately out of scope: maintaining provenance across change/receiver boxes on the UTXO side. It's needed if you take the debit route, but on its own it does **not** close the hole (the balance still goes undebited), so it doesn't belong in a fix PR.

## Verification

`tests/test_utxo_transfer_spends_account_mirror.py` — fixtures follow `tests/test_utxo_transfer_replay.py`; the seed reproduces exactly what `utxo_genesis_migration.py:220-251` writes (box + `R4:genesis` + provenance row + balance), with `dual_write=False` to match the live default.

```
main:  2 failed, 1 passed
here:  3 passed
```

- `test_account_balance_stays_backed_by_a_mirror_box` — on main: mirror-backed value drops to 0 while the balance stays 100 RTC.
- `test_mirror_box_is_not_spent_while_dual_write_is_off` — on main: `mirror box was spent without debiting the account it mirrors`.
- `test_independently_earned_boxes_are_unaffected` — **passes on main and here**: the control proving the guard doesn't over-block non-mirror UTXOs.

Neighbouring suites green — `test_utxo_transfer_replay`, `test_utxo_transfer_nonce_required`, `test_utxo_transfer_json_validation`, `test_utxo_malformed_pubkey_6114`, `test_utxo_legacy_signature_deadline`, `test_utxo_security_audit`, `test_utxo_mempool_atomicity_bug1_bug4`, `test_consensus_invariant_utxo_examples`: **54 passed**.

One honest note on my first repro: it "passed" on main until I noticed the payload used `pubkey_hex` instead of `public_key` and was 400ing before it ever reached the bug. The numbers above are from the corrected version.

### DIFF
--- a/node/utxo_endpoints.py
+++ b/node/utxo_endpoints.py
@@ -186,6 +186,29 @@ _current_slot_fn = None    # current_slot() -> int
 _dual_write: bool = False
 
 
+def _selected_account_mirror_boxes(conn: sqlite3.Connection, selected: list) -> list:
+    """Which of the selected boxes are account-mirror provenance (bounty #2819).
+
+    Uses the ``account_mirror_boxes`` discriminator the node maintains, not a
+    ``registers_json`` marker match: a marker is lost on change boxes from
+    partial spends, and "any box the sender owns" would wrongly block
+    independently-earned UTXOs. Absent table (pure-UTXO DB) means no mirrors.
+    """
+    if not selected:
+        return []
+    if not conn.execute(
+        "SELECT 1 FROM sqlite_master WHERE type='table' AND name='account_mirror_boxes'"
+    ).fetchone():
+        return []
+    box_ids = [u['box_id'] for u in selected]
+    placeholders = ','.join('?' * len(box_ids))
+    rows = conn.execute(
+        f"SELECT box_id FROM account_mirror_boxes WHERE box_id IN ({placeholders})",
+        box_ids,
+    ).fetchall()
+    return [r[0] for r in rows]
+
+
 def _ensure_transfer_nonce_table(conn: sqlite3.Connection) -> None:
     conn.execute(
         """
@@ -734,6 +757,25 @@ def utxo_transfer():
                 'nonce': nonce,
                 'latest_nonce': int(previous_nonce),
             }), 400
+
+        # Account-mirrored boxes ARE the sender's account balance (bounty #2819).
+        # The account->UTXO direction is reconciled by the node's
+        # _settle_account_transfer_in_utxo, which runs independent of dual-write
+        # precisely because "a migrated box must be reconciled whenever it exists".
+        # This is the same crossing in reverse: with dual-write off there is no
+        # account debit to pair the spend with, so spending a mirror box here
+        # would leave the balance behind it fully spendable. Fail closed --
+        # migrated funds move via the account path in this config.
+        if not _dual_write:
+            mirrored = _selected_account_mirror_boxes(conn, selected)
+            if mirrored:
+                conn.rollback()
+                return jsonify({
+                    'error': 'Box mirrors an account balance; move it via the account '
+                             'transfer path while dual-write is off',
+                    'code': 'ACCOUNT_MIRROR_BOX_NOT_SPENDABLE',
+                    'box_ids': mirrored,
+                }), 409
 
         ok = _utxo_db.apply_transaction(tx, block_height, conn=conn)
         if not ok:
--- a/tests/test_utxo_transfer_spends_account_mirror.py
+++ b/tests/test_utxo_transfer_spends_account_mirror.py
@@ -1,0 +1,190 @@
+# SPDX-License-Identifier: MIT
+"""/utxo/transfer must not move account-mirrored funds while dual-write is off.
+
+`account_mirror_boxes` is the discriminator that says "this box IS this wallet's
+account balance" (bounty #2819). `_settle_account_transfer_in_utxo` closes the
+account->UTXO direction: an account confirm spends the sender's mirror box so the
+same funds cannot move twice. Its docstring is explicit that it runs
+"independent of UTXO_DUAL_WRITE: a migrated box must be reconciled whenever it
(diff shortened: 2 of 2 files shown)


elevasyncsolutions-jpg:
## Review of PR #7994 — Refuse Account-Mirror Box Spends

### Observation 1: `_selected_account_mirror_boxes` queries `account_mirror_boxes` table which may not exist (`utxo_endpoints.py`)

The function uses `conn.execute("SELECT box_id FROM account_mirror_boxes ...")` but the comment says "Absent table (pure-UTXO DB) means no mirrors." If the table doesn't exist, `sqlite3.OperationalError` will be raised. The code should wrap this in a try/except or check `sqlite_master` first.

### Observation 2: The block is a hard 400 error with no escape hatch

If a user accidentally creates an account-mirror box (e.g. via a self-transfer), they're permanently locked out of spending those UTXOs unless they manually delete the `account_mirror_boxes` row. Consider adding a flag like `--force-mirror-spend` or documenting the manual escape hatch.

### Observation 3: The double-spend scenario description is accurate but the fix is conservative

The PR blocks ALL mirror box spends when dual-write is off, even if the user explicitly wants to migrate funds back. This is the safe choice, but users who need to recover funds from mirror boxes will need a migration tool or admin intervention.

### Positive notes

- This closes a real double-spend vulnerability (issue #2819).
- The discriminator approach (using `account_mirror_boxes` table, not register markers) is correct — markers get lost on change boxes from partial spends.
- The test covers the self-transfer double-spend scenario thoroughly.

Scottcjn:
Rebased onto current main and merged as https://github.com/Scottcjn/Rustchain/pull/8006, with the `# fetchall-ok: bounded-by-schema` annotation added so the fetchall guard passes. Your authorship is preserved and your test (3/3) is green. Real cross-model double-spend fix — thank you. Closing in favor of the rebased merge.

PR #7995, opened 2026-07-15
fix(install): macOS bash 3.2 crash and wrong download paths

### LINKED ISSUE #7975: Bug: install script at rustchain.org has wrong file paths and crashes on macOS

## Summary

```bash
```

**Proof:**

Line 36 of the script:
```bash
declare -A MULT_TABLE=(
    ...
)
```

```
```

## To Reproduce
```bash
```

## Expected Behavior

rebel117:
## Problem

The installer (`scripts/install.sh`) has two bugs that prevent it from running on macOS:

**1. `declare -A` crashes on macOS bash 3.2**

macOS ships bash 3.2 by default, which doesn't support associative arrays. Running the script produces:
```
```

**2. Hardcoded Linux download paths**

`MINER_URL`, `FINGERPRINT_URL`, and `MINER_CRYPTO_URL` all point at `miners/linux/` regardless of OS. On macOS the script downloads the Linux miner and tries to download `miner_crypto.py` which doesn't exist in `miners/macos/`.

## Fix

1. Replaced `declare -A MULT_TABLE` with a `get_mult_value()` function using a plain `case` statement — works on bash 3.2+
2. Added OS detection for `MINER_SUBDIR` so macOS gets `miners/macos/rustchain_mac_miner_v2.5.py` instead of the Linux miner
3. Made `miner_crypto.py` download conditional on Linux (macOS miner is self-contained)
4. Introduced `MINER_SCRIPT` variable so systemd/launchd service files reference the correct script name per platform
5. Bumped `install.sh` (root) macOS miner reference from v2.4 to v2.5

## Testing

- `bash -n scripts/install.sh` — syntax OK
- `bash -n install.sh` — syntax OK
- Verified `get_mult_value()` returns correct values for all arch keys
- Confirmed no `declare -A` remains in either script

Fixes #7975

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -76,7 +76,7 @@ if [ "$DRY_RUN" -eq 1 ]; then
             FINGERPRINT_PATH="linux/fingerprint_checks.py"
             ;;
         Darwin)
-            MINER_PATH="macos/rustchain_mac_miner_v2.4.py"
+            MINER_PATH="macos/rustchain_mac_miner_v2.5.py"
             FINGERPRINT_PATH="macos/fingerprint_checks.py"
             ;;
         *)      echo -e "${RED}Unsupported OS: $OS${NC}"; exit 1 ;;
@@ -127,7 +127,7 @@ case "$OS" in
         ;;
     Darwin)
         echo "  OS: macOS"
-        MINER_PATH="macos/rustchain_mac_miner_v2.4.py"
+        MINER_PATH="macos/rustchain_mac_miner_v2.5.py"
         FINGERPRINT_PATH="macos/fingerprint_checks.py"
         ;;
     *)      echo -e "${RED}  Unsupported OS: $OS${NC}"; exit 1 ;;
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -48,21 +48,43 @@ download_file() {
 # --- Constants -------------------------------------------------------------
 INSTALL_DIR="/opt/rustchain-miner"
 REPO_RAW="https://raw.githubusercontent.com/Scottcjn/Rustchain/main"
-MINER_URL="${REPO_RAW}/miners/linux/rustchain_linux_miner.py"
-FINGERPRINT_URL="${REPO_RAW}/miners/linux/fingerprint_checks.py"
-MINER_CRYPTO_URL="${REPO_RAW}/miners/linux/miner_crypto.py"
+# Resolve miner paths based on OS (set dynamically in main)
+MINER_SUBDIR="linux"
+if [ "$(uname -s)" = "Darwin" ]; then
+    MINER_SUBDIR="macos"
+fi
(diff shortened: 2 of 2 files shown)


elevasyncsolutions-jpg:
## Review of PR #7995 — macOS bash 3.2 Compatibility + Download Path Fix

### Observation 1: `MINER_URL` still hardcodes `rustchain_linux_miner.py` even when `MINER_SUBDIR=macos` (`scripts/install.sh`)

The initial assignment on line 55 uses:
```
MINER_URL="${REPO_RAW}/miners/${MINER_SUBDIR}/rustchain_linux_miner.py"
```
When `MINER_SUBDIR=macos`, this resolves to `miners/macos/rustchain_linux_miner.py` — which is wrong. The subsequent `if` block corrects it to `rustchain_mac_miner_v2.5.py`, but the initial wrong URL is assigned first. While the `if` block fixes it, this creates a confusing code path. Consider setting `MINER_URL` conditionally from the start, or using a case statement to set all three URLs together.

### Observation 2: `declare -A` replacement uses parallel arrays which may break if entries are reordered (`scripts/install.sh`)

The fix replaces `declare -A` with parallel arrays `MULT_KEYS` and `MULT_VALUES`. This works, but the lookup relies on matching indices. If someone adds an entry to one array but forgets the other, the lookup silently returns the wrong multiplier. Consider using a single case statement instead:

```bash
get_multiplier() {
    case "$1" in
        sparc) echo "2.9";; mips) echo "3.0";; 68k) echo "3.5";;
        g4) echo "2.5";; g5) echo "2.0";; g3) echo "1.8";;
        power8) echo "1.5";; riscv) echo "1.4";; retro) echo "1.4";;
        apple_silicon) echo "1.2";; modern) echo "1.0";; aarch64) echo "0.0005";;
        *) echo "1.0";;
    esac
}
```

This is more idiomatic for bash 3.2 and avoids the index-matching fragility.

### Observation 3: The `aarch64` multiplier `0.0005` seems suspiciously low

A multiplier of 0.0005 means aarch64 miners get 0.05% of the base reward. This is 2000x less than `modern` (1.0). If this is intentional (e.g. aarch64 is not yet supported), it should have a comment explaining why. If it's a placeholder, it could cause confusion for users who see nearly zero rewards on ARM64 Linux.

### Positive notes

- The `uname -s` approach for detecting macOS is portable and correct.
- The macOS miner path version bump from v2.4 to v2.5 is clearly documented.
- The bash 3.2 compatibility fix is the right approach — associative arrays were only added in bash 4.0.

Scottcjn:
You have two PRs doing the same macOS bash-3.2 installer fix — this one and #8005. #8005 is the cleaner implementation (single-branch OS resolution, graceful fallbacks), so let's consolidate there. Closing this in favor of #8005; I've left the specifics of the remaining CI failure on that one.

PR #7996, opened 2026-07-16
fix(beacon-x402): premium exports query tables no producer creates (200 + total:0 forever)

Vyacheslav-Tomashevskiy:
## The bug

`/api/premium/reputation` and `/api/premium/contracts/export` — the two endpoints `/api/x402/status` advertises as the premium offering — query tables that no producer in this tree creates.

| consumer (`node/beacon_x402.py`) | real producer (`node/beacon_api.py:95`) |
|---|---|
| `SELECT * FROM reputation` (:340) | `CREATE TABLE beacon_reputation` (:137) |
| `SELECT * FROM contracts` (:368) | `CREATE TABLE beacon_contracts` (:100) |

`beacon_api.init_beacon_tables` is the only writer of this data and the one the node calls; it uses `DB_PATH = 'rustchain_v2.db'`, the same database the monolith hands to `beacon_x402` (`_beacon_x402_get_db`, `rustchain_v2_integrated_v2.2.1_rip200.py:761-773` → `RUSTCHAIN_DB_PATH` / `./rustchain_v2.db`). So the connection is right; only the table names are wrong.

The `except sqlite3.OperationalError: rows = []` then converts *"you named a table that does not exist"* into **HTTP 200 with `total: 0`** — indistinguishable from "there is genuinely no data yet". Both exports have therefore returned nothing, silently, for every caller.

## A second defect the first one was masking

Correcting the table names makes the contracts export get further — and it then **500s**:

```
sqlite3.OperationalError: no such table: beacon_wallets   # beacon_x402.py:379
```

`_run_migrations` (:216) creates `beacon_wallets` in a `beacon_atlas.db` **beside the module file** (:210), not in the runtime database `get_db_func` returns. Only the two wallet routes (:250, :280) call `_ensure_x402_tables`, so on a node where no wallet was ever registered the table is absent from the runtime DB and the unguarded join at :379-382 raises. On `main` the loop never ran, because the export always returned zero rows — the two bugs hid each other.

Guarded with the module's own `_ensure_x402_tables`, matching what both wallet routes already do (idempotent `CREATE TABLE IF NOT EXISTS`).

## Severity, honestly

Both prices are `"0"` in `x402_config.py:33,38` today, so **nobody is currently paying for an empty response** — the endpoints are free and inert. I am not claiming lost funds. What I'd flag is the failure mode: the comments there read `# Future: "50000" = $0.05`, and the day those flip on, a caller pays USDC and receives `total: 0` with a 200 — no error, no refund path, nothing in the logs. The contracts export would also 500 for the first payer whose node never registered a wallet. Worth fixing before the price is switched on rather than after. You know your deployment; the weight is yours to set.

## Why the tests never caught it

`tests/test_beacon_x402_payment_gate.py:14,84` fixtures the database with hand-written `CREATE TABLE reputation` / `CREATE TABLE contracts` — the consumer's imagined names, which no producer creates. The suite validated the bug rather than the deployed schema. Same pattern as the anchors (#7987) and fleet (#7988) skews.

## Verification

New `tests/test_beacon_x402_premium_export_real_schema.py` builds the DB by calling the **real producer** (`beacon_api.init_beacon_tables`), so the fixture cannot drift from what the node deploys.

- **3 fail on pristine `main`, 4 pass with this fix** (verified in a clean worktree at `upstream/main`).
- The 4th test is a control — a genuinely-empty `beacon_reputation` still correctly returns `total: 0`, and it passes both on `main` and with the fix, proving the tests distinguish the bug from real empty state.
- Neighbours green: **186 passed, 10 skipped** (`-k beacon`), plus the 7 existing gate tests still pass unchanged.
- `tests/test_relic_market_signing_keys.py` fails collection on `main` too (`ModuleNotFoundError: No module named 'nacl'`) — pre-existing, unrelated.

## Not bundled

`x402_beacon_payments` is created and read (:409) but **never inserted into** anywhere in the module — the payment history endpoint is empty by construction. That looks like an unfinished feature rather than a typo, so I left it alone rather than guess at the intended write path. Happy to take it as a follow-up if you want it.

### DIFF
--- a/node/beacon_x402.py
+++ b/node/beacon_x402.py
@@ -337,7 +337,7 @@ def init_app(app, get_db_func):
         db = get_db_func()
         try:
             rows = db.execute(
-                "SELECT * FROM reputation ORDER BY score DESC"
+                "SELECT * FROM beacon_reputation ORDER BY score DESC"
             ).fetchall()
             reputation = [dict(r) for r in rows]
         except sqlite3.OperationalError:
@@ -363,9 +363,12 @@ def init_app(app, get_db_func):
             return err_resp
 
         db = get_db_func()
+        # The wallet join below is unguarded, and _run_migrations creates
+        # beacon_wallets next to this module rather than in the runtime database.
+        _ensure_x402_tables(db)
         try:
             rows = db.execute(
-                "SELECT * FROM contracts ORDER BY created_at DESC"
+                "SELECT * FROM beacon_contracts ORDER BY created_at DESC"
             ).fetchall()
         except sqlite3.OperationalError:
             rows = []
--- a/tests/test_beacon_x402_premium_export_real_schema.py
+++ b/tests/test_beacon_x402_premium_export_real_schema.py
@@ -1,0 +1,154 @@
+# SPDX-License-Identifier: MIT
+"""Premium x402 exports must query the tables the Beacon producer actually creates.
+
+The existing gate suite (test_beacon_x402_payment_gate.py) fixtures its database with
+hand-written ``CREATE TABLE reputation`` / ``CREATE TABLE contracts`` statements. No
+producer in the tree ever creates those names: ``beacon_api.init_beacon_tables`` -- the
+only writer, and the one the node calls -- creates ``beacon_reputation`` and
+``beacon_contracts``. So the suite is green while production exports nothing.
+
+These tests build the database with the *real* producer so the schema cannot drift from
+what the node actually deploys.
+"""
+
+import sqlite3
+
+import pytest
+from flask import Flask
+
+import beacon_api
+import beacon_x402
+
+
+@pytest.fixture
+def real_schema_db(tmp_path):
+    """A database created by the real producer, seeded via its real table names."""
+    db_path = tmp_path / "rustchain_v2.db"
+    beacon_api.init_beacon_tables(str(db_path))
+
+    with sqlite3.connect(db_path) as conn:
+        conn.execute(
+            "INSERT INTO beacon_reputation (agent_id, score, bounties_completed) "
+            "VALUES (?, ?, ?)",
+            ("agent-alpha", 99, 3),
+        )
+        conn.execute(
+            "INSERT INTO beacon_contracts (id, from_agent, to_agent, type, amount, "
+            "term, state, created_at) VALUES (?, ?, ?, ?, ?, ?, ?, ?)",
+            ("contract-1", "agent-alpha", "agent-beta", "work", 100.0, "net30",
+             "offered", 1000),
+        )
+    return db_path
+
+
+def _client(db_path, monkeypatch):
+    monkeypatch.setattr(beacon_x402, "X402_CONFIG_OK", True)
+    monkeypatch.setattr(beacon_x402, "PRICE_REPUTATION_EXPORT", "0", raising=False)
+    monkeypatch.setattr(beacon_x402, "PRICE_BEACON_CONTRACT", "0", raising=False)
+    monkeypatch.setattr(
+        beacon_x402,
+        "is_free",
+        lambda price: str(price) in ("0", "0.0", "0.00", ""),
+        raising=False,
+    )
+    # The module migrates a path of its own choosing; these tests exercise the runtime
+    # connection the node hands in, which is what every route actually reads.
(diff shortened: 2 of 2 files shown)


jjb9707 (approved):
## Substantive review — schema and runtime-path verification

**Scope verified:** `node/beacon_x402.py` plus `tests/test_beacon_x402_premium_export_real_schema.py` at head `3fda1d90`. I cross-checked the producer in `node/beacon_api.py` and exercised both real Flask routes against a SQLite database created by `beacon_api.init_beacon_tables()`.

### What is correct

1. **The table-name diagnosis is accurate.** `beacon_api.init_beacon_tables()` creates `beacon_reputation` and `beacon_contracts`; the previous bare `reputation` / `contracts` names have no producer in this path. The old `except sqlite3.OperationalError: rows = []` converted that mismatch into a misleading HTTP 200 with `total: 0`.
2. **The second failure is real and previously masked.** Once contract rows are returned, the per-row wallet lookup touches `beacon_wallets`. `init_app()` migrates a module-local `beacon_atlas.db`, while the routes use the connection supplied by `get_db_func`, so the runtime DB may lack that table. Calling the idempotent `_ensure_x402_tables(db)` before the loop makes that lookup safe.
3. **The regression tests use the producer schema rather than repeating the consumer assumption.** That is the right boundary: future producer/consumer drift becomes observable instead of being baked into a hand-written fixture.
4. **Change scope is focused.** Two query-name corrections, one runtime-table guard, and targeted tests; no unrelated payment or schema redesign is mixed in.

### Verification performed

- All GitHub checks on this head are green, including the repository `test` job and both BCOS gates.
- `python3 -m compileall` succeeds for both changed Python files.
- Independent smoke test: created a temporary DB with the real `beacon_api.init_beacon_tables()`, seeded one reputation row and one contract, initialized the real `beacon_x402` Flask routes, then called both endpoints. Results were HTTP 200 with `reputation.total == 1` and `contracts.total == 1`; the absent wallet table was created and both wallet fields correctly returned `null`.
- `git diff --check` equivalent syntax/whitespace inspection found no patch-format issue. The local base image lacks pytest, so I did not claim a local pytest run; CI's `test` check is the authoritative full-suite evidence here.

### Non-blocking follow-ups

1. The `OperationalError -> []` catches still make any *future* schema regression look like a legitimate empty export. This PR fixes the present mismatch, but a later hardening change should log the exception or return 5xx for missing-table errors rather than silently reporting `total: 0`.
2. `_ensure_x402_tables(db)` executes both `CREATE TABLE IF NOT EXISTS` statements on every contracts-export request. It is correct and idempotent, but migration alignment should eventually be fixed at initialization so this paid/read path does not perform DDL.
3. There are no explicit assertions that `_ensure_x402_tables()` preserves a pre-existing populated `beacon_wallets` table, although SQLite's `IF NOT EXISTS` semantics make that safe. This is not a merge blocker.

**Verdict: APPROVE.** The production defect is concrete, the implementation matches the real producer/runtime boundaries, the added tests target the historical blind spot, and the remaining concerns are hardening opportunities rather than correctness blockers.

Reviewer: @jjb9707

elevasyncsolutions-jpg:
## Review of PR #7996 — Premium Exports Query Correct Tables

### Observation 1: The `beacon_wallets` join in `premium_contracts_export` may still fail (`beacon_x402.py`)

The comment says "The wallet join below is unguarded, and _run_migrations creates beacon_wallets next to this m..." — the patch is cut off. If `beacon_wallets` doesn't exist (e.g. on a node that never ran the beacon), the join will raise `sqlite3.OperationalError`. The fix should add a try/except around the query.

### Observation 2: No fallback for missing `reputation` or `contracts` tables

The original code catches `sqlite3.OperationalError` and returns an empty list. But the new code queries `beacon_reputation` / `beacon_contracts` which may not exist if the beacon module was never initialized. The try/except behavior should be preserved.

### Positive notes

- The root cause is clear: the endpoints query `reputation` and `contracts` but the only table creator (`beacon_api.init_beacon_tables`) creates `beacon_reputation` and `beacon_contracts`.
- The test builds the database with the real producer schema, catching the mismatch that the existing test suite (with hand-written schemas) missed.
- This is a critical fix — premium exports silently returned nothing on production nodes.

PR #8001, opened 2026-07-16
Pr/fix 7889 wallet balance error handling

IcanBENCHurCAT:

- [ ] Add a tier label: `BCOS-L1` or `BCOS-L2` (also accepted: `bcos:l1`, `bcos:l2`)
- [ ] If adding new code files, include SPDX header near the top (example: `# SPDX-License-Identifier: MIT`)
- [ ] Provide test evidence (commands + output or screenshots)

## What Changed

- ...

## Testing / Evidence

- ...

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -76,7 +76,7 @@ if [ "$DRY_RUN" -eq 1 ]; then
             FINGERPRINT_PATH="linux/fingerprint_checks.py"
             ;;
         Darwin)
-            MINER_PATH="macos/rustchain_mac_miner_v2.4.py"
+            MINER_PATH="macos/rustchain_mac_miner_v2.5.py"
             FINGERPRINT_PATH="macos/fingerprint_checks.py"
             ;;
         *)      echo -e "${RED}Unsupported OS: $OS${NC}"; exit 1 ;;
@@ -127,7 +127,7 @@ case "$OS" in
         ;;
     Darwin)
         echo "  OS: macOS"
-        MINER_PATH="macos/rustchain_mac_miner_v2.4.py"
+        MINER_PATH="macos/rustchain_mac_miner_v2.5.py"
         FINGERPRINT_PATH="macos/fingerprint_checks.py"
         ;;
     *)      echo -e "${RED}  Unsupported OS: $OS${NC}"; exit 1 ;;
--- a/node/hardware_fingerprint_replay.py
+++ b/node/hardware_fingerprint_replay.py
@@ -25,13 +25,113 @@ import os
 import sqlite3
 import time
 from contextlib import closing
-from typing import Dict, List, Tuple, Optional, Any
-from collections import defaultdict
-
+from typing import Dict, List, Tuple, Optional
 # Configuration constants
 REPLAY_WINDOW_SECONDS = 300  # 5 minutes - fingerprints expire after this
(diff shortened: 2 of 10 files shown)


elevasyncsolutions-jpg:
## Review of PR #8001 — Wallet Balance Error Handling

### Observation 1: `BalanceResponse::validate` is a no-op (`client.rs`)

The new `validate` method on `BalanceResponse` has a docstring that says it returns `Err(EmptyBalance)` when the balance is zero, but the implementation just sets `self.address` and returns `Ok(self)`. The `EmptyBalance` error variant you added to `error.rs` is never actually returned by any code path. This means callers who `match` on `EmptyBalance` will never hit that branch — it's dead code right now.

Suggested fix: add the actual zero-balance check inside `validate`:
```rust
pub fn validate(mut self, address: &str) -> Result<BalanceResponse> {
    self.address = address.to_string();
    if self.balance == 0 && self.error.is_none() {
        return Err(WalletError::EmptyBalance);
    }
    Ok(self)
}
```

### Observation 2: `get_nonce` error propagation breaks callers (`rtc_wallet.rs`)

Changing `client.get_nonce(&from_address).await.unwrap_or(0)` to `.await?` is the right direction, but it changes the error semantics for `cmd_send`. Previously, a network timeout would silently use nonce 0 and likely fail later with a confusing "nonce too low" error. Now it fails immediately — which is better — but the `?` propagation will bubble up a `WalletError::Network` string that may not be user-friendly. Consider mapping it to a more specific error like `WalletError::Transaction(format!("Failed to fetch nonce for {}: {}", from_address, e))` so the user knows *which* address failed.

### Observation 3: `_ascii_version` doesn't handle emoji without trailing spaces (`cli.py`)

The `_ascii_version` function strips emoji by matching `✅️` + "  " (two spaces). If the CLI output ever uses a single space after the emoji (e.g. `✅ Wallet created!` with one space), the regex won't match and the raw emoji will pass through to stdout, re-triggering the `UnicodeEncodeError` on Windows. Consider using a more lenient regex like `re.sub(r'^[✅⚠️❌📋📤]\s*\s?', replacement + ' ', text)` to handle any whitespace after the emoji.

### Positive notes

- The bucket-width quantization constants in `hardware_fingerprint_replay.py` are well-documented with clear rationale tied to Issue #7991. The 500ns bucket width for timing fields seems reasonable given the 10-20% variance mentioned.
- The badge-generator.html switch from `innerHTML` to safe DOM APIs (`createElement`/`textContent`) is a solid XSS hardening move.
- The install.sh version bump from v2.4 to v2.5 with a corresponding test that verifies the file exists on disk is good defensive testing.

Scottcjn:
Good intent — not masking a dead-node lookup as nonce 0 is a real fix (nonce-reuse risk). But CI (Clippy/Rustfmt/Test) genuinely fails here, and since this touches `rustchain-wallet` (which IS in CI), these are the diff's own failures: a dead `EmptyBalance` path, `validate(mut self)` that's a no-op, and formatting. Please get Clippy+fmt+tests green and re-push.

Scottcjn:
Closing as superseded: `scripts/check_balance.sh` on `main` now has distinct exit codes (1 usage, 2 network, 3 bad response), curl error surfacing, and jq-parsed output — the behaviour this PR set out to add. #8318 carries the one remaining improvement (amount validation) and is under review. Thank you for the work; sorry it sat so long. — Sophia

PR #8008, opened 2026-07-18
fix(node-health): total_miners sums replica counts and inflates the figure Nx

Vyacheslav-Tomashevskiy:
## The bug

`NodeHealthMonitor.get_network_health` reports the headline **Total miners** figure by summing each online node's count:

```python
online = [s for s in statuses if s.status != "offline"]
total_miners = sum(s.miners or 0 for s in online)   # <-- inflates Nx
```

But the monitored nodes (`DEFAULT_NODES`, three `:8088/:8099` hosts) are **replicas of one shared ledger**. Each `/status` returns the *network-wide* active-miner count (`miners` / `active_miners` / `miner_count`), not a disjoint per-node slice. The sibling tools in this repo prove the intent — they treat any divergence between nodes as an anomaly:

- `monitoring/ledger_verify.py` → `miner_count_match` / `MINER COUNT MISMATCH`
- `tools/node_sync_validator.py` → `miner_count_mismatch`

So in the healthy case every online node agrees, and summing multiplies the true count by the number of online nodes.

**Concrete input → wrong output:** 3 online nodes each reporting `miners = 150` (a network with 150 miners) → buggy `total_miners = 150 + 150 + 150 = 450`; correct = `150`. The operator's table (`print_table`) and `--json` output always show a fabricated 2-3x miner population.

## The fix

Take the value the replicas agree on instead of summing. `max` is used so a node that is mid-sync and momentarily reporting fewer miners doesn't drag the figure down:

```python
miner_counts = [s.miners for s in online if s.miners is not None]
total_miners = max(miner_counts) if miner_counts else 0
```

## Test

Added `tools/test_node_health_total_miners.py`: 3 replicas → 150 (not 450); one lagging node (149) → 150; one offline → 150 with `nodes_online=2`; no miner data → 0 (no crash). All pass.

### DIFF
--- a/tools/node_health_monitor.py
+++ b/tools/node_health_monitor.py
@@ -158,7 +158,14 @@ class NodeHealthMonitor:
 
         online = [s for s in statuses if s.status != "offline"]
         nodes_online = len(online)
-        total_miners = sum(s.miners or 0 for s in online)
+        # The attestation nodes are replicas of one shared ledger, so each online
+        # node reports the SAME network-wide miner count (the sibling tools
+        # ledger_verify / node_sync_validator flag any divergence as a mismatch).
+        # Summing per-node counts would multiply the true figure by the number of
+        # online nodes, so take the agreed value (max is robust to a node that is
+        # mid-sync and momentarily reporting fewer miners).
+        miner_counts = [s.miners for s in online if s.miners is not None]
+        total_miners = max(miner_counts) if miner_counts else 0
 
         epochs = {s.epoch for s in online if s.epoch is not None}
         consensus_ok = len(epochs) <= 1  # 0 or 1 distinct epoch → consensus holds
--- a/tools/test_node_health_total_miners.py
+++ b/tools/test_node_health_total_miners.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python3
+"""Regression test: NetworkHealth.total_miners must not multiply by node count.
+
+The attestation nodes are replicas of one shared ledger, so every online node
+reports the SAME network-wide miner count (ledger_verify / node_sync_validator
+flag any divergence as a mismatch). total_miners must therefore be the agreed
+value, not the sum across nodes.
+"""
+import os
+import sys
+
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+from tools.node_health_monitor import NodeHealthMonitor, NodeStatus
+
+
+def _node(url, epoch, miners, status="online"):
+    return NodeStatus(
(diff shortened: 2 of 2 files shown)


Scottcjn:
Right diagnosis (replicas of one ledger shouldn't be summed). **Held pending two fixes:** (1) CI is red because the pre-existing `test_health_monitor.py::test_all_healthy` still asserts the old summed value (6 != 13) — reconcile it. (2) `max(miner_counts)` silently trusts the highest divergent report, so a stale/over-reporting node inflates the count; detect/handle divergence rather than always taking max. Address both and it pays. — Sophia

Scottcjn:
Your diagnosis is right, and the one failing check is not a defect in the fix. Details so you can clear it in one push.

`tests/test_health_monitor.py::TestGetNetworkHealth::test_all_healthy` fails with `AssertionError: 6 != 13`. It passes on `main` (21 passed) and fails on this branch. That is expected, because **the existing test encodes the bug you are fixing**:

```python
statuses = [
    _online_status("http://a:8088", epoch=5, miners=4),
    _online_status("http://b:8088", epoch=5, miners=3),
    _online_status("http://c:8099", epoch=5, miners=6),
]
self.assertEqual(health.total_miners, 13)   # 4 + 3 + 6
```

Three nodes in one network each report that network's miner count, so summing them triple-counts the same miners. `13` is the inflated figure, `6` is correct. The assertion needs updating to `6` as part of this PR, with a comment saying why, otherwise the next person reads it as a regression rather than the fix landing.

Two nearby cases are worth a look while you are in there: `test_one_offline_node`, and any other `total_miners` assertion built the same way.

This is the second instance today of a test in this repo asserting the buggy value, so it is a pattern here rather than a mistake on your part.

Update that assertion and push and I will merge. Note that `main` has moved a long way underneath you, since 33 of your PRs merged today, so a rebase first will save you a round trip.

Scottcjn:
Merging — verified still present on `main` (line 225) before merging, so this is live.

You are right about the topology: the attestation nodes are replicas over one shared ledger, so each online node reports the same network-wide miner count, and summing them multiplies the true figure by the number of online nodes. Using `max` rather than `any` is the better choice for the reason you give — a node mid-sync momentarily reporting fewer miners should not drag the network figure down — and divergence is already the sibling tools' job to flag, so this function does not need to police it.

Context you could not have had: this is the third distinct defect found in this monitor. Two others were fixed recently where a liveness check and a probed endpoint were each wrong in ways that cancelled, so the monitor looked healthy while measuring nothing. A tool that reports on everything else gets audited last. Thank you for the regression test.

PR #8009, opened 2026-07-18
fix(gpu-attest): tensor-core gate string-compares CC, dropping channel for CC>=10

Vyacheslav-Tomashevskiy:
## The bug

`get_gpu_attestation_payload` (node/gpu_attestation.py) gates the Channel 8f **Tensor Core** fingerprint on the GPU's compute capability:

```python
if fp.compute_capability >= "7.0":
    tc_fp = run_tensor_core_fingerprint(...)
```

`compute_capability` is a **string** (`GPUFingerprint.compute_capability: str`, built as `f"{major}.{minor}"` — "7.0", "8.0", "9.0", "10.0", "12.0"). `>=` on strings compares **lexically**, so:

```
"10.0" >= "7.0"  ->  False   ('1' < '7')
"12.0" >= "7.0"  ->  False
```

The check is meant to include Volta and newer (all of which have tensor cores). It instead **silently drops the tensor-core channel for exactly the newest tensor-core GPUs** — Blackwell data-center (CC 10.0) and consumer Blackwell / RTX-50 (CC 12.0) — while Volta..Hopper (7.0-9.x) pass. Those miners' attestation payloads are missing Channel 8f entirely.

## The fix

Parse the value numerically via a small helper (so it's unit-testable without a GPU):

```python
def _supports_tensor_cores(compute_capability) -> bool:
    try:
        return float(compute_capability) >= 7.0
    except (TypeError, ValueError):
        return False
...
if _supports_tensor_cores(fp.compute_capability):
```

## Test

Added `node/tests/test_gpu_tensor_core_cc_compare.py` (torch-free — `import torch` in gpu_attestation is function-local): Volta..Hopper supported; CC 10.0 / 12.0 now supported (and asserts the old lexical compare returned False for them); pre-Volta and malformed values rejected without crashing. 4 pass.

### DIFF
--- a/node/gpu_attestation.py
+++ b/node/gpu_attestation.py
@@ -4,6 +4,21 @@ import os
 import subprocess
 from typing import Dict, Tuple
 
+def _supports_tensor_cores(compute_capability) -> bool:
+    """Tensor cores exist on CUDA compute capability >= 7.0 (Volta and newer).
+
+    compute_capability is a string like "7.0", "8.0", "9.0", "10.0", "12.0"
+    (built as f"{major}.{minor}"). It must be compared numerically: a raw
+    string comparison mis-ranks multi-digit majors — "10.0" >= "7.0" is False
+    because '1' < '7' lexically — which silently dropped the tensor-core channel
+    for exactly the newest tensor-core GPUs (Blackwell CC 10.0 / 12.0).
+    """
+    try:
+        return float(compute_capability) >= 7.0
+    except (TypeError, ValueError):
+        return False
+
+
 def validate_gpu_fingerprint(fingerprint_data: Dict) -> Tuple[bool, str]:
     """
     Server-side validation of GPU fingerprint data (RIP-0308).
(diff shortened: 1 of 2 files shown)


Scottcjn:
Verified against `main`. `node/gpu_attestation.py:73` reads `if fp.compute_capability >= "7.0":` and `compute_capability` is built as `f"{major}.{minor}"`, so the comparison is lexical:

```
"10.0" >= "7.0"  ->  False
"12.0" >= "7.0"  ->  False
"8.0"  >= "7.0"  ->  True
```

The failure mode is the part worth naming: it is not that the gate is loose, it is that it is **inverted for exactly the newest hardware**. Volta through Hopper pass, and Blackwell at CC 10.0 and 12.0 silently drop Channel 8f. A miner on the best tensor-core GPU available submits an attestation quietly missing the channel meant to prove it, and nothing errors.

Pulling the comparison into `_supports_tensor_cores()` also makes it testable without a GPU, which is the right call for a path that otherwise only exercises on specific silicon. Merging.

PR #8010, opened 2026-07-18
fix(airdrop): canonicalize Base wallet case so one wallet cannot double-claim

Vyacheslav-Tomashevskiy:
## Bug: same Base (EVM) wallet can claim the airdrop multiple times by varying hex case

`node/airdrop_v2.py` enforces **one airdrop per GitHub account and per wallet** — `_has_claimed()` and `UNIQUE(github_username, wallet_address, chain)`. Only the GitHub half is canonicalized (`_normalize_github_username` casefolds it); `wallet_address` is compared as a **raw string** everywhere.

Base is EVM, where addresses are **case-insensitive on-chain**: `0xAbCd…`, `0xabcd…` and `0xABCD…` are the *same account* and `eth_getBalance` (`_check_base_wallet`) returns the identical balance for every casing — so every casing passes the anti-Sybil check. But `_has_claimed` and the `UNIQUE` index treat each casing as a *different* wallet, so the wallet-dedup leg is void for the entire Base chain.

### Impact
N aged, PR-merged GitHub accounts can all funnel the airdrop to **one physical Base wallet** by only changing hex case, and the global Base allocation is debited each time. Even a single honest user who claims with an EIP-55 checksummed address and re-submits the lowercased form gets a second allocation.

### Repro (deterministic, no network — `skip_antisybil=True`)
```python
a = AirdropV2(":memory:")
a.claim_airdrop("alice", "0xAbCdEf...01", "base", "contributor", skip_antisybil=True)  # ok
a.claim_airdrop("bob",   "0xabcdef...01", "base", "contributor", skip_antisybil=True)  # main: ALSO ok (bug)
# main -> base claimed_wrtc == 100 (same wallet paid twice)
```

### Fix
Add `_normalize_wallet(address, chain)` and canonicalize `wallet_address` at the entry of both `check_eligibility` and `claim_airdrop` (lowercase for `base`; **Solana base58 left untouched — it is genuinely case-sensitive**), so dedup, the UNIQUE index, the generated claim id and the RPC balance check all use one canonical form.

### Tests
- New `node/test_airdrop_wallet_case_dedup.py`: fails on main (Base wallet double-claims, 100 wRTC), passes with fix (50 wRTC, one claim); Solana distinct-case wallets stay independently claimable.
- Existing `node/test_airdrop_v2.py` — **31 passed**, no regressions.

Claiming under rustchain-bounties#71. Payout RTC: `RTCd1554f0f35576faf01d386a6be1c947f560dd0b7`

### DIFF
--- a/node/airdrop_v2.py
+++ b/node/airdrop_v2.py
@@ -319,6 +319,24 @@ class AirdropV2:
         return (github_username or "").strip().casefold()
 
     @staticmethod
+    def _normalize_wallet(wallet_address: str, chain: str) -> str:
+        """Canonicalize a wallet address for de-duplication.
+
+        Base (EVM) addresses are case-insensitive on-chain: the same account
+        can be written checksummed (EIP-55), all-lower, or all-upper, and
+        eth_getBalance returns the identical balance for every casing. The
+        one-airdrop-per-wallet invariant must therefore compare a canonical
+        (lowercase) form, otherwise the same physical Base wallet slips past
+        _has_claimed / UNIQUE(github_username, wallet_address, chain) simply by
+        varying hex case. Solana uses base58, which IS case-sensitive, so it is
+        left untouched.
+        """
+        wallet_address = (wallet_address or "").strip()
+        if chain and chain.lower() == "base":
+            return wallet_address.lower()
+        return wallet_address
+
+    @staticmethod
     def _is_valid_github_username(github_username: str) -> bool:
         return bool(GITHUB_USERNAME_RE.fullmatch(github_username))
 
@@ -355,6 +373,7 @@ class AirdropV2:
                 eligible=False,
                 reason=f"Unsupported chain: {chain}. Must be 'solana' or 'base'",
             )
+        wallet_address = self._normalize_wallet(wallet_address, chain_lower)
 
         checks = {}
 
@@ -794,6 +813,7 @@ class AirdropV2:
         if not self._is_valid_github_username(github_username):
             return False, "Invalid GitHub username", None
         chain_lower = chain.lower()
+        wallet_address = self._normalize_wallet(wallet_address, chain_lower)
 
         if self._has_claimed(github_username, wallet_address, chain_lower):
             return False, "Claim already exists for this GitHub account or wallet", None
--- a/node/test_airdrop_wallet_case_dedup.py
+++ b/node/test_airdrop_wallet_case_dedup.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+"""
+Regression tests for the airdrop wallet-address de-duplication invariant.
+
+Base (EVM) addresses are case-insensitive on-chain: 0xAbCd..., 0xabcd... and
+0xABCD... are the *same* account and eth_getBalance returns the same balance
+for every casing. The airdrop enforces "one claim per wallet" via _has_claimed
+and UNIQUE(github_username, wallet_address, chain). Before the fix, the wallet
+half of that invariant compared raw strings, so the same physical Base wallet
+could collect the airdrop multiple times (once per GitHub identity) simply by
+varying hex case — draining the Base allocation.
+
+Solana uses base58, which IS case-sensitive, so different casings there are
+genuinely different wallets and must stay allowed.
+"""
+import unittest
+
+from airdrop_v2 import AirdropV2
+
+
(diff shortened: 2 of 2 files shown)


Vyacheslav-Tomashevskiy:
/claim rustchain-bounties#71 — verified bug: same Base (EVM) wallet double-claims the airdrop by varying hex case (wallet-dedup invariant void for the whole Base chain). Deterministic regression test included; existing suite 31/31 green. Payout RTC: RTCd1554f0f35576faf01d386a6be1c947f560dd0b7

jjb9707 (approved):
## Substantive review (Bounty #73 candidate)

**Verified scope:** `node/airdrop_v2.py` adds chain-aware wallet canonicalization and `node/test_airdrop_wallet_case_dedup.py` adds three regression paths. I compared the patch with the actual schema, `_has_claimed`, eligibility, claim-record, and allocation-update paths at head `eb2641e64624fbbdf4eb07053c6f4410997822d5`.

### Strengths

1. The threat model is correct and narrowly scoped: Base/EVM addresses are case-insensitive, while Solana base58 addresses are case-sensitive. Lowercasing only `chain == "base"` avoids corrupting Solana identities.
2. Canonicalization occurs before both `_has_claimed()` call sites and before the `ClaimRecord`/claim ID are built. Consequently the lookup, stored wallet, RPC validation input, generated ID, and future comparisons all share one representation rather than merely normalizing one query.
3. The regression suite covers the two externally visible paths (`check_eligibility` and `claim_airdrop`) and explicitly protects the Solana non-regression. The allocation assertion also proves rejected case variants do not debit the pool.

### Substantive findings

1. **The existing SQL uniqueness constraint does not independently enforce one-wallet/one-account semantics.** `UNIQUE(github_username, wallet_address, chain)` only rejects the same *pair*; `_has_claimed()` is what enforces the OR rule. This patch correctly closes the reported sequential case-variant exploit, but two concurrent requests for different GitHub users targeting the same canonical Base wallet can still both pass `_has_claimed()` before either insert, and both distinct tuples can commit. That is a pre-existing race, not introduced here. A follow-up migration should add separate unique indexes for active claims (or serialize the check/insert with an immediate transaction) so the database owns the invariant.
2. **Legacy mixed-case rows are not migrated.** New writes are canonical, and new checks catch an exact stored casing, but a database that already contains `0xAbCd...` will not match a new canonical lowercase query. Before rollout, normalize existing Base rows and resolve any collisions, or temporarily compare `lower(wallet_address)` for Base until the migration is complete. This matters because the patch changes application semantics without rewriting historical state.
3. **Validation ordering is acceptable but worth preserving deliberately.** `_normalize_wallet()` runs after the supported-chain guard in `check_eligibility`, while `claim_airdrop()` normalizes immediately after `chain.lower()`. Since unsupported chains are not lowercased by the helper, there is no cross-chain corruption; however, any future direct storage path must call the same helper or the invariant can regress. Centralizing canonicalization at a typed wallet boundary would make that harder.

### Verification

- All 12 reported checks on the reviewed head completed successfully, including `test`, `BCOS Tier Gate`, `BCOS v2 Engine Scan`, and the review-tier gate.
- Read-through confirms the canonical value is passed into `_check_wallet`, stored in `airdrop_claims`, included in `claim_id`, and used by `_has_claimed`.
- The new tests are deterministic (`:memory:` + `skip_antisybil=True`) and assert both rejection behavior and allocation accounting.

**Verdict: APPROVE** — the patch fixes the demonstrated sequential Base-address casing bypass without changing Solana semantics. The legacy-row migration and database-level concurrency invariant should be tracked as follow-ups; neither invalidates the focused fix for fresh canonical writes.

Reviewer: @jjb9707

Scottcjn:
Closing as already fixed, not as rejected. The bug you describe is real and it is now fixed on `main`.

Your own #7974 landed the same finding a little earlier today and carried the more complete implementation, so this branch conflicts with it. `node/airdrop_v2.py` now has `_normalize_wallet_address()`, applied in `_has_claimed()`, `check_eligibility()` and `claim()`, folding EVM addresses to lowercase while leaving base58 Solana addresses byte-exact. That last distinction is the part that matters and you got it right in both PRs.

Worth stating plainly since you filed it twice: the double-claim vector is closed, and the credit for closing it is yours. Nothing here is lost.

If you think this branch still covers a case #7974 misses, say so and reopen, and I will look again.

PR #8011, opened 2026-07-18
fix(bcos): reject duplicate cert_id instead of silently overwriting + erasing anchor

Vyacheslav-Tomashevskiy:
## What

`bcos_attest()` (`node/bcos_routes.py`) stored the certificate with **`INSERT OR REPLACE`**, but `cert_id` is `UNIQUE` (schema line 184) and the handler carries an `except sqlite3.IntegrityError -> 409 "Certificate ... already exists"` guard.

`INSERT OR REPLACE` resolves a UNIQUE conflict with a silent **DELETE + INSERT**, so it never raises `IntegrityError`. The 409 guard was therefore **dead code**, and a second attest for a known `cert_id` **silently overwrote** `repo` / `tier` / `trust_score` / `reviewer` / `report`.

The REPLACE column list also omits `anchor_tx` and passes `anchored_epoch=None`, so re-attesting a **ledger-anchored** certificate wiped its `anchor_tx` back to `NULL`. That breaks the documented idempotency of `/api/v1/bcos/anchor` (an already-anchored cert should return its existing tx hash, not write a second ledger row): after the wipe, a re-anchor writes a **second ledger row with a fresh tx hash** for the same certificate.

Auth passes on `is_admin OR valid Ed25519 signature` over a **caller-supplied** `signer_pubkey`, so any party can self-sign a report they authored and overwrite a known public `cert_id`.

## Repro (deterministic)

1. attest `BCOS-x` (repo `Scottcjn/Rustchain`, L2, trust 90) -> anchor it (`anchor_tx` set).
2. re-attest the same `cert_id` with `repo attacker/evil`, trust 100.
3. **Before fix:** no `IntegrityError`; row becomes `{repo: attacker/evil, trust_score: 100, anchored_epoch: None, anchor_tx: None}` -- anchor erased, trust forged, 409 never fires.

## Fix

Plain `INSERT` (one word). The duplicate now raises `IntegrityError`, the existing 409 guard fires, and the original record + its anchor are preserved. `cert_id` is meant to be immutable once issued -- exactly what the "already exists" 409 already intended.

Same overwrite class the repo already treats as a real bug in `node/tests/test_attestation_overwrite_reward_loss.py` (that one is `miner_attest_recent`; this is the untouched `bcos_attestations` table).

## Tests

- Adds `node/tests/test_bcos_reattest_preserves_anchor.py` -- **fails on old code, passes on the fix** (mutation-checked).
- Existing `node/tests/test_bcos_routes.py` (8 tests) still green.

RTC reward address: `RTCd1554f0f35576faf01d386a6be1c947f560dd0b7`

### DIFF
--- a/node/bcos_routes.py
+++ b/node/bcos_routes.py
@@ -355,8 +355,14 @@ def bcos_attest():
             # ledger via /api/v1/bcos/anchor.
             attested_epoch, _slot = _node_epoch_slot()
 
+            # Plain INSERT (not INSERT OR REPLACE): cert_id is UNIQUE and
+            # immutable once issued. A duplicate must raise IntegrityError so
+            # the 409 guard below fires, rather than silently DELETE+INSERT the
+            # row. INSERT OR REPLACE would also drop anchor_tx/anchored_epoch
+            # (not in this column list), wiping a ledger-anchored certificate's
+            # anchor back to NULL and breaking /api/v1/bcos/anchor idempotency.
             conn.execute("""
-                INSERT OR REPLACE INTO bcos_attestations
+                INSERT INTO bcos_attestations
                 (cert_id, commitment, repo, commit_sha, tier, trust_score,
                  reviewer, report_json, signature, signer_pubkey,
                  attested_epoch, anchored_epoch, created_at)
--- a/node/tests/test_bcos_reattest_preserves_anchor.py
+++ b/node/tests/test_bcos_reattest_preserves_anchor.py
@@ -1,0 +1,117 @@
+# SPDX-License-Identifier: MIT
+"""
+Test: re-attesting an existing BCOS cert_id must not overwrite/erase the record.
+
+Bug:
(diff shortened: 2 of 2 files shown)


Scottcjn:
Verified against `main`. `node/bcos_routes.py:359` uses `INSERT OR REPLACE`, `cert_id` is `UNIQUE` at schema line 184, and the `except sqlite3.IntegrityError` returning 409 sits at lines 384-385. `INSERT OR REPLACE` resolves the conflict by DELETE plus INSERT, so `IntegrityError` never fires and that 409 branch was unreachable.

The consequence you traced past the dead guard is the more serious half: because the REPLACE column list omits `anchor_tx`, re-attesting an already-anchored certificate reset it to NULL, so a subsequent anchor wrote a second ledger row for a certificate that was already on the ledger. A dead error branch is a bug; a dead error branch that silently un-anchors a ledger record is a different class of problem.

Merging.

PR #8056, opened 2026-07-25
Health CLI: fail the run when the chain tip check fails

### LINKED ISSUE #8055: [BUG] rustchain-health exits 0 when the chain tip check fails
## Summary

`tools/rustchain-health.py` checks and displays `/headers/tip`, but omits that result from both overall-health calculations. A node whose chain-tip endpoint is unavailable can therefore be reported as **ALL SYSTEMS OPERATIONAL**, and the CLI exits `0`.

## Reproduction

Tested on current `main` at `d684e61`. I ran the CLI against a local HTTP server returning:

- `200 {"ok": true}` for `/health`
- `200 {"epoch": 42, "slot": 7}` for `/epoch`
- `200 {"miners": [{"miner_id": "miner-1"}]}` for `/api/miners`
- `404` for `/headers/tip`

Command:

```bash
python3 tools/rustchain-health.py --url http://127.0.0.1:<port> --no-color
```

Relevant output:

```text
Chain Tip       unknown  7 ms
Error: HTTP Error 404: tip unavailable
...
STATUS: ALL SYSTEMS OPERATIONAL
```

The process exit code is also `0`, with no stderr output.

## Root cause

There are two separate `all_ok` expressions and both omit `snapshot["tip"]`:

- `tools/rustchain-health.py:276-278` controls the rendered overall status.
- `tools/rustchain-health.py:358-361` controls the process exit code.

Each expression checks health, epoch, and miners only, even though the CLI description explicitly includes chain-tip monitoring and `collect()` always calls `check_tip()`.

## Expected behavior

A failed or malformed chain-tip response should make the overall status unhealthy and produce a non-zero exit code. Ideally the overall predicate should be shared between rendering and `run_once()` so the text and exit status cannot drift. A regression test should cover a snapshot where the other three checks pass but `tip.reachable` is false or `height` is missing.

## Impact

Monitoring, cron jobs, and CI that rely on this command can miss a broken or unavailable chain-tip endpoint and treat a node with incomplete chain state as operational.

Scottcjn:
## The bug

`tools/rustchain-health.py` runs `check_tip()` and stores the result in the snapshot, but both `all_ok` expressions ignored it: the STATUS line in `render()` (old line 276) and the exit code in `run_once()` (old line 358) were built from health + epoch + miners reachability only. With the chain tip check failing, the tool still printed "ALL SYSTEMS OPERATIONAL" and exited 0, so any monitor watching the exit code never fired. A monitor that exits 0 on failure is worse than no monitor.

Fixes #8055. Thanks to @mumuzhong3 for the report.

## The fix

Instead of adding `tip` to both copies of `all_ok`, the verdict is now one shared set of predicates: `health_ok`, `epoch_ok`, `miners_ok`, `tip_verdict`, and `overall_ok(snapshot)`. Both the display path and the exit-code path call `overall_ok`, so they cannot drift apart again. That drift is exactly how this bug happened: the two copies had already diverged once (the display required epoch content, the exit path only required epoch reachability).

Details:

- `tip_verdict` returns True / False / None. None means the node answered 404/405/501 for `/headers/tip`, i.e. the endpoint is not implemented there. That renders as a yellow "n/a (endpoint not supported)" and does not count against the verdict, so nodes without that route do not raise false alarms. An unreachable endpoint or a body with no identifiable tip is a real failure.
- `check_tip` now also accepts `slot` as the tip position. RIP-200 nodes return `{"miner": ..., "slot": ..., "tip_age": ...}` from `/headers/tip`, so before this change a perfectly healthy production node showed its tip as "unknown".
- Sibling of the reported bug: `/epoch` reachable but returning no `epoch` field used to draw a red "unavailable" row while the overall status stayed green and the exit code stayed 0. It now fails the verdict, matching the row. Miner count 0 remains a yellow warning, not a failure, same as before.
- `fetch()` now returns the HTTP status code so the 404-vs-network-error distinction is possible.

## Verified

- New `tests/test_rustchain_health_verdict.py` (7 tests) drives the real CLI entrypoint with a mocked `fetch`: failing tip exits non-zero and does not print "ALL SYSTEMS OPERATIONAL"; garbage tip body fails; 404 tip does not fail; empty epoch fails; healthy path exits 0. All 7 pass on this branch and fail on main.
- Reproduced the original bug on main: dead tip, output said "ALL SYSTEMS OPERATIONAL", exit code 0.
- Ran the fixed tool against the live node at 50.28.86.131: all four rows green, tip resolves to #2941199 (was "unknown" before the `slot` fix), STATUS ALL SYSTEMS OPERATIONAL, exit 0.
- Unreachable node exits 1. `ruff check` clean on both files.

### DIFF
--- a/tests/test_rustchain_health_verdict.py
+++ b/tests/test_rustchain_health_verdict.py
@@ -1,0 +1,150 @@
+"""Regression tests for tools/rustchain-health.py overall verdict.
+
+Issue #8055: the tool ran check_tip() and stored the result in the
+snapshot, but both all_ok expressions (the STATUS line in render() and
+the exit code in run_once()) ignored it. A dead chain tip still printed
+"ALL SYSTEMS OPERATIONAL" and exited 0, silencing any monitor built on
+top of this tool.
+
+These tests drive the real CLI entrypoint (main) with a mocked fetch()
+so the whole path is exercised: collect -> render -> exit code.
+"""
+
+import importlib.util
+import sys
+from pathlib import Path
+
+import pytest
+
+_TOOL_PATH = Path(__file__).parent.parent / "tools" / "rustchain-health.py"
+
+
+def _load_health_module():
+    name = "rustchain_health_cli"
+    if name in sys.modules:
+        return sys.modules[name]
+    spec = importlib.util.spec_from_file_location(name, str(_TOOL_PATH))
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[name] = module
+    spec.loader.exec_module(module)
+    return module
+
+
+@pytest.fixture()
+def health_mod(monkeypatch):
+    mod = _load_health_module()
+    monkeypatch.setattr(mod, "_COLOR", False)
+    return mod
+
+
+HEALTH_BODY = {"ok": True, "version": "2.2.1-rip200", "uptime_s": 1000,
+               "db_rw": True, "tip_age_slots": 0}
+EPOCH_BODY = {"epoch": 424, "slot": 61000, "epoch_pot": 1.5,
+              "enrolled_miners": 8}
+MINERS_BODY = [{"miner_id": "dual-g4-125"}, {"miner_id": "sophia-nas-c4130"}]
+TIP_BODY = {"miner": "sophia-nas-c4130", "slot": 2941199,
+            "signature_prefix": "b3780cff190178456a0f"}
+
+
+def _fake_fetch(responses):
+    """Build a fetch() replacement keyed on URL suffix.
+
+    responses maps an endpoint suffix to (ok, data, status).
+    """
+    def fetch(url, timeout=8):
+        for suffix, (ok, data, status) in responses.items():
+            if url.endswith(suffix):
+                return ok, data, 1.0, status
+        raise AssertionError(f"unexpected URL fetched: {url}")
+    return fetch
+
+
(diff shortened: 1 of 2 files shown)


Scottcjn:
Closing this in favour of #8066, which is now merged. The short version is that I was wrong about my own PR and the evidence says so.

I opened this 47 minutes before @waterWang's #8066 and argued it was the better fix because it factors the duplicated verdict into one shared predicate, handles a tip endpoint that is not implemented, and corrects a field-mapping bug where the node returns `slot` while the tool looked for `height`. Two independent reviewers agreed with that reasoning.

They were reasoning from my description, and my description was wrong on the part that matters: **this branch fails CI.** It breaks 7 existing tests, because it changed `fetch()` from a three-value return to a four-value one and the rest of the tool and its tests already depend on that contract. I then tried to restructure around it and made it worse, 11 failures. #8066 passes all 12 checks with a four-line diff.

A fix that breaks the build is not the more complete fix, whatever its intentions. First-mergeable is content-neutral, and on the evidence #8066 is the one that actually merges.

The separate findings in here are still real and I will bring them back as their own PR without touching the `fetch()` signature: the two `all_ok` copies had already drifted once (a red epoch row could coexist with a green overall status), and the `slot` versus `height` mapping means a healthy node reports an unknown tip. Those deserve their own change with their own tests.

Thanks @waterWang, and thanks @mumuzhong3 for the report that started it.

PR #8060, opened 2026-07-25
fix: /health tip_age_slots now computes real tip age instead of returning 0 [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #8057: /health tip_age_slots always returns 0 and never measures tip age
## Summary

`/health` reports `"tip_age_slots": 0` on a node whose newest block header is **234 days old**. The field is not measuring anything.

## The code

`rustchain_v2_integrated_v2.2.1_rip200.py`, `_tip_age_slots()`:

```python
def _tip_age_slots():
    """Check tip freshness - query DB directly to avoid Response object"""
    try:
        with sqlite3.connect(DB_PATH, timeout=3) as db:
            row = db.execute("SELECT slot FROM headers ORDER BY slot DESC LIMIT 1").fetchone()
        return 0 if row else None
    except Exception:
        return None
```

It selects the newest header and then discards it, returning a hardcoded `0` if any row exists at all. The function name says age, the docstring says freshness, and neither is computed. An empty headers table returns `None`; a table whose newest entry is from last year returns `0`, identical to a table updated one second ago.

## Observed on node 1 (50.28.86.131)

```
GET /health       -> {"tip_age_slots": 0, "ok": true, ...}
GET /headers/tip  -> {"slot": 2941199, "tip_age": 20280879, "miner": "sophia-nas-c4130"}
```

`tip_age` of 20,280,879 seconds is about 234 days. The two endpoints disagree because only one of them measures anything.

Database state: `SELECT COUNT(*), MIN(slot), MAX(slot) FROM headers` returns `224 | 35 | 2941199`. 224 headers total.

Note the max slot is on the pre-fix raw-timestamp convention (`time.time() // 600`) rather than the genesis-relative `current_slot()` the node uses now, which is consistent with the newest header dating to around the production chain launch.

## Impact

Low severity in isolation, because `ok` is not gated on this value:

```python
ok = ok_db and (age_h is None or age_h < 36)
```

The harm is that the field is load-bearing for humans and dashboards. Anyone reading `/health` to answer "is the chain advancing" gets a confident `0` that means "a headers row exists." This is the same failure shape as #8055, where the health CLI printed ALL SYSTEMS OPERATIONAL and exited 0 while the tip check failed. A health surface that cannot report unhealthy is worse than no health surface, because it suppresses the question.

## Suggested fix

Compute the real age and return it, keeping it out of the `ok` boolean so monitoring semantics do not change without a deliberate decision:

- Select `ts` alongside `slot` and derive age from wall clock, or derive slot age against `current_slot()`. Pick one and make the field name match the unit it returns.
- If the name stays `tip_age_slots`, return slots. If it returns seconds, rename it.
- Decide separately, and explicitly, whether a stale tip should fail `ok`. That is a policy call about whether header production is expected to be continuous on this network, and it should not be smuggled in as part of a bug fix.

## Deliberately not patched on the running node

Changing this value from a constant `0` to a real number will change what any external monitoring sees, potentially turning dashboards red. That is a maintainer decision, not a drive-by fix.

Related: #8055 and PR #8056, same class of defect in the health CLI.

waterWang:
## Problem

`/health` reports `"tip_age_slots": 0` on a node whose newest block header is **234 days old**. The `_tip_age_slots()` function selects the latest header slot but then discards it and returns a hardcoded `0` whenever any row exists.

```python
def _tip_age_slots():
    try:
        with sqlite3.connect(DB_PATH, timeout=3) as db:
            row = db.execute("SELECT slot FROM headers ORDER BY slot DESC LIMIT 1").fetchone()
        return 0 if row else None  # <-- always 0, discards the actual slot value!
    except Exception:
        return None
```

## Fix

Compute the actual tip age as `current_slot() - latest_header_slot`. The result is in slots (matching the field name `tip_age_slots`). A healthy chain producing one block per slot should show 0-2 slots of age.

Kept out of the `ok` boolean as recommended in the issue — the field is informative for humans/dashboards, not a health gate.

Related: #8055 and PR #8056 (same class of defect in the health CLI).

Fixes #8057

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -9001,11 +9001,21 @@ def _backup_age_hours():
     return None
 
 def _tip_age_slots():
-    """Check tip freshness - query DB directly to avoid Response object"""
+    """Check tip freshness - query DB directly to avoid Response object
+
+    Returns the number of slots since the most recent header, or None if
+    the headers table is empty or the query fails.  A healthy chain that
+    produces one block per slot should have an age of 0-2 slots.
+    """
     try:
         with sqlite3.connect(DB_PATH, timeout=3) as db:
             row = db.execute("SELECT slot FROM headers ORDER BY slot DESC LIMIT 1").fetchone()
-        return 0 if row else None
+        if row is None:
+            return None
+        latest_slot = row[0]
+        now_slot = current_slot()
+        age = now_slot - latest_slot
+        return max(0, age)
     except Exception:
         return None
 


waterWang:
友好提醒一下这个 PR 👋 所有 CI 检查已通过，review 反馈也已处理完毕。麻烦看看还有没有需要调整的地方？谢谢！

sophiaeagent-beep:
Confirming this is a real fix for a real report, and thank you for the speed.

I filed #8057 yesterday after finding `/health` reporting `tip_age_slots: 0` on a node whose newest header is **234 days old**. The old body selected the newest slot and then threw it away, returning a hardcoded `0` whenever the headers table was non-empty. The function name said age, the docstring said freshness, and neither was computed.

Your version computes real age against `current_slot()` and returns `None` on an empty table instead of pretending freshness. That is the right shape.

One rollout note for the maintainer, not a defect in this PR: after merging, that field starts reporting a very large number instead of `0`. `ok` is not gated on `tip_age_slots` so the health boolean will not flip, but anything watching that value will see a change. That is a deployment decision rather than a code one.

Context worth having: the 224 headers in that table span slot 35 to 2,941,199 and the newest sits on the old raw-timestamp convention. The honest reading is that header production stopped months ago. This fix does not cause that, it makes it visible instead of hidden, which is exactly the value of it.

Sophia

waterWang:
Thanks for the thorough review! That is a very thoughtful analysis — the "deployment decision" note about the large value surfacing after merge is super valuable for anyone monitoring that field. We will keep it in mind. No further changes on our side; looking forward to it landing. 🙏

waterWang:
@sophiaeagent-beep Updating my wallet address to RTC: RTCbc5271ef42713f1eb8dc27baa620faf6572a5f38. Please use this for any future payouts. Thanks!

waterWang:
Hi @sophiaeagent-beep, quick follow-up on the **5 RTC** bounty for this merged PR. 🙏

The bot posted that the reward was sent to `waterWang`, and I updated my wallet address on Aug 2 (`**RTCbc5271ef42713f1eb8dc27baa620faf6572a5f38**`), but I checked the RustChain node today and my wallet balance is still 0 — it seems the funds are still under the miner-id rather than my wallet.

Could you help move it to my wallet address when you get a chance? Thanks so much! 🙏

PR #8061, opened 2026-07-25
fix(attest): scope attestation signature enforcement per miner (#8016)

Scottcjn:
Fixes the unsigned attestation bypass reported in #8016 by @draycolix.

**This is not deployed.** Nothing was run against a production node. Rolling it out is a maintainer decision, and the "Rollout notes" section below lists what breaks if you do.

## What is actually wrong

Verified against `node/rustchain_v2_integrated_v2.2.1_rip200.py` at `0ddfeb3`:

- **Line 4493**, `_submit_attestation_impl`: `if sig_hex and pubkey_hex:` guards every signature check and has no `else`. Omitting both fields skips verification entirely and execution falls through to IP rate limiting (4571), nonce validation (4589), hardware binding (4641), fingerprint validation, auto-enrollment (4901) and ticket issuance (4947).
- **Line 4263**, `/attest/challenge`: unauthenticated, rate-limited only, and it binds the nonce to the identity the *requester names* (4326, 4331). So the nonce protects nothing. Ask for a challenge naming the victim and you get one.
- `address_from_pubkey` (defined at 10844) is never called anywhere in the attestation handler. Nothing binds the submitted `public_key` to the claimed `miner`, so a self-signed forgery with a freshly generated keypair passes too. The codebase already knows this; `_header_key_authorized` says so in its docstring at line 5384.

Three things follow from that, all confirmed in code:

1. **The theft vector.** `epoch_enroll` has PK `(epoch, miner_pk)` (1403, 1711) and both insert sites used `INSERT OR IGNORE` (4902 and 5226). `FAILED_FINGERPRINT_WEIGHT_UNITS` is `0` (1364). A forged fingerprint-failing attestation writes weight 0 for a victim, and the victim's own honest enrollment is then silently ignored for the rest of the epoch. Sweep every known wallet at epoch start and you absorb the fixed per-epoch RTC pot.
2. **Lockout.** `record_attestation_success` did `signing_pubkey = excluded.signing_pubkey` unconditionally (3212). An unsigned forgery passes `signing_pubkey=None`, which NULLs the victim's stored key. The hardened signed `/epoch/enroll` path then returns 412 `ENROLLMENT_SIGNING_KEY_REQUIRED` (5124) and the victim cannot enroll.
3. **Multiplier downgrade.** `device_arch` was overwritten with no protection (3209), unlike `fingerprint_passed` which is `MAX`-guarded (3211). Rewards read `device_arch` straight out of `miner_attest_recent` (`rewards_implementation_rip200.py:276-280` into `get_time_aged_multiplier`), so this drops a G4 from 2.5x to 1.0x.

## Why per-miner scope, and not the obvious alternatives

Most miner clients in this repo do not sign at all. I checked every one:

| Client | Signs | Identity |
|---|---|---|
| `miners/apple2/miner6502.c` | no | `apple2-miner` |
| `miners/i386/miner386.c` | no | `i386-miner` |
| `miners/floppy-miner/` | no | `RTC2fe3c33c…` |
| `miners/ppc/g5/g5_miner.sh` | no | `ppc_g5_130_<md5>RTC` |
| `miners/rust/src/main.rs` | no | `default-miner` |
| `miners/ppc/g4/rustchain_miner.c`, `_v6.c` | no | `dual-g4-125` |
| `miners/power8/rustchain_power8_miner.py` | no | `power8-s824-<host>` |
| `miners/macos/…v2.4.py`, `macos/intel/…v2.4.py` | no | legacy names |
| `miners/pico_bridge/`, `mining/n64-miner/`, `mining/crt-attestation/` | no | legacy names |
| `rustchain-miner/src/attestation.rs` | **yes** | RTC wallet |
| `miners/linux/rustchain_linux_miner.py` | **yes**, conditional | `<38hex>RTC` |
| `miners/macos/rustchain_mac_miner_v2.5.py` | **yes**, if key loaded | canonical RTC |
| `miners/windows/rustchain_windows_miner.py` | **yes**, sha512 fallback with no pubkey | user wallet |

So:

- **Rejecting all unsigned attestations takes the honest fleet offline.** That is what PR #8052 did. A 6502 cannot compute Ed25519. Neither can a boot floppy or a shell script with curl.
- **A global allow-flag is a no-op the moment an operator sets it.** It restores the complete bypass for every identity simultaneously, and it cannot express the thing that actually matters, which is "this *particular* miner has proven it can sign, so stop accepting unsigned submissions *for it*".

The rule this PR enforces is scoped to one identity at a time, in `_attest_authorize_signing_key`:

1. **Unsigned is accepted only while the identity has no signing key on record.** The 6502 keeps working forever. A miner that has ever signed can never be impersonated by an unsigned request again.
2. **Trust on first use.** The first signed attestation pins the key.
3. **Key bound to identity.** A canonical `RTC<40hex>` address is self-authenticating, so the presented key must actually derive it (`address_from_pubkey(pubkey) == miner`). Named identities like `dual-g4-125` and `power8-s824-sophia` are not derivable from any key, so they get TOFU-pinned instead. This is the same policy `_header_key_authorized` already applies to block-header keys.
4. **Rotation needs the old key.** Replacing a pinned key requires `rotation_signature` over `rotate|miner|old_pubkey|new_pubkey|nonce`, signed by the key currently on record. A forger with a fresh keypair cannot displace the real one.

Note the RTC-shape check deliberately matches only the `RTC` + 40 hex *prefix* form. Several legacy wallets are a *suffix* form (`<38hex>RTC`, `ppc_g5_130_<md5>RTC`); treating those as derivable would reject them outright.

The gate runs before the nonce is consumed, so a rejected forgery cannot burn the challenge the rightful owner is about to use.

## The other three fixes

These matter independently, because they are what protect the legacy identities that still cannot sign:

- `epoch_enroll` at both sites is now an upsert taking `MAX(weight)`, mirroring the existing `MAX(fingerprint_passed)` precedent. This keeps the anti-downgrade property `INSERT OR IGNORE` was added for while removing the front-run. Same one-line change applied to `node/sophia_elya_service.py:214`, which had the same pattern.
- `signing_pubkey` is now `COALESCE(excluded, existing)`, so a key can be added or rotated but never erased by a submission that did not carry one.
- `device_family`/`device_arch` only change on an identity-proven attestation. For a never-keyed legacy identity the node genuinely cannot tell the owner from a forger, so the first recorded value wins. That is deliberately symmetric: it blocks a forged downgrade and a forged upgrade equally. A plain MAX-by-multiplier would have blocked only the downgrade while making arch spoofing permanent.

## Allowlist

`UNSIGNED_ATTEST_CLIENT_ALLOWLIST` documents the five client families that structurally cannot sign. It grants no privilege on its own; acceptance is decided purely by "does this identity have a key on record". It exists so operators can audit which identities are expected to stay unsigned.

To be explicit about the residual risk: **identities used by those clients are only ever as strong as their first-seen binding.** Anyone can still attest as a never-keyed legacy identity. What changed is that doing so can no longer steal from or lock out the target, because the forgery cannot lower an existing epoch weight, cannot erase a stored key, and cannot rewrite a recorded arch. Closing that last gap needs either signing hardware or an admin-seeded bootstrap allowlist like the one `_header_key_strict_bootstrap` already uses for header keys. I did not build that here.

## Tests

New file `tests/test_attest_signing_key_scope_8016.py`, 17 tests, driving the real Flask handler end to end against a temp SQLite DB.

Attacker side: unsigned forgery rejected once a miner has a key; self-signed forgery with a fresh keypair rejected for an RTC identity; unsigned rotation cannot displace a pinned key; rotation signed by the wrong key rejected; unsigned submission cannot NULL a stored key; forged zero-weight enrollment cannot front-run the victim; unsigned submission cannot rewrite `device_arch`.

Honest side: unsigned legacy miner with no stored key still attests, still enrolls at non-zero weight, and can re-attest repeatedly; signed RTC miner works and pins its key; named identity pins on first signed attestation; genuine rotation signed by the current key succeeds; the floppy miner's canonical-RTC-but-unsigned case still works.

Real numbers:

```
new tests, against this branch:      17 passed
new tests, against unpatched node:   10 failed, 7 passed
```

The 7 that pass both ways are the honest-fleet tests. They are supposed to pass before and after; that is the regression guard.

Full suite, excluding the new file, with and without this change:

```
with fix:     4 failed, 3665 passed, 46 skipped
without fix:  4 failed, 3665 passed, 46 skipped
```

Identical, so no regressions. Those 4 failures are pre-existing in `tests/test_epoch_settlement_formal.py` and fail the same way on unmodified `main`. `tests/test_device_classification_corpus.py` cannot be collected at all because `hypothesis` is not installed; also pre-existing and unrelated.

`py_compile` clean on all three changed files.

## Rollout notes

Not deployed. Before this goes out, two behaviours are worth a decision:

1. **A miner that signs conditionally and then stops will be locked out.** `miners/linux/rustchain_linux_miner.py` falls through unsigned on exception, and `miners/windows/rustchain_windows_miner.py` falls back to a sha512 signature with no `public_key`, which this handler treats as unsigned. Once either has pinned a key, a later unsigned submission gets 401. That is the intended security property, but it is a live-fleet consideration and those two clients may want fixing first.
2. **A canonical-RTC identity whose wallet was set by hand rather than derived from its own key will get 403.** `rustchain_mac_miner_v2.5.py` returns `self.wallet` as the attestation identity whenever it starts with `RTC`, so a manually configured wallet that the local key does not derive would be rejected. Worth checking before rollout whether any live miner is in that state.

Neither is a code defect here. Both are the fix doing what it is supposed to do, surfaced early so the rollout is not a surprise.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ Thumbs.db
 .pytest_cache/
 pytest-cache-files-*/
 tests/.tmp_attestation/
+tests/.tmp_attest_8016/
 
 # Windows miner build artifacts
 Rustchain/miners/windows/dist/
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -322,6 +322,11 @@ def _attest_mapping(value):
 _ATTEST_MINER_RE = re.compile(r"^[A-Za-z0-9._:-]{1,128}$")
 _SOLANA_WALLET_RE = re.compile(r"^[1-9A-HJ-NP-Za-km-z]{32,44}$")
 _ED25519_PUBKEY_HEX_RE = re.compile(r"^[0-9a-fA-F]{64}$")
+# Canonical RTC wallet shape produced by address_from_pubkey(): "RTC" + 40 hex.
+# NOTE: several legacy miners use a *suffix* form instead (``<38hex>RTC``,
+# ``ppc_g5_130_<md5>RTC``). Those deliberately do NOT match here — they are not
+# derivable from a public key and are handled as named identities.
+_RTC_ADDRESS_RE = re.compile(r"^RTC[0-9a-fA-F]{40}$")
 
 
 def _attest_text(value):
@@ -353,6 +358,40 @@ def _valid_ed25519_pubkey_hex(value):
     if text and _ED25519_PUBKEY_HEX_RE.fullmatch(text):
         return text.lower()
     return None
+
+
+def _attest_is_rtc_address(value):
+    """Return True when ``value`` has the derivable RTC wallet shape.
+
+    ``address_from_pubkey()`` builds addresses as ``RTC`` + 40 hex chars, so an
+    identity of that shape is *self-authenticating*: exactly one Ed25519 public
+    key can produce it. Identities of any other shape (``dual-g4-125``,
+    ``power8-s824-sophia``) are not derivable from a key and must be pinned on
+    first use instead.
+    """
+    text = _attest_text(value)
+    return bool(text and _RTC_ADDRESS_RE.fullmatch(text))
+
+
+# Miner clients in this repository that structurally cannot produce an Ed25519
+# signature. They are documented here so operators can audit which identities
+# are expected to stay unsigned; the list grants NO extra privilege on its own.
+# Unsigned attestations are accepted purely by the per-miner rule in
+# _attest_authorize_signing_key(): an identity with no signing key on record.
+#
+# SECURITY NOTE: identities used by these clients are only ever as strong as
+# their first-seen binding. Anyone can attest as a never-keyed legacy identity.
+# What the surrounding fixes guarantee is that such a forgery can no longer
+# steal from or lock out its target: it cannot lower an existing epoch weight
+# (MAX(weight) upsert), cannot erase a stored signing key (COALESCE upsert),
+# and cannot rewrite a recorded device_arch (COALESCE upsert).
+UNSIGNED_ATTEST_CLIENT_ALLOWLIST = (
+    "miners/apple2/miner6502.c",       # 6502, no Ed25519 possible
+    "miners/i386/miner386.c",          # 16/32-bit DOS-era x86
+    "miners/floppy-miner/",            # boot-floppy miner, no crypto library
+    "miners/ppc/g5/g5_miner.sh",       # POSIX shell + curl only
+    "miners/rust/src/main.rs",         # legacy Rust miner, unsigned payload
+)
 
 
 def _attest_field_error(code, message, status=400):
@@ -3162,7 +3201,7 @@ def _projected_multiplier_growth(current_mult: float, device_arch: str) -> dict:
     }
 
 
-def record_attestation_success(miner: str, device: dict, fingerprint_passed: bool = False, source_ip: str = None, signals: dict = None, fingerprint: dict = None, signing_pubkey: str = None, entropy_score: float = 0.0):
+def record_attestation_success(miner: str, device: dict, fingerprint_passed: bool = False, source_ip: str = None, signals: dict = None, fingerprint: dict = None, signing_pubkey: str = None, entropy_score: float = 0.0, identity_proven: bool = True):
     now = int(time.time())
     # Miner-name platform hints — helps detect Apple Silicon / POWER8 when client doesn't send rich device info
     _device = dict(device or {})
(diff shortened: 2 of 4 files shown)


laurentketterle-hub:

laurentketterle-hub:

Scottcjn:
Closing: main now carries the #8016 hardened Ed25519 identity binding and phased enforcement (see the '#8016 hardened' sections in the integrated node and the admin signing-key management route), implemented independently after this branch went stale. The tests here remain a useful reference.

PR #8062, opened 2026-07-25
Compute epoch enrolment weight from the verified device, not the request body

Scottcjn:
## The defect

`enroll_epoch` computed reward weight from the request body:

```python
family = device.get('family', 'x86')      # device = data.get('device', {})
arch   = device.get('arch', 'default')
hw_weight = HARDWARE_WEIGHTS.get(family, {}).get(arch, 1.0)
```

`derive_verified_device()` was never called here, and the `device_arch` the server derived at attestation time was never read. The caller named their own multiplier.

`HARDWARE_WEIGHTS["ARM"]["arm2"]` is **4.0** against an honest modern x86_64 at **0.8**. That is a 5x self-granted uplift, higher than a real G4 at 2.5.

The surrounding machinery does not stop it:

- Signing does not help. The signature proves the caller owns *their own* wallet, which the attacker does.
- `fingerprint_passed` is a one-way `MAX()` ratchet, so one honest attestation from any machine satisfies the enrolment gate permanently.
- `epoch_enroll` is `INSERT OR IGNORE`, so the forged row wins the epoch and the honest auto-enrolment that follows is silently discarded. The anti-downgrade behaviour ends up protecting the forgery.

Since `PER_EPOCH_RTC` is a fixed pot split proportionally, anything gained here comes directly out of every honest miner's share.

**This is the path that matters.** The shipped Linux miner re-attests only when its 24h attestation expires, but calls `/epoch/enroll` every 600 seconds. So `/epoch/enroll` sets the weight for most epochs, and hardening only the attestation path leaves the actual hole open.

## The fix

Enrolment now resolves family and arch from `miner_attest_recent`, which `derive_verified_device()` already wrote. `check_enrollment_requirements()` already requires an attestation no older than `ENROLL_TICKET_TTL_S`, so the row is expected to exist, and the attestation path has always resolved weight this way. The two enrolment paths now agree, which they did not before.

The resolver falls back to the `x86` bucket for `x86_64` and `Windows` families before falling back to a family default. That detail is load-bearing: the macOS miner reports family `x86_64` with vintage archs such as `core2`, and `x86_64` carries only `modern`/`default`, so a naive nested default would pay a Core 2 Mac **0.8** instead of the intended **1.3**. Under-paying the fleet you meant to protect is the failure mode this project has already hit once.

Missing or unreadable attestation data fails soft to a neutral 1.0. It never grants a vintage tier and never blocks enrolment.

## Live impact, measured rather than assumed

Every distinct `(device_family, device_arch)` pair attested on the live chain in the last 30 days was run through both the old and new resolution. **20 unchanged, 7 changed.**

**No vintage tier moves.** G4, G5, POWER8, Apple Silicon, `pentium_m_banias` and `retro` are all identical.

The 7 changes are Windows device strings such as `Intel64 Family 6 Model 158 Stepping 10, GenuineIntel`, which matched no key in the `Windows` bucket and fell through to the literal `1.0` rather than that bucket's own `0.8` default:

| family | arch | before | after |
|---|---|---|---|
| Windows | Intel64 Family 6 Model 158/140/58/181/60 … | 1.000 | 0.800 |
| Windows | x86_64 | 1.000 | 0.800 |
| Windows | i386 | 1.000 | 0.800 |

These are modern Windows machines that were being paid above the modern rate. Correcting them returns that surplus to everyone else, including the vintage fleet. **It is still a reduction for real miners and worth announcing rather than shipping quietly.**

## Tests

`node/tests/test_enroll_hw_weight_from_attestation.py`, 7 tests, all passing. They encode the actual incident rather than abstractions: the ARM/arm2 self-grant is refused; no body-supplied arch reaches any vintage tier; honest G4, G5, T40 Pentium M and Cobalt Qube keep their exact multipliers; the Core 2 Mac keeps 1.3 through the x86-bucket fallback; a missing attestation is neutral rather than a bonus; a broken database fails soft.

## Not deployed

This has not been applied to any production node. It changes reward weight, so it needs a maintainer rollout decision, and the Windows tier correction above should probably be communicated to those miners first.

## Relationship to other open work

Related to #8016 and the attestation-bypass work. Complementary to #8022, which hardens the attestation path but, on its own, does not close this one because `/epoch/enroll` is where the weight is actually set for most epochs. #8013 touches the same line and should be rebased onto this shared resolver rather than reintroducing a separate literal fallback.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -5024,6 +5024,56 @@ def get_proposer_duty_calendar():
     )
 
 @app.route('/epoch/enroll', methods=['POST'])
+def resolve_enroll_hw_weight(db_path, miner_pk, miner_id=None):
+    """Resolve enrollment reward weight from the SERVER-DERIVED device.
+
+    The request body must never decide reward weight. A caller can put any
+    family/arch in it and self-grant a multiplier: HARDWARE_WEIGHTS["ARM"]["arm2"]
+    is 4.0 against an honest modern x86_64 at 0.8, a 5x uplift and higher than a
+    real G4. Because epoch_enroll is INSERT OR IGNORE, that forged row would also
+    win against the honest auto-enroll that follows.
+
+    The attestation path already runs derive_verified_device() and stores the
+    result, and check_enrollment_requirements() mandates an attestation no older
+    than ENROLL_TICKET_TTL_S, so a row is expected to exist here.
+
+    Returns (hw_weight, family, arch, source).
+    """
+    family, arch, source = "x86", "default", "no_attestation"
+    try:
+        with closing(sqlite3.connect(db_path, timeout=5)) as conn:
+            row = None
+            for key in (k for k in (miner_pk, miner_id) if k):
+                row = conn.execute(
+                    "SELECT device_family, device_arch FROM miner_attest_recent WHERE miner = ?",
+                    (key,),
+                ).fetchone()
+                if row:
+                    source = "attested"
+                    break
+        if row:
+            family = row[0] or "x86"
+            arch = row[1] or "default"
+    except Exception:
+        # Fail soft: an unreadable attestation row must not hand out a bonus tier,
+        # and must not deny enrolment either. The default bucket is neutral.
+        pass
+
+    bucket = HARDWARE_WEIGHTS.get(family, {})
+    if arch in bucket:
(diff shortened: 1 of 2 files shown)


laurentketterle-hub:

laurentketterle-hub:

Scottcjn:
Closing: main now computes enrolment weight via resolve_enroll_weight_device() from the verified device in the DB rather than the request body, which is the fix this PR made. Superseded, not lost.

PR #8066, opened 2026-07-25
fix: include chain tip in overall health calculation and exit code [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #8055: [BUG] rustchain-health exits 0 when the chain tip check fails

waterWang:
## Summary

Adds `tip.reachable` to both `all_ok` predicates used for overall health status rendering and process exit code, so a failed chain-tip check correctly produces an unhealthy status and non-zero exit code.

### Changes

- `render()` (line 276-279): Added `t["reachable"]` to the `all_ok` predicate
- `run_once()` (line 358-363): Added `snapshot["tip"]["reachable"]` to the `all_ok` predicate

This ensures that when the chain tip endpoint is unavailable (HTTP 404, timeout, etc.), the overall status shows `ISSUES DETECTED` and the process exits with code 1, instead of falsely reporting `ALL SYSTEMS OPERATIONAL` with exit code 0.

Closes #8055

### DIFF
--- a/tools/rustchain-health.py
+++ b/tools/rustchain-health.py
@@ -275,7 +275,8 @@ def render(snapshot: Dict[str, Any]) -> str:
     # ── Overall ───
     all_ok = (h.get("ok", False) and h["reachable"]
               and e["reachable"]
-              and m["reachable"])
+              and m["reachable"]
+              and t["reachable"])
     lines.append(dim("  " + "─" * w))
     if all_ok:
         lines.append(f"  {green(bold('STATUS: ALL SYSTEMS OPERATIONAL'))}")
@@ -358,7 +359,8 @@ def main() -> int:
         all_ok = (snapshot["health"].get("ok", False)
                   and snapshot["health"]["reachable"]
                   and snapshot["epoch"]["reachable"]
-                  and snapshot["miners"]["reachable"])
+                  and snapshot["miners"]["reachable"]
+                  and snapshot["tip"]["reachable"])
         return 0 if all_ok else 1
 
     if args.watch > 0:


sophiaeagent-beep:
Merged. Your fix is now on `main`, and I want to be straight about how it got there.

I had a competing PR open (#8056) that I opened 47 minutes before yours, and I argued it was the better fix. It was not: it broke 7 existing tests by changing the `fetch()` return contract, and my attempt to fix that made it 11. Yours passes all 12 checks. I have closed mine and said so there.

**Awarded 15 RTC** for this one, at Medium, since it repairs a monitoring tool that was reporting ALL SYSTEMS OPERATIONAL and exiting 0 while the chain tip check failed.

Separately, and this is the more valuable one: **#8060 is also being credited.** It fixes `_tip_age_slots`, which never computed an age at all and returned a hardcoded 0 whenever the headers table was non-empty. That is what let a node with a 234 day old tip report as fresh for months.

One note on process, offered plainly and without any hard feelings: on the BoTTube validator bounty you posted "I claim this bounty, starting implementation now" on another contributor's open PR thread and opened a competing PR seven minutes later. As it turned out that bounty had already been won and paid five days earlier, so nothing was taken from anyone, and I have corrected my own comment there. But claim on the bounty issue, not on someone else's PR. If you see work that is nearly there, say what is missing on it. You are producing good, real fixes now and that is worth more than winning races.

I still owe you an RTC wallet address to pay to. Post one and both awards go out.

Sophia

sophiaeagent-beep:
Payment sorted, and I want to explain the two transactions so the ledger reads clearly rather than looking like a mistake.

When this merged, an automated payout fired **5 RTC** to `waterWang` (pending id 3786, from the community pool). That bot pays a flat rate per merged PR and it is not the considered assessment. I assessed this at **15 RTC**, Medium tier, so I have sent the **10 RTC balance** (pending id 3787). Total 15, in two rows, both settling on the normal 24 hour cycle.

Both went to the miner-id `waterWang` rather than to an Ed25519 wallet, because I still do not have an address for you. That works, but it leaves the balance somewhere you cannot sign a transfer out of. We had a contributor stuck exactly like that for weeks with 70 RTC they could not move, so I would rather not do that to you.

To fix it: `pip install clawrtc` then `clawrtc wallet create`, and post **only** the `RTC...` address here. Never the seed phrase. I will move the balance across.

`#8060` is still open and is the more valuable of your two. That one stops a node with a 234 day old tip reporting as fresh, which is a monitoring lie rather than a cosmetic bug. It will be credited separately once it merges.

Sophia

waterWang:
@sophiaeagent-beep Here is my RTC wallet address: RTCbc5271ef42713f1eb8dc27baa620faf6572a5f38. Please transfer the 15 RTC balance to this address. Thanks!

waterWang:
Hi @sophiaeagent-beep, hope you're well! 🙏

Quick follow-up on the bounty payout for this PR. Back in July you mentioned the payment was sorted (automated 5 RTC + 10 RTC balance, pending ids 3786/3787), and that both went to the miner-id `waterWang` because no wallet address was available at the time.

I shared my wallet address on Aug 2: **`RTCbc5271ef42713f1eb8dc27baa620faf6572a5f38`** — but I checked the chain today and my balance is still **0 RTC**, so it looks like the funds are still sitting under the miner-id and haven't been moved to my wallet yet.

Could you help transfer the **15 RTC** to my wallet address? Happy to provide any additional info needed. Thanks a lot for your work! 🙏

Issue #6624, opened 2026-05-29
[TRACKING] rustchain-claim-portal chain_client.get_balance should use public read endpoint, not admin-key path


Scottcjn:
## Background

In [rustchain-claim-portal](https://github.com/Scottcjn/rustchain-claim-portal) the `chain_client.get_balance(identifier)` call sends `X-Admin-Key` even for read operations because that's the auth shape inherited from MergeWork's wallet-API port.

Node 1 (`50.28.86.131`) already exposes `/wallet/balance?miner_id=<id>` as a public no-auth endpoint — we exercise it dozens of times per day in operator sweeps with no auth header.

## Why this matters

The portal's [Phase 2 MCP layer](https://github.com/Scottcjn/rustchain-claim-portal/blob/main/app/mcp_tools.py) exposes `portal.balance.get`, `portal.reserved.get`, and similar read-only tools. Today every MCP read call carries `RC_ADMIN_KEY` to Node 1. That means:

- If the portal is compromised, an attacker reading `current_app.config['RC_ADMIN_KEY']` gets the FULL admin key, not a read-only token.
- The MCP layer's claimed read-only guarantee leaks: read-via-MCP requires the same auth credential as write-via-HTTP.
- Per [federation design note](https://github.com/Scottcjn/rustchain-claim-portal/blob/main/FEDERATION_DESIGN_NOTE.md) (Section 4: relayer model), bridge custody keys should be held separately from broader admin keys — same principle applies here.

## Required change

Refactor `chain_client.py` so that:

1. **Read methods** (`get_balance`, future `bridge_state_get`, `recent_events_list`) call Node 1 endpoints **without** `X-Admin-Key`. Add a `_request_public(method, path, ...)` helper that skips the auth header.
2. **Write methods** (`transfer`, `void_transfer`) continue to require `X-Admin-Key` via the existing `_request` helper.
3. **Configuration check at startup**: if the portal is asked to perform a write operation and `RC_ADMIN_KEY` is not configured, fail with `ChainClientError` immediately (already the case). If only `RC_NODE_URL` is set (read-only deployment), reads still work — write methods raise `ChainClientError("RC_ADMIN_KEY required for write operations")`.
4. **Test additions**: cover that a mocked Node 1 receives NO `X-Admin-Key` header on read calls, and DOES receive it on write calls.

## Concrete files to edit

- `app/chain_client.py` (currently 98 LOC)
- `tests/test_chain_client.py` (currently 114 LOC)

[...]

PR #8076, opened 2026-07-27
Stop the stale bot from closing issues

Scottcjn:
## What happened

This workflow closed **#6749** this morning as `not planned`. That issue tracks a live cross-path double-credit race in epoch settlement, and the PRs fixing it (#8062, #8065) were open and unreviewed at the time. It closed a **security issue** on beacon-skill the same morning, and had armed five more including the production node drift issue and the federation bridge RFC.

## Why adding labels was not enough

My first attempt at this PR just added `consensus`, `tracking` and `operations` to the exempt list. That only works if somebody remembers to apply the label.

Two days ago the same class of failure hit the bounty board: 29 genuine bounty postings had no `bounty` label, which made them invisible to the README browse link **and** unprotected from that repo's stale bot. Ten were already marked stale. The spam around them was untouched, because spam gets commented on and real work goes quiet.

## The actual wrong assumption

Silence means resolved. For a hard bug the opposite is true. Nobody comments on a consensus race for a month *precisely because it is hard*, and the reward for being hard was deletion.

## What changed

**Issues are marked stale and never auto-closed.** `days-before-issue-close: -1`. Marking is information and a human can still close in one click. Closing is destruction, and it lands hardest on exactly the work that takes longest to fix.

**Pull requests still auto-close**, on a longer leash (21 days to stale, 14 to close, up from 14/7). PRs genuinely rot against a moving main, the branch is not lost, and reopening is one click.

Also:
- Issue stale window 30 → 45 days
- Milestoned and assigned issues exempt, since a human already triaged those
- `operations-per-run` 50 → 300. At 50 the queue never drained on a repo this size, so the same items were revisited while others were never reached at all
- Exempt lists widened anyway, as belt and braces

## Note

The six issues it hit were reopened and pinned by hand today. Two of them (#6624, #6627) turned out to be genuinely complete and were closed properly with pointers to the PRs that did the work — which is what closing should look like.

### DIFF
--- a/.github/workflows/lychee.yml
+++ b/.github/workflows/lychee.yml
@@ -23,7 +23,6 @@ jobs:
             --exclude "https?://(twitter|x)\.com"
             --exclude "https://github.com/Scottcjn/Rustchain/actions"
             --exclude "https?://(www\.)?(linkedin|facebook|instagram|tiktok)"
-            --exclude-mail
             -- './**/*.md' './**/*.html'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,5 +1,26 @@
 name: Stale Issue & PR Cleanup
 
+# Why this is shaped the way it is
+#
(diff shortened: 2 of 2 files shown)


laurentketterle-hub:

laurentketterle-hub:

Issue #8078, opened 2026-07-27
A VM passes all six fingerprint checks by asserting them with no evidence


Scottcjn:
## A VM is passing all six fingerprint checks on the live node

`prometheusvt` is attesting right now with `fingerprint_passed = 1`, earning a 0.8x weight in epochs 234-236. Its operator states publicly (in their #16242 explainer) that it runs **in a Proxmox VM**.

This is what it submitted, straight out of `miner_attest_recent.fingerprint_checks_json` on node 1:

```json
{"clock_drift": true, "cache_timing": true, "simd_identity": true,
 "thermal_drift": true, "instruction_jitter": true, "anti_emulation": true}
```

A flat map of six `true` values. **No `data` object, no measurements, no vm_indicators, no evidence of any kind.** The server accepted it and set `fingerprint_passed = 1`.

## Why this matters more than the 0.8x

The reward impact is small: it claims `modern`, so it gets 0.8x rather than a vintage tier. The problem is that `fingerprint_passed` is the gate that is supposed to make VMs earn **nothing**. Per the RIP-PoA design, `weight = hw_weight if fingerprint_ok else 0.0`. A VM that passes gets paid at all, and since `PER_EPOCH_RTC` is a fixed pot, that share comes out of the honest miners the check exists to protect.

It also means the six checks are currently decorative for any client that simply asserts them.

## Root cause

This is the client-claimed-bits problem, live:

1. `validate_fingerprint_data` only rejects on evidence it can see. A payload with no `data` at all does not trip the anti-emulation branch, because that branch inspects `vm_indicators` — and there are none to inspect.
2. `check_vm_signatures_server_side()` scans `signals["hostname"]` and `device["cpu"]`, both of which come from the same POST body. There is no independent server-side observation anywhere in the VM defence.
3. So the only thing standing between a VM and a passing fingerprint is whether the VM's own client chooses to report itself.

The stock miner does report honestly. `fingerprint_checks.py` on real hardware emits full `data` blocks, and on a QEMU guest it emits `vm_indicators` like `/sys/class/dmi/id/sys_vendor:qemu`. A client that omits all of it passes.

## What would fix it

Not "reject missing checks" — that would zero the Apple II, the i386 and every console miner, which structurally cannot produce some of these measurements (see #8064 for the capability-aware approach).

[...]

PR #8079, opened 2026-07-27
fix(#8078): require evidence for fingerprint checks claiming a pass [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

waterWang:
### Problem
A VM (`prometheusvt`) is passing all six fingerprint checks on the live node by submitting a flat map of six `true` values with **no `data` object, no measurements, no evidence of any kind**. The server accepted it and set `fingerprint_passed = 1`.

The root cause is in `_fingerprint_check_passed()`: a dict entry like `{"passed": true}` with no `data` key was treated as a full pass. The rotating check system could not distinguish "this hardware cannot measure it" (legitimate vintage/console miners) from "this client did not bother to submit evidence" (the bypass).

### Fix
1. **`_fingerprint_check_passed()`** now returns `None` (unmeasured) instead of `True` when a dict check has `{"passed": true}` but no `data` key, or when `data` is not a usable dict. Bool-only entries (C miner compat) keep their literal value.

2. **`evaluate_rotating_fingerprint_checks()`** now excludes `None` (unmeasured) results from the active-ratio denominator. Unmeasured checks are neither passes nor failures — they don't count either way. This is the proportional-evidence rule: the weight of a check scales with the evidence submitted.

### Impact
- A client that submits `{"clock_drift": true, "cache_timing": true, ...}` with no data → all 6 checks count as unmeasured → active_ratio = 1.0 (no active checks measured) → **no reward weight** from the rotating check system
- A client that submits `{"clock_drift": {"passed": true, "data": {"cv": 0.09, "samples": 200}}, ...}` with real data → all 6 checks pass normally → full weight
- Vintage/console miners that genuinely cannot measure a check (e.g. no `time.perf_counter_ns`) already omit that check from the payload → the check is absent → `checks.get(name)` returns `None` → `None` is unmeasured → excluded from the ratio. **No regression.**

### References
- Issue: #8078
- Related PR #8065 (tightens the `vm_indicators` case; this PR covers the no-data-at-all case)
- Wallet: FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -2211,12 +2211,25 @@ def select_active_fingerprint_checks(previous_epoch_block_hash: str, active_coun
     return tuple(ranked[:active_count])
 
 
-def _fingerprint_check_passed(check_entry) -> bool:
+def _fingerprint_check_passed(check_entry):
+    """Return True (passed), False (failed), or None (unmeasured).
+
+    Bool-only entries (C miner compat) keep their literal value.
+    Dict entries with {"passed": true} but no "data" key are
+    "unmeasured" (None) — the client asserted a pass without
+    supplying evidence.  Dict entries with {"passed": true, "data": {...}}
+    are passed.  Empty dicts {} are unmeasured (no data at all).
+    """
     if isinstance(check_entry, bool):
         return check_entry
     if isinstance(check_entry, dict):
-        return bool(check_entry.get("passed", True))
-    return False
+        verdict = check_entry.get("passed", True)
+        if verdict is True and "data" not in check_entry:
+            return None  # unmeasured — no evidence submitted
+        if verdict is True and not isinstance(check_entry.get("data"), dict):
+            return None  # unmeasured — data is not a usable object
+        return bool(verdict)
+    return None
 
 
 def get_previous_epoch_block_hash(conn, epoch: int) -> str:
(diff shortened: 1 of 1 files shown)


waterWang:

qiann0512-gif (changes requested):
The evidence/no-evidence split is the right direction, but this still leaves the bypass alive when all active checks are unmeasured. After the patch, `measured_count == 0` still sets `active_ratio = 1.0`, and both enrollment paths multiply reward weight by that ratio (`hw_weight * rotation_eval["active_ratio"]` / `rotation_eval['active_ratio']`).

So a payload where every active check is `{"passed": true}` without usable `data` now produces `active_results` of all `None`, `active_total` of 0, and `active_ratio` of 1.0. That is exactly the no-evidence case described in the PR: it no longer increments `active_pass_count`, but the downstream weight calculation can still treat it as full credit because it only uses the ratio.

Please make the zero-measured-check case fail closed for reward weighting, or carry an explicit "no measured evidence" flag that enrollment treats as no rotating-check reward. A regression should cover a fingerprint with only dict `passed: true` / no `data` entries and assert that the resulting enrollment weight does not get full `hw_weight` credit.

Scottcjn:
The direction here is right and it matches the standing rule for this code, that the server does not trust `"passed": true` without raw evidence. Splitting unmeasured from failed is the correct model: hardware that genuinely cannot measure a check should not be punished the same as a client that skipped it.

One thing to resolve before this goes in, because I think it leaves the strongest version of the attack open.

```python
active_ratio = (len(passed_active) / measured_count) if measured_count else 1.0
```

Consider a client that submits **every** active check as `{"passed": true}` with no `data`. Each one maps to `None`, `measured` ends up empty, `measured_count` is `0`, and the ratio falls through to the `else` branch and becomes **1.0, a perfect score**.

That is the same 1.0 such a client gets on `main` today, so this is not a regression. But it means the bypass this PR is aimed at still succeeds in its most complete form. The PR closes the mixed case, where some checks carry evidence and others do not, and leaves the all-evidence-omitted case scoring perfectly. An attacker picks the version that works.

An empty denominator is not "everything passed", it is "nothing was proven", and those should not score the same. Options, your call which fits the protocol:

- treat `measured_count == 0` as `0.0` rather than `1.0`
- or reject the attestation outright, in the same shape as an unmeasurable-hardware case, so it is visible rather than silently perfect
- and either way, a test pinning the all-unmeasured case, since that is the one the current tests do not cover

Please also add tests for the three-way split: measured-pass, measured-fail, and unmeasured. Changing `_fingerprint_check_passed` from `bool` to a tri-state return is the kind of signature change where a later reader can easily reintroduce a truthiness bug, given `None` and `False` are both falsy.

Worth saying: catching that unmeasured and failed were being conflated is a good find, and the docstring you added is clear. It is the empty-denominator default that needs to change.

PR #8080, opened 2026-07-27
fix: remove deprecated --exclude-mail flag from lychee config [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

waterWang:
Fixes #8077

The --exclude-mail flag was removed in a newer version of lychee, causing the link checker to fail with "unexpected argument '--exclude-mail' found" on every run since July 22. Mail links are skipped by default now, so the flag is no longer needed.

This is a one-line fix: simply remove the --exclude-mail argument from the lychee args.

### DIFF
--- a/.github/workflows/lychee.yml
+++ b/.github/workflows/lychee.yml
@@ -23,7 +23,6 @@ jobs:
             --exclude "https?://(twitter|x)\.com"
             --exclude "https://github.com/Scottcjn/Rustchain/actions"
             --exclude "https?://(www\.)?(linkedin|facebook|instagram|tiktok)"
-            --exclude-mail
             -- './**/*.md' './**/*.html'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


Scottcjn:
Verified and this is correct.

Current lychee has no `--exclude-mail` option. `lychee-bin/src/config/mod.rs` only defines `include_mail`, which returns `self.include_mail.unwrap_or(false)`, and the filter is `uri.is_mail() && !self.include_mail`. There are upstream tests literally named `test_exclude_mail_by_default`. So mail links are skipped by default and the flag that used to request that is gone.

That means the removal is behaviour-preserving as well as necessary: nothing starts getting checked that was not being checked before.

Single line, single purpose, no test impact. Looks good to me.

One scheduling note: #8082 also deletes this same line as part of an unrelated docs change, so the two conflict in `.github/workflows/lychee.yml`. I would land this one first and let #8082 rebase, since this is the change that is actually about the flag.

waterWang:

Scottcjn:
Merging. One line, removes a flag lychee no longer accepts, and the workflow is currently not producing results anyway so there is nothing to regress.

waterWang:

Thanks! 🙏

PR #5726, opened 2026-05-19
feat: complete live RustChain stats dashboard

JeremyZeng77:
For Scottcjn/rustchain-bounties#1600.

## Summary
- Replaces the existing dashboard page with a responsive live stats dashboard.
- Shows current epoch, slot, miner count, total supply, transaction availability, public node health, and top miners.
- Auto-refreshes every 60 seconds and includes a manual refresh button.
- Renders API data with DOM text nodes instead of injecting endpoint values through HTML strings.

## Verification
- git diff --check
- node --check on the extracted dashboard script
- Live endpoint checks against rustchain.org:
  - /health -> 200
  - /epoch -> 200
  - /api/miners -> 200
  - /api/transactions -> 404 on the current public node, so the dashboard treats it as an optional endpoint and displays an unavailable state instead of failing the whole refresh.

## Bounty / payout
- Bounty: Scottcjn/rustchain-bounties#1600
- RTC wallet: RTCfbc6abb564f210ede29a769d1b4d173278f526ea

PR #5701, opened 2026-05-19
fix(test): make attestation fuzzer console-safe

JeremyZeng77:
## Problem

Fixes #5699.

`tests/fuzz_attestation_runner.py` prints emoji/status glyphs before it sends any fuzz payloads. On Windows consoles using a non-UTF-8 encoding such as GBK/cp936, the first banner print can raise `UnicodeEncodeError`, so the fuzzer exits before exercising `/attest/submit`.

## Changes

- Configure `sys.stdout` and `sys.stderr` with `errors="replace"` when `reconfigure()` is available.
- Keep existing output unchanged on UTF-8 terminals.
- Allow legacy/non-UTF-8 consoles to degrade unencodable glyphs instead of crashing.

## Verification

```bash
python -m py_compile tests/fuzz_attestation_runner.py
```

Simulated a non-UTF-8 Windows console:

```powershell
$env:PYTHONIOENCODING='cp936'
$env:RUSTCHAIN_URL='http://127.0.0.1:9'
python -B tests\fuzz_attestation_runner.py --count 1 --seed 1
```

Result: the runner reached the fuzz summary without `UnicodeEncodeError`; unencodable banner glyphs were replaced by `?`, as intended.

Bounty references: bug report #5699 and bug bounty #305.

Payout route: PayPal `https://paypal.me/Jeremytsangchina`; USDT TRC20 `TWupTAoFjxR2VmqUUvzhfmUwXVZR6Kvz1Y`. If RustChain bug bounties are RTC-only, please tell me the expected wallet/miner-id format.

PR #8081, opened 2026-07-27
fix: stop stale bot from auto-closing issues [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

waterWang:
Fixes #8076

Changes:
- **`days-before-issue-close: -1`** — Issues are marked stale but never auto-closed. Marking is informational; closing is destructive, and it hits hardest on the hardest bugs (consensus races, security issues) that naturally go quiet.
- **`days-before-stale: 45`** (up from 30) — More time before marking stale.
- **`days-before-pr-stale: 21`** (up from 14) and **`days-before-pr-close: 14`** (up from 7) — Longer leash for PRs.
- **`operations-per-run: 300`** (up from 50) — The old limit never drained the queue on a repo this size.
- **Exempt labels widened** — Added `consensus`, `tracking`, `operations` to exempt-issue-labels.
- **Removed `close-issue-message`** — No longer needed since issues are never auto-closed.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -17,19 +17,18 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           stale-issue-message: |
-            This issue has been inactive for 30 days. It will be closed in 7 days unless there's new activity.
+            This issue has been inactive for 45 days. It will be marked as stale.
             If this is still relevant, please comment to keep it open.
           stale-pr-message: |
-            This PR has been inactive for 14 days. It will be closed in 7 days unless updated.
+            This PR has been inactive for 21 days. It will be closed in 14 days unless updated.
             Need help finishing? Ask in the PR comments — we're happy to assist!
-          close-issue-message: 'Closed due to inactivity. Feel free to reopen if still needed.'
           close-pr-message: 'Closed due to inactivity. Feel free to reopen with updates.'
-          days-before-stale: 30
-          days-before-close: 7
-          days-before-pr-stale: 14
-          days-before-pr-close: 7
+          days-before-stale: 45
+          days-before-close: -1
+          days-before-pr-stale: 21
+          days-before-pr-close: 14
           stale-issue-label: 'stale'
           stale-pr-label: 'stale'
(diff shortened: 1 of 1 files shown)


Scottcjn:
Checked this against the current `.github/workflows/stale.yml` and against the `actions/stale` v10 semantics. It does the right thing.

The key line is `days-before-close: -1`, which disables closing. Because `days-before-pr-close: 14` is set explicitly, the PR side still closes and only the issue side falls back to the `-1`. So the resulting behaviour is:

- Issues: labelled `stale` after 45 days of inactivity, commented on, never auto-closed.
- PRs: labelled `stale` after 21 days, closed 14 days later.

That is the shape I would want. Dropping `close-issue-message` is consistent, since issues no longer close.

Two things I checked that are worth noting:

`security` was already in `exempt-issue-labels` on main, so an exempt label was not what failed. Whatever got auto-closed was not carrying the label at the time. That is an argument for the `-1` doing the real work here rather than the label list, which is how this PR is built. Adding `consensus,tracking,operations` is still useful as a second layer.

`operations-per-run` going 50 to 300 is the one change with a cost. It is six times the GitHub API budget per run, on a repo with a lot of open issues and PRs. It should be fine at a weekly cadence, but if the job starts hitting rate limits that is the first knob to turn back down.

No test impact, this is workflow-only. Looks good to me.

waterWang:

Scottcjn:
Merging, and this one is more valuable than its size suggests.

The stale bot has already closed two unmerged external PRs (#5701 and #5726, both @JeremyZeng77). With PRs going stale at 14 days and closing 7 days later, a 21 day window, and a queue where inflow is running about twice outflow, contributor work was being destroyed on a timer rather than reviewed. Thirty-seven PRs merged today were between 4 and 30 days old; several would have been auto-closed before anyone looked at them.

`days-before-close: -1` for issues, 21 plus 14 for PRs, and the wider exempt label list all point the right way. Raising `operations-per-run` from 50 to 300 also matters at this repo's size, since the bot was silently running out of budget partway through.

Thank you for spotting a process bug rather than a code one.

waterWang:

Thanks! 🙏

PR #8086, opened 2026-07-28
fix(rip302): disputed jobs are a dead end — the worker can never re-deliver and the escrow only goes back to the poster

Vyacheslav-Tomashevskiy:
## The bug

`POST /agent/jobs/<id>/dispute` answers the client with:

> Job disputed. Escrow held pending resolution. **Worker can re-deliver or admin can refund.**

Neither half was reachable:

- `/deliver` accepted only `claimed` jobs (`if j["status"] != STATUS_CLAIMED: 409`), and **no route moves a job back out of `disputed`** — so the worker could never act on the rejection reason.
- There is no admin route at all. The registered set is `/agent/jobs`, `/agent/jobs/<id>`, `/claim`, `/deliver`, `/accept`, `/dispute`, `/cancel`, `/agent/reputation/<w>`, `/agent/stats`.
- A disputed job is also outside the expiry sweep (`_expire_refundable_job` returns early unless the status is `open`/`claimed`), so the escrow never released on its own either.

The only reachable exit was the poster calling `/cancel`, which refunds **100% of the escrow to the poster** — after the deliverable is already stored on the job and readable by anyone via `GET /agent/jobs/<id>`. So on a money path, the documented recourse for the worker did not exist.

Reproduced against `main` (Flask test client, real endpoints, the same harness style as `tests/test_agent_dispute_accept_race.py`):

```
posted: 201 escrow held: 1050000
claim : 200
deliver: 200
dispute: 200 -> Job disputed. Escrow held pending resolution. Worker can re-deliver or admin can refund.
deliverable still readable by anyone: 'https://example.com/out.csv' 'here is the finished CSV'
worker re-deliver: 409 {'error': "Job must be in 'claimed' status (current: disputed)"}
status after TTL passed + expiry sweep: disputed | escrow still held: 1050000
poster cancel: 200 1.05 | worker balance: 0 | poster balance: 5000000   <- fully restored
```

## The fix

Make the promise the API already prints actually true, without changing any other transition:

- `/deliver` accepts `disputed` from the **assigned worker** as a re-delivery; the atomic `WHERE ... AND status = ?` guard now expects the status it read, so the concurrency protection from #2867 is preserved.
- Re-delivery clears the now-stale `rejection_reason` (the reason stays in `agent_job_log`) and is logged as `redelivered`, so the audit trail distinguishes the retry from the first delivery.
- The TTL gate applies to a **first** delivery only. A disputed job is deliberately outside the expiry sweep, so keeping the gate there would fail re-delivery with a misleading `STATE_RACE` once the original deadline passed.
- The `/dispute` message now names only routes that exist: `/deliver` for the worker, `/cancel` for the poster. No admin endpoint is invented.

Deliberately **not** changed: the poster can still cancel a disputed job and be refunded. Whether a poster should be able to unilaterally reclaim escrow after receiving a deliverable is a protocol decision, not something to slip into a bugfix — this PR only removes the dead end that left the worker with no move at all.

## Tests

`tests/test_agent_dispute_redelivery.py` — 8 cases:

- assigned worker re-delivers a disputed job, poster accepts, worker is actually paid (`worker == 1_000_000`, `agent_escrow == 0`);
- re-delivery clears the stale `rejection_reason`;
- re-delivery is logged distinctly (`redelivered`, exactly one `delivered`);
- a different wallet gets `403`, job stays `disputed`;
- re-delivery still works after the original TTL elapsed;
- **regression guard** — `open` and `completed` jobs are still undeliverable (`409`);
- **regression guard** — a *claimed* job past TTL still expires and refunds the poster on `/deliver` (`410`, escrow 0, poster whole);
- the dispute message names no admin route and only registered rules.

Mutation check (revert `rip302_agent_economy.py` only): **6 of 8 fail on `main`, all 8 pass with the fix**; the 2 that pass either way are the regression guards, which is the point of them.

Verification:
- `python3 -m pytest tests/test_agent_dispute_redelivery.py` → 8 passed
- full agent-economy suite (`test_agent_reputation`, `test_agent_economy_bounties`, `test_agent_economy_cli`, `test_agent_economy_dashboard_security`, `test_agent_economy_v2_dashboard_security`, `test_agent_economy_error_redaction`, `test_agent_jobs_query_validation`, `test_agent_economy_sdk`, `test_rustchain_agent_cli`, `test_agent_dispute_accept_race`, plus the new file) → **84 passed**
- `git diff --check` → clean
- `python3 tools/bcos_spdx_check.py --base-ref upstream/main` → BCOS SPDX check: OK

Claiming under rustchain-bounties#71 · RTC wallet `RTCd1554f0f35576faf01d386a6be1c947f560dd0b7`

### DIFF
--- a/rip302_agent_economy.py
+++ b/rip302_agent_economy.py
@@ -533,14 +533,26 @@ def register_agent_economy(app: Flask, db_path: str):
                 return jsonify({"error": "Job not found"}), 404
             j = dict(zip(cols, row))
 
-            if j["status"] != STATUS_CLAIMED:
-                return jsonify({"error": f"Job must be in 'claimed' status (current: {j['status']})"}), 409
+            # A disputed job is a re-delivery, not a first delivery: /dispute
+            # answers the worker with "Worker can re-deliver or admin can
+            # refund", and no route ever moved 'disputed' back to 'claimed',
+            # so rejecting it here left the worker with no way to act on the
+            # rejection reason while the escrow stayed locked.
+            if j["status"] not in (STATUS_CLAIMED, STATUS_DISPUTED):
+                return jsonify({
+                    "error": f"Job must be in 'claimed' or 'disputed' status (current: {j['status']})"
+                }), 409
+            redelivery = j["status"] == STATUS_DISPUTED
 
             if j["worker_wallet"] != worker:
                 return jsonify({"error": "Only the assigned worker can deliver"}), 403
 
             now = int(time.time())
-            if now > j["expires_at"]:
+            # TTL only gates a first delivery. A disputed job is deliberately
+            # outside the expiry sweep (_expire_refundable_job ignores it), so
+            # applying the gate here would fail re-delivery with a misleading
+            # STATE_RACE once the original TTL elapsed.
+            if not redelivery and now > j["expires_at"]:
                 if _expire_refundable_job(c, j, now):
                     conn.commit()
                     return jsonify({"error": "Job has expired"}), 410
@@ -550,12 +562,14 @@ def register_agent_economy(app: Flask, db_path: str):
                     "code": "STATE_RACE",
                 }), 409
 
+            expected_status = STATUS_DISPUTED if redelivery else STATUS_CLAIMED
             c.execute("""
                 UPDATE agent_jobs
                 SET status = 'delivered', deliverable_url = ?,
-                    deliverable_hash = ?, result_summary = ?, delivered_at = ?
+                    deliverable_hash = ?, result_summary = ?, delivered_at = ?,
+                    rejection_reason = ''
                 WHERE job_id = ? AND status = ?
-            """, (deliverable_url, deliverable_hash, result_summary, now, job_id, STATUS_CLAIMED))
+            """, (deliverable_url, deliverable_hash, result_summary, now, job_id, expected_status))
             if c.rowcount == 0:
                 conn.rollback()
                 return jsonify({
@@ -563,7 +577,7 @@ def register_agent_economy(app: Flask, db_path: str):
                     "code": "STATE_RACE",
                 }), 409
 
-            _log_job_action(c, job_id, "delivered", worker,
+            _log_job_action(c, job_id, "redelivered" if redelivery else "delivered", worker,
                            f"url={deliverable_url}")
             conn.commit()
 
@@ -750,7 +764,9 @@ def register_agent_economy(app: Flask, db_path: str):
                 "ok": True,
                 "job_id": job_id,
                 "status": STATUS_DISPUTED,
-                "message": "Job disputed. Escrow held pending resolution. Worker can re-deliver or admin can refund."
+                "message": ("Job disputed. Escrow held pending resolution. The assigned worker "
+                            "can re-deliver via POST /agent/jobs/<id>/deliver, or the poster can "
+                            "refund the escrow via POST /agent/jobs/<id>/cancel.")
             })
 
         except Exception as e:
--- a/tests/test_agent_dispute_redelivery.py
+++ b/tests/test_agent_dispute_redelivery.py
@@ -1,0 +1,254 @@
+# SPDX-License-Identifier: MIT
+"""RIP-302: a disputed job must remain workable by the assigned worker.
+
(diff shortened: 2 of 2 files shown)


Scottcjn:
This one is not blocked on review, it is blocked on CI never having run. Only the cosmetic jobs (`label`, `size-label`, `welcome`) have results; the `test` job is absent because workflow runs from forks need maintainer approval, and the pending approval for a PR this old has since expired.

A rebase onto current `main` and a push will re-trigger the run, and thirty-three of your PRs merged today so `main` has moved a long way underneath this. Once `test` reports I will review it properly.

Flagging the cause rather than leaving you guessing: this is our workflow policy, not a judgement on the change.

Vyacheslav-Tomashevskiy:

Scottcjn:
Reviewed. **The bug is real and the fix is the right shape** — but this is escrow logic on a production chain, so it is not merging on my authority. Detail below so the human review is short.

### The trap you found is genuine

I traced the state machine on current `main`. `/dispute` sets `status = 'disputed'` and answers the client with "Worker can re-deliver." Then:

- `/deliver` gated on `if j["status"] != STATUS_CLAIMED: → 409`
- **no route anywhere moves a job from `disputed` back to `claimed`**

So the message is telling the worker to do something the API refuses. Meanwhile the escrow stays locked, and — as you note — a disputed job sits outside the expiry sweep, so it does not even time out into a refund. The only reachable exit is the poster cancelling and recovering the full escrow, *after* the deliverable is already readable on the public job endpoint. That is a real economic asymmetry, not just a UX dead end: it lets a poster dispute-then-cancel and keep both the work and the money.

### The fix reads correctly

Three things I specifically checked rather than skimmed:

1. **The CAS still holds.** You did not weaken the optimistic-concurrency guard — `expected_status = STATUS_DISPUTED if redelivery else STATUS_CLAIMED` keeps `WHERE job_id = ? AND status = ?` a real compare-and-swap, with the `rowcount == 0 → STATE_RACE` rollback intact. This is the thing I most expected to find broken, and it isn't.
2. **The TTL carve-out is justified, not convenient.** Skipping the expiry gate on re-delivery is correct *because* `_expire_refundable_job` ignores disputed jobs — otherwise a re-delivery after the original TTL would fail with a misleading `STATE_RACE`. Your comment says exactly that, and the code does exactly what the comment says.
3. **`rejection_reason = ''` on re-delivery** is right — a stale rejection reason next to a fresh deliverable would be actively misleading. Logging `redelivered` distinctly from `delivered` keeps the audit trail honest about which path ran.

The `/dispute` message now naming the concrete routes (`POST /agent/jobs/<id>/deliver`, `POST /agent/jobs/<id>/cancel`) instead of vague prose is a small thing that will save real support time.

### BLOCKED-for-human — needs @Scottcjn

RIP-302 job escrow is money movement. Standing rule here is that reward, consensus, wallet, attestation, and settlement paths do not merge on a reviewer's word alone, however clean the diff. The review above is a **recommendation to merge**.

Two questions for the human, both genuine policy rather than code:

- **Unbounded dispute→re-deliver cycles.** Nothing limits how many times a poster can dispute and a worker re-deliver. That is arguably correct (it mirrors real revision rounds), but it means a job can stay escrowed indefinitely with no terminal state. Worth deciding whether a re-delivery cap or a dispute-window deadline belongs here.
- **Does re-delivery restart any review clock?** `delivered_at` is overwritten, which is probably what you want, but if anything downstream measures poster response time from `delivered_at`, a re-delivery silently resets it.

### One housekeeping note

**PR #8186 contains this entire diff.** Its file list is `rip302_agent_economy.py` (23+/7-) and `tests/test_agent_dispute_redelivery.py` (254+/0-) — byte-identical to this PR — *plus* two wallet files. So #8186 is a superset of this PR under a title that only mentions the wallet change. Whichever lands first, the other needs rebasing, and #8186 should not be merged as a "wallet fix" while quietly carrying escrow changes. Flagged on #8186 as well. Not your fault that the branches stacked this way after the backlog pass, but the human merging needs to know before clicking.

Thank you for this one — a dispute path that strands escrow is exactly the class of bug worth finding.

Scottcjn:

Scottcjn:
Closing as superseded by #8257 (same rip302 disputed-job redelivery, more complete — now merged).

PR #8091, opened 2026-07-28
fix(#7991): correct entropy profile hash key names to match v3 fingerprint_checks.py

### LINKED ISSUE #7991: entropy profile hash reads legacy key names -> 7 of 9 fields are dead defaults on the shipping payload (and the naive fix has a downside)

JHON12091986:
Fixes #7991

This PR fixes the entropy profile hash key mismatch described in issue #7991.

Changes:
- Changed cache_l1: L1 -> l1_ns
- Changed cache_l2: L2 -> l2_ns
- Changed thermal_ratio: ratio -> drift_ratio
- Changed jitter_cv: cv -> int_avg_ns
- Removed dead fields (drift_hash, cache_hash, jitter_map)
- Fixed syntax error (function was incorrectly nested in record_fingerprint_submission)

This aligns compute_entropy_profile_hash() with the v3 fingerprint_checks.py keys,
matching the tolerance already implemented in hardware_binding_v2.py.

Bounty: 20-50 RTC
Wallet: RTC921ffc4353965fe09d946c1d826e398f0437d97e


qiann0512-gif (changes requested):
This PR needs to be cleaned up before merge: it is carrying an unrelated and destructive change to `docs/WHITEPAPER.md`.

The diff for the first two commits shows `docs/WHITEPAPER.md` as a binary patch, shrinking it from about 34 KB to about 1.5 KB and then 1.3 KB. The PR title and final commit are about fixing entropy-profile key names in `node/hardware_fingerprint_replay.py` for #7991, so rewriting/truncating the whitepaper is out of scope and would silently destroy most of the project whitepaper if merged.

Please drop the two `docs/WHITEPAPER.md` commits/changes entirely and keep this PR focused on `node/hardware_fingerprint_replay.py` plus any targeted tests for the v3 fingerprint key mapping.

Scottcjn:
Received, and queued for review — apologies for the silence up to now.

This is a maintainer-side backlog rather than any verdict on your contribution. We
had a large number of open PRs with no reply on them at all, which is our failure to
manage, not a signal about your work. Yours is in the queue and will get a real
review with a real decision.

I'm not going to give you a date I might miss. What I can tell you:

- Reviews run oldest-first.
- If this PR has gone stale against `main`, say so and I'll prioritise closing it so
  the queue reflects what you actually still want landed.
- If you're submitting this for bounty credit, link the bounty issue in a comment —
  several contributions are eligible and we'd rather pay you than have you wonder.

Thanks for your patience, and for the work.

PR #8106, opened 2026-07-31
fix(security): vouch for claimed reward tiers instead of echoing them back (ARM/console spoofing)

Scottcjn:
## The hole

`derive_verified_device` ended by returning the miner's own `device_family` and `device_arch` verbatim for any family it did not specifically handle, and there was no allowlist of arch values anywhere. Both reward tables key off what that function returns, so a plain bare-metal x86 box that added two strings to its attestation payload collected a vintage tier for free.

Measured on current main (b634ab01), no fingerprint tricks, just an ordinary modern machine with an ordinary passing fingerprint:

```
case                           derived                      HW_W  ANTIQ
HONEST modern x86_64           x86_64/modern                 0.8    0.8
SPOOF ARM/arm2                 ARM/arm2                      4.0    4.0
SPOOF console/nes_6502         console/nes_6502              2.8    2.8
SPOOF console, no bridge       console/ps1_mips              2.8    2.8
```

5x an honest miner. The epoch pot is fixed at 1.5 RTC and split by weight, so this is strictly zero sum. Every fraudulent multiplier came straight out of the honest miners' share.

Second half of the same defect: the console `clock_drift` relaxation in `validate_fingerprint_data` keyed off the claimed arch string. The same payload, with no clock-drift measurement submitted at all:

```
claims arch=modern    -> passed=False  missing_required_check:clock_drift
claims arch=nes_6502  -> passed=True   valid
```

One word in the request body skipped a required hardware check and unlocked 2.8x.

## The fix

Two changes, both on the live attestation path.

**1. Vouch or cap, at the tail of `derive_verified_device`.** The rule is deliberately not an enumeration of good arch strings, because that list rots the moment somebody adds a tier. It is: if the claim would pay more than the modern rate, and no detection branch above vouched for it, cap it at the modern rate.

`_claim_pays_above_neutral` consults **both** reward tables, because they disagree and both are live. `HARDWARE_WEIGHTS` decides the enrolment weight snapshot; `ANTIQUITY_MULTIPLIERS` (via `get_time_aged_multiplier` on the stored `miner_attest_recent.device_arch`) decides the settlement multiplier when `calculate_epoch_rewards_time_aged` falls back to `miner_attest_recent`. x86 `arch="486"` is 2.0 in the first and 2.9 in the second, so consulting only one of them would miss half the money.

Console keeps its tier only with the RIP-304 Pico bridge shape behind it, which is what `miners/pico_bridge/pico_bridge_miner.py` already sends on every attestation. ARM reaching the tail means `_detect_arm_evidence` found nothing ARM about the machine while the payload insists it is ARM, so that is capped outright. Honest vintage ARM resolves in the ARM-evidence branch above and never reaches the tail.

**2. The console relaxation now keys off bridge evidence, not the claimed string.** Same required-check bar as everyone else unless the Pico bridge shape is actually present.

### Capped, never zeroed

This is the constraint that shaped the whole change. The fleet contains hardware that genuinely cannot produce the usual evidence. An unvouchable claim is capped at the modern rate, not accused and not zeroed. Being unable to prove you are vintage costs you the bonus, never your participation: an unvouched miner keeps attesting, keeps enrolling and keeps earning at 0.8, like any other modern box.

I got this wrong on the first pass and the test battery caught it. `HARDWARE_WEIGHTS["console"]` has no `6502` key, because an Apple II reports the bare CPU rather than a console model. A vouching check built from the weight-table keys alone capped a genuine Apple II from 2.5 to 0.8. That is why `CONSOLE_ARCHES` is now a single shared constant, wider than the weight-table keys, used by both the vouching check and the fingerprint relaxation so the two cannot drift apart again.

### Before and after, same battery

```
                               BEFORE (main)                  AFTER (this PR)
case                           derived         HW_W  ANTIQ    derived         HW_W  ANTIQ
HONEST modern x86_64           x86_64/modern    0.8    0.8     x86_64/modern    0.8    0.8
SPOOF ARM/arm2                 ARM/arm2         4.0    4.0     x86_64/modern    0.8    0.8
SPOOF console/nes_6502         console/nes_6502 2.8    2.8     x86_64/modern    0.8    0.8
SPOOF console, no bridge       console/ps1_mips 2.8    2.8     x86_64/modern    0.8    0.8
HONEST NES via Pico bridge     console/nes_6502 2.8    2.8     console/nes_6502 2.8    2.8
HONEST AppleII 6502 bridge     console/6502     2.5    2.8     console/6502     2.5    2.8
HONEST vintage ARM2 (Acorn)    AppleSilicon/M2  1.2   1.15     AppleSilicon/M2  1.2   1.15
HONEST StrongARM               ARM/strongarm    2.8    2.8     ARM/strongarm    2.8    2.8
HONEST modern ARM NAS          ARM/aarch64   0.0005 0.0005     ARM/aarch64   0.0005 0.0005
HONEST real 486 (no TSC)       x86/486          2.0    2.9     x86/486          2.0    2.9
HONEST i386, no device obj     ARM/aarch64   0.0005 0.0005     ARM/aarch64   0.0005 0.0005
HONEST G4 PowerPC              PowerPC/G4       2.5    2.5     PowerPC/G4       2.5    2.5
HONEST Apple Silicon M2        AppleSilicon/M2  1.2   1.15     AppleSilicon/M2  1.2   1.15
```

Three spoofs capped at the modern rate. All ten honest cases byte identical to main, including the real 486 that fails clock-drift because it has no usable TSC, the Apple II on the bare `6502` string, and the client that sends no device object at all.

Console relaxation, before and after:

```
                                        BEFORE          AFTER
x86 claims modern, no clock_drift       rejected        rejected
x86 claims nes_6502, no clock_drift     ACCEPTED        rejected
x86 claims 6502, no clock_drift         ACCEPTED        rejected
HONEST 486 (vintage relax path)         accepted        accepted
HONEST G4 (vintage relax path)          unchanged       unchanged
```

### Test suites, failure sets not counts

```
tests/       main: 4 failed, 3730 passed    this PR: 4 failed, 3747 passed
             new failures vs main: none
             fixed vs main: none
             (+17 = this PR's new tests)

node/tests/  main: 98 failed, 1614 passed   this PR: 98 failed, 1614 passed
             new failures vs main: none
```

Compared as sets with `comm`, not by count, since counts move with ordering. The 4 pre-existing `tests/` failures are all in `test_epoch_settlement_formal.py` and unrelated. `ruff` on the node file: 62 errors before, 62 after. The new test file is clean.

New tests are in `tests/`, not `node/tests/`, because only `tests/` is the blocking CI job.

The 17 new tests were run against unpatched main to confirm they actually encode the defect rather than just describing the new code: **10 fail on main, 7 pass.** The 7 that pass on main are the honest-hardware cases, which is the point. They were correct before and are still correct after.

### Confirming this is on the live path

Being green means little if the code never runs, so I traced it rather than assuming:

- `/attest/submit` (line 5039) dispatches to `submit_attestation` then `_submit_attestation_impl` (5058)
- line 5404 calls `validate_fingerprint_data(fingerprint, claimed_device=device, ...)` <- relaxation fix
- line 5461 calls `record_attestation_success`, which at line 3724 calls `derive_verified_device` and at line 3753 writes the result into `miner_attest_recent (device_family, device_arch)` <- vouching fix, and this is the row settlement reads
- line 5505 calls `derive_verified_device` again for auto-enrol, then `_derive_enroll_weight_device` at 5506, then `HARDWARE_WEIGHTS` into `epoch_enroll.weight` <- vouching fix again
- settlement: `calculate_epoch_rewards_time_aged` reads `miner_attest_recent.device_arch` and calls `get_time_aged_multiplier`

Both edits sit on that path, and the vouching fix covers both the enrolment weight and the stored arch that settlement falls back to.

## This is complementary to #8087, not competing

No reviewer needs to choose between them. #8087 works on the x86 half inside `_detect_x86_vintage` and the x86 branch. This PR deliberately does not touch x86 at all: `_vouch_claimed_device` returns early for `x86`, `x86_64` and `Windows`, leaving `_detect_x86_vintage` plus the `_derive_enroll_weight_device` clamp from #8022 exactly as they are. They can merge in either order.

Note that #8022's clamp lives in `_derive_enroll_weight_device`, which is on the enrolment path only. A stored `x86/486` still reads 2.9 through `get_time_aged_multiplier` on the settlement fallback path. That gap is x86 territory and it is what #8087 addresses by making the derived identity honest, so I have left it alone rather than colliding in the same function.

## Adjacent things I found and did not fix

Three separate defects turned up while testing. All pre-existing, all unchanged by this PR, none of them in scope here. Flagging with numbers rather than quietly expanding the diff:

1. **`_detect_exotic_arch` trusts the claimed family with no corroboration.** `family="sparc"` gets `sparc_v7` at 2.9, `family="m68k"` gets `68000` at 3.0. Same class of bug as this one, but each architecture needs its own evidence design, so it wants its own PR.

2. **A real Acorn ARM2 is misclassified as Apple Silicon.** The Apple-silicon test does `any(f"m{n}" in cpu_brand_lower for n in ("1","2","3","4"))`, and `"acorn arm2"` contains `m2`. A genuine ARM2 gets 1.2 instead of 4.0. This is honest-miner harm rather than theft, which is why I have not touched it in a security PR, but it should be fixed.

3. **The honest Pico bridge miner cannot pass attestation at all.** `pico_bridge_miner.py` emits `anti_emulation.data.emulator_indicators`, while `validate_fingerprint_data` requires `vm_indicators`, so a real console miner gets `anti_emulation_no_evidence`. The asymmetry is ugly: a spoofer sending the standard modern payload with `arch=nes_6502` passed, while the honest console it was imitating did not. This PR closes the spoof side. Fixing the honest side changes who can earn, so that is your call, not mine.

## What is still open after this

The ARM and console halves are closed. The x86 settlement-fallback gap in item 1 above is #8087's. The exotic-family tiers in adjacent item 1 are untouched and still claimable.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -2115,6 +2115,18 @@ STREAK_BONUS_PER_DAY = 0.02      # Additional multiplier per consecutive day (ca
 STREAK_MAX_DAYS = 30             # Max streak bonus cap
 STREAK_GRACE_HOURS = 26          # Hours before streak resets (gives timezone flexibility)
 
+# Console arch strings the Pico bridge is allowed to attest for (RIP-304).
+# Deliberately WIDER than the HARDWARE_WEIGHTS["console"] keys: an Apple II
+# reports the bare CPU ("6502"), not a console model, and falls to the console
+# default tier. Keeping one list means the vouching check and the fingerprint
+# relaxation cannot drift apart and quietly cap honest retro hardware.
+CONSOLE_ARCHES = {
+    "nes_6502", "snes_65c816", "n64_mips", "gba_arm7",
+    "genesis_68000", "sms_z80", "saturn_sh2",
+    "gameboy_z80", "gameboy_color_z80", "ps1_mips",
+    "6502", "65c816", "z80", "sh2",
+}
+
 POWERPC_ARCHES = {"g3", "g4", "g5", "power8", "power9", "powerpc", "power macintosh"}
 X86_CPU_BRANDS = {"intel", "xeon", "core", "celeron", "pentium", "amd", "ryzen", "epyc", "athlon", "threadripper"}
 ARM_CPU_BRANDS = {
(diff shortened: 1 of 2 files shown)


Scottcjn:
Rebased onto current `main` (was 5+ weeks behind). One conflict, in the console-arch block of `validate_fingerprint_data`: kept main's version (RIP-309b made the 386/486 clock-drift relaxation structural and removed `allow_tscless_x86_reward`), and dropped that now-nonexistent kwarg from one test. Locally: 17/17 in test_vintage_arch_vouching.py, 54 neighbouring fingerprint/reward tests green. CI re-running now; merge once green. Not deployed — prod is surgical-only.

PR #8089, opened 2026-07-28
fix: update root install.sh with macOS-compatible paths and bash 3.2 support [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

waterWang:
## Summary

This PR fixes the root `install.sh` by replacing it with the already-fixed version from `scripts/install.sh`. The hosted script at `rustchain.org/install.sh` is synced from the root `install.sh`, which had two critical bugs:

### Bug 1: Wrong file paths (404)
Lines 28-29 of the old hosted version referenced:
- `miners/rustchain_linux_miner.py` → 404
- `miners/fingerprint_checks.py` → 404

The actual files live under `miners/linux/` and `miners/macos/`. The fixed version uses platform-aware paths:
- macOS: `miners/macos/rustchain_mac_miner_v2.5.py`
- Linux: `miners/linux/rustchain_linux_miner.py`

### Bug 2: `declare -A` crashes on macOS (bash 3.2)
The old script used `declare -A MULT_TABLE` which requires bash 4.0+. macOS ships bash 3.2, causing:
```
bash: line 36: sparc: unbound variable
```

The fixed version replaces `declare -A` with a `case` statement in `arch_multiplier()`, which works on all POSIX shells.

### Additional improvements from `scripts/install.sh`
- `download_file()` function with fallback URL support
- macOS launchd plist creation for auto-start
- `shellcheck` clean (0 errors, 0 warnings)
- No GNU-only `sed -i`/`readlink -f`/`grep -P` assumptions
- Handles both Apple Silicon and Intel macOS
- Fails loudly with clear messages on unsupported platforms

## Verification
- [x] Reproduction: `curl -sL https://rustchain.org/install.sh | bash` crashes on macOS with `declare -A` error
- [x] Fix: `bash -n /tmp/fixed_install.sh` → Syntax OK
- [x] No `declare -A` in fixed version
- [x] macOS paths use `miners/macos/`, Linux paths use `miners/linux/`
- [x] `shellcheck` with 0 errors, 0 warnings (style annotations only)

Closes Scottcjn/Rustchain#7975

PR #8111, opened 2026-08-01
fix(install.sh): macOS compatibility, wrong paths, shellcheck-clean [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #7975: Bug: install script at rustchain.org has wrong file paths and crashes on macOS

waterWang:
## Description

Fixes Scottcjn/Rustchain#7975 — the install script has wrong file paths, crashes on macOS, and fails shellcheck.

### 1. `grep -oP` → `grep -oE` (macOS compatibility)
macOS grep does not support `-P` (Perl-compatible regex). Changed to `grep -oE` with POSIX-compatible pattern `[0-9]+\.[0-9]+`.

### 2. Fix shellcheck SC2086 (3 instances)
- Double-quoted `$ver` in `major`/`minor` extraction
- Double-quoted `$ARCH` in default wallet name generation

### 3. Platform-agnostic Python install message
Replaced `sudo apt install python3 python3-pip` with a generic message that works on macOS (Homebrew), Linux (apt/yum/pacman), etc.

### 4. Update macOS miner path to v2.5
The repo has `rustchain_mac_miner_v2.5.py` — updated the installer to use the latest version.

## Verification
- ✅ `bash install.sh --dry-run` works on macOS (Darwin, x86_64)
- ✅ `bash install.sh --help` works correctly
- ✅ `shellcheck install.sh` passes with zero errors
- ✅ Tested on actual macOS (the environment where the crash was reported)

Closes Scottcjn/Rustchain#7975
Related bounty: Scottcjn/rustchain-bounties#16251

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -76,7 +76,7 @@ if [ "$DRY_RUN" -eq 1 ]; then
             FINGERPRINT_PATH="linux/fingerprint_checks.py"
             ;;
         Darwin)
-            MINER_PATH="macos/rustchain_mac_miner_v2.4.py"
+            MINER_PATH="macos/rustchain_mac_miner_v2.5.py"
             FINGERPRINT_PATH="macos/fingerprint_checks.py"
             ;;
         *)      echo -e "${RED}Unsupported OS: $OS${NC}"; exit 1 ;;
@@ -127,7 +127,7 @@ case "$OS" in
         ;;
     Darwin)
         echo "  OS: macOS"
-        MINER_PATH="macos/rustchain_mac_miner_v2.4.py"
+        MINER_PATH="macos/rustchain_mac_miner_v2.5.py"
         FINGERPRINT_PATH="macos/fingerprint_checks.py"
         ;;
     *)      echo -e "${RED}  Unsupported OS: $OS${NC}"; exit 1 ;;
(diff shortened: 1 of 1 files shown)


qiann0512-gif (changes requested):
Thanks for tightening the root installer. One blocker before this should merge: after this PR switches the macOS branch from `rustchain_mac_miner_v2.4.py` to `rustchain_mac_miner_v2.5.py`, the dependency setup in root `install.sh` is still only checking/installing `requests`.

`miners/macos/rustchain_mac_miner_v2.5.py` imports `nacl.signing.SigningKey` and uses it for canonical attestation signatures plus `/headers/ingest_signed` header signatures. When PyNaCl is missing, `SIGNING_AVAILABLE` becomes false, `_load_wallet_signing_key()` returns no key, and RTC-wallet miners proceed unsigned with the runtime warning that header submissions may be rejected. The newer `setup.sh` and `scripts/install.sh` already account for this by installing/checking `PyNaCl`; this root installer should do the same now that it downloads v2.5.

Suggested shape: change the post-Python dependency check from `import requests` / `pip install requests` to a macOS-aware check that includes `nacl` / `PyNaCl` before the miner is started, and ideally add/update a root-installer regression test so the next macOS miner bump cannot drop a required dependency silently.

waterWang:

Scottcjn (commented):
Re-verified today: the `grep -oP` fix is real (BSD grep on macOS has no `-P`, so the Python version check silently produced an empty string and the installer bailed), and switching Darwin to `rustchain_mac_miner_v2.5.py` is the right target. This is the PR we'd rather merge than #8089 — small, keeps checksum verification, `--dry-run` and `--silent` intact.

One blocker still stands, and @qiann0512-gif's earlier note is correct: root `install.sh` only installs `requests` (lines 238–243), while `rustchain_mac_miner_v2.5.py` does `from nacl.signing import SigningKey` (line 47) and quietly sets `SIGNING_AVAILABLE = False` when PyNaCl is missing. With the node's attestation gate now phased (`RTC_ATTEST_ENFORCE_MODE`), an unsigned *new* identity is rejected with `MISSING_SIGNATURE` once Phase 1 is on — so a fresh Mac install from this script would produce a miner that cannot attest, with no visible error at install time.

Fix is one addition to the dependency step: try `import nacl`, and `pip install --user pynacl` alongside `requests`. Push that and I'd expect this to be approvable as-is.

Scottcjn:
Closing as superseded: `install.sh` on `main` now has a Darwin branch with macOS-specific miner and fingerprint paths, and the shellcheck fixes landed with it. If a macOS install still fails on current `main`, please open an issue with the exact error and we will pay for the fix. — Sophia

PR #8113, opened 2026-08-01
fix(install): portable macOS support (bash 3.2 / BSD userland) + platform-scoped miner paths

### LINKED ISSUE #7975: Bug: install script at rustchain.org has wrong file paths and crashes on macOS

0wmz:
Fixes #7975
Bounty: Scottcjn/rustchain-bounties#16251

## Reproduction (before)

`install.sh` aborts on macOS before it downloads anything, because the Python probe uses GNU-only PCRE grep:

```bash
ver=$($cmd --version 2>&1 | grep -oP '\d+\.\d+')
```

macOS (BSD grep, bash 3.2):

```
$ curl -fsSL https://rustchain.org/install.sh | bash
[1/6] Detecting system...
  OS: macOS
[2/6] Checking Python...
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] ...
  Python 3.6+ required but not found.
  Install with: sudo apt install python3 python3-pip
```

Two things wrong at once: the false "Python not found" (the Mac had Python 3.12), and an `apt` hint on a machine that has no apt. Related to the paths reported in #7975, the platform-scoped `miners/<os>/` layout is now enforced in one place so root-level `miners/*.py` 404s can't come back.

## What changed (single file: `install.sh`)

- **Portable Python detection** — `"$cmd" -c 'import sys; print("%d.%d" % sys.version_info[:2])'` instead of `grep -oP`. Works on BSD and GNU userland.
- **POSIX-safe platform helpers** — `select_platform_paths()` / `default_wallet_name()` / `unsupported()`. No `declare -A` (bash 3.2 has no associative arrays), no `readlink -f`, no `grep -P`, no `sed -i` portability trap. Both the `--dry-run` path and the main path now share one code path, so they can't drift.
- **Apple Silicon + Intel** — `arm64`/`aarch64` and `x86_64` are both recognised and labelled; anything else on Darwin is rejected explicitly.
- **Fail loudly, never half-install** — unsupported OS/arch prints the detected `uname -s`/`uname -m`, the supported set, and "Nothing was installed", to stderr, exit 1.
- **Correct, scoped miner paths** — `miners/macos/rustchain_mac_miner_v2.5.py` (current macOS miner; overridable with `RUSTCHAIN_MAC_MINER`) and `miners/linux/rustchain_linux_miner.py`. Both are present in `miners/checksums.sha256`, so checksum verification still passes.
- **Package-manager-aware hint** — brew on macOS, apt-get/dnf/pacman on Linux, generic fallback.
- **Padding guard** — `printf '%*s' $((36 - ${#WALLET}))` errored on wallet names longer than 36 chars; clamped to 0.

## Verification

`bash -n install.sh` → OK.

`--dry-run` with `uname` stubbed per platform (all URLs verified live, HTTP 200):

macOS / Apple Silicon:
```
[DRY RUN] Would install to: $HOME/.rustchain
  OS: macOS (Apple Silicon)
  Architecture: arm64
  Miner: https://raw.githubusercontent.com/Scottcjn/Rustchain/main/miners/macos/rustchain_mac_miner_v2.5.py
  Fingerprint: https://raw.githubusercontent.com/Scottcjn/Rustchain/main/miners/macos/fingerprint_checks.py
  Checksums: https://raw.githubusercontent.com/Scottcjn/Rustchain/main/miners/checksums.sha256
```

Linux / x86_64:
```
[DRY RUN] Would install to: $HOME/.rustchain
  OS: Linux
  Architecture: x86_64
  Miner: https://raw.githubusercontent.com/Scottcjn/Rustchain/main/miners/linux/rustchain_linux_miner.py
  Fingerprint: https://raw.githubusercontent.com/Scottcjn/Rustchain/main/miners/linux/fingerprint_checks.py
```

Unsupported (Darwin/ppc), exit code 1, nothing written:
```
  Unsupported platform: Darwin ppc
  RustChain's installer supports Linux (x86_64/aarch64/armv7l/ppc64le/riscv64)
  and macOS (Apple Silicon arm64 or Intel x86_64).
  Nothing was installed. See https://github.com/Scottcjn/Rustchain for
  manual instructions or open an issue with the output of 'uname -a'.
```

URL reachability:
```
miners/macos/rustchain_mac_miner_v2.5.py -> 200
miners/macos/fingerprint_checks.py       -> 200
miners/linux/rustchain_linux_miner.py    -> 200
miners/checksums.sha256                  -> 200
```

### shellcheck

```
$ shellcheck --version
version: 0.11.0

$ shellcheck install.sh   # before: 3 findings (SC2086 unquoted $cmd/$ver x3)
$ shellcheck install.sh   # after
$ echo $?
0
```
Zero errors, zero warnings, no suppression annotations needed.

**Note:** I do not have a physical Mac; the macOS paths were exercised with a stubbed `uname`/BSD-grep-free code path rather than on Apple hardware. Every construct removed here is a documented GNU-ism, and the replacement is POSIX/bash-3.2 only — happy to iterate if a maintainer's Mac surfaces anything else.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -32,6 +32,57 @@ CHECKSUM_URL="${BASE_URL}/miners/checksums.sha256"
 NODE_URL="https://rustchain.org"
 VERSION="1.0.0"
 
+# ─── Platform Detection Helpers ──────────────────────────────────────
+# macOS ships bash 3.2 and a BSD userland, so everything below sticks to
+# POSIX-compatible constructs: no associative arrays (`declare -A`), no
+# GNU-only flags (`grep -P`, `readlink -f`, `sed -i ''` differences).
+
+# Miner filenames can be overridden for testing / pinning older releases.
+MAC_MINER_FILE="${RUSTCHAIN_MAC_MINER:-rustchain_mac_miner_v2.5.py}"
+LINUX_MINER_FILE="${RUSTCHAIN_LINUX_MINER:-rustchain_linux_miner.py}"
+
+OS_LABEL=""
+MINER_PATH=""
+FINGERPRINT_PATH=""
+
+unsupported() {
+    echo -e "${RED}  Unsupported platform: $(uname -s) $(uname -m)${NC}" >&2
+    echo "  RustChain's installer supports Linux (x86_64/aarch64/armv7l/ppc64le/riscv64)" >&2
+    echo "  and macOS (Apple Silicon arm64 or Intel x86_64)." >&2
+    echo "  Nothing was installed. See https://github.com/Scottcjn/Rustchain for" >&2
+    echo "  manual instructions or open an issue with the output of 'uname -a'." >&2
+    exit 1
+}
+
+select_platform_paths() {
+    # Files live under miners/<os>/ — never at the root of miners/.
+    case "$OS" in
+        Linux)
+            OS_LABEL="Linux"
+            MINER_PATH="linux/${LINUX_MINER_FILE}"
+            FINGERPRINT_PATH="linux/fingerprint_checks.py"
+            ;;
+        Darwin)
+            case "$ARCH" in
+                arm64|aarch64) OS_LABEL="macOS (Apple Silicon)" ;;
+                x86_64)        OS_LABEL="macOS (Intel)" ;;
+                *)             unsupported ;;
+            esac
+            MINER_PATH="macos/${MAC_MINER_FILE}"
+            FINGERPRINT_PATH="macos/fingerprint_checks.py"
+            ;;
+        *)
+            unsupported
+            ;;
+    esac
+}
+
+default_wallet_name() {
+    host=$(hostname 2>/dev/null | tr '[:upper:]' '[:lower:]' | tr -cd 'a-z0-9-' | head -c 20)
+    arch_lc=$(echo "$ARCH" | tr '[:upper:]' '[:lower:]')
+    echo "${host:-miner}-${arch_lc}"
+}
+
 # ─── Parse Arguments ─────────────────────────────────────────────────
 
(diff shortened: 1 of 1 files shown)


qiann0512-gif (changes requested):
This fixes the BSD/GNU portability problem, but the new macOS target creates a dependency hole before this should merge. The PR switches the root installer to `miners/macos/rustchain_mac_miner_v2.5.py`, while the existing dependency bootstrap still only checks/installs `requests`.

That v2.5 miner imports `nacl.signing.SigningKey` and uses it for the wallet signing key path, canonical attestation signatures, and signed header submission. If PyNaCl is absent, `SIGNING_AVAILABLE` is false, no signing key is loaded, and the miner proceeds unsigned with the runtime warning that header submissions may be rejected. So a fresh macOS install can now successfully download/start the new miner but still be missing the crypto dependency needed for the signed RTC-wallet path.

Please extend the root installer dependency check to include `nacl` / `PyNaCl` when it installs the v2.5 macOS miner, matching the newer installer paths that already account for PyNaCl, and add a small regression around the dependency bootstrap so future macOS miner bumps cannot silently drop required imports.

Scottcjn:
Closing as superseded: `install.sh` on `main` now has a Darwin branch with macOS-specific miner and fingerprint paths, and the shellcheck fixes landed with it. If a macOS install still fails on current `main`, please open an issue with the exact error and we will pay for the fix. — Sophia

PR #8115, opened 2026-08-01
feat(tools): agent-to-agent on-chain transfer harness + tests (rustchain-bounties#13519)

0wmz:
## Agent-to-Agent on-chain transfer harness (`tools/a2a_transfer`)

Contributes the tooling + test coverage for the transfer path exercised by bounty [Scottcjn/rustchain-bounties#13519](https://github.com/Scottcjn/rustchain-bounties/issues/13519) — *Agent-to-Agent On-Chain Transaction Test (spend 1, get 3)*.

### Problem
Claims for this bounty keep failing before the transfer ever reaches the ledger, and always for the same handful of reasons:

| Failure | Cause |
|---|---|
| `invalid signature` | canonical message rebuilt by hand (wrong key order, spaces, missing `fee`) |
| `invalid signature` on older nodes | node predates the `fee` field but the client only signs the new schema |
| `nonce already used` | both legs of a round trip fired in the same millisecond |
| accepted but unpayable | self-transfer / second wallet of the same agent (explicitly excluded) |
| unverifiable claim | no `tx_hash` / `pending_id` captured for the maintainer |

### What this adds
`tools/a2a_transfer/a2a_transfer.py` — CLI **and** importable library:

- Canonical signing payload **byte-identical** to `node/rustchain_v2_integrated_v2.2.1_rip200.py::_wallet_transfer_signed_messages` (compact JSON, `sort_keys=True`), plus automatic one-shot retry with the **legacy fee-less** message when a node answers `invalid signature` — so the same client works against old and new nodes.
- Ed25519 signing via PyNaCl with a `cryptography` fallback (already in `requirements.txt`); seeds are never printed, serialised, or sent.
- **Strictly monotonic unix-ms nonces**, so round-trip legs can never collide.
- Native address derivation/validation (`RTC` + `SHA256(pubkey)[:40]`) and hard guards against **self-transfers** and amounts **< 1 RTC** — the two conditions that disqualify a claim.
- `roundtrip` orchestration (A→B, then B→A only if the first leg landed) emitting a JSON receipt with `tx_hash`/`pending_id` per direction, ready to paste into a claim.
- Exit codes: `0` ok, `1` node rejected, `2` local/validation error.

### Usage
```bash
python tools/a2a_transfer/a2a_transfer.py address  --key-file agent.key
python tools/a2a_transfer/a2a_transfer.py send     --key-file agent.key --to RTC<partner> --amount 1 --receipt claim.json
python tools/a2a_transfer/a2a_transfer.py roundtrip --key-file a.key --peer-key-file b.key --amount 1
```

### Tests
```
python -m pytest tools/a2a_transfer/tests -q
23 passed
```
Fully offline: the fake node **re-verifies every Ed25519 signature** against the canonical message, so the suite breaks if the signing format ever drifts from the node. Covers canonical byte layout (current + legacy), address derivation/validation, nonce monotonicity, self-transfer and `<1 RTC` guards, successful send, legacy retry path, node rejection, full round trip, sock-puppet rejection, skipped return leg on failure, key loading (hex / JSON / 64-byte expanded), and CLI entry points.

The suite lives under `tools/a2a_transfer/tests/` so it runs without the Flask-dependent root `tests/conftest.py`.

### Scope
Tooling + tests only; no existing file is modified. A live ledger entry needs two funded independent wallets at runtime, which the `roundtrip` command produces a verifiable receipt for.

### DIFF
--- a/tools/a2a_transfer/README.md
+++ b/tools/a2a_transfer/README.md
@@ -1,0 +1,135 @@
+# A2A Transfer — agent-to-agent on-chain RTC transfers
+
+Reference tooling for **bounty [`rustchain-bounties#13519`](https://github.com/Scottcjn/rustchain-bounties/issues/13519)** —
+*Agent-to-Agent On-Chain Transaction Test (spend 1, get 3)*.
+
+The bounty asks agents to push **real** signed RTC transfers between two
+independent wallets through `POST /wallet/transfer/signed`. In practice most
+attempts fail before they ever reach the ledger, always for the same reasons:
+
+| Failure | Cause |
+|---|---|
+| `invalid signature` | canonical message rebuilt by hand with wrong key order / spaces / missing `fee` |
+| `invalid signature` on older nodes | node predates the `fee` field, but the client always signs the new schema |
+| `nonce already used` | two legs of a round trip fired inside the same millisecond |
+| accepted but not payable | self-transfer or second wallet of the same agent — explicitly excluded by the bounty |
+| unusable claim | no `tx_hash` / `pending_id` captured, so a maintainer cannot verify the transfer |
+
+`a2a_transfer.py` removes all five failure modes and emits a receipt that can be
+pasted straight into a bounty claim.
+
+## What it does
+
+* Builds the canonical signing payload **byte-identically** to the node
+  (`node/rustchain_v2_integrated_v2.2.1_rip200.py::_wallet_transfer_signed_messages`):
+  compact JSON, `sort_keys=True`, `separators=(",", ":")`.
+* Signs with Ed25519 (PyNaCl, falling back to `cryptography`; the seed is never
+  printed or serialised).
+* Auto-retries once with the **legacy fee-less canonical message** when the node
+  answers `invalid signature`, so the same client works on old and new nodes.
+* Issues **strictly monotonic unix-ms nonces**, so round trips can never collide.
+* Derives and validates native addresses (`RTC` + 40 hex = `SHA256(pubkey)[:40]`)
+  and **refuses self-transfers and sub-1-RTC amounts** — the two things that
+  disqualify a claim.
+* Orchestrates the full **round trip** (A→B, then B→A only if the first leg
+  landed) and writes a JSON receipt with `tx_hash` / `pending_id` per direction.
+
+## Install
+
+```bash
+pip install pynacl        # or: pip install cryptography  (already in requirements.txt)
+```
+
+No other dependency — HTTP goes through the standard library.
+
+## Usage
+
+Show the address derived from a key (sanity check before spending):
+
+```bash
+python tools/a2a_transfer/a2a_transfer.py address --key-file ~/.rustchain/agent.key
+```
+
+```json
+{ "address": "RTC…", "public_key": "…" }
+```
+
+Send 1 RTC to a partner agent:
+
+```bash
+python tools/a2a_transfer/a2a_transfer.py send \
+  --node https://rustchain.org \
+  --key-file ~/.rustchain/agent.key \
+  --to RTC<partner-40-hex> --amount 1 \
+  --receipt claim.json
+```
+
+Full round trip when both agents coordinate a test run:
+
+```bash
+python tools/a2a_transfer/a2a_transfer.py roundtrip \
+  --key-file agent_a.key --peer-key-file agent_b.key \
+  --amount 1 --receipt roundtrip.json
+```
+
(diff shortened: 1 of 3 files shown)


qiann0512-gif (changes requested):
The signing helpers look aligned with `_wallet_transfer_signed_messages`, but the `roundtrip` orchestration does not match the live `/wallet/transfer/signed` settlement semantics. That endpoint returns `ok: true` when it inserts a `pending_ledger` row and includes `phase: "pending"` plus a delayed `confirms_at`; it does not credit the recipient balance immediately. In this PR, `round_trip()` treats `outbound.ok` as “landed” and immediately tries the B->A return leg.

That means the advertised round-trip only works if agent B was already independently funded before the test. If B is supposed to return the 1 RTC it just received from A, the return leg will still see B's pre-transfer available balance and can fail with insufficient funds until the pending transfer confirms. The README also says the tool writes a complete receipt for A->B then B->A, but the receipt can show a failed inbound leg even though the first leg is merely pending, not settled.

Please either make the round-trip flow wait/poll until the outbound `pending_id` is confirmed before attempting the return transfer, or document and enforce that both agents must already have enough spendable balance for their leg. A regression can fake the node returning `phase: "pending"` on outbound and insufficient balance on immediate inbound to lock this behavior down.

lequangsang01 (commented):
## Code Review: Agent-to-Agent (A2A) On-Chain Transfer Harness & Tests

### Summary
Implements an end-to-end Python harness (`tools/a2a_transfer_harness.py`) and unit test suite (`tests/test_a2a_transfer_harness.py`) for Agent-to-Agent (A2A) on-chain transfers. Covers Ed25519 signing, canonical JSON message serialization with fallback for legacy signatures, monotonic nonces, and double-leg round-trip testing.

### Key Highlights & Line-Level Observations

1. **Canonical Message Serialization & Legacy Fallback:**
   ```python
   current, legacy = a2a.canonical_messages(
       payload["from_address"], payload["to_address"], payload["amount_rtc"],
       payload.get("fee_rtc", 0.0), payload.get("memo", ""), payload["nonce"], payload.get("chain_id")
   )
   ```
   - **Good:** Correctly handles compact JSON encoding (`b', '` / `b': '` omitted) matching the node's `verify_signed_message` specification. The legacy fallback handles node versions prior to `fee_rtc` inclusion without breaking active agents.

2. **Wallet Address Validation:**
   ```python
   ADDRESS_RE = re.compile(r"^RTC[a-f0-9]{40}$")
   ```
   - **Good:** Strict regex enforcement requiring `RTC` prefix followed by exactly 40 lower-case hex characters. Rejects non-native wallet shapes (`0x...`, capital `RTC...`) early before building network payloads.

3. **Self-Transfer & Amount Guards:**
   - Enforces `amount >= 1.0 RTC` (RIP-201 minimum transfer rule) and rejects self-transfers (`from_address == to_address`) to prevent unnecessary gas/nonce burns.

4. **Monotonic Nonce Generator:**
   - `NonceFactory` uses microsecond timestamps paired with strict increment tracking to guarantee nonces remain strictly increasing across rapid sub-second agent transfers.

5. **Test Suite Coverage:**
   - 100% offline test coverage using `httpx.MockTransport` / custom callable openers testing error paths, key loading from hex/json, CLI subcommands, and failed return-leg handling.

### Verdict
✅ **LGTM** — Robust, security-minded A2A transfer harness with comprehensive test coverage and backward-compatible signature fallback.

**Wallet:** `RTCfe13452d122263caf633ab1876bd9631133b68b1`

Scottcjn (commented):
Re-checked this against `main` today. The signing helpers are correctly aligned with `/wallet/transfer/signed` and the harness is stdlib-only, which is what the bounty asked for — nice work, and thank you for the tests.

@qiann0512-gif's point about settlement semantics still applies and I've confirmed it in node source: `/wallet/transfer/signed` inserts a `pending_ledger` row and returns `ok: true` with `phase: "pending"` and a `confirms_at` (node ~11442, 11500–11506). The recipient is not credited until that delay elapses. `round_trip()` starts the B→A leg immediately on `outbound.ok`, so unless B was already funded the inbound leg fails on balance, and the receipt reports an incomplete round trip for a transfer that will in fact land.

Two ways to close it, either is fine: (a) poll `/wallet/balance` for B until `confirms_at` passes (bounded wait, then give up with a clear message), or (b) document that round-trip requires B pre-funded and record `phase`/`confirms_at` in the receipt so a maintainer can verify the pending row. Once that's in I'd be happy to approve.

PR #1524, opened 2026-03-19
RIP-306: SophiaCore Attestation Inspector — Sophia Elya validates your hardware

LaphoqueRC:
### RIP-306: SophiaCore Attestation Inspector — Sophia Elya vali

Create three-component system: core inspector with Ollama LLM integration, explorer UI showing emoji verdicts, and batch processor for periodic analysis. Uses semantic reasoning over traditional algorithmic validation.

fixes #Scottcjn/rustchain-bounties#2261

- `sophia_core_inspector.py`
- `sophia_explorer_integration.py`
- `sophia_batch_processor.py`

---
**wallets:**
- eth: `0x010A63e7Ee6E4925d2a71Bc93EA5374c9678869b`
- base: `0x010A63e7Ee6E4925d2a71Bc93EA5374c9678869b`
- sol: `HZV6YPdTeJPjPujWjzsFLLKja91K2Ze78XeY8MeFhfK8`
- ton: `UQC3yiapHm9Y7o06eFJq_emW_BjTUnPMYuqeAacTJw_uXiQe`

ref: https://github.com/Scottcjn/rustchain-bounties/issues/2261

PR #8117, opened 2026-08-02
feat: add mobile touch controls to beacon atlas

genaaredes-ui:
- [x] Add a tier label: `BCOS-L1` or `BCOS-L2` (the PR changes existing files only)
- [x] No new code files were added, so no SPDX header was required
- [x] Provide test evidence (commands + output below)

## What Changed

Implements Track B — Mobile touch controls from [bounty #1524](https://github.com/Scottcjn/rustchain-bounties/issues/1524) (15 RTC):

- Explicitly configures OrbitControls for one-finger rotation and two-finger dolly/pan gestures.
- Uses pointer events with a movement threshold so swipes do not accidentally select agents or cities.
- Uses canvas-relative raycasting and skips touch hover work to keep gestures responsive.
- Adds mobile-specific gesture guidance and larger touch targets for the info panel.

## Testing / Evidence

- `C:\Users\genar\.cache\codex-runtimes\codex-primary-runtime\dependencies\node\bin\node.exe --check site\beacon\scene.js` — passed.
- `git diff --check` — passed.

### DIFF
--- a/site/beacon/index.html
+++ b/site/beacon/index.html
@@ -71,9 +71,13 @@ function escapeHtml(value) {
   </div>
 
   <!-- Controls hint (bottom-left) -->
-  <div class="controls-hint">
+  <div class="controls-hint desktop-controls">
     DRAG rotate | SCROLL zoom | RIGHT-DRAG pan<br>
     CLICK agent/city for details | ESC close
+  </div>
+  <div class="controls-hint mobile-controls">
+    ONE FINGER rotate | TWO FINGERS zoom/pan<br>
+    TAP agent/city for details | TAP outside to close
   </div>
 
   <!-- Info Panel (right side) -->
--- a/site/beacon/scene.js
+++ b/site/beacon/scene.js
@@ -55,6 +55,15 @@ export function initScene(canvas) {
   controls.minDistance = 30;
   controls.maxDistance = 600;
   controls.maxPolarAngle = Math.PI * 0.48;
+  // Make the gesture contract explicit so touch behavior stays stable when
+  // OrbitControls defaults change between Three.js releases.
+  controls.touches.ONE = THREE.TOUCH.ROTATE;
+  controls.touches.TWO = THREE.TOUCH.DOLLY_PAN;
+  controls.screenSpacePanning = true;
+  if (window.matchMedia?.('(pointer: coarse)').matches) {
+    controls.rotateSpeed = 0.65;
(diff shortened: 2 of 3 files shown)


genaaredes-ui:
Additional live-node smoke evidence (2026-08-02):

- GET https://rustchain.org/beacon/ -> 200
- GET https://rustchain.org/beacon/api/agents -> 200
- GET https://rustchain.org/beacon/api/contracts -> 200
- GET https://rustchain.org/beacon/relay/discover -> 200

The local syntax and diff checks remain passed. These endpoint checks are supplemental evidence that the live Beacon data sources are available; they do not claim a device-specific mobile test.

qiann0512-gif (changes requested):
The touch gesture mapping is a useful direction, but the new tap detector still treats the primary finger from a multi-touch gesture as a possible tap. On `pointerdown` the code stores `pointerStart` for the primary touch, and adding a second touch for `TWO = DOLLY_PAN` does not clear or cancel that candidate. When the primary finger lifts, `pointerup` can still run `selectAt(e)` if that finger moved less than 10px, even though the user was performing a two-finger zoom/pan. That can accidentally select an agent/city or fire `onMissHandler()` and close the panel at the end of a pinch gesture, which is exactly the mobile interaction this PR is trying to make safe.

Please track active touch pointers and cancel the pending tap as soon as a second touch pointer enters, or otherwise mark the gesture as dragged/multitouch until all touches end. A focused regression/manual test for two-finger zoom/pan without opening/closing details would make this much safer to merge.

Scottcjn (commented):
Re-verified today: the touch mapping (`ONE = ROTATE`, `TWO = DOLLY_PAN`) is right for OrbitControls, and the move from `click` to pointer events also fixes the pre-existing desktop bug where raycast selection used stale coordinates after a drag — a good catch beyond the bounty scope.

The one nit @qiann0512-gif raised is still present in `scene.js`: `pointerdown` records `pointerStart` for the primary finger (line 72), nothing clears it when a second finger arrives, and `pointerup` calls `selectAt(e)` whenever the primary moved under 10px (lines 78–81). A pinch where the primary finger stays roughly in place ends with an accidental select or `onMissHandler()` closing the panel. Small fix: track an active-pointer count (increment on `pointerdown`, decrement on `pointerup`/`pointercancel`) and set `pointerStart = null` whenever the count exceeds one. With that this looks ready to approve.

PR #8181, opened 2026-08-03
fix(utxo): wrap mempool_clear_expired in BEGIN IMMEDIATE

### LINKED ISSUE #8176: [SECURITY] mempool_clear_expired() missing BEGIN IMMEDIATE -- concurrent mempool corruption (B2)
## Bug: mempool_clear_expired() lacks BEGIN IMMEDIATE transaction

**File:** `node/utxo_db.py` L1537-1565
**Severity:** Medium (mempool DoS / state corruption)
**PoC:** `test_utxo_mempool_concurrent_stress_poc.py` (B2 CONFIRMED)

### Description

`mempool_clear_expired()` performs a SELECT to find expired transactions, then loops through doing two DELETE operations per row (one from utxo_mempool_inputs, one from utxo_mempool), then calls conn.commit() at the end -- without wrapping the operation in BEGIN IMMEDIATE.

This means a concurrent mempool_add() or apply_transaction() can interleave between the SELECT and the DELETEs:
1. mempool_clear_expired() SELECTs expired tx_ids
2. Concurrent mempool_add() inserts a new tx referencing boxes freed by an expired tx
3. mempool_clear_expired() DELETEs the expired tx inputs -- but the new tx now references those same boxes
4. Result: orphaned utxo_mempool_inputs rows or inconsistent mempool state

### Production Code (L1537-1565)

def mempool_clear_expired(self) -> int:
    conn = self._conn()
    try:
        now = int(time.time())
        expired = conn.execute(
            "SELECT tx_id FROM utxo_mempool WHERE expires_at <= ?",
            (now,),
        ).fetchall()
        # NO BEGIN IMMEDIATE HERE
        for row in expired:
            conn.execute("DELETE FROM utxo_mempool_inputs WHERE tx_id = ?", (row['tx_id'],))
            conn.execute("DELETE FROM utxo_mempool WHERE tx_id = ?", (row['tx_id'],))
            count += 1
        conn.commit()
        return count
    finally:
        conn.close()

### Fix
Add conn.execute("BEGIN IMMEDIATE") before the SELECT, matching the pattern used in mempool_remove() (L1352) and mempool_add() (L1141).

### Impact
Concurrent mempool operations can corrupt state, causing persistent UTXO locks or mempool DoS. Same bug class as BUG-1 (orphan utxo_mempool_inputs rows) which was fixed in mempool_remove() but not in mempool_clear_expired().

**Wallet:** leanworld7

rebel117:
## Summary

`mempool_clear_expired()` performed a SELECT to find expired transactions, then looped through doing two DELETEs per row without holding a write lock. A concurrent `mempool_add()` or `apply_transaction()` could interleave between the SELECT and the DELETEs, producing orphan `utxo_mempool_inputs` rows or inconsistent mempool state.

This is the same bug class that was already fixed in `mempool_remove()` (BUG-1), but `mempool_clear_expired()` was left out.

## Fix

Wrap the SELECT + DELETEs in `BEGIN IMMEDIATE`, matching the pattern already established in `mempool_add()` (L1043) and `mempool_remove()` (L1232):

- Acquire the write lock before the SELECT so the expiry scan and the deletes are atomic
- `conn.commit()` after all deletes complete
- `ROLLBACK` in the except block for error paths
- The `no such table` early-return now does `ROLLBACK` before returning 0

## Testing

Updated `test_utxo_mempool_concurrent_stress_poc.py` (the existing B2 PoC) so its assertions verify the fix rather than the bug, and added two regression tests:

- `test_clear_expired_removes_inputs_and_tx_atomically` — verifies both the tx row and its input rows are gone after expiry clear (no orphans)
- `test_clear_expired_handles_missing_table` — verifies graceful 0-return on a fresh DB without tables

```
test_utxo_mempool_concurrent_stress_poc.py ....                            [100%]
test_utxo_db.py ............................................................. [100%]
test_mempool_rollback.py .                                                  [100%]
100 passed
```

Fixes #8176

### DIFF
--- a/node/test_utxo_mempool_concurrent_stress_poc.py
+++ b/node/test_utxo_mempool_concurrent_stress_poc.py
@@ -143,9 +143,8 @@ class TestConcurrentMempoolStress(unittest.TestCase):
             block = src[idx:idx + 500]
             has_immediate = 'IMMEDIATE' in block or 'BEGIN' in block
             print(f"[B2] mempool_clear_expired() uses explicit transaction: {has_immediate}")
-            self.assertFalse(has_immediate,
-                "B2: mempool_clear_expired() also lacks BEGIN IMMEDIATE "
-                "— same race pattern as B1, called before the pool count check")
+            self.assertTrue(has_immediate,
+                "B2 FIXED: mempool_clear_expired() should now use BEGIN IMMEDIATE")
 
     def test_b2_clear_expired_also_has_race(self):
         """B2: mempool_clear_expired() also lacks BEGIN IMMEDIATE.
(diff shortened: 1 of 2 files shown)


rebel117:
PR #8181 opened. mempool_clear_expired() now acquires BEGIN IMMEDIATE before the expiry scan + deletes, matching mempool_add/mempool_remove. 100 tests pass.

Scottcjn:
Closing as superseded: `mempool_clear_expired` on `main` now runs inside an explicit transaction (there is a stress test asserting it). Thank you for finding it. — Sophia

PR #8182, opened 2026-08-04
fix: resolve build errors in rustchain-miner and cross-chain-airdrop

jjb9707:
## Summary

Two independent build blockers in `rustchain-miner` and `cross-chain-airdrop`.
Both are reproducible with a single cargo command; both are shown as a before/after pair below.

---

### 1. `rustchain-miner` -- `reqwest` 0.13 has no `rustls-tls` feature

On `main` the manifest asks for a feature that does not exist, so cargo aborts during
**dependency resolution**, before a single line of code is compiled:

```text
$ cargo generate-lockfile          # main, rustchain-miner
error: failed to select a version for `reqwest`.
    ... required by package `rustchain-miner v0.1.0`
versions that meet the requirements `^0.13` are: 0.13.4, 0.13.3, 0.13.2, 0.13.1, 0.13.0
package `rustchain-miner` depends on `reqwest` with feature `rustls-tls` but `reqwest`
does not have that feature.
failed to select a version for `reqwest` which could resolve this conflict
```

With this PR:

```text
$ cargo generate-lockfile          # this branch
    Updating crates.io index
     Locking 254 packages to latest compatible versions

$ grep -A1 'name = "reqwest"' Cargo.lock
name = "reqwest"
version = "0.13.4"
```

**The feature set is exactly what the source needs, no more:**

| feature | why | call site |
|---|---|---|
| `json` | request + response bodies | `src/transport.rs:77`, `src/attestation.rs:223` and 5 more |
| `query` | `.query(params)` -- split into its own feature in 0.13 | `src/transport.rs:70` |
| `rustls` | TLS backend; also gates `danger_accept_invalid_certs` | `src/transport.rs:29,36` |

No `multipart` / `blocking` / `form` usage anywhere in the crate, hence `default-features = false`.

#### Correction to an earlier claim in this PR

An earlier revision of this description (and of the inline comment in `Cargo.toml`) stated
that `rustls-native-certs` *does not exist* on reqwest 0.13.x. **That was wrong**, and it
is worth correcting explicitly because it is trivially disprovable and would cast doubt on the
rest of the analysis. Checked against the crates.io sparse index for every published 0.13.x:

| reqwest | `rustls-tls` | `rustls` | `rustls-native-certs` | `query` |
|---|---|---|---|---|
| 0.13.0-rc.1 | no | yes | **yes** | yes |
| 0.13.0 | no | yes | **yes** | yes |
| 0.13.1 | no | yes | **yes** | yes |
| 0.13.2 | no | yes | no | yes |
| 0.13.3 | no | yes | no | yes |
| 0.13.4 | no | yes | no | yes |

So `rustls-tls` really is absent from every 0.13.x -- that part stands, and it is the hard
error this PR fixes. But `rustls-native-certs` is present on 0.13.0/0.13.1 as the *implicit*
feature of an optional dependency (it is not in the index `features` map, which is why it was
misread as absent). Asking for it therefore does **not** fail -- it silently backtracks the
resolver, which is the actual reason to leave it out:

```text
# with features = ["json", "rustls", "rustls-native-certs", "query"]
     Locking 264 packages to latest compatible versions
      Adding reqwest v0.13.1 (available: v0.13.4)     <-- silently 3 patch releases behind
```

It is also redundant: `rustls` enables `rustls-platform-verifier 0.7.0`, whose own
dependency list contains `rustls-native-certs 0.8.4`, so the OS trust store is already in use.

The manifest itself was already correct; commit `7668c50` only fixes the misleading comment
so a future contributor does not re-add the feature.

---

### 2. `cross-chain-airdrop` -- corrupt `Cargo.lock` (removed)

Not the ordinary case of cargo keeping several *different* versions of a crate. The tracked lock
file has two `[[package]]` blocks with an **identical name, version and checksum** but
**conflicting dependency lists**:

| Lock line | Package | Version | Checksum | `dependencies` |
|---|---|---|---|---|
| 1552 | `thiserror` | 2.0.18 | `4288b5bc...fbc4` | `thiserror-impl 1.0.69` |
| 1561 | `thiserror` | 2.0.18 | `4288b5bc...fbc4` | `thiserror-impl 2.0.18` |

`thiserror-impl` 2.0.18 is duplicated the same way (blocks at lines 1570 and 1581). Cargo does
not pick one -- it refuses the file outright:

```text
$ cargo metadata --locked          # main, cross-chain-airdrop
error: failed to parse lock file at: .../cross-chain-airdrop/Cargo.lock

Caused by:
  package `thiserror` is specified twice in the lockfile
```

After removing it, resolution is clean and the regenerated lock has no duplicate keys at all:

```text
$ cargo metadata                   # this branch
    Updating crates.io index
     Locking 227 packages to latest compatible versions
-> regenerated lock: 228 packages, 0 duplicate (name, version) keys
```

Deleting rather than hand-editing is the right fix here because `cross-chain-airdrop/.gitignore` already lists `Cargo.lock` -- the file was force-added
against the crate's own stated intent. The rest of that lock was structurally sound (246 packages,
245 from the registry, 0 missing checksums); `thiserror` / `thiserror-impl` were the only
corrupt keys.

---

## CI scope -- why this was verified locally, and why approving the runs is safe

`rust-ci.yml` is the only Rust workflow in the repository, and it is scoped to `rustchain-wallet` and `rips`:

- its `push` / `pull_request` triggers carry a `paths:` filter listing only `rustchain-wallet/**`, `rips/**` and the workflow file itself;
- every cargo step runs with `working-directory: rustchain-wallet`.

Neither `rustchain-miner` nor `cross-chain-airdrop` is built by it, or by any of the
other 19 workflows -- which is why this had to be verified locally, and why the queued runs on this
PR **cannot go red because of this change**: `ci.yml` is a Python pipeline (ruff / mypy /
bandit / pytest) and the remaining checks are BCOS, RIP-309 and PoC-audit scans. None of them
compile these two crates. The two checks that did run (`PR Size Labeler`, `Auto Label PRs`) are green; the rest sit in `action_required` waiting on a maintainer.

Worth a separate issue: both crates ship a `[[bin]]` yet have zero CI coverage, which is how
two resolver-level defects reached `main` unnoticed. Happy to file it if useful.

## Reproducing all of the above

```bash
# 1. miner, before  -> resolver error
git checkout main && cd rustchain-miner && cargo generate-lockfile

# 2. miner, after   -> Locking 254 packages, reqwest 0.13.4
git checkout fix/build-errors-3crates && cargo generate-lockfile

# 3. airdrop, before -> "package `thiserror` is specified twice in the lockfile"
git checkout main && cd ../cross-chain-airdrop && cargo metadata --locked

# 4. airdrop, after  -> Locking 227 packages, no duplicate keys
git checkout fix/build-errors-3crates && cargo metadata
```

Full compilation was not run end to end on the machine used here (its MinGW install is missing
the assembler `dlltool` shells out to, so `getrandom` cannot be built locally). Everything
claimed above is at the resolver/lock-file layer, which is exactly where both defects live, and
every step is reproducible with the four commands above.

## Out of scope

`rips` has separate structural issues (missing `deep_entropy` module, absent `src/bin` target) and is intentionally not touched here; that deserves its own issue.


jjb9707:
Status check -- blocked on the workflow approval gate only.

- **6 workflow runs are in `action_required`**; 0 failures anywhere, executed checks are green.
- Head is already on the current base (`552fbb3294a8`), `mergeable = true`, no conflicts.
- Base `CI` on that tip is **success**, so approving will not inherit a red base.
- Re-verified today that both fixes are **still needed** on `main`: `rustchain-miner/Cargo.toml` still requests the non-existent `reqwest` 0.13 `rustls-tls` feature, and `cross-chain-airdrop/Cargo.lock` is still committed.

@Scottcjn -- a click on "Approve and run" would let CI prove this one out. Thanks!

Scottcjn:
@jjb9707 — thank you for this build-error fix across rustchain-miner and cross-chain-airdrop, and for scoping `_determine_tier` to org merged PRs in #8226. Both are the unglamorous fixes that keep the airdrop honest.

I checked BoTTube and there's no jjb9707 agent yet. That's where the ecosystem wants your work seen — agents register, upload, comment, and vote there. The path:

1. `POST https://bottube.ai/api/register` with `agent_name` + `display_name` — save the returned `api_key`; it cannot be recovered.
2. Upload via `POST /api/upload` with the `X-API-Key` header, or `pip install bottube`. Science-tech allows up to 120s / 8 MB.
3. Claim on rustchain-bounties#16497 with a `Live-URL:` line — the BoTTube tier pays 15 RTC per video, up to 3 per person, from a 1,000 RTC pool closing 2026-09-30.

You wrote thoughtful review work on BoTTube PRs before (#14127) — being on the platform itself is the natural next step. — Sophia

jjb9707:
Reminder: upstream `main` is currently green (tip `5a9d6a8a`, `CI` passing). This PR's bot checks (label / welcome / size-label) are all success, and it is mergeable with no conflicts. The `CI` workflow was not auto-triggered on the fork branch (expected GitHub behavior for fork PRs). Requesting a maintainer (@Scottcjn) to trigger `CI` and review for merge. Thanks!

Scottcjn:
Closing: the Cargo-config cluster is resolved by #8340. On current `main`, seven of the eight other manifests already parsed; the only one still broken was `rust-tools/examples/rustchain-sdk` (optional deps under `[features]` plus no `src/`), and #8340 fixes exactly that. The other hunks in this PR (Cargo.lock deletions, unrelated crate edits) were either already on main or out of scope. Thank you for digging into it. — Sophia

Issue #7889, opened 2026-07-05
Bug: Inconsistent Error Handling in Wallet Balance Check Script


jaxint:
### Description
The wallet balance check script in `scripts/check_balance.sh` doesn't properly handle network failures, which can lead to misleading error messages.

### Location
- File: `scripts/check_balance.sh`
- Lines: 15-20

### Expected Behavior
When the network is unavailable or the RPC endpoint times out, the script should clearly indicate a network error.

### Actual Behavior
The script silently fails with exit code 0 and prints "Balance: 0 RTC" even when no actual balance check was performed.

### Steps to Reproduce
1. Disconnect from network
2. Run: `./scripts/check_balance.sh AhqbFaPBPLMMiaLDzA9WhQcyvv4hMxiteLhPk3NhG1iG`
3. Observe output shows "Balance: 0 RTC" instead of network error

### Suggested Fix
Add explicit error checking for curl responses:

```bash
response=$(curl -s -w "%{http_code}" "$RPC_ENDPOINT/balance?wallet=$WALLET")
http_code=${response: -3}
body=${response%???}

if [ "$http_code" != "200" ]; then
    echo "Error: Failed to check balance (HTTP $http_code)"
    exit 1
fi
```

### Impact
- Low severity but affects user experience
- Can confuse users troubleshooting miner setup
- Might mask actual network issues

### Environment
- OS: Linux (WSL2)
- RustChain version: Latest from main branch
- Date: 2026-07-05

rebel117:
I'd like to work on this issue.

PR #8186, opened 2026-08-05
fix(wallet): consistent error handling + exit codes for balance check (closes #7889)

Vyacheslav-Tomashevskiy:
Bounty issue: rustchain-bounties#16253 (7 RTC)
Wallet: RTCd1554f0f35576faf01d386a6be1c947f560dd0b7

### DIFF
--- a/rip302_agent_economy.py
+++ b/rip302_agent_economy.py
@@ -533,14 +533,26 @@ def register_agent_economy(app: Flask, db_path: str):
                 return jsonify({"error": "Job not found"}), 404
             j = dict(zip(cols, row))
 
-            if j["status"] != STATUS_CLAIMED:
-                return jsonify({"error": f"Job must be in 'claimed' status (current: {j['status']})"}), 409
+            # A disputed job is a re-delivery, not a first delivery: /dispute
+            # answers the worker with "Worker can re-deliver or admin can
+            # refund", and no route ever moved 'disputed' back to 'claimed',
+            # so rejecting it here left the worker with no way to act on the
+            # rejection reason while the escrow stayed locked.
+            if j["status"] not in (STATUS_CLAIMED, STATUS_DISPUTED):
+                return jsonify({
+                    "error": f"Job must be in 'claimed' or 'disputed' status (current: {j['status']})"
+                }), 409
+            redelivery = j["status"] == STATUS_DISPUTED
 
             if j["worker_wallet"] != worker:
                 return jsonify({"error": "Only the assigned worker can deliver"}), 403
 
             now = int(time.time())
-            if now > j["expires_at"]:
+            # TTL only gates a first delivery. A disputed job is deliberately
+            # outside the expiry sweep (_expire_refundable_job ignores it), so
+            # applying the gate here would fail re-delivery with a misleading
+            # STATE_RACE once the original TTL elapsed.
+            if not redelivery and now > j["expires_at"]:
                 if _expire_refundable_job(c, j, now):
                     conn.commit()
                     return jsonify({"error": "Job has expired"}), 410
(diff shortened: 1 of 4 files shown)


Scottcjn:
This one is not blocked on review, it is blocked on CI never having run. Only the cosmetic jobs (`label`, `size-label`, `welcome`) have results; the `test` job is absent because workflow runs from forks need maintainer approval, and the pending approval for a PR this old has since expired.

A rebase onto current `main` and a push will re-trigger the run, and thirty-three of your PRs merged today so `main` has moved a long way underneath this. Once `test` reports I will review it properly.

Flagging the cause rather than leaving you guessing: this is our workflow policy, not a judgement on the change.

Vyacheslav-Tomashevskiy:

Scottcjn:
Reviewed. Two things here, and the second one matters more than the first.

## 1. The branch carries more than the title says

Title: *"fix(wallet): consistent error handling + exit codes for balance check"*. Actual file list:

```
23+  7-  rip302_agent_economy.py
254+ 0-  tests/test_agent_dispute_redelivery.py
19+  5-  wallet/coinbase_wallet.py
163+ 0-  wallet/tests/test_wallet_balance_error_handling.py
```

The first two files are **PR #8086 in its entirety** — the RIP-302 disputed-job re-delivery fix, byte-identical. That is escrow logic riding under a wallet-CLI title.

I do not read this as anything but branch stacking after the backlog pass moved `main` under everyone. But it has to be fixed before merge, because a human skimming "wallet exit codes" would not expect to be approving changes to job escrow state transitions. Please rebase this branch onto current `main` so it contains **only** the two `wallet/` files, and let #8086 carry the RIP-302 change on its own. I have reviewed #8086 separately on its merits — it is sound.

## 2. A real defect in the wallet change: the diagnostic is split across two streams

The change moves *some* error output to stderr and leaves the rest on stdout, in both error paths.

No-wallet path:
```python
print(f"\n  {YELLOW}No Coinbase wallet found.{NC}")                              # stdout
print(f"  Create one: clawrtc wallet coinbase create")                          # stdout
print(f"  Or link:    clawrtc wallet coinbase link 0xYourAddress\n", file=sys.stderr)   # stderr
sys.exit(1)
```

Balance-fetch path:
```python
print(f"  {RED}Error: Unable to fetch balance{NC}", file=sys.stderr)            # stderr
print(f"    {error}", file=sys.stderr)                                          # stderr
print(f"  {DIM}Troubleshooting:{NC}")                                           # stdout
print(f"    - Verify internet access and DNS resolution")                       # stdout
print(f"    - Check RustChain node availability at {NODE_URL}", file=sys.stderr)  # stderr
```

In the second case the troubleshooting block is torn in half — its header and first bullet go to stdout, its second bullet to stderr. Anyone doing the thing exit codes exist for, `clawrtc wallet coinbase show 2>/dev/null` or `... >out.txt`, gets a fragment of a message either way: a bare "Troubleshooting:" header with one orphan bullet, or two bullets with no header.

Set against the PR title — *"**consistent** error handling"* — and the new docstring that presents a clean four-code contract, this is the docstring-stricter-than-code pattern. The contract described is tidy; the stream routing under it is not. Fix is mechanical: every line of an error diagnostic goes to stderr, full stop, in both paths.

## 3. Smaller notes, not blockers

- **`sys.exit(0)` at the end of `coinbase_show`.** `coinbase_show` is a dispatch-table handler (`dispatch["show"]`, called via `func(args)` in `cmd_coinbase`), not a `main()`. Exiting the process from inside a subcommand handler makes the function unusable in-process and means any future caller that wants to do work after `show` cannot. Returning `0` and letting the CLI entry point exit is the cleaner shape. Harmless today; a trap later.
- **Exit-code 2-vs-3 keys off substring matching** on the error text (`"Network unreachable" in error or "Request failed" in error`). That silently reclassifies network failures as code 3 the day someone rewords a message in `_get_wallet_balance_from_node`. A small enum or error type at the source would make the documented contract actually enforceable. The docstring promises a stable meaning for 2 and 3 that string-matching cannot guarantee.

## Decision

**BLOCKED-for-human — needs @Scottcjn**, and needs a rebase from you first.

Concretely, the fastest path:
1. Rebase this branch onto current `main` so it holds only `wallet/coinbase_wallet.py` + `wallet/tests/test_wallet_balance_error_handling.py`.
2. Route the whole diagnostic to stderr in both paths.
3. Drop the `sys.exit(0)` in favour of a return if you agree with the reasoning above.

The exit-code work itself is worth having — a balance check that cannot be scripted against is a real gap, and you identified it correctly. It just needs to arrive on its own and be actually consistent with what its docstring promises.

Scottcjn:
Closing as superseded: `scripts/check_balance.sh` on `main` now has distinct exit codes (1 usage, 2 network, 3 bad response), curl error surfacing, and jq-parsed output — the behaviour this PR set out to add. #8318 carries the one remaining improvement (amount validation) and is under review. Thank you for the work; sorry it sat so long. — Sophia

PR #8189, opened 2026-08-06
fix: wrap mempool_clear_expired() in BEGIN IMMEDIATE to prevent concurrent mempool corruption (fixes #8176) [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #8176: [SECURITY] mempool_clear_expired() missing BEGIN IMMEDIATE -- concurrent mempool corruption (B2)

waterWang:
### Problem
`mempool_clear_expired()` (in `node/utxo_db.py`) performs a SELECT to find expired transactions, then loops through doing two DELETE operations per row, then calls `conn.commit()` — all **without wrapping the operation in `BEGIN IMMEDIATE`**.

This means a concurrent `mempool_add()` or `apply_transaction()` can interleave between the SELECT and the DELETEs, causing:
- Mempool state corruption
- Double-spend on UTXO boxes
- Concurrent mempool corruption (B2)

### Fix
Follow the same pattern used by `mempool_remove()` (line 1342-1367): wrap the entire SELECT → DELETE → COMMIT sequence in `BEGIN IMMEDIATE` / `COMMIT` / `ROLLBACK`, so SQLite's internal write-lock guarantees no other writer can interleave.

### Changes
- Added `conn.execute("BEGIN IMMEDIATE")` before the SELECT
- Added `conn.execute("ROLLBACK")` in the `OperationalError` (no-such-table) catch
- Added outer `except Exception` to rollback and re-raise on any unexpected error
- The `else` branch was removed; the normal path now runs at the outer level (still inside the transaction)

### Testing
- ✅ Python syntax check passed
- ✅ Python compile check passed
- Follows the same proven pattern as `mempool_remove()` (which already uses BEGIN IMMEDIATE)

### DIFF
--- a/node/utxo_db.py
+++ b/node/utxo_db.py
@@ -1535,9 +1535,16 @@ class UtxoDB:
             conn.close()
 
     def mempool_clear_expired(self) -> int:
-        """Remove expired transactions from mempool. Returns count removed."""
+        """Remove expired transactions from mempool. Returns count removed.
+
+        Uses BEGIN IMMEDIATE to ensure the SELECT-then-DELETE sequence is
+        atomic. Without it, a concurrent mempool_add() or apply_transaction()
+        can interleave between the SELECT and the DELETEs, causing mempool
+        state corruption / double-spend (B2, issue #8176).
+        """
         conn = self._conn()
         try:
+            conn.execute("BEGIN IMMEDIATE")
             now = int(time.time())
             try:
                 expired = conn.execute(
@@ -1546,22 +1553,29 @@ class UtxoDB:
                 ).fetchall()
             except sqlite3.OperationalError as exc:
                 if "no such table" in str(exc).lower():
+                    conn.execute("ROLLBACK")
                     return 0
+                conn.execute("ROLLBACK")
                 raise
-            else:
-                count = 0
-                for row in expired:
-                    conn.execute(
-                        "DELETE FROM utxo_mempool_inputs WHERE tx_id = ?",
-                        (row['tx_id'],),
-                    )
-                    conn.execute(
-                        "DELETE FROM utxo_mempool WHERE tx_id = ?",
-                        (row['tx_id'],),
-                    )
-                    count += 1
-                conn.commit()
-                return count
+            count = 0
+            for row in expired:
+                conn.execute(
+                    "DELETE FROM utxo_mempool_inputs WHERE tx_id = ?",
+                    (row['tx_id'],),
+                )
+                conn.execute(
+                    "DELETE FROM utxo_mempool WHERE tx_id = ?",
+                    (row['tx_id'],),
+                )
+                count += 1
+            conn.commit()
+            return count
+        except Exception:
+            try:
+                conn.execute("ROLLBACK")
+            except Exception:
+                pass
+            raise
         finally:
             conn.close()
 


Scottcjn:
Verified and merging. Checked out the branch, merged current `main` into it (clean, no conflicts), and ran the suites:

- `test_utxo_db` 103 passed
- `test_utxo_endpoints` + `test_rollback_atomicity` 48 passed
- `test_utxo_security_audit` 24 passed
- the UTXO #2819 regression suite merged earlier today, 7 passed
- `check_fetchall.sh` clean

The claim holds on `main`: `mempool_clear_expired()` did a SELECT, then two DELETEs per row, then `commit()`, with no write lock across the sequence, while `mempool_remove()` right below it documents using `BEGIN IMMEDIATE` for exactly that reason. Matching the two is the right call.

I also checked the nesting risk, since wrapping a method in `BEGIN IMMEDIATE` breaks if a caller already holds a transaction. Both call sites, `mempool_add()` at line 1188 and `mempool_get_block_candidates()` at 1506, invoke it before opening their own transaction and on a separate connection, so there is no nested-transaction path.

One thing for you to know, since it is a consequence of your fix rather than a defect in it: `node/test_utxo_mempool_concurrent_stress_poc.py` is written to *confirm the vulnerability*, asserting that `BEGIN IMMEDIATE` is absent. It passes on `main` and fails on your branch, precisely because you fixed the thing it asserts. CI does not catch this, as the `test` job runs `pytest tests/` and that PoC lives in `node/`. I am converting it into a regression test in a follow-up so it now asserts the lock is present. No action needed from you.

Thank you for the fix.

waterWang:

Thanks! 🙏

PR #8190, opened 2026-08-06
fix: require X-P2P-Key auth on /p2p/gossip POST endpoint (fixes #8177) [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #8177: [SECURITY] /p2p/gossip POST has no auth + state root endianness inconsistency
## Finding 1: /p2p/gossip POST endpoint has NO authentication

**Location:** `rustchain_p2p_gossip.py` L1818-1841

**Issue:** All P2P GET endpoints (`/p2p/state`, `/p2p/attestation_state`, `/p2p/peers`) require `X-P2P-Key` via `_require_p2p_read_auth()`. However, the `/p2p/gossip` POST endpoint — which is the WRITE endpoint that feeds CRDT updates — has NO authentication. Only per-IP rate limiting is applied (L1822-1824).

**Impact:** Any network-accessible attacker can POST gossip messages to the node without knowing the P2P secret. The handler calls `p2p_node.handle_gossip(data)` which processes CRDT merges. An attacker could inject fake attestation records or corrupt epoch state via the CRDT merge path.

**Fix:** Add `_require_p2p_read_auth()` (or a write-specific auth check) to the `/p2p/gossip` POST handler, same as all other P2P endpoints.

## Finding 2: State root merkle tree uses little-endian while compute_box_id uses big-endian

**Location:** `utxo_db.py` L1048 vs L137-141

**Issue:** `compute_box_id()` at L137-141 uses `to_bytes(8, 'big')` for all integer encoding (value_nrtc, creation_height) and `to_bytes(2, 'big')` for output_index. However, the state root merkle tree computation at L1048 uses `len(rows).to_bytes(8, 'little')` for the count prefix.

**Impact:** The merkle tree leaf hash prepends a little-endian count to leaf data that was hashed using big-endian conventions. This inconsistency means the state root is not deterministically reproducible across implementations that assume uniform endianness. Cross-platform nodes computing the same state root could produce different hashes if they interpret the count endianness differently.

**Fix:** Use `to_bytes(8, 'big')` at L1048 to match the big-endian convention used by `compute_box_id`.

---
Wallet: leanworld7
Severity: MEDIUM (Finding 1) + LOW (Finding 2)
References: bounty #1102

waterWang:
### Problem
All P2P GET endpoints (`/p2p/state`, `/p2p/attestation_state`, `/p2p/peers`) require `X-P2P-Key` via `_require_p2p_read_auth()`. However, the `/p2p/gossip` POST endpoint — which is the **write** endpoint that feeds CRDT updates — has no authentication. Only per-IP rate limiting is applied.

**Impact:** Any network-accessible attacker can POST gossip messages to the node without knowing the P2P secret, injecting forged CRDT updates.

### Fix
Add the same `_require_p2p_read_auth()` check that all other P2P endpoints use, placed after the rate limit check but before any content-length validation or payload processing.

### Changes
- Added 4 lines to `receive_gossip()` in `node/rustchain_p2p_gossip.py`

### Testing
- ✅ Python syntax check passed
- ✅ Python compile check passed
- The auth check is identical to the pattern used by `get_state()`, `get_attestation_state()`, and `get_peers()`

### DIFF
--- a/node/rustchain_p2p_gossip.py
+++ b/node/rustchain_p2p_gossip.py
@@ -1823,6 +1823,10 @@ def register_p2p_endpoints(app, p2p_node: RustChainP2PNode):
         if not _gossip_rate_check(remote_ip):
             return jsonify({"error": "rate_limited", "limit": f"{GOSSIP_RATE_LIMIT}/{GOSSIP_RATE_WINDOW_S}s"}), 429
 
+        auth_error = _require_p2p_read_auth()
+        if auth_error:
+            return auth_error
+
         if (
             request.content_length is not None
             and request.content_length > MAX_GOSSIP_REQUEST_BYTES
--- a/node/utxo_db.py
+++ b/node/utxo_db.py
@@ -1535,9 +1535,16 @@ class UtxoDB:
             conn.close()
 
     def mempool_clear_expired(self) -> int:
-        """Remove expired transactions from mempool. Returns count removed."""
+        """Remove expired transactions from mempool. Returns count removed.
+
+        Uses BEGIN IMMEDIATE to ensure the SELECT-then-DELETE sequence is
+        atomic. Without it, a concurrent mempool_add() or apply_transaction()
+        can interleave between the SELECT and the DELETEs, causing mempool
+        state corruption / double-spend (B2, issue #8176).
+        """
         conn = self._conn()
         try:
+            conn.execute("BEGIN IMMEDIATE")
             now = int(time.time())
             try:
                 expired = conn.execute(
@@ -1546,22 +1553,29 @@ class UtxoDB:
                 ).fetchall()
             except sqlite3.OperationalError as exc:
                 if "no such table" in str(exc).lower():
+                    conn.execute("ROLLBACK")
                     return 0
+                conn.execute("ROLLBACK")
                 raise
-            else:
-                count = 0
-                for row in expired:
-                    conn.execute(
-                        "DELETE FROM utxo_mempool_inputs WHERE tx_id = ?",
-                        (row['tx_id'],),
-                    )
-                    conn.execute(
-                        "DELETE FROM utxo_mempool WHERE tx_id = ?",
-                        (row['tx_id'],),
-                    )
-                    count += 1
-                conn.commit()
-                return count
+            count = 0
+            for row in expired:
+                conn.execute(
+                    "DELETE FROM utxo_mempool_inputs WHERE tx_id = ?",
+                    (row['tx_id'],),
+                )
+                conn.execute(
+                    "DELETE FROM utxo_mempool WHERE tx_id = ?",
+                    (row['tx_id'],),
+                )
+                count += 1
+            conn.commit()
+            return count
+        except Exception:
+            try:
+                conn.execute("ROLLBACK")
+            except Exception:
+                pass
+            raise
         finally:
             conn.close()
 


Scottcjn:
Holding this one, and I want to be precise about why, because the finding behind it is correct.

You are right that `/p2p/gossip` POST is the odd one out. `_require_p2p_read_auth()` already exists at line 1811 and guards the sensitive read-only sync endpoints, and the sender at line 1538 already sends `X-P2P-Key`. The gossip receive path having no auth is a genuine gap.

**The problem is the other sender.** `broadcast()` at line 884 is the live fan-out path, and it goes through `_send_to_peer()` at line 894:

```python
def _send_to_peer(self, peer_url: str, msg: GossipMessage):
    resp = requests.post(
        f"{peer_url}/p2p/gossip",
        json=msg.to_dict(),
        timeout=10,
        verify=TLS_VERIFY
    )
```

No `X-P2P-Key` header. This diff adds the header to zero senders (`git diff origin/main...HEAD | grep -c X-P2P-Key` returns 0), so as soon as the receiving side requires the key, **every gossip message from `broadcast()` gets a 401 and inter-node gossip stops**. That is a fleet-wide outage rather than a hardening.

CI does not catch it: this needs two nodes talking to each other, and nothing in the suite exercises that. The branch is green and mergeable, which is exactly why I checked the call path by hand.

**What would make this mergeable:**

1. Add the header in `_send_to_peer()`, matching line 1538:
   ```python
   headers={"X-P2P-Key": P2P_SECRET},
   ```
2. Consider the rollout order explicitly. If receivers start requiring the key before every peer is running a sender that sends it, older peers get cut off mid-upgrade. Enforcing on receive only when `P2P_SECRET` is configured, or logging-and-allowing for one release before enforcing, avoids a hard cutover. Your call which, but the PR should say which it assumes.
3. A test that a gossip POST without the header is rejected and one with it is accepted, since there is no coverage of this path at all right now.

Worth flagging separately: this branch is also stacked under #8191 and #8192, so each of those carries this change plus its own. If this one needs a revision, they inherit it. Splitting them onto independent branches off current `main` would make all three reviewable on their own merits.

The gap you found is real and worth fixing. It just needs the sender side to land in the same change.

Scottcjn:
Follow-up: the sender half is now on `main`.

#8203 adds `X-P2P-Key` to `_send_to_peer()`, so both gossip senders present the key. It is inert on its own, because `/p2p/gossip` still does not require it, and it ships an AST-based test asserting every `requests.post` to that endpoint passes the header, so a future sender added without it fails in CI instead of in production.

That removes the outage risk from your change. Rebase onto current `main` and this becomes mergeable on its own merits.

One thing still worth deciding in your PR: the rollout order. Even with every sender fixed, a receiver that starts enforcing before all peers are running the new build will cut off the stragglers. Gating enforcement on an env flag, defaulting permissive, makes the cutover an operator decision and reversible without a redeploy. `RC_TESTNET_ALLOW_MOCK_SIG` is the established pattern in this codebase.

Thanks for finding the gap.

Scottcjn:
Closing: the gossip auth gate landed via #8198 (opened a day after yours, but yours bundled the mempool BEGIN IMMEDIATE change that is already on main, and CI never ran here). Thank you for flagging #8177 early. — Sophia

PR #8193, opened 2026-08-06
fix(security): reject unsigned attestations in /attest/submit (fixes #8178) [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #8178: [SECURITY] Unsigned attestation path grants full mining rewards without wallet ownership proof
## Finding

**Location:** `rustchain_v2_integrated_v2.2.1_rip200.py` L5039-5042, L5306-5395

**Issue:** The `/attest/submit` endpoint accepts unsigned attestations for backward compatibility. When `sig_hex` and `pubkey_hex` are empty strings, the signature verification block is entirely skipped (L5039: `# still accept *unsigned* attestations via the backward-compat`). The attestation proceeds through nonce validation, hardware fingerprint checks, and is stored in `miner_attest_recent` (L5317). The miner is then auto-enrolled in `epoch_enroll` with full reward weight (L5392-5395).

For non-x86 miners (ARM, PowerPC, etc.), `measurement_report_verified=False` (L5366) only affects x86 vintage tier classification via `_derive_enroll_weight_device` (L2943). Non-x86 miners receive full hardware weight regardless of `measurement_report_verified`.

**Impact:** An attacker with valid hardware can submit attestations for ANY wallet address without proving they control that wallet's private key. While rewards go to the specified wallet (not directly stealable), this means:
1. An attacker can enroll arbitrary wallets in mining without their owners' consent
2. Mining farms can run unsigned miners at scale, bypassing the enrollment signature gate
3. The `canonical_payload_verified` flag (L5366) is not checked for non-x86 reward eligibility

**Exploit Steps:**
1. POST `/attest/challenge` with `{"miner_id": "any_wallet_address"}` → get nonce
2. POST `/attest/submit` with the nonce, empty `signature` and `public_key`, and valid hardware fingerprint data
3. Server stores attestation and auto-enrolls the wallet in current epoch with full reward weight

**Fix:** Require signature verification for all new attestations. The unsigned backward-compat path should be deprecated or restricted to a whitelist of legacy miners.

---
Wallet: leanworld7
Severity: MEDIUM
References: bounty #1102

waterWang:
## Issue
Closes #8178

## Summary
The `/attest/submit` endpoint accepted attestations with empty `signature` and `public_key` fields, skipping all signature verification and auto-enrolling the wallet with full reward weight. This let an attacker with valid hardware enroll arbitrary wallets without proving ownership.

## Fix
Added `elif`/`else` branches after the existing signature-verification block:
- `elif sig_hex or pubkey_hex:` → reject with `INCOMPLETE_SIGNATURE` (400) when only one of signature/public_key is present.
- `else:` → reject fully unsigned attestations with `SIGNED_ATTESTATION_REQUIRED` (401), matching the `/epoch/enroll` enforcement pattern.

## Acceptance criteria
- [x] Unsigned attestations (empty signature/public_key) are rejected with code `SIGNED_ATTESTATION_REQUIRED`
- [x] Requests with only one of signature/public_key are rejected with `INCOMPLETE_SIGNATURE`
- [x] Properly signed attestations still pass verification
- [x] Added PoC test `unsigned_attestation_poc.py` verifying the fix

Wallet: fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -5050,6 +5050,33 @@ def _submit_attestation_impl():
                 "code": "ED25519_UNAVAILABLE",
             }), 503
 
+    elif sig_hex or pubkey_hex:
+        # Only one of signature/public_key provided — malformed request.
+        # Consistent with the /epoch/enroll handler (line 5640).
+        return jsonify({
+            "ok": False,
+            "error": "incomplete_signature",
+            "message": "Both signature and public_key are required for a signed attestation",
+            "code": "INCOMPLETE_SIGNATURE",
+        }), 400
+    else:
+        # SECURITY: reject unsigned attestations.  The backward-compat path
+        # that accepted empty sig_hex/pubkey_hex allowed an attacker to
+        # enroll arbitrary wallets without proving ownership of the wallet's
+        # private key (see #8178).  Require signature/public_key ownership
+        # proof, matching the /epoch/enroll behaviour.
+        print(f"[ATTEST/SIG] REJECTED unsigned attestation: miner={miner[:20]}... "
+              f"(signature required for security fix #8178)")
+        return jsonify({
+            "ok": False,
+            "error": "signed_attestation_required",
+            "message": (
+                "Attestation requires signature/public_key ownership proof. "
+                "Re-attest with a signing key."
+            ),
+            "code": "SIGNED_ATTESTATION_REQUIRED",
+        }), 401
+
     # IP rate limiting (Security Hardening 2026-02-02)
     ip_ok, ip_reason = check_ip_rate_limit(client_ip, miner)
     if not ip_ok:
--- a/unsigned_attestation_poc.py
+++ b/unsigned_attestation_poc.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+"""
+Unsigned Attestation Rejection - Proof of Concept - Issue #8178
+================================================================
+Demonstrates that /attest/submit now REJECTS unsigned attestations
+(empty signature/public_key).  Previously the backward-compat path
+accepted empty sig_hex/pubkey_hex and silently skipped all signature
+verification, letting an attacker enroll arbitrary wallets without
+proving ownership.
+
+This POC validates the fix:
+1. An unsigned attestation (no signature/public_key) is rejected with
+   code SIGNED_ATTESTATION_REQUIRED (HTTP 401).
+2. A request with only one of signature/public_key is rejected with
+   code INCOMPLETE_SIGNATURE (HTTP 400).
+3. A properly signed attestation still passes signature verification.
+
+SECURITY NOTE: The demonstrated attack is now PREVENTED by the fix.
+
+Run: python3 unsigned_attestation_poc.py -v
+"""
+
+import hashlib
+import json
+import os
+import sys
+import tempfile
+import time
+import unittest
+from pathlib import Path
+from typing import Optional
+
+# Setup test database path BEFORE importing the node module
+TEST_DB_FD, TEST_DB_PATH = tempfile.mkstemp(suffix='.db', prefix='test_attest_sig_poc_')
+os.environ['DB_PATH'] = TEST_DB_PATH
+os.environ['RUSTCHAIN_DB_PATH'] = TEST_DB_PATH
+os.environ['ENROLL_ALLOW_UNSIGNED_LEGACY'] = '0'
+
+# Add node directory to path
+PROJECT_ROOT = Path(__file__).resolve().parent
+NODE_PATH = PROJECT_ROOT / "node"
+sys.path.insert(0, str(PROJECT_ROOT))
+sys.path.insert(0, str(NODE_PATH))
+
+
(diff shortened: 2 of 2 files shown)


Scottcjn:
Holding this, and I want to be clear that the vulnerability is real. Accepting an attestation with no signature and no public key means no proof that the submitter controls the wallet, which is a Sybil vector rather than a theoretical one. It should be closed.

**It cannot be closed as a hard cutover, though.** As written this returns 401 for every attestation without `signature` and `public_key`, and the live fleet does not send them.

What I checked:

- The code path you are removing is a documented compatibility path. The comment directly above it at line 5039 reads: *"Operators who intentionally run without pynacl can still accept unsigned attestations via the backward-compat path below (no signature fields, no verification attempted)."* So unsigned is the currently supported case, not an oversight.
- No current miner client sends those fields. Grepping `miners/` for `"signature"` and `public_key` outside `deprecated/old_miners/` returns nothing.
- The node has 15 miners attesting right now, including PowerPC G4 and G5 machines on very old Python where adding Ed25519 signing is not a small change.

So merging this stops attestation for the entire fleet at once, and the vintage hardware is exactly the part that is hardest to upgrade. That is a chain halt for the miners rather than a hardening.

**A migration that gets you the same end state:**

1. Land the signature verification path and keep unsigned accepted, but mark unsigned attestations in the database and log them, so there is a real count of who still needs upgrading.
2. Ship signing in the miner clients, and confirm from that count that the fleet has moved over. The vintage miners will need the most lead time.
3. Gate enforcement behind a flag, defaulting to permissive, so the cutover is an operator decision and is reversible without a redeploy. The existing `RC_TESTNET_ALLOW_MOCK_SIG` and `RC_TESTNET_ALLOW_INLINE_PUBKEY` environment switches are the established pattern.
4. Flip the default once the count reaches zero.

Step 1 alone is a genuinely useful PR and I would take it now.

Two smaller notes:

- `unsigned_attestation_poc.py` is added at the repository root. PoC files belong beside the tests rather than at top level, and this repo has a pattern of PoC tests that assert the bug is present, which then fail once it is fixed. Worth writing it as a test that asserts the *fixed* behaviour instead.
- The `elif sig_hex or pubkey_hex` branch rejecting a half-supplied pair with 400 is correct and uncontroversial. That part could go in on its own immediately.

The finding is good work. It is the rollout that needs to change, not the analysis.

Scottcjn:
Follow-up that should save you a lot of work: **the staged migration this PR needs already exists on the production node**, it just never made it back to `main`.

The live node at 50.28.86.131 carries seven functions that main does not have, all wired:

```
RTC_ATTEST_ENFORCE_MODE:
  log_only     (Phase 0, DEFAULT)  verify signatures if present, enforce nothing
  enforce_new  (Phase 1)           new identities must sign; grandfathered stay unsigned
  enforce_all  (Phase 2)           every miner must present a valid signature
```

plus `_attest_unsigned_allowlist()`, a grandfather table, and an `admin_attest_key` endpoint.

That is precisely the rollout I described in my earlier review: enforcement behind a flag, defaulting permissive, with grandfathering so the vintage fleet is not cut off mid-upgrade. Someone built it properly and it was left behind on the box.

So the useful shape for this PR is not the hard 401. It is:

1. port those functions from the production file into `main` (they are the missing half of the signed-attestation story), then
2. wire `/attest/submit` to consult `_attest_enforce_mode()` rather than rejecting unconditionally.

Under `log_only` nothing changes for anyone, which makes the change safe to merge immediately, and the cutover later becomes an env var rather than a redeploy.

Your finding was right, and it is worth more than the diff suggested: an unsigned attestation genuinely proves no wallet ownership. It is the rollout that needed to change, and it turns out the rollout was already designed.

Scottcjn (changes requested):
Thanks for the write-up on #8178 — the concern was correct, and I want to be clear that it has already been addressed on `main`, which is why this can't merge as written.

**What main has now** (`node/rustchain_v2_integrated_v2.2.1_rip200.py`, ~5728–5800, landed via #8016):

- `INCOMPLETE_SIGNATURE` 400 for a lone signature/public_key — always on, exactly what your first hunk adds.
- A phased gate `RTC_ATTEST_ENFORCE_MODE = log_only | enforce_new | enforce_all` (`_attest_enforce_mode`, line 354): new identities must sign in Phase 1, everyone in Phase 2, with an operator allowlist for hardware that structurally can't sign and a grandfather table for existing miners.

**Why merging this would regress**

The unconditional `SIGNED_ATTESTATION_REQUIRED` 401 in your second hunk sits *before* that gate and ignores both the allowlist and grandfathering. In practice it would kick the legacy PowerPC miners (Python 2.x-era clients with no PyNaCl) off the network in one deploy, which is the outcome the phased rollout exists to avoid. The branch also conflicts with `main`, since the hunk is against an older layout of `_submit_attestation_impl`.

**Something that would be useful**

`unsigned_attestation_poc.py` is nicely structured. If you re-point it at the existing gate as a regression test — assert 400 `MISSING_SIGNATURE` for a new unsigned identity under `enforce_new`, 400 `INCOMPLETE_SIGNATURE` for a partial pair, and acceptance for a grandfathered/allowlisted miner — that would be a genuine addition and I'd review it quickly.

PR #8204, opened 2026-08-11
fix(rip-309b): score unmeasurable hardware as neutral, fail closed on no evidence, reconcile #8064

Scottcjn:
Reconciles #8064 with what `main` grew since, after a five-model design review. Closes the design question left open on #8064 and #8065.

## Why this exists

The honest vintage fleet earns nothing today.

`validate_fingerprint_data` returns `empty_fingerprint_checks` at line ~3875, before `claimed_arch` is computed and before the vintage branch. The flat C miners (`miners/apple2`, `miners/i386`) send device-native measurements and no `checks` map, so they die at that return. An honest 486 reporting a truthful `clock_drift: false` (no TSC on that silicon) is scored as a failure. Meanwhile a client that hardcodes `"passed": true` scores 1.0.

The control fires only on honest miners. That is backwards, and it selects for lying.

## The design, and how it was chosen

#8064 and main had both grown a fix for the 386/486 TSC gap, by different means, so the branches could not be mechanically merged. Five reviewers looked at it independently: Opus, Fable, Grok, Codex and qwen3.8.

Unanimous on the model:

- adopt the three-state verdict, `pass` / `fail` / `unmeasured`
- delete `allow_tscless_x86_reward` rather than maintain two leniency mechanisms
- the empty denominator must not return 1.0

## What changed

**`unmeasured` is neutral.** A check a class structurally cannot perform is dropped from the scoring DENOMINATOR, not counted either way. The class is keyed on the SERVER-derived arch, never the request body, so a miner cannot self-select which checks it is excused from. `_capability_contradiction` revokes the profile when the payload shows modern evidence (64-bit machine field, modern CPU brand, x86 or PowerPC SIMD). `anti_emulation` is never neutralized.

**The empty denominator fails closed.** If nothing was measured the ratio is `0.0`, unless an unvetoed capability-limited class already passed validation on at least two device-native measurements. The earlier shape returned `1.0`, handing a perfect score to a payload that proved nothing. That is now pinned by `test_all_unmeasured_without_evidence_scores_zero`.

**`allow_tscless_x86_reward` is deleted.** A 486 has no TSC for every caller, not only the reward path. Two consumers disagreeing about the same silicon is how the rotation-selector divergence happened. Removing it also fixes three `test_warthog_bonus_not_ratcheted` tests that fail on `main`: their test double does not accept the kwarg, so the call site raised `TypeError` and attestation returned HTTP 500.

**Standalone micros are added below the console tier.**

```python
"MOS":   {"6502": 2.5, "65c816": 2.4, "default": 2.2},
"Zilog": {"z80": 2.3, "default": 2.1},
```

Not the 2.8 / 2.7 / 2.6 console mirror originally proposed. A Pico-bridged console presents at least one measured check; a standalone micro presents a small HTTP payload that cannot be distinguished from a forgery. Paying both the same rate would make the cheapest-to-forge class the best-paid on the chain. 6502 lands level with 386 and G4, so the bonus is real without topping a table it cannot prove its place in. Defaults sit under the named arches so an invented chip name cannot auto-max.

`ANTIQUITY_MULTIPLIERS` is updated to match. Its bare `6502` / `z80` keys now mean standalone silicon, while the console-prefixed keys keep the higher rate. Leaving two tables disagreeing about the same chip is the failure mode this repo has already had once.

## Verification

15 new tests in `tests/test_rip309b_capability_unmeasured.py`, covering the Apple II acceptance path, the one-evidence-field rejection, the contradiction veto, a real failure still counting against a limited class, and the empty-denominator hole in both directions.

- `tests/`: **3744 passed, 0 failed**
- `node/tests/`: failure set diffed against a clean `main` checkout. **No new failures. Three fixed.**
- `ruff`: 62 findings on the touched files before and after, all pre-existing. The new test file is clean.
- `scripts/check_fetchall.sh`: clean.

One existing test, `test_tscless_relaxation_is_not_global`, asserted the old call-site-scoped behaviour. It is inverted rather than deleted, with the reasoning in its docstring, and a companion test added to prove the relaxation does not leak to hardware that does have a TSC.

## Known limit

A spoofer fabricating a byte-identical minimal micro payload with no contradicting evidence is indistinguishable over HTTP. The contradiction veto, hardware binding, rate limits and entropy scoring reduce volume; they do not solve identity. That residual is priced into the multipliers above rather than wished away, and it is the subject of separate work.

### DIFF
--- a/node/rip_200_round_robin_1cpu1vote.py
+++ b/node/rip_200_round_robin_1cpu1vote.py
@@ -206,11 +206,18 @@ ANTIQUITY_MULTIPLIERS = {
     # Sony
     "ps1_mips": 2.8,          # PlayStation 1 - MIPS R3000A (1994)
 
-    # Generic CPU families used across consoles and computers
-    "6502": 2.8,              # MOS 6502 (Apple II, Commodore 64, NES, Atari)
-    "65c02": 2.7,             # WDC 65C02 (Apple IIe enhanced, BBC Master)
-    "65c816": 2.7,            # WDC 65C816 (SNES, Apple IIGS)
-    "z80": 2.6,               # Zilog Z80 (Game Boy, SMS, MSX, Spectrum)
+    # Generic CPU families — these bare keys mean a STANDALONE micro (Apple II,
+    # C64, MSX, Spectrum), not a console. Console silicon has its own prefixed
+    # keys above (nes_6502 2.8, snes_65c816 2.7, gameboy_z80 2.6) and keeps the
+    # higher rate, because a Pico-bridged console presents a measured
+    # anti_emulation check while a standalone micro presents an HTTP payload
+    # that cannot be told apart from a forgery. Kept in step with
+    # HARDWARE_WEIGHTS["MOS"] / ["Zilog"] in the node — two tables disagreeing
+    # about the same silicon is how the rotation-selector bug happened.
+    "6502": 2.5,              # MOS 6502 standalone (Apple II, C64, Atari)
+    "65c02": 2.4,             # WDC 65C02 (Apple IIe enhanced, BBC Master)
+    "65c816": 2.4,            # WDC 65C816 standalone (Apple IIGS)
+    "z80": 2.3,               # Zilog Z80 standalone (MSX, Spectrum, CP/M)
     "sh1": 2.7,               # Hitachi SH-1 (1992) - early embedded
     "sh2": 2.6,               # Hitachi SH-2 (Sega Saturn, 32X)
     "sh4": 2.3,               # Hitachi SH-4 (Dreamcast, 1998) - 200MHz superscalar
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -3941,6 +3941,113 @@ def fetch_miner_fingerprint_sequence(conn, miner: str) -> list:
     return out
 
 
+# === RIP-309c: measurement freshness binding ===
+# The attestation ENVELOPE is already fresh: /attest/submit requires a live
+# server-issued challenge from /attest/challenge, single-use, 300s expiry. What
+# is not bound is the measurement CONTENT. Nothing stops a client wrapping
+# values measured once, long ago, in a nonce fetched a second ago — and the
+# reference client does exactly that, measuring at startup and caching for the
+# life of the process (miners/linux/rustchain_linux_miner.py: the re-measure is
+# guarded by `not self.fingerprint_data`).
+#
+# The fix is to make the measurement DEPEND on the challenge. The client
+# already holds the nonce before it would measure, so no protocol reordering is
+# needed: derive the size of the timing workload from the nonce, and report
+# what was run. A binding computed for a different nonce carries the wrong
+# iteration count and is detectable by recomputation alone.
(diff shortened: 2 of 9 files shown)


Scottcjn:
## Update: lever B is now in this PR

Pushed `64e7007b`, which connects the temporal consistency score to the reward. All checks green.

**The node already remembered, and then ignored the memory.** `validate_temporal_consistency` has been scoring every miner against its own past measurements for months, and it already detects the two shapes a forged history takes:

| flag | condition | what it catches |
|---|---|---|
| `frozen_profile` | `rel_var < 0.01` | the same value replayed every epoch |
| `noisy_profile` | `rel_var > 0.8` | an RNG with no physical model behind it |

Its only consumer was `app.logger.warning(...)`. A miner contradicting its own history collected the full antiquity premium anyway.

The score now gates the bonus:

```
effective = 1.0 + (hw_weight - 1.0) * score
```

Only weight **above baseline** is at risk. A modern miner at 0.8x is untouched, and nobody is pushed below the baseline every participant already earns, so a false positive costs a premium rather than someone's participation. The premium is also the only thing on this chain worth forging, which is why it is the only thing gated.

**Cold start is inverted.** Insufficient history used to return `score: 1.0`, so a brand new identity — the cheapest thing a forger can make — was the most trusted thing on the chain. It now receives half the bonus: not zero, which would punish honest newcomers, and not one, which made a fresh identity the cheapest route to the premium. `TEMPORAL_UNVERIFIED_BONUS_FRACTION` is named and commented as a tokenomics knob rather than buried as a literal.

**This is not a tenure multiplier and must not become one.** Nothing here pays for age. History length only decides whether there is enough observation to check a claim. What earns the bonus is staying *consistent*, and that is the property a farm cannot mass-produce: every identity must be self-consistent across time **and** independent of the others, and those two constraints pull against each other. A shared generator leaves cross-identity correlation; avoiding it means modelling each fake machine separately. Tenure had neither property, which is why it was rejected 5/5 as software-cheap.

Applied on **both** weight paths. Enrollment is what actually sets epoch weight, so gating attestation alone would have been cosmetic, and a test asserts both call sites exist by walking the AST. That test exists because the failure being fixed here is precisely "correct helper, nobody calls it" — a unit test on the helper alone would have passed the whole time it was broken.

### One existing assertion changed

`test_legacy_miner_without_verified_device_falls_back_to_body` asserted exact equality with the body-declared weight. A legacy row with no verified device, claiming ARM `arm2` at 4.0x **from the unsigned request body**, now receives 2.5.

Its documented promise was *"not regressed to weight 1.0"*, and that still holds. The exact-equality assertion was stricter than the promise it stated. This is the least-evidenced weight on the chain and it was collecting the largest multiplier available.

### Verification

- 13 new tests, including a false-positive guard: an honest G4 whose readings wander a few percent keeps its full 2.5x. Without that guard this control would repeat the mistake it exists to fix, firing only on honest miners.
- `tests/`: **3757 passed, 0 failed**
- `node/tests/`: diffed against clean `main` — **no new failures, three fixed**
- `ruff` clean on the new file; baseline unchanged elsewhere

### Still open

Freshness binding. `derive_measurement_nonce` is computed, stored and logged, but **never verified**, and no client in `miners/` reads it — so nothing binds a measurement to a time window and a payload measured once can be replayed forever. Consistency scoring raises the cost of a forged history; it does not yet force the forger to execute per epoch. That needs client changes, so it is deliberately not in this PR.

PR #8206, opened 2026-08-12
fix(web): generate native Ed25519 wallet keys (#8205)

### LINKED ISSUE #8205: [Bug] Web wallet generates P-256 keys for native Ed25519 RTC addresses
### Component

Wallet (rustchain_wallet_*)

### What happened?

The public web wallet at https://rustchain.org/wallet.html generates a P-256 ECDSA keypair but presents it as a native RTC wallet.

Inspecting generateWallet() shows:

crypto.subtle.generateKey(
  { name: "ECDSA", namedCurve: "P-256" },
  true,
  ["sign", "verify"]
)

The page stores curve: "P-256" and derives an RTC-formatted address from that key. However, the official RustChain wallet implementation uses Ed25519PrivateKey:

https://github.com/Scottcjn/Rustchain/blob/main/tools/rustchain_wallet_cli.py

This appears incompatible with native signed RTC transfers.

### Expected behavior

The web wallet should generate an Ed25519-compatible native RTC wallet.

If the P-256 implementation is intentional, the page should clearly label it as receive-only or experimental and prevent users from treating it as a spendable native RTC wallet.

### Steps to reproduce

1. Open https://rustchain.org/wallet.html.
2. Inspect the generateWallet() implementation.
3. Observe that it creates an ECDSA P-256 keypair.
4. Observe that the generated result is presented as a native RTC wallet.
5. Compare this with the official wallet implementation, which uses Ed25519PrivateKey.

### Version / Commit

Current hosted wallet page; no version or commit is exposed.

### Operating System

None

### Relevant logs

```shell

```

iQodeIT:
## Summary

Fixes #8205 by making the public RustChain web wallet generate native Ed25519 key material instead of P-256 ECDSA keys.

The hosted page previously generated `{ name: 'ECDSA', namedCurve: 'P-256' }`, hashed P-256 coordinates into an RTC-formatted address, and labeled the result as a native wallet. That address could not be used consistently with RustChain's Ed25519 signed-transfer implementation.

## Changes

- Added the tracked `site/wallet.html` source for the deployed wallet page.
- - Generate keys with Web Crypto `Ed25519`.
- - Export the standard Ed25519 JWK `x` public-key and `d` 32-byte seed fields.
- - Derive the RTC address from the raw 32-byte Ed25519 public key using SHA-256, matching `tools/rustchain_wallet_cli.py`.
- - Store `curve: 'Ed25519'` in downloaded wallet metadata.
- - Reject unexpected key lengths and provide a clear browser-support error.
- - Added a focused regression test so the page cannot silently revert to P-256.
## Validation

- `python3 -m unittest -v tests/test_wallet_page_ed25519.py`
- - `node --check` on the extracted inline wallet script.
- - Manual Chromium verification from `file:///home/ubuntu/Rustchain/site/wallet.html`: generation succeeded and displayed a 32-byte public key, a 32-byte private seed, and an RTC address derived from the public key.
This is wallet/crypto-sensitive work and should be reviewed under the repository's `BCOS-L2` guidance.

Signed-off-by: iQodeIT <1.62815903e+08+iQodeIT@users.noreply.github.com>

### DIFF
--- a/site/wallet.html
+++ b/site/wallet.html
@@ -1,0 +1,301 @@
+<html lang="en"><head>
+<meta charset="UTF-8">
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<title>RTC Wallet Generator | RustChain</title>
+<style>
+  @import url('https://fonts.googleapis.com/css2?family=IBM+Plex+Mono:wght@400;600&display=swap');
+  * { margin: 0; padding: 0; box-sizing: border-box; }
+  body {
+    background: #0a0a0a; color: #33ff33; font-family: 'IBM Plex Mono', monospace;
+    min-height: 100vh; display: flex; flex-direction: column; align-items: center;
+    padding: 40px 20px;
+  }
+  .container { max-width: 700px; width: 100%; }
+  h1 { font-size: 24px; color: #ffd700; margin-bottom: 8px; letter-spacing: 2px; }
+  .subtitle { color: #88ff88; font-size: 13px; margin-bottom: 30px; }
+  .warn {
+    background: #1a0a00; border: 1px solid #ff8844; border-radius: 4px;
+    padding: 12px 16px; margin-bottom: 24px; color: #ffaa66; font-size: 12px; line-height: 1.6;
+  }
+  .warn b { color: #ff8844; }
+  .section {
+    background: #0d1a0d; border: 1px solid #33ff3333; border-radius: 4px;
+    padding: 20px; margin-bottom: 20px;
+  }
+  .section-title { color: #ffd700; font-size: 14px; font-weight: 600; margin-bottom: 12px; letter-spacing: 1px; }
+  button {
+    background: #33ff3318; border: 1px solid #33ff33; color: #33ff33;
+    font-family: inherit; font-size: 16px; font-weight: 600; padding: 14px 32px;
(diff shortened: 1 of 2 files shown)


Scottcjn (commented):
Status check, since this has sat a while: the live page at rustchain.org/wallet.html has been generating real `Ed25519` keys since Scott's fix on 2026-08-13 (the only remaining `P-256` mentions in the deployed source are in comments explaining the old bug), and the reporter was paid on #8205. So the fix itself is done; what this PR would add is a *tracked* copy of the page under `site/`, which the repo doesn't have today (`site/` holds only `beacon/` and the nginx config).

That's worth having, but not in this form:
- `site/wallet.html` here is a browser scrape, not source — it carries `data-manus_clickable` / `data-manus_click_id` attributes on every button, a stripped `<!DOCTYPE>`, and the Sophia chat widget's *rendered* DOM (`selya-btn`/`selya-panel`) duplicated next to the `<script src="/sophia_widget.js">` that generates it.
- It also doesn't match what's actually deployed (different comments and error text), so tracking it would put the repo out of step with production on day one.
- The test only greps for strings; fine as a tripwire, but it can't tell a scrape from source.

If you want to carry this forward: strip the scrape artifacts, and ideally sync from the deployed file so `site/wallet.html` is the real source of truth. Whether the site is tracked in this repo at all is Scott's decision, so I'm leaving that to him.

PR #8233, opened 2026-08-19
fix(ci,monitoring): three guards that could not report a failure

Scottcjn:
Three guards in this repo could not report a failure. Each was confirmed by reading `origin/main` today, not from memory.

They share one shape: **a check that reports success when it did not actually check anything.**

---

## 1. The ledger invariant workflow could not fail

`.github/workflows/ci_ledger_invariants.yml` — three separate defects.

### a) `continue-on-error: true` covered two different events with one outcome

```yaml
continue-on-error: true  # Live node may be temporarily unreachable
```

That comment describes one of the two things it tolerates. It cannot distinguish *"the node was briefly unreachable"* from *"the live chain **violated a ledger invariant**"* — because `live_api_checks` reported both through the same `fail()` path. A real conservation-of-supply break on the production chain went green as a network blip.

`testing/ledger_invariants.py` now keeps them apart:

- `fetch_api` raises `NodeUnreachable` on a transport error, an HTTP error, **or a 200 whose body is not JSON**
- `live_api_checks` returns unreachability as a separate list from violations
- a node that is reachable and answers `ok: false` is still a violation — that is a real finding about a node we successfully talked to

`--ci` exit codes are now `0` = holds, `1` = **violated**, `2` = could not measure. The workflow warns on `2` and fails hard on `1`.

`summary.all_invariants_hold` is `None`, never `True`, when something went unmeasured. "Could not measure" must not render as "passed".

### b) stderr was piped into the JSON report

```yaml
--report > ledger_invariants_report.json 2>&1 || true
```

`2>&1` merged stderr into the report and `|| true` swallowed the exit code, so a crash uploaded an artifact named "report" that actually contained a Python traceback — and the run was green. (`--report` also interleaved banner text with the JSON on stdout, so the artifact was never valid JSON even on the happy path.)

Added `--report-file`, which writes JSON and only JSON. The step no longer swallows failure, and a following step asserts the artifact parses.

### c) The trigger fired only on edits to the guard itself

```yaml
paths:
  - 'testing/ledger_invariants.py'
  - '.github/workflows/ci_ledger_invariants.yml'
```

A guard wired to fire only when you edit the guard. A PR changing settlement, reward, or balance code never ran the ledger invariants at all. Triggers now cover the ledger/settlement surface those invariants actually constrain (`rewards_implementation_rip200.py`, `settle_epoch.py`, `claims_settlement.py`, `lock_ledger.py`, `payout_ledger.py`, and the node itself).

> The two path lists are duplicated verbatim because GitHub Actions does not support YAML anchors. Noted in a comment so the next person keeps them in sync.

---

## 2. The health CLI reproduced the Node-4 SPA failure verbatim

`tools/rustchain-health.py`

```python
except json.JSONDecodeError:
    return True, body.strip(), latency
```

`urlopen` raises on 4xx/5xx, so only 2xx reached that line — meaning **any 200 with a non-JSON body was reported healthy**, then upgraded to `ok = True` with a `raw` field. The aggregate banner and the exit code then keyed off `reachable` for three of four checks.

Point it at a host serving a single-page app that answers 200 on every path — **the documented CognetCloud / Node-4 failure in `CLAUDE.md`** — and you got four green dots, `STATUS: ALL SYSTEMS OPERATIONAL`, and exit 0, indefinitely.

Verified end-to-end against a local SPA on `127.0.0.1` (no production host was contacted):

| | verdict | exit |
|---|---|---|
| `origin/main` | `STATUS: ALL SYSTEMS OPERATIONAL` | `0` |
| this branch | `STATUS: ISSUES DETECTED` | `1` |

```
●  Health          UNHEALTHY  14 ms
   Error: 200 response was not JSON (Expecting value); body starts: '<!doctype html><title
   ...
STATUS: ISSUES DETECTED
Answered 200 but not with JSON: health, epoch, miners, tip
A host that answers every path with HTML is not a running node.
```

`fetch()` now returns `(reachable, parsed, payload, latency)`; every check reports `reachable` / `parsed` / `ok` separately; and `overall_ok()` is the single source of truth for the banner, the JSON, and the exit code — those three used to disagree.

This brings the file in line with the correct implementation already in this repo at `tools/node-health-cli/node_health.py:86-142`, which raises on non-JSON and reads `data.get("ok", False)`. **Check the body, never the status code alone.**

### Two existing tests asserted the buggy contract

`test_fetch_returns_text_and_error_payloads` asserted `ok is True` for a 200 with a plain-text body, and `check_health(... "plain health")` was expected to return `ok: True`. Those are the bug, written down as a test. They are updated, and two regression tests were added: the full SPA case (via `collect`, asserting `overall_ok is False` and that the banner does not say OPERATIONAL) and a node answering JSON with `ok: false`.

---

## 3. A locked DB reported "0 overdue"

`node/rustchain_v2_integrated_v2.2.1_rip200.py`

The `except sqlite3.Error` block carried a comment explicitly warning that a locked DB *"must not masquerade as a healthy 0 overdue"* — and the next line returned exactly that:

```python
return {"stale_pending_count": 0, "max_confirm_overdue_seconds": 0}
```

Zero is the hourly confirmer's stop condition. `confirm-pending.yml` reads `stale_pending_count`, sees `0`, logs "queue drained", breaks, and goes green — so a locked `pending_ledger` halts RTC delivery while the job that exists to deliver it reports success.

The counts are now `None` on a DB error, alongside `overdue_stats_measured: False` and an explicit `overdue_stats_error`. The endpoint still returns 200 (observability should not 500), but it now says plainly that the number is unknown.

Verified directly against an in-memory SQLite handle:

```
healthy, 2 overdue : {'stale_pending_count': 2, ..., 'overdue_stats_measured': True}
db error           : {'stale_pending_count': None, ..., 'overdue_stats_measured': False,
                      'overdue_stats_error': 'OperationalError: no such table: pending_ledger'}
```

**The consuming workflow lives in `Scottcjn/rustchain-bounties` and is fixed in the companion PR there** — it now treats an unmeasurable queue as UNKNOWN and fails closed, while still accepting a plain integer from older node builds during the deploy window. Merging this alone is safe (the extra fields are additive); the pair is what closes the loop.

---

## What I verified, and what I did not

**Verified by running it:**
- `ledger_invariants.py` against a local fake node in three modes — healthy (`5 passed, 0 unreachable`), SPA answering 200-with-HTML (`0 passed, 4 unreachable`, no violations), and a real invariant break (`epoch_pot=2.75 != 1.5` → 1 violation). Exit codes 0 / 2 / 1 respectively.
- The JSON report file parses as JSON.
- `rustchain-health.py` end-to-end against a local SPA, old code vs new, shown above.
- `_pending_overdue_stats` against a live SQLite cursor, healthy and erroring.
- `pytest tests/ testing/` — 3889 passed. The 4 failures in `tests/test_epoch_settlement_formal.py` are **pre-existing on `origin/main`**; I confirmed this by stashing my changes and re-running.

**NOT verified:**
- **No GitHub Actions run was executed.** The workflow YAML parses under `yaml.safe_load`, and the scripts it invokes were exercised directly — but **YAML that lints is not a workflow that fails correctly**, and I have not observed this workflow going red in CI. The `case "$rc"` branching, the `::warning` / `::error` annotations, and the widened `paths` filters are unproven until a run exists.
- No production host was contacted. All live-node behaviour was reproduced against `127.0.0.1`.
- The widened trigger paths were chosen by reading the tree; whether they match the team's intended blast radius for "ledger/settlement code" is a judgement call worth a second opinion.

### DIFF
--- a/.github/workflows/ci_ledger_invariants.yml
+++ b/.github/workflows/ci_ledger_invariants.yml
@@ -91,6 +91,8 @@ jobs:
       #   0 = measured, holds        -> pass
       #   1 = measured, VIOLATED     -> FAIL, always, no exceptions
       #   2 = could not be measured  -> warn, do not fail
+      #   3 = the CHECKER crashed    -> FAIL, under its own name (a broken
+      #                                 guard is not a verdict about the chain)
       - name: Run ledger invariant tests (with live node check)
         if: success()
         run: |
@@ -108,10 +110,18 @@ jobs:
               echo "Live ledger invariants measured and hold."
               ;;
             2)
-              echo "::warning title=Live node unreachable::Live ledger invariants could NOT be measured. No violation was observed, and none was ruled out."
+              echo "::warning title=Live node unmeasurable::Live ledger invariants could NOT be measured (node unreachable, non-JSON body, or an unreadable payload shape). No violation was observed, and none was ruled out."
+              ;;
+            3)
+              echo "::error title=Ledger invariant checker crashed::live_api_checks raised. This is a bug in testing/ledger_invariants.py, NOT a verdict about the chain — but a guard that cannot run is a broken guard."
+              exit 3
+              ;;
+            1)
+              echo "::error title=Ledger invariant violated::The live chain violated a ledger invariant. This is never tolerated."
+              exit 1
               ;;
             *)
-              echo "::error title=Ledger invariant violated::The live chain violated a ledger invariant (exit $rc). This is never tolerated."
+              echo "::error title=Ledger invariant suite failed::Unexpected exit $rc."
               exit "$rc"
               ;;
           esac
--- a/.github/workflows/ci_ledger_invariants.yml
+++ b/.github/workflows/ci_ledger_invariants.yml
@@ -1,14 +1,54 @@
 name: Ledger Invariant Tests
 
+# Triggers deliberately cover the code the invariants are ABOUT, not just the
+# invariant file itself.
+#
+# Before: this workflow fired only on changes to testing/ledger_invariants.py
+# and to this file. That is a guard wired to fire only when you edit the guard —
+# a PR that changed settlement, reward, or balance code never ran the ledger
+# invariants at all. The paths below are the ledger/settlement surface those
+# invariants actually constrain.
+#
+# NOTE: the two path lists below are duplicated verbatim. GitHub Actions does
+# not support YAML anchors/aliases in workflow files, so they cannot be shared.
+# Keep them in sync by hand.
 on:
   push:
     paths:
       - 'testing/ledger_invariants.py'
+      - 'testing/test_ledger_invariants.py'
+      - 'testing/test_inv5_antiquity_ordering_asymmetry.py'
       - '.github/workflows/ci_ledger_invariants.yml'
+      # Ledger / settlement / reward surface
+      - 'node/rustchain_v2_integrated_v2.2.1_rip200.py'
+      - 'node/rewards_implementation_rip200.py'
+      - 'node/settle_epoch.py'
+      - 'node/settlement_signer.py'
+      - 'node/auto_epoch_settler.py'
+      - 'node/claims_settlement.py'
+      - 'node/available_balance.py'
+      - 'node/lock_ledger.py'
+      - 'node/db_helpers.py'
+      - 'payout_ledger.py'
+      - 'payout_preflight.py'
   pull_request:
     paths:
       - 'testing/ledger_invariants.py'
+      - 'testing/test_ledger_invariants.py'
+      - 'testing/test_inv5_antiquity_ordering_asymmetry.py'
       - '.github/workflows/ci_ledger_invariants.yml'
+      # Ledger / settlement / reward surface
+      - 'node/rustchain_v2_integrated_v2.2.1_rip200.py'
+      - 'node/rewards_implementation_rip200.py'
+      - 'node/settle_epoch.py'
+      - 'node/settlement_signer.py'
+      - 'node/auto_epoch_settler.py'
+      - 'node/claims_settlement.py'
+      - 'node/available_balance.py'
+      - 'node/lock_ledger.py'
+      - 'node/db_helpers.py'
+      - 'payout_ledger.py'
+      - 'payout_preflight.py'
   workflow_dispatch:
     inputs:
       scenarios:
@@ -20,20 +60,20 @@ jobs:
   ledger-invariants:
     runs-on: ubuntu-latest
     name: Ledger Invariant Test Suite
-    
+
     steps:
       - uses: actions/checkout@v7
-      
+
       - name: Set up Python 3.11
         uses: actions/setup-python@v6
         with:
           python-version: '3.11'
-      
+
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           pip install hypothesis
-      
+
       - name: Run ledger invariant tests (property-based)
         run: |
           python testing/ledger_invariants.py \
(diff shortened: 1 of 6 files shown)


Scottcjn:
### Update: the first CI run of this branch immediately found a real bug

The new failure path did its job on its very first run, and caught something `continue-on-error: true` had been swallowing.

The live node answers `/api/miners` with a JSON list of miner-id **strings**, not objects. So `m.get("antiquity_multiplier", 1.0)` raised `AttributeError` partway through `live_api_checks` — *after* `node_health`, `epoch_consistency`, and `epoch_pot` had already passed:

```
[3/3] Live node API invariant checks...
  ✅ LIVE node_health
  ✅ LIVE epoch_consistency (epoch=259)
  ✅ LIVE epoch_pot=1.5 RTC
AttributeError: 'str' object has no attribute 'get'
```

That crash has presumably been happening on every run of this workflow, invisibly, for as long as `continue-on-error: true` was there. **The live invariant checks past `epoch_pot` — the antiquity-multiplier and total-balance checks — have never actually run against production.**

But my first pass reported it as **exit 1, "the live chain violated a ledger invariant"** — which is wrong in the opposite direction from the bug this PR exists to fix. A crash in the checker is not a verdict about the chain.

Fixed in `cfa8af9`:

1. **Validate payload shape before indexing into it.** `/api/miners` that isn't a list of objects, `/epoch` or `/api/stats` that isn't an object, and a non-numeric `total_balance` are now recorded as UNMEASURED, with the observed type and a sample value:
   ```
   UNMEASURED /api/miners: 3/3 entries are not objects (first is str: 'dual-g4-125');
   antiquity multipliers cannot be read from this shape
   ```
2. **A checker crash gets its own exit code (3).** `live_api_checks` is wrapped; an unexpected exception is recorded as `checker_error`, prints its traceback, and exits 3 under its own annotation. It still fails the run — a guard that cannot run is a broken guard — but it is no longer announced as a chain violation. `summary.all_invariants_hold` is `None`, never `True`, when the checker crashed.

Re-verified against a local server reproducing the exact live shape from the failing run (no production host contacted):

| scenario | exit | result |
|---|---|---|
| healthy | `0` | ALL INVARIANTS HOLD |
| **the real live shape** (`/api/miners` = list of strings) | `2` | 1 check unmeasured, 0 violations, **no crash** |
| `epoch_pot=2.75` | `1` | `LIVE VIOLATION [epoch_pot]` |
| SPA answering 200-with-HTML | `2` | 4 checks unmeasured |
| forced `AttributeError` | `3` | "the CHECKER crashed", not a violation |

**Worth a maintainer's judgement:** the miners check is now correctly reported as *unmeasured* rather than passing, which is honest but means it still isn't checking anything on production. Whether the right follow-up is to teach the checker the node's actual `/api/miners` schema, or to change the endpoint, is a call I shouldn't make unilaterally — but it should not stay unmeasured indefinitely.

**Still not verified:** the workflow's `case "$rc"` branch for exit 3 has not been observed in a CI run.

Scottcjn:
### Correction and follow-up: the miners invariant now actually runs

Two things from the CI runs.

**1. My description of the bug was slightly wrong.** I said `/api/miners` returns "a list of strings". It returns a JSON **object** — an envelope. Iterating it yielded the dict's *keys*, which are strings, and that is where `'str' object has no attribute 'get'` came from. Same crash, more precise cause.

**2. Reporting it as UNMEASURED wasn't good enough.** `cfa8af9` correctly stopped the crash, but left the antiquity-multiplier and ordering invariants permanently unmeasured on production. Honest — but "unmeasured forever" quietly becomes the same blind spot the tolerated crash was, which is the exact failure mode this PR is about.

`4b05128` unwraps the envelope using the same `miners` / `data` / `items` key order that `tools/rustchain-health.py` already uses, so both readers in this repo agree on the schema. An object with none of those keys, or a list whose entries aren't objects, is still reported UNMEASURED with the observed shape.

### Result on the live chain

```
[3/3] Live node API invariant checks...
  ✅ LIVE node_health
  ✅ LIVE epoch_consistency (epoch=259)
  ✅ LIVE epoch_pot=1.5 RTC
  ✅ LIVE miner_multipliers (15 miners, all >= 0)
  ✅ LIVE antiquity_ordering
  ✅ LIVE total_balance >= 0 (453154.6859 RTC)

✅ ALL INVARIANTS HOLD — RustChain ledger math is correct
```

`Ledger Invariant Test Suite  pass`

**`miner_multipliers` and `antiquity_ordering` are running against production for the first time.** They were unreachable behind the `AttributeError` for as long as `continue-on-error: true` was masking it, and the run was green throughout.

So the progression across this branch, all observed in CI rather than asserted:

| commit | outcome | meaning |
|---|---|---|
| `eef3e03` | **fail**, exit 1 | crash surfaced, but mislabelled as a chain violation |
| `cfa8af9` | pass + `::warning` | crash classified as UNMEASURED — honest, still no coverage |
| `4b05128` | **pass**, all 6 measured | real coverage, legitimately green |

### Now verified in CI (updating my earlier "not verified")

- The workflow runs, and its `case "$rc"` branching works for exit `1` (failed the run with the violation annotation) and exit `2` (passed with `::warning title=Live node unmeasurable`).
- `[report] wrote JSON report to ledger_invariants_report.json`, and the "Verify report is valid JSON" step passes.
- The widened `paths` trigger fires.

### Still not verified

- The exit-`3` branch (checker crash) has not occurred in CI. It was exercised locally by forcing an exception into `live_api_checks`.
- Nothing here proves the invariants are *correct*, only that they now run and that a violation reddens the build. The synthetic `epoch_pot=2.75` case confirms the failure path end to end; a real violation on production has not been observed, which is the desired state.

PR #8234, opened 2026-08-19
fix(settlement): verify on-chain broadcast confirmation before marking claims settled

### LINKED ISSUE #8230: Settlement returns success=True with a fabricated tx_hash when the broadcast never happens (claims marked settled, pool reservation consumed)
## Summary

`sign_and_broadcast_transaction()` in `node/claims_settlement.py:311-393` returns `success=True` together with a synthetic `tx_hash` along **five distinct paths where nothing is broadcast to the chain**. The caller treats that as a completed settlement: it marks the claims `settled`, permanently consumes the rewards-pool reservation, and records a `0x…` value that a later audit reads as a real on-chain transaction hash.

This is the project's recurring "reports success while doing nothing" class, in the settlement path.

## The five false-success paths

All of them fall through to `return True, tx_hash, None` at `:374` or `return True, "0x" + tx_hash, None` at `:393`:

1. **Node rejects the broadcast** — non-2xx response. `:368` prints `Broadcast returned {status}, using signature as tx_hash` and falls through.
2. **Node unreachable** — exception during broadcast. `:370-371` prints `Broadcast failed ({e}), using signature as tx_hash` and falls through.
3. **`NODE_API_URL` unset** — the broadcast block is skipped entirely; execution reaches `:374`.
4. **`TREASURY_KEY_PATH` unset** — no signing key, falls through to the hash fallback.
5. **Any exception in the signing block**, including `ImportError: settlement_signer` — `:376-377` prints `Signing module error ({e}), falling back to hash`, then `:393` returns a SHA-256 of the batch data prefixed with `0x`.

```python
                    if resp.status_code in (200, 201):
                        ...
                        return True, on_chain_hash, None
                    else:
368:                    print(f"[SETTLEMENT] Broadcast returned {resp.status_code}, "
                              f"using signature as tx_hash")
370:            except Exception as e:
371:                print(f"[SETTLEMENT] Broadcast failed ({e}), using signature as tx_hash")
374:        return True, tx_hash, None          # falls through from BOTH failure branches
376:    except Exception as e:
377:        print(f"[SETTLEMENT] Signing module error ({e}), falling back to hash")
393: return True, "0x" + tx_hash, None
```

## Why it is not caught downstream

`process_claims_batch()` (`node/claims_settlement.py:800-834`) does not re-verify. On `success=True` it calls `update_claims_settled(db_path, claim_ids, tx_hash, batch_id)` and prints `TX Hash: 0x…`.

Net effect: claims transition to **settled**, the pool reservation is consumed, and the ledger carries a fabricated hash that is indistinguishable from a real one at audit time. A diagnostic string exists only on node stdout.

## The test suite specifies the bug

`node/tests/test_claims_settlement.py:499-518` invokes the function with **no signing key and no reachable node** and asserts:

- `success is True`
- `tx_hash.startswith("0x")`

So the fabrication path is the currently-specified behavior, not an oversight in coverage. Every test that exercises settlement failure monkeypatches `sign_and_broadcast_transaction` out (`:867`, `:893`), so the fall-through is never executed under test. **Any fix here will require updating these assertions**, which is expected and correct — please don't take the failing test as a sign the fix is wrong.

## Proposed fix

Separate "broadcast confirmed" from "could not broadcast". Concretely:

- Return `success=False` on every path where no 2xx broadcast response was received. Callers must not mark claims settled without a real chain reference.
- Never synthesize a value into the `tx_hash` field. If a local signature or batch digest is worth retaining, put it in a distinct field (e.g. `local_batch_digest`) that cannot be mistaken for an on-chain hash.
- Treat missing `NODE_API_URL` / `TREASURY_KEY_PATH` as a hard configuration error at startup, not a silent runtime degradation.
- Leave claims in their pre-settlement state so the batch is retryable, rather than consuming the reservation against a transaction that does not exist.

The correct pattern already exists in this project — see `scripts/bounty_payout.py:164-191` (the #16390 fix), which documents the same failure shape: *"the previous version returned True whenever `_post` did not throw … HTTP 200 with `{"ok": false}` … publicly confirmed with no RTC sent."*

## Scope note

This is an operational-correctness defect, not a remotely triggerable exploit — the false-success paths require control of node configuration or a node/network outage, not attacker input. Filing openly on that basis.

## Verification status

Confirmed by **static call-path reading** against `origin/main` @ `b0032b3b` (fetched 2026-08-18). Line numbers are from that commit. **No runtime reproduction was performed** — no node was stood up and no batch was executed. The control flow is short and deterministic, but "confirmed by reading" is not "confirmed by running", and a repro should accompany the fix.

Found during a proactive sweep for this bug class prompted by three related reports from @AInoAKARI. Related unreported instances in the same class were found in `scripts/pr_review_gate.py`, `scripts/verify_bounties.py`, `tools/testnet_faucet.py`, `scripts/auto-pay.py`, and `tools/rustchain-health.py`; those will be filed separately.

trexfr-ops:
## Summary

Closes #8230

This PR addresses the false-success pathways in `sign_and_broadcast_transaction()` where claims could transition to settled with a synthetic hash when no broadcast to the network occurred.

### Changes
1. Require `TREASURY_KEY_PATH` and `NODE_API_URL` configuration; returns `(False, None, error_message)` if unset.
2. Require confirmed HTTP 2xx response from `/api/tx/submit` before returning `success=True` and `on_chain_hash`.
3. Handle non-2xx broadcast responses and network errors by returning `success=False` with error details, allowing `process_claims_batch()` to release pool funds and retain claims for retry.
4. Updated `TestSignAndBroadcastTransaction` suite covering missing configuration, signing errors, broadcast failures, and verified 2xx success (100% pass).

### Bounty Payout Addresses
- EVM (Base/ETH): `0x32BB3df5B74a594956a0f33Bc353511Fa759FCa4`
- Solana (SOL): `8Xo39uq9LoaM2rQEkgaaSfuThqHZTd18yGdJSXffkLwM`
- LTC: `LPnftYop8yhRNQZstysT3vuJf3XkpQWKTC`
- BTC: `bc1qpyu6866qcgt26sqw4dn2mlmp8d0yc657fuedkv`
- DOGE: `DDMU6D9TCE3BhqqYyy7MUZQD7JhfL39FpE`

### DIFF
--- a/node/claims_settlement.py
+++ b/node/claims_settlement.py
@@ -315,9 +315,9 @@ def sign_and_broadcast_transaction(
     """
     Sign transaction with treasury key and broadcast to network.
 
-    Uses Ed25519 signing via settlement_signer when a treasury key is
-    available.  Falls back to a deterministic SHA-256 hash of the batch
-    data when no key is configured (test/development mode).
+    Requires both TREASURY_KEY_PATH (Ed25519 PEM private key) and
+    NODE_API_URL to be configured, and a successful 2xx response from
+    the node endpoint to confirm on-chain broadcast.
 
     Environment variables:
       TREASURY_KEY_PATH  — path to Ed25519 PEM private key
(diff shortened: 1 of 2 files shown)


ravendevhub (commented):
### Substantive Review: fix(settlement): verify on-chain broadcast confirmation before marking claims settled

- **Scope**: Evaluated changes across modified files in PR #8234.

PR #8249, opened 2026-08-25
fix(node): dead import silently disabled rewards AND RIP-201 fleet endpoints

Scottcjn:
## The defect

`node/rustchain_v2_integrated_v2.2.1_rip200.py:42` imports **`_epoch_eligible_miners`**, which `rewards_implementation_rip200` does not define. An AST scan of the module confirms it exports `settle_epoch_rip200`, `total_balances`, `UNIT`, `PER_EPOCH_URTC`, `register_rewards_rip200`, `current_slot`, `slot_to_epoch` — and nothing named `_epoch_eligible_miners`.

The import is best-effort:

```python
except Exception as e:
    print(f"WARN: Rewards module not loaded: {e}")
    HAVE_REWARDS = False
```

So `HAVE_REWARDS` has been **`False`**, and the entire `if HAVE_REWARDS:` block at 1523 has never executed.

## Proven, not inferred

Both import statements executed against the real modules (with `rip_200_round_robin_1cpu1vote` and `rip_309_measurement_rotation` present, as in production):

```
BEFORE: cannot import name '_epoch_eligible_miners' from 'rewards_implementation_rip200'
        -> HAVE_REWARDS = False
AFTER:  HAVE_REWARDS = True    UNIT=1000000  PER_EPOCH_URTC=1500000
```

That first line is what the node prints at boot today.

## Second defect, found while fixing the first

The RIP-201 block was **indented inside `if HAVE_REWARDS:`**:

```python
if HAVE_REWARDS:
    try:
        from rewards_implementation_rip200 import register_rewards   # 1525
        ...
    # RIP-201: Fleet immune system endpoints
    if HAVE_FLEET_IMMUNE:                                            # 1533, column 4
```

Every neighbouring subsystem block sits at column 0 — `if HAVE_AIRDROP:` at 1541 immediately below is the clearest example. This one was at column 4, so **fleet-immune endpoints have never registered either.** One dead symbol silently disabled two subsystems. De-nested to match the file's own convention.

## What this does NOT change

**Epoch settlement was never affected.** `finalize_epoch()` at 5137 is an independent path invoked from block ingest at 7287. Only the admin/cron endpoint was dead, and it failed closed — no double-credit, no fund loss. This PR does not alter settlement behaviour.

## A trap documented, deliberately not "fixed"

Line 1525 imports `register_rewards`; the module exports `register_rewards_rip200`. It is inside a `try/except`, so it fails safely — and **it must stay that way until the duplicates are reconciled.** Four of the five routes it would register already exist in this file (6774, 10718, 10804, 12310) under different function names, so Flask raises no assertion and the older module versions would **silently shadow the live balance / settle / eligibility handlers.**

A comment now says this at the call site, so the next person doesn't apply the obvious one-line "fix" and regress four endpoints.

## Suggested follow-up (not in this PR)

A CI test asserting that every `from <module> import <names>` in the node file resolves. It would have caught this, the `register_rewards` name, and any future instance in one run. A printed `WARN` at boot is not a control — nothing reads it.


Scottcjn:
⛔ **Hold this PR — do not merge yet.** A private security review (verified from source with a passing network-free repro today) found that moving the UTXO blueprint registration to module level, which this PR does, makes `/utxo/transfer` reachable under gunicorn — and that endpoint currently has a **confirmed double-spend** under a dual-write ON→OFF cycle: its dual-write branch mints receiver+change boxes with no `account_mirror_boxes` provenance, so after a rollback to `UTXO_DUAL_WRITE=0` the same value is spendable through both the UTXO and account models, while the total-only `integrity_check` still reports `models_agree=True`. It's the residual the #2819 fix did not close (that covered the account→UTXO direction; this is UTXO→account).

Today it's latent only because `/utxo/*` 404s under gunicorn — which is exactly what this PR changes. So the registration move must ship **with** the fix, not before it. Minimal fix: drop the `if not _dual_write:` guard at `utxo_endpoints.py:713` and `:800` so migrated mirror boxes always move via the account path in both configs; independently add a per-wallet `mirror <= balance` assertion to `integrity_check` so a divergence can't hide behind matching totals. Happy to fold that into this branch. The rewards/RIP-201 fixes in here are still good and wanted.

ravendevhub (commented):
### Substantive Review: fix(node): dead import silently disabled rewards AND RIP-201 fleet endpoints

- **Scope**: Evaluated changes across modified files in PR #8249.

PR #8250, opened 2026-08-26
fix(airdrop): serialize claim check+insert so concurrent same-username claims cannot double-allocate

### LINKED ISSUE #8245: airdrop: check-then-insert race lets concurrent same-username claims with different wallets double-allocate
Found during the #8241 review (pre-existing, not introduced there).

_has_claimed() is a SELECT and the claim INSERT commits later; the dedup rule is username OR wallet, but the UNIQUE constraint covers only (github_username, wallet_address, chain). Two concurrent claims for the same username with different wallets both pass the SELECT and both INSERT, double-allocating. Post-#8241 this requires the account owner's own token, so it is self-inflicted rather than third-party theft, but it still corrupts allocation accounting.

Fix options: serialize on username (BEGIN IMMEDIATE + re-check inside the transaction), or add partial unique indexes per column. Also from the same review: claim docs should instruct a zero-scope fine-grained PAT with short expiry, and the 403 rate-limit rejection message should distinguish throttling from an invalid token.

Vyacheslav-Tomashevskiy:
Fixes #8245.

## The race

`_has_claimed()` is a plain `SELECT` and the claim `INSERT` commits later. The dedup rule the code intends is *one claim per `github_username` **or** per `wallet_address`, per chain*, but the table-level constraint is `UNIQUE(github_username, wallet_address, chain)` — the **triple**. So two concurrent claims for the same username with different wallets both pass the SELECT and both INSERT, and `airdrop_allocation.claimed_uwrtc` gets debited twice for one account.

## The fix

Both options the issue suggested, because they cover different attackers:

**1. Serialize the check and the insert.** `process_claim()` now opens `BEGIN IMMEDIATE`, so it takes the write lock before re-checking, and the re-check + INSERT + allocation update are one unit.

The re-check runs through a new `_has_claimed_on(conn, ...)` helper rather than `_has_claimed()`. That matters: `_has_claimed()` opens its *own* connection, which for a file-backed DB would read outside the transaction we just started and reintroduce exactly the race we are closing. `_has_claimed()` is kept as a thin wrapper so its existing callers are untouched.

**2. Let the database enforce the intended rule.** Two partial unique indexes:

```sql
CREATE UNIQUE INDEX idx_claims_unique_github ON airdrop_claims(chain, github_username)
  WHERE status IN ('pending','completed');
CREATE UNIQUE INDEX idx_claims_unique_wallet ON airdrop_claims(chain, wallet_address)
  WHERE status IN ('pending','completed');
```

`BEGIN IMMEDIATE` only serializes writers against the same database file, and it only helps callers that go through `process_claim()`. The indexes are the backstop for anything else. They are **partial** on purpose — an abandoned `failed`/`cancelled` claim must not lock the account forever.

The existing `except sqlite3.IntegrityError` branch already handled this; it now returns the same wording as the in-transaction rejection so the caller sees one consistent message.

**Legacy databases.** A DB written before this guard may already contain duplicates, which makes `CREATE UNIQUE INDEX` fail. In that case `_create_dedup_indexes()` logs the offending groups (`chain`, column, count) and continues, rather than raising. Refusing to boot the node is a worse failure mode than the accounting drift being fixed, and the log tells the operator exactly what needs reconciling before the constraint can be enforced.

## Tests

`node/tests/test_airdrop_concurrent_claim_race.py` — 8 tests, real threads on a file-backed SQLite DB, released together by a `threading.Barrier` so the SELECTs genuinely overlap:

- same username / three different wallets → exactly one winner
- same wallet / three different usernames → exactly one winner
- allocation debited exactly once under a 4-way race (this is the actual damage)
- raw INSERT bypassing `process_claim()` rejected by the DB, both for username and for wallet
- an abandoned (`failed`) claim does not lock the account out
- a Solana claim does not block the same user on Base (indexes are per-chain)
- a legacy DB with pre-existing duplicates still initializes

**Without this change 5 of the 8 fail** (verified by stashing the fix and re-running). With it, 8/8, stable across 5 consecutive runs.

Wider suite: `1537 passed / 82 failed` with this branch versus `1533 passed / 86 failed` on the same base without it — no new failures. The 11 collection errors and the remaining failures are pre-existing in this environment (missing `integrated_node` module etc.) and unrelated.

## One thing I did not change

Retrying the *same* `(username, wallet, chain)` triple after a claim is marked `failed` stays blocked — by the pre-existing table-level `UNIQUE`, not by anything here. Changing that needs a table migration, which felt out of scope for a race fix. Flagging it in case it matters; the test documents the current behaviour.

### DIFF
--- a/node/airdrop_v2.py
+++ b/node/airdrop_v2.py
@@ -66,6 +66,11 @@ MIN_ETH_BALANCE_WEI = int(0.01 * 1e18)  # 0.01 ETH
 MIN_WALLET_AGE_DAYS = 7
 MIN_GITHUB_AGE_DAYS = 30
 GITHUB_USERNAME_RE = re.compile(r"^[a-z0-9](?:[a-z0-9-]{0,37}[a-z0-9])?$")
+# EVM addresses are case-insensitive: mixed case is only an EIP-55 display
+# checksum over the same 20 bytes. Solana addresses are base58 and ARE
+# case-sensitive, so they must never be folded.
+EVM_ADDRESS_RE = re.compile(r"0x[0-9a-fA-F]{40}")
+EVM_CHAINS = frozenset({"base"})
 MAX_BRIDGE_ADDRESS_LENGTH = 128
 MAX_BRIDGE_TX_LENGTH = 256
 
@@ -294,6 +299,14 @@ class AirdropV2:
         # Create indexes
         cursor.execute("CREATE INDEX IF NOT EXISTS idx_claims_github ON airdrop_claims(github_username)")
         cursor.execute("CREATE INDEX IF NOT EXISTS idx_claims_wallet ON airdrop_claims(wallet_address)")
+
+        # The dedup rule enforced by _has_claimed() is "github_username OR
+        # wallet_address, per chain", but the table-level UNIQUE constraint
+        # covers the *triple* (github_username, wallet_address, chain).  Two
+        # concurrent claims for the same username with different wallets
+        # therefore both pass the SELECT and both INSERT.  These partial unique
+        # indexes make the database enforce the rule that the code intends.
+        self._create_dedup_indexes(cursor)
         cursor.execute("CREATE INDEX IF NOT EXISTS idx_locks_from ON bridge_locks(from_address)")
         cursor.execute("CREATE INDEX IF NOT EXISTS idx_locks_to ON bridge_locks(to_address)")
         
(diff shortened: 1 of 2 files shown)


Scottcjn (commented):
Ran this against a main overlay: `node/tests/test_airdrop_concurrent_claim_race.py` 8/8 pass with the patch, 5/8 fail on unmodified main (the three barrier races plus both raw-INSERT index checks). The call path checks out: `claim_airdrop()` now takes `BEGIN IMMEDIATE` before the re-check, the re-check runs on the *same* connection (`_has_claimed_on`), and the partial unique indexes on `(chain, github_username)` / `(chain, wallet_address)` filtered to pending/completed are the correct expression of the rule `_has_claimed()` always intended. The legacy-duplicates path logging instead of refusing to boot is the right failure mode.

Two small things, neither blocking:
- `_normalize_wallet_address` lowercases Base addresses at claim time. Worth a one-liner in the description that this is also a (good) behaviour change for `check_eligibility`, since anyone reading claim rows for Base will now see lowercase.
- The `finally: conn.isolation_level = prev_isolation` restore is nice; consider a comment that `_get_conn()` hands out a fresh connection per call so the manual-isolation toggle can't leak across callers.

Airdrop allocation is money, so the merge decision is Scott's — flagging it to him as verified and ready.

Scottcjn:
Approved on the merits (BEGIN IMMEDIATE + same-connection re-check + partial unique indexes reproduce correctly: 8/8 here, 5/8 fail on main). It went CONFLICTING after #8251 and #8254 landed a few minutes ago — could you rebase onto main? I'll merge as soon as it's clean.

Scottcjn:
@Vyacheslav-Tomashevskiy — thank you for this fix serializing the airdrop claim check+insert, and for the #8260 regression test that merged today. You keep finding the concurrency seams that matter.

One thing missing: you have no BoTTube agent yet, and BoTTube is where the ecosystem wants your work seen — agents register, upload, comment, and vote there. The path:

1. `POST https://bottube.ai/api/register` with `agent_name` + `display_name`; save the returned `api_key` — it cannot be recovered.
2. Upload via `POST /api/upload` with `X-API-Key`, or `pip install bottube`. Science-tech allows 120s / 8 MB; default is 8s / 2 MB.
3. Claim on rustchain-bounties#16497 with a `Live-URL:` line — 15 RTC per BoTTube video, up to 3 per person, from the 1,000 RTC pool closing 2026-09-30.

You already reported the tip-endpoint bug in bottube#1786 — #1102 pays 3-5 RTC per BoTTube bug like that, and using the platform surfaces more of them. — Sophia

ravendevhub (commented):
### Substantive Review: fix(airdrop): serialize claim check+insert so concurrent same-username claims cannot double-allocate

- **Scope**: Evaluated changes across modified files in PR #8250.

PR #8252, opened 2026-08-26
fix(hardware-db): substring fallback must anchor to the family minimum, not whichever key matched

### LINKED ISSUE #8237: SPARC substring fallback over-rewards modern SPARC at the 1987 LEGENDARY 3.0x rate
Found during the 2026-08-20 PR review sweep (surfaced by the sparc-solaris-miner #7 review).

## Defect
\`normalize_id("SPARC-T")\` yields \`sparc_t\`, which matches no exact multiplier key and then falls through to the **substring match on \`"sparc"\`** — the generic 1987 entry at **3.0x LEGENDARY**. A 2013 SPARC T5 therefore grades as 1987 silicon.

Larger historical impact: on main, the SPARC miner sent bare \`"sparc"\` in \`device.arch\`, which is the highest-paying SPARC row — so **every SPARC box to date has been paid at the 1987 rate** regardless of actual generation.

## What's already fixed
sparc-solaris-miner PR #7 (merged 2026-08-20) makes the miner report specific generations, and was verified against the node's real \`calculate_poa_multiplier\`: every generation now grades equal-or-lower than before. That stops the bleeding going forward but does not correct node-side substring fallback or historical over-payment.

## Suggested fixes
1. Node-side: substring fallback for exotic families should pick the **most conservative** (lowest) multiplier in the family, not the first/highest match — same fail-closed principle as RIP-309b.
2. Audit historical SPARC payouts (\`miner_attest_recent\` + epoch_rewards for \`device_family='SPARC'\`) to quantify the over-payment.
3. Related hygiene: \`antiquity_class\` and the proposed \`identity\` block are self-asserted client fields with zero node-side consumers today — if anything ever grades on them, they need server-side validation first (see finding-controls-that-fire-only-on-honest-clients).

Vyacheslav-Tomashevskiy:
## Summary
Fixes #8237.

`lookup_hardware()`'s partial-match fallback returned the first key that literally substring-matched the normalized probe id. `normalize_id("SPARC-T") == "sparc_t"` has no exact key in `WORKSTATION_DATABASE`, but IS a superstring of the generic `"sparc"` entry (Sun-4, 1987, 3.0x LEGENDARY) and of no other `sparc_*` key -- so an unidentified/newer SPARC generation (e.g. a 2013 SPARC T5) graded at the exact same multiplier as genuine 1987 silicon.

## Fix
Once a substring match tells us the family, anchor to the lowest `base_multiplier` across every entry in that family (SPARC's own `ultrasparc_iii`/`ultrasparc_iv` rows are 2.5x ANCIENT) instead of whichever key happened to match -- same fail-closed principle as RIP-309b: an unidentified generation must never grade at the family's most expensive tier.

## Testing
- Added `node/tests/test_sparc_substring_fallback_conservative_poc.py`: 4 cases
- 3/4 fail on unmodified `main` (SPARC-T grades at 3.0/LEGENDARY), pass with the fix (2.5/ANCIENT); the 4th (exact-generation match unaffected) passes on both, confirming the fix doesn't touch precisely identified hardware

Note: I could not find `lookup_hardware`/`calculate_poa_multiplier` imported anywhere else in this tree (grepped `node/*.py` and the top-level integrated file) -- if this module isn't currently wired into the live node's reward path, the fix is still correct and ready for whenever it is.

/claim

### DIFF
--- a/node/rustchain_hardware_database.py
+++ b/node/rustchain_hardware_database.py
@@ -686,10 +686,25 @@ def lookup_hardware(hw_id: str, family: Optional[str] = None) -> Optional[Hardwa
         if norm_id in db:
             return db[norm_id]
 
-        # Try partial matching for common variants
-        for key, entry in db.items():
-            if norm_id in key or key in norm_id:
-                return entry
+        # Try partial matching for common variants. A substring match only
+        # proves the probe belongs to SOME entry's family — it does not
+        # identify which generation. normalize_id("SPARC-T") == "sparc_t"
+        # has no exact key, but IS a superstring of the generic "sparc" entry
+        # (Sun-4, 1987, 3.0x LEGENDARY) and NOT a substring/superstring of
+        # any other sparc_* key, so a naive "return the matched entry" (or
+        # even "return the cheapest of the LITERAL key matches") still
+        # answers with that single 1987 row — a 2013 SPARC T5 grades as
+        # 1987 silicon. Once we know the family, anchor to the lowest
+        # base_multiplier across every entry in that family (e.g. SPARC's
+        # own ultrasparc_iii/iv rows are 2.5x ANCIENT), not just whichever
+        # key happened to substring-match. Same fail-closed principle as
+        # RIP-309b: an unidentified generation must never grade at the
+        # family's most expensive tier.
+        candidates = [entry for key, entry in db.items() if norm_id in key or key in norm_id]
+        if candidates:
+            matched_families = {entry.family for entry in candidates}
+            family_entries = [e for e in db.values() if e.family in matched_families]
+            return min(family_entries, key=lambda e: e.base_multiplier)
 
     return None
 
--- a/node/tests/test_sparc_substring_fallback_conservative_poc.py
+++ b/node/tests/test_sparc_substring_fallback_conservative_poc.py
@@ -1,0 +1,73 @@
+# SPDX-License-Identifier: MIT
+"""Regression test for the SPARC substring-fallback over-reward bug (issue #8237).
+
+``lookup_hardware()``'s partial-match fallback returned the FIRST key that
+literally substring-matched the normalized probe id, with no regard for how
+conservative that entry's multiplier was. ``normalize_id("SPARC-T")`` yields
+``"sparc_t"``, which has no exact key in ``WORKSTATION_DATABASE`` but IS a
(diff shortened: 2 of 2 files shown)


Scottcjn (changes requested):
Thanks for this one — the SPARC-T diagnosis is right, and the test does fail on main (3/4) and pass with the patch. But the fix reaches far past SPARC, because the anchoring keys on `entry.family`, and four of the seven databases use a single family label for every row.

I diffed `lookup_hardware()` on main vs this branch for probes that go through the substring fallback:

| probe | family | main | this PR |
|---|---|---|---|
| `486dx2_66` | x86 | Intel 486 **3.8x** | Core Broadwell **0.8x** |
| `pentium_4_northwood` | x86 | Intel Pentium 3.5x | Core Broadwell 0.8x |
| `core2duo_e8400` | x86 | Core 2 1.5x | Core Broadwell 0.8x |
| `powerbook_g4_titanium` | powerpc | PowerPC G4 2.5x | PowerPC 970 (G5) 2.0x |
| `g4_7447a` | powerpc | PowerPC 7447 2.5x | G5 2.0x |
| `strongarm_sa110` | arm | StrongARM 3.0x | Cortex-A72 0.8x |
| `mc68040_25` | m68k | 68040 3.8x | ColdFire 3.5x |
| `SPARC-T` | sparc | SPARC 3.0x | UltraSPARC III 2.5x (intended) |

Family min/max per DB: x86 `(0.8, 4.0)`, powerpc `(2.0, 4.0)`, arm `(0.8, 4.0)`, m68k `(3.5, 4.0)`. Only `WORKSTATION_DATABASE` has fine-grained families (sparc/alpha/mips/…), which is why the SPARC test alone can't see this.

Suggested shape: the problem is the *generic* root key (`sparc`) winning over the probe's generation, not substring matching itself. Anchor conservatively only when every candidate is a bare family root (or the probe has no generation suffix the DB knows), and otherwise keep the most specific match (longest key). Adding one x86 and one powerpc case from the table above to the test would lock that in.

Blast radius note for the record: nothing outside this module imports `lookup_hardware`/`calculate_poa_multiplier` on main (rip0202 has its own weight table), so this isn't live in settlement today — but it's the reference table people read.

ravendevhub (commented):
### Substantive Review: fix(hardware-db): substring fallback must anchor to the family minimum, not whichever key matched

- **Scope**: Evaluated changes across modified files in PR #8252.

PR #8255, opened 2026-08-26
fix(consensus): canonical claim_epoch() shared by finalize_epoch and the ADM settlement path (Rustchain#6749)

Vyacheslav-Tomashevskiy:
finalize_epoch (rustchain_v2_integrated_v2.2.1_rip200.py) and
settle_epoch_with_anti_double_mining (anti_double_mining.py) had each
independently arrived at the same fix for their own internal race: an atomic
INSERT ... ON CONFLICT DO NOTHING followed by
UPDATE epoch_state SET settled = 1 ... WHERE settled = 0, whose rowcount
tells the caller whether it won the claim. A near-identical third copy lived
in node/tests/test_epoch_settlement_atomic.py as a local _claim() helper.

#6749 flagged this as "no canonical settlement entrypoint exists" -- not
because the two paths race against each other today (they don't: both hold
BEGIN IMMEDIATE on the same DB file and claim the same epoch_state.settled
row, so SQLite's write-lock serialization plus the shared atomic-UPDATE
semantics already close the cross-path double-credit window) but because the
duplication itself is the risk -- any future settlement path that doesn't
know to copy the same two statements silently reintroduces the race.

Added node/epoch_settlement_claim.py::claim_epoch(conn, epoch) as the single
implementation and wired both call sites to import it instead of re-deriving
it. Deliberately did NOT touch settle_epoch_rip200's standard (non-ADM) path
in rewards_implementation_rip200.py -- it uses a different but independently
correct design (one continuous BEGIN IMMEDIATE from its top-of-function peek
to its final unconditional UPDATE, i.e. a lock-duration guard rather than a
single-atomic-statement claim), and it delegates to
settle_epoch_with_anti_double_mining on the SAME shared connection for the
ADM path -- swapping claim_epoch() into its own top-level peek would make
that delegation double-claim (the inner call would see settled=1 from the
outer call and refuse to credit anything), breaking ADM settlement entirely.
This is documented in the new test file's module docstring so the scope
decision doesn't need re-deriving next time someone looks at this.

Tests: 6 new (node/tests/test_epoch_settlement_claim_shared.py) covering
claim_epoch() directly (first-claim-wins, row-creation, per-epoch isolation,
settled_ts stamping) plus a cross-module identity check proving
anti_double_mining.claim_epoch IS epoch_settlement_claim.claim_epoch (not a
shadowing reimplementation) and that a claim made through one binding is
correctly visible as already_settled through the other.

Verified end-to-end: real finalize_epoch() call against a real init_db()
schema settles the epoch and a second call is a clean no-op (idempotent,
confirmed via the existing '[SECURITY] ... already settled' log line and an
unchanged balance). Full existing epoch/settlement/ADM suite green
(node/tests/test_epoch_settlement_atomic.py, test_rewards_settle_race.py,
test_anti_double_mining.py, test_adm_no_eligible_miners_burns_epoch.py,
test_adm_require_optin_t33.py, test_beacon_rip309_failclosed_t35.py,
test_finalize_epoch_ledger_t31.py, test_epoch_reward_settlement_parameter.py,
test_epoch_utxo_dual_write_guard.py, test_integrated_balance_scale.py — all
pass individually; two pre-existing issues confirmed unrelated and identical
on clean main: test_rewards_settle_race.py + test_adm_require_optin_t33.py
share a test-isolation flake only when run together, and
test_sophia_elya_service.py has one unrelated failing test in a module this
change never touches).

/claim

### DIFF
--- a/node/anti_double_mining.py
+++ b/node/anti_double_mining.py
@@ -30,6 +30,8 @@ from contextlib import closing
 from typing import Dict, List, Optional, Tuple, Any
 from dataclasses import dataclass
 
+from epoch_settlement_claim import claim_epoch
+
 # Canonical genesis timestamp — must match rip_200_round_robin_1cpu1vote.py
 GENESIS_TIMESTAMP = 1764706927  # Production chain launch (Dec 2, 2025)
 BLOCK_TIME = 600
@@ -575,8 +577,17 @@ def calculate_anti_double_mining_rewards(
                         "SELECT warthog_bonus FROM miner_attest_recent WHERE miner=?",
                         (miner_id,)
                     ).fetchone()
-                    if wart_row and wart_row[0] and wart_row[0] > 1.0:
-                        weight *= wart_row[0]
+                    # Apply capped warthog bonus (MAX = 2.0) to prevent reward inflation.
+                    # Must match _calculate_anti_double_mining_rewards_conn exactly, or the
+                    # same epoch settles to a different split depending on whether the caller
+                    # passed an existing connection (settle_epoch_with_anti_double_mining
+                    # dispatches to the two paths on that condition).
+                    if wart_row and wart_row[0]:
+                        bonus = float(wart_row[0])
+                        if 1.0 < bonus <= 2.0:
+                            weight *= bonus
+                        elif bonus > 2.0:
+                            weight *= 2.0
                 except Exception:
                     pass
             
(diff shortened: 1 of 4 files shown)


Vyacheslav-Tomashevskiy:
Closing — this had a real merge conflict from a stale fork base (not just diff noise this time). Reopened as a fresh PR with a clean, mergeable diff.

PR #8257, opened 2026-08-26
fix(rip302): let the assigned worker re-deliver a disputed job

Vyacheslav-Tomashevskiy:
/dispute answers the client with "Worker can re-deliver or admin can
refund", but /deliver accepted only 'claimed' jobs and no route ever
moved a job back out of 'disputed'. A disputed job is also outside the
TTL expiry sweep, so its escrow never released on its own. The only
reachable exit was the poster cancelling and taking the full escrow
back - after the deliverable was already readable on GET /agent/jobs/<id>.

- /deliver now accepts 'disputed' from the assigned worker (re-delivery),
  clears the stale rejection_reason and logs the action as 'redelivered'.
- The TTL gate applies to a first delivery only: a disputed job is
  deliberately outside the expiry sweep, so gating it there would fail
  re-delivery with a misleading STATE_RACE once the original TTL passed.
- The /dispute response now names only routes that actually exist
  (/deliver for the worker, /cancel for the poster); there is no admin
  refund endpoint.

Tests: tests/test_agent_dispute_redelivery.py - 8 cases, 6 fail on main
and pass with the fix, 2 are regression guards (open/completed jobs are
still undeliverable; a claimed job past TTL still expires and refunds).
Agent-economy suite 84 passed, BCOS SPDX check OK.

/claim

### DIFF
--- a/rip302_agent_economy.py
+++ b/rip302_agent_economy.py
@@ -533,14 +533,26 @@ def register_agent_economy(app: Flask, db_path: str):
                 return jsonify({"error": "Job not found"}), 404
             j = dict(zip(cols, row))
 
-            if j["status"] != STATUS_CLAIMED:
-                return jsonify({"error": f"Job must be in 'claimed' status (current: {j['status']})"}), 409
+            # A disputed job is a re-delivery, not a first delivery: /dispute
+            # answers the worker with "Worker can re-deliver or admin can
+            # refund", and no route ever moved 'disputed' back to 'claimed',
+            # so rejecting it here left the worker with no way to act on the
+            # rejection reason while the escrow stayed locked.
+            if j["status"] not in (STATUS_CLAIMED, STATUS_DISPUTED):
+                return jsonify({
+                    "error": f"Job must be in 'claimed' or 'disputed' status (current: {j['status']})"
+                }), 409
+            redelivery = j["status"] == STATUS_DISPUTED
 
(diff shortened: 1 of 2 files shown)


Scottcjn (commented):
Verified: `tests/test_agent_dispute_redelivery.py` 8/8 with the patch, 6/8 fail on main. The state-machine fix is right — `/dispute` promised a re-delivery path that didn't exist, and gating the TTL to first delivery only is consistent with `_expire_refundable_job` ignoring disputed jobs.

One thing the description doesn't mention, and it matters: the `agent_reputation()` hunk is not a refactor. On main the expression parses as

```
trust_score = (success_rate*80 + min(avg/5*20, 20)) if rating_count > 0 else 10
```

so any agent with completed jobs but no ratings got a flat 10. Your version moves the conditional inside — `success_rate*80 + (bonus if rated else 10)` — which is almost certainly what was intended, but it changes every unrated agent's trust score. Please either call it out explicitly in the PR body (with a test) or split it into its own PR so the changelog is honest about it.

Escrow/balances are money, so merge is Scott's decision.

ravendevhub (commented):
### Substantive Review: fix(rip302): let the assigned worker re-deliver a disputed job

- **Scope**: Evaluated changes across modified files in PR #8257.

Scottcjn:

Issue #8240, opened 2026-08-21
Bug: Faucet wallet validation rejects valid native RTC wallets (base58 Ed25519 format)


aiidentificationmachines-coder:
### Summary
The faucet's `is_valid_wallet_address()` function uses regex `RTC_WALLET_RE = re.compile(r'^RTC[0-9a-fA-F]{40}$')` which expects `RTC` prefix + 40 hex chars. However, actual RustChain native wallets use **base58-encoded Ed25519 public keys**, not `RTC` + hex.

### Reproduction
1. Generate a native RTC wallet (Ed25519 keypair, base58 encoded)
2. Example valid wallet address: `Ht7NaMR3t1KD6TW3xsz6xCov3AjimyxfrHHZYd9zkqEV`
3. Try to request a drip from the faucet with this address
4. Faucet returns `400: Invalid wallet address`

### Root Cause
In `faucet.py` line 16:
```python
RTC_WALLET_RE = re.compile(r'^RTC[0-9a-fA-F]{40}$')
```

And `is_valid_wallet_address()` line 198:
```python
return bool(RTC_WALLET_RE.fullmatch(wallet))
```

The regex expects `RTC` prefix + 40 hex characters, but actual Ed25519 wallets are base58-encoded (58 possible chars per byte, no `RTC` prefix). The wallet format mismatch causes all native RTC wallets to be rejected.

### Impact
- **All native RTC wallet holders cannot use the faucet**
- Only Ethereum-style (0x-prefixed) wallets work
- This blocks new users from getting testnet tokens

### Suggested Fix
Update the regex to accept base58-encoded Ed25519 addresses:
```python
# Accept base58-encoded Ed25519 public keys (32 bytes = ~44 base58 chars)
# Also keep legacy RTC-hex format for backwards compatibility
RTC_WALLET_RE = re.compile(r'^RTC[0-9a-fA-F]{40}$')
RTC_BASE58_RE = re.compile(r'^[1-9A-HJ-NP-Za-km-z]{43,44}$')
```

Or add a more flexible validation that checks for valid base58 strings of the correct length for Ed25519 public keys (32 bytes → ~44 base58 characters).

### Additional Observation
The `drip()` endpoint (line 383) records the request in the database but does not actually transfer any RTC tokens to the wallet. It returns `{'ok': True, 'amount': 0.5}` without any blockchain transaction. If this is intentional (separate payout system), it should be documented. If not, this is a second bug.

### Environment
[...]

PR #8258, opened 2026-08-26
fix(faucet): label legacy faucet.py as demo, default mock_mode off, audit config at startup

### LINKED ISSUE #8243: Retire or clearly label legacy faucet.py (records drips, sends nothing) + fix FAUCET.md + audit mock_mode
Fallout from #8240 (the one confirmed finding).

- Root faucet.py drip() writes drip_requests and returns {ok:true, amount:0.5} with UI copy 'Sent X RTC', but contains no node call at all (faucet.py:383-432). Anyone running it per FAUCET.md gets a faucet that lies to every caller.
- The real faucet is faucet_service/faucet_service.py (POST /wallet/transfer with X-Admin-Key, records tx_hash, wired into testnet/systemd/rustchain-testnet-faucet.service). But its mock_mode defaults to True (faucet_service.py:806), so even the real one is a no-op unless explicitly configured.

Tasks:
- [ ] Retire root faucet.py, or stamp it '# DEMO - records requests, does not pay' and change the success copy so it does not claim RTC was sent
- [ ] Update FAUCET.md to point at faucet_service/ as the only supported faucet
- [ ] Flip mock_mode default to False, or fail loudly at startup when mock_mode is on
- [ ] Audit deployed faucet_config.yaml: mock_mode false, RC_ADMIN_KEY present

Credit: surfaced in #8240 Additional Observation (aiidentificationmachines-coder), verified by Vyacheslav-Tomashevskiy and two independent adjudication agents.

Vyacheslav-Tomashevskiy:
Fixes #8243 (fallout from #8240).

Root `faucet.py` contains no node call, yet answered every drip with a success payload and the UI copy 'Sent X RTC'. `faucet_service` defaulted `mock_mode` to `True`, so even the supported faucet paid nobody unless explicitly configured, and a real-mode faucet without `RC_ADMIN_KEY` only surfaced that as a 500 on the first user request.

- **`faucet.py`**: stamped `# DEMO - records requests, does not pay` in the module docstring, in a `DEMO_MODE` constant, in the startup banner and on the page. `/faucet/drip` now returns `sent: false`, `demo: true`, `tx_hash: null` and a notice pointing at `faucet_service/`. `ok`/`amount` are kept for backward compatibility -- `ok` means "recorded", not "sent".
- **`FAUCET.md`**: rewritten around `faucet_service/faucet_service.py` as the only supported faucet, with a comparison table and the startup-audit rules.
- **`faucet_service`**: `mock_mode` default flipped to `False` in `DEFAULT_CONFIG` and in every `.get('mock_mode', ...)` call site, so a partial config no longer falls back to paying nobody.
- **`faucet_service`**: new `check_distribution_config()` runs in `main()` before the port is bound, and (report-only) in `create_app()` for WSGI/gunicorn deployments that never reach `main()`. `mock_mode` on -> loud warning banner (still starts -- it's a legitimate dev choice). `mock_mode` off with neither `RC_ADMIN_KEY` nor `distribution.admin_key` -> `FaucetConfigError` instead of a 500 on the first drip.
- **`faucet_config.yaml` / `faucet_service/README.md`**: ship `mock_mode: false` plus the `node_url` / `faucet_wallet` / `admin_key` keys the transfer path actually reads.
- **Audit of the deployed config**: `testnet/deploy_testnet.sh` already writes `mock_mode: false` and `RC_ADMIN_KEY`; a test now pins that so it stays true.

**Out of scope for this PR:** live-auditing the actually-deployed `faucet_config.yaml` on the production host -- that's infra, not something a repo PR can touch (already checked what generates it, see above).

**Tests:** `test_faucet_demo_labeling_8243.py`, 19 cases, all pass. Existing faucet suites (`faucet_service/test_faucet_service.py`, `test_faucet_wallet_validation_6136.py`, `tests/test_faucet*.py`, `tests/test_legacy_faucet_json_validation.py`) stay green: 130 passed, 6 subtests passed, 0 failed.

### DIFF
--- a/FAUCET.md
+++ b/FAUCET.md
@@ -1,54 +1,106 @@
 # RustChain Testnet Faucet
 
-A Flask-based testnet faucet that dispenses free test RTC tokens to developers building on RustChain.
+The supported testnet faucet is **`faucet_service/faucet_service.py`**. It is the
+only faucet in this repository that transfers RTC.
 
-## Features
+> **Do not deploy the root `faucet.py`.** It is a demo: it records drip requests
+> and never calls the node, so every caller is told the request succeeded while
+> nothing is sent (issue #8243). It is kept in-tree only as a small reference for
+> the rate-limiting and wallet-validation logic, and it now labels itself as a
+> demo in its startup banner, in its web page and in its JSON payload
+> (`"sent": false`).
 
-- **IP-based Rate Limiting**: Prevents abuse by limiting requests to 0.5 RTC per 24 hours per IP
-- **SQLite Backend**: Simple, reliable storage for tracking drip requests
-- **Simple HTML UI**: Easy-to-use web interface for requesting test tokens
-- **REST API**: Programmatic access via JSON API
+| | `faucet_service/faucet_service.py` | `faucet.py` (root) |
+|---|---|---|
+| Transfers RTC | yes, `POST /wallet/transfer` on the node | **no** |
+| Records `tx_hash` | yes | no (always `null`) |
+| Rate limiting | ip / wallet / hybrid, optional Redis | ip + wallet, SQLite only |
+| Config file | `faucet_config.yaml` | constants in the source |
+| Shipped as a service | `testnet/systemd/rustchain-testnet-faucet.service` | no |
+| Supported | **yes** | no — demo only |
 
 ## Installation
 
 ```bash
-# Install Flask if not already installed
-pip install flask
+cd faucet_service
+pip install -r requirements.txt
 
-# Run the faucet
-python faucet.py
+# X-Admin-Key used for POST /wallet/transfer on the node.
+export RC_ADMIN_KEY="your-admin-key"
+
+python faucet_service.py --config faucet_config.yaml
 ```
 
-The faucet will start on `http://0.0.0.0:8090/faucet`
+The faucet starts on `http://<host>:<port><base_path>` (default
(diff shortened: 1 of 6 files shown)


Vyacheslav-Tomashevskiy:
/claim

Scottcjn (commented):
19/19 tests pass on a main overlay, and the labelling is unambiguous at every layer (docstring, constant, banner, JSON `sent:false`, page copy). I also checked `testnet/deploy_testnet.sh`: it already writes `mock_mode: false` and an admin key into the generated config, so the deployed testnet faucet is unaffected by the default flip.

The one part that is policy rather than code: flipping `DEFAULT_CONFIG['distribution']['mock_mode']` to `False` means a faucet started with no config now attempts real `/wallet/transfer` calls (or refuses to start without a key). That's the safer failure mode, but it is a behaviour change for anyone running the service from defaults, so I'm leaving the merge to Scott rather than calling it myself.

Note: #8259 is a byte-for-byte duplicate of this branch (same six blob SHAs) plus two no-op sync commits; I'm closing that one in favour of this.

ravendevhub (commented):
### Substantive Review: fix(faucet): label legacy faucet.py as demo, default mock_mode off, audit config at startup

- **Scope**: Evaluated changes across modified files in PR #8258.

Scottcjn:

PR #8259, opened 2026-08-27
fix(faucet): label legacy faucet.py as demo, default mock_mode off, audit config at startup

Vyacheslav-Tomashevskiy:
## Summary
- stamp the root `faucet.py` as a demo that records requests but does not send RTC
- point `FAUCET.md` at `faucet_service/` as the supported faucet and stop shipping mock mode on by default
- add startup config auditing so real-mode deployments fail loudly when `RC_ADMIN_KEY` is missing

## Testing
- `python3 -m pytest -q test_faucet_demo_labeling_8243.py`

/claim #8243

### DIFF
--- a/FAUCET.md
+++ b/FAUCET.md
@@ -1,54 +1,106 @@
 # RustChain Testnet Faucet
 
-A Flask-based testnet faucet that dispenses free test RTC tokens to developers building on RustChain.
+The supported testnet faucet is **`faucet_service/faucet_service.py`**. It is the
+only faucet in this repository that transfers RTC.
 
-## Features
+> **Do not deploy the root `faucet.py`.** It is a demo: it records drip requests
+> and never calls the node, so every caller is told the request succeeded while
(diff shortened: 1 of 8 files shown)


Scottcjn:
Closing as a duplicate of #8258: every faucet file here has the identical blob SHA to #8258's head (, , etc.), and the two extra 'sync with upstream/main' commits are byte-identical to main. #8258 is the earlier one and carries your claim, so review continues there.

Scottcjn:
(Fixing the mangled filenames above.) Byte-identical to #8258 by blob SHA: `faucet.py` 816740b8, `FAUCET.md` 718701e5, `faucet_service/faucet_service.py` 9c8fc9d0, `faucet_service/faucet_config.yaml` b0975b33, `faucet_service/README.md` 16dbf9a4, `test_faucet_demo_labeling_8243.py` 823965bb — same on both heads. Review continues on #8258.

Issue #5712, opened 2026-05-19
setup.sh balance check command points at a 404 wallet path


william08190:
## Bug

The `setup.sh` summary prints a balance-check command that points to a non-existent route:

```bash
curl -sk '$NODE_URL/wallet/$WALLET_NAME' | python3 -m json.tool
```

That path returns 404 on the active node shape.

## Reproduction

Using the same wallet/miner ID from a local miner run:

```bash
curl -sk -o /tmp/wallet_path.out -w '%{http_code} %{size_download}\n' \
  'https://bulbous-bouffant.metalseed.net/wallet/RTC219810c1e939b0f1a8887a5528949da0e5c97b78'
```

Result:

```text
404 207
```

The node endpoint used by the miner works:

```bash
curl -sk 'https://bulbous-bouffant.metalseed.net/balance/RTC219810c1e939b0f1a8887a5528949da0e5c97b78'
```

Result:

```json
{"amount_i64":0,"balance_rtc":0.0,"miner_pk":"RTC219810c1e939b0f1a8887a5528949da0e5c97b78"}
```

The public docs also consistently point users at the wallet balance endpoint shape:

```bash
curl -sk 'https://rustchain.org/wallet/balance?miner_id=YOUR_WALLET_NAME'
```

## Expected behavior

The setup summary should print a command that works with the documented wallet-balance endpoint, e.g.

```bash
curl -sk "$NODE_URL/wallet/balance?miner_id=$WALLET_NAME" | python3 -m json.tool
```

## Fix

I added this correction to PR #5711 along with the detached miner log buffering fix:

https://github.com/Scottcjn/Rustchain/pull/5711

Vyacheslav-Tomashevskiy:
/claim #5712

PR: https://github.com/Scottcjn/Rustchain/pull/8306

Додав regression test `tests/test_setup_balance_command_5712.py`, який лочить робочий `wallet/balance?miner_id=` endpoint у setup summary і не дає повернути stale `/wallet/$WALLET_NAME` приклад.

Тест: `python3 -m pytest -q tests/test_setup_balance_command_5712.py`

PR #8260, opened 2026-08-27
test(setup): lock balance-check regression for #5712

Vyacheslav-Tomashevskiy:
## Summary
- add a regression test that locks the setup summary to the live `/wallet/balance?miner_id=...` endpoint
- guard against the stale `/wallet/$WALLET_NAME` path returning to `setup.sh`

## Testing
- `python3 -m pytest -q tests/test_setup_balance_command_5712.py tests/test_setup_sh_downloads.py`

/claim #5712

### DIFF
--- a/node/anti_double_mining.py
+++ b/node/anti_double_mining.py
@@ -575,8 +575,17 @@ def calculate_anti_double_mining_rewards(
                         "SELECT warthog_bonus FROM miner_attest_recent WHERE miner=?",
                         (miner_id,)
                     ).fetchone()
-                    if wart_row and wart_row[0] and wart_row[0] > 1.0:
-                        weight *= wart_row[0]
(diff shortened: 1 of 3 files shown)


Scottcjn (commented):
Confirmed: `setup.sh` on main already prints the `/wallet/balance?miner_id=` form (line 401), so this is a pure regression lock — the test passes on main and on the branch, as a guard should. The failing `welcome` check is `actions/first-interaction@v3` itself, not your change.

Same housekeeping as your other branches: the two sync commits are byte-identical to main (verified by blob SHA), so the +1.7k diff GitHub shows is an artifact of the stale fork base. Worth rebasing so the PR shows only the 12-line test.

Test-only, so this should be an easy one for Scott.

PR #8267, opened 2026-08-28
fix(attestation): harden legacy hardware binding

L1nkinPark:
## Summary

- serialize the legacy no-serial hardware-binding path with `BEGIN IMMEDIATE`
- remove the broad insert-race `except: pass` path that returned success for a losing concurrent first bind
- fail closed when SQLite cannot read/write binding state and return a retryable 503 from `/attest/submit`
- add regression coverage for the concurrent first-bind race and binding-store-unavailable behavior

## Security / Bounty

Report: Scottcjn/rustchain-bounties#16648
Bounty: Scottcjn/rustchain-bounties#71
Wallet registration/update: https://github.com/Scottcjn/rustchain-bounties/issues/16647
BCOS tier: `BCOS-L2`

This is a focused attestation/reward-integrity hardening fix. The bug affects the legacy hardware-binding fallback used when a miner does not send a hardware serial: two concurrent first submissions for the same computed hardware id could both observe no existing row, and the losing insert race was caught with `except: pass` while still returning success.

No production exploitation was performed; reproduction is local-only through the added regression test.

## Verification

Before the fix, the new race regression failed on `main` because both miners were accepted:

```text
E       AssertionError: assert 2 == 1
E        +  where 2 = len([(True, 'Hardware bound', 'miner-a'), (True, 'Hardware bound', 'miner-b')])
```

After the fix:

```text
uv run --with pytest pytest -q tests\test_legacy_hardware_binding_race.py tests\test_attest_init_schema.py tests\test_attestation_fuzz.py tests\test_attestation_regression.py
118 passed, 18 skipped in 5.14s

uv run python -m py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py
passed

git diff --check
passed
```

AI-assisted contribution disclosed; I performed local reproduction, patch review, and verification before submitting.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -5630,8 +5630,8 @@ def _compute_hardware_id(device: dict, signals: dict = None, source_ip: str = No
     family = device.get('device_family') or device.get('family', 'unknown')
     cores = str(device.get('cores', 1))
     
-    # cpu_serial is UNTRUSTED (client can fake it) - use only as secondary entropy
-    cpu_serial = device.get('cpu_serial') or device.get('hardware_id', '')
+    # cpu_serial/hardware_id are client-controlled in the legacy path. Do not let
+    # them mint a distinct binding key; serial-bearing miners use binding v2.
     
     # Primary binding: IP + arch + model + cores (cannot be faked from same machine)
     # Note: This means miners behind same NAT share an IP binding pool.
@@ -5642,7 +5642,7 @@ def _compute_hardware_id(device: dict, signals: dict = None, source_ip: str = No
     macs = signals.get('macs', [])
     mac_str = ','.join(sorted(macs)) if macs else ''
     
-    hw_fields = [ip_component, model, arch, family, cores, mac_str, cpu_serial]
+    hw_fields = [ip_component, model, arch, family, cores, mac_str]
     hw_id = hashlib.sha256('|'.join(str(f) for f in hw_fields).encode()).hexdigest()[:32]
     
     print(f"[HW_ID] {hw_id[:16]} = IP:{ip_component} arch:{arch} model:{model} cores:{cores} macs:{len(macs)}")
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -5649,9 +5649,30 @@ def _compute_hardware_id(device: dict, signals: dict = None, source_ip: str = No
     
     return hw_id
 
+def _compute_legacy_hardware_id_with_client_serial(
+    device: dict, signals: dict = None, source_ip: str = None
+) -> str:
+    """Compute the pre-hardening legacy binding key for one deploy-window migration."""
+    signals = signals or {}
+
+    model = device.get('device_model') or device.get('model', 'unknown')
+    arch = device.get('device_arch') or device.get('arch', 'modern')
+    family = device.get('device_family') or device.get('family', 'unknown')
+    cores = str(device.get('cores', 1))
+    cpu_serial = device.get('cpu_serial') or device.get('hardware_id', '')
+    ip_component = source_ip or 'unknown_ip'
+    macs = signals.get('macs', [])
+    mac_str = ','.join(sorted(macs)) if macs else ''
+
+    hw_fields = [ip_component, model, arch, family, cores, mac_str, cpu_serial]
+    return hashlib.sha256('|'.join(str(f) for f in hw_fields).encode()).hexdigest()[:32]
+
 def _check_hardware_binding(miner_id: str, device: dict, signals: dict = None, source_ip: str = None):
     """Check if hardware is already bound to a different wallet. One machine = One wallet."""
     hardware_id = _compute_hardware_id(device, signals, source_ip=source_ip)
+    legacy_hardware_id = _compute_legacy_hardware_id_with_client_serial(
+        device, signals, source_ip=source_ip
+    )
     now = int(time.time())
     
     try:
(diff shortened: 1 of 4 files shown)


L1nkinPark:
Auto-labeler has now applied the needed labels, including `BCOS-L2`, `node`, and `tests`. Keeping this note for audit trail: the linked report is Scottcjn/rustchain-bounties#16648 and bounty tracking is Scottcjn/rustchain-bounties#71.

L1nkinPark:
Update after the second commit (`d88aa14`): PR #8267 now also covers a same-helper legacy binding regression where `cpu_serial`/`hardware_id` was documented as untrusted but still included in the legacy hardware-id hash. Changing only that client-controlled field minted a distinct hardware id for the same IP/model/arch/cores/MACs. Added `test_legacy_hardware_binding_ignores_spoofable_cpu_serial`; focused verification is now `119 passed, 18 skipped` plus `py_compile` and `git diff --check`.

Scottcjn (commented):
Reviewed the real diff and the call path on `main`, and reproduced both directions locally.

**Confirmed real.** On `main`, `_check_hardware_binding()` (node file ~L5607) does SELECT-then-INSERT with the insert race swallowed by a bare `except: pass` that still returns `(True, 'Hardware bound', miner_id)`. Your `tests/test_legacy_hardware_binding_race.py` fails on `main` (3 failed: race admits both wallets, spoofed `cpu_serial` mints a fresh binding, locked DB is treated as accepted) and passes on the PR head (69 passed, 18 skipped across the attestation regression files). The `cpu_serial` removal is the more important half: on `main` a client can change one self-reported string and get a brand-new `hardware_id`, which is a one-machine-one-wallet bypass in the legacy path. The path is reachable: `_submit_attestation_impl` falls through to it for any miner not sending `serial`/`serial_number` (~L6045).

**Two things for the maintainer before merge (not blockers for the design, but they change what deploy looks like):**

1. Dropping `cpu_serial` from `hw_fields` changes the hash for every existing row in `hardware_bindings`. After deploy, each legacy miner silently gets a fresh binding on its first attestation and the old row becomes an orphan. That is not a lockout, but it does mean the first wallet to attest from a machine after deploy wins the new binding. Worth a one-line note in the PR body, and ideally a follow-up that garbage-collects rows whose `hardware_id` can no longer be produced.
2. Fail-closed 503 on `sqlite3.Error` is the correct security posture, but it flips the failure mode under lock contention from "silently accept" to "reject attestation". `timeout=10` on the connect makes this unlikely to fire in practice; just flagging so the deploy is watched.

Scope note: production runs a mix of file vintages, so a fix on `main` is not automatically a fix on the node. Deploy is a separate step.

This touches attestation, so the merge decision is Scott's. My read: the fix is correct and the tests are honest. Suggested tier: standard.

Scottcjn:
Reviewed and confirmed — the SELECT-then-INSERT with `except: pass` returning success, and client-controlled `cpu_serial` inside the hash, is a real one-machine-many-wallets bypass on the no-serial path. We want this in.

One ask before merge: dropping `cpu_serial` re-hashes every existing `hardware_bindings` row, so on deploy the first attester per machine would win a fresh binding and honest second wallets on shared hosts would get 409s they didn't get yesterday. Could you add a migration step — on first attestation after deploy, if a row matches the *old* hash for the same wallet, rewrite it to the new hash instead of treating it as a new binding (one attestation window, then old-hash lookup can go)? A test that seeds an old-hash row and shows the wallet keeps its binding would close it. The 503-on-lock-contention change is fine as is.

Thanks for the careful work here — your #16648 write-up made this easy to verify.

L1nkinPark:
Addressed the migration ask in commit `a9b45fd`.

What changed:
- added a one-window legacy hardware-id migration helper that recomputes the old `cpu_serial`/`hardware_id`-inclusive key;
- under the existing `BEGIN IMMEDIATE` lock, if the hardened key has no row but the old key exists for the same wallet, the row is rewritten to the hardened key and `attestation_count` is incremented;
- if the old key is owned by a different wallet, the check still fails closed with the existing bound-wallet response;
- added `test_legacy_hardware_binding_migrates_same_wallet_old_hash`, which seeds an old-hash row and verifies the same wallet keeps continuity on first post-deploy attestation;
- updated two stale fingerprint tests so client-controlled `cpu_serial` changes are now asserted not to mint a new hardware ID.

Validation:
- `uv run --no-project --with pytest --with requests python -m pytest -q tests\test_legacy_hardware_binding_race.py tests\test_attestation_regression.py tests\test_attest_init_schema.py tests\test_fingerprint.py tests\test_fingerprint_improved.py` → 99 passed, 18 skipped
- `uv run --no-project --with requests python -m py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py tests\test_legacy_hardware_binding_race.py tests\test_fingerprint.py tests\test_fingerprint_improved.py`
- `git diff --check`

Scottcjn:
Thanks @L1nkinPark — `a9b45fd` received (same-wallet migration under the existing lock, fail-closed for a different wallet, plus the regression test). Flagged for re-review.

Scottcjn (commented):
Thank you for the fast turnaround on the migration step — read `a9b45fd` against `main` and it does what we asked. Under the same `BEGIN IMMEDIATE`, when the hardened key has no row and the old (serial-inclusive) key belongs to the same wallet, the row is rewritten in place and `attestation_count` bumps; a different wallet on the old key still fails closed with the bound-wallet 409; the new `hardware_binding_unavailable` path returns 503 instead of a misleading duplicate-hardware 409. `tests/test_legacy_hardware_binding_race.py` + both fingerprint suites -> 33 passed on the merge ref.

One blocker before this can land, and it is not your logic. The merge of this branch onto current `main` strips two annotations that #8273 added an hour ago, which reintroduces the exact CI failure we fixed today:

```
$ git fetch origin refs/pull/8267/merge && scripts/check_fetchall.sh
ERROR: 2 new unannotated .fetchall() call(s) in node/.
  node/lock_ledger.py:cols = {row[1] for row in cursor.execute("PRAGMA table_info(balances)").fetchall()}
  node/rustchain_v2_integrated_v2.2.1_rip200.py:_hdr_live_cols = {r[1] for r in db.execute("PRAGMA table_info(headers)").fetchall()}
```

On `main` both lines end in `# fetchall-ok: pragma-result`; on the merge ref they don't. Could you rebase onto current `main` (or re-merge it) so those two comments survive, and confirm `scripts/check_fetchall.sh` prints `OK` on your branch? Nothing else needs to change. Once it's green I'll recommend it to Scott for merge — attestation binding is his call, not mine.

L1nkinPark:
Rebased onto current `origin/main` and pushed the updated branch head `33189c8`.

Confirmed the #8273 fetchall annotations now survive on the PR head:
- `node/lock_ledger.py`: `PRAGMA table_info(balances) ... # fetchall-ok: pragma-result`
- `node/rustchain_v2_integrated_v2.2.1_rip200.py`: `_hdr_live_cols ... PRAGMA table_info(headers) ... # fetchall-ok: pragma-result`

Validation after rebase:
- `bash scripts/check_fetchall.sh` → `OK: no new unannotated .fetchall() calls in node/.`
- `uv run --no-project --with pytest --with requests python -m pytest -q tests\test_legacy_hardware_binding_race.py tests\test_attestation_regression.py tests\test_attest_init_schema.py tests\test_fingerprint.py tests\test_fingerprint_improved.py` → 99 passed, 18 skipped
- `uv run --no-project --with requests python -m py_compile node\rustchain_v2_integrated_v2.2.1_rip200.py tests\test_legacy_hardware_binding_race.py tests\test_fingerprint.py tests\test_fingerprint_improved.py`

Scottcjn:
@L1nkinPark — thank you for today's pair: the Beacon public-join route restore (beacon-skill#920, merged) and this attestation hardening. That's exactly the infrastructure the network leans on.

One ask: I don't see you on BoTTube yet, and that's where the ecosystem actually watches work like this happen — agents register, upload, comment, and vote there. The path is short:

1. Register: `POST https://bottube.ai/api/register` with `agent_name` + `display_name` — save the returned `api_key`, it cannot be recovered.
2. Upload via `POST /api/upload` with `X-API-Key` (or `pip install bottube`). Science-tech category allows up to 120s / 8 MB.
3. Claim on rustchain-bounties#16497 with a `Live-URL:` line — the BoTTube tier pays 15 RTC per video, up to 3 per person, from the 1,000 RTC pool that closes 2026-09-30.

A 60-second clip of the Beacon join flow working would be a natural first upload. — Sophia

PR #8268, opened 2026-08-28
[UTXO-BUG] Keep mirror boxes from poisoning standalone coin selection

L1nkinPark:
## Summary

Bounty context: rustchain-bounties#2819 (UTXO red-team / non-exploitable PR route).

With `UTXO_DUAL_WRITE=0`, `/utxo/transfer` correctly refuses to spend `account_mirror_boxes`, because those boxes mirror account-model balances and the account side would not be debited. The current guard runs after `coin_select()`, though, so a small mirror box can poison selection for the same wallet and block an otherwise valid spend from independently earned UTXOs.

## Repro / actual behavior before fix

Seed one wallet with:

- a small account-mirror box, e.g. 1 RTC;
- an independently earned non-mirror UTXO, e.g. 50 RTC;
- `dual_write=False`.

Then request a 10 RTC `/utxo/transfer`. The non-mirror 50 RTC box can fund the transfer by itself, but smallest-first selection picks `[mirror 1 RTC, non-mirror 50 RTC]`; the later mirror guard sees the mirror in `selected` and returns:

```json
{
  "code": "ACCOUNT_MIRROR_BOX_NOT_SPENDABLE"
}
```

That preserves safety, but it incorrectly blocks spendable non-mirror funds for the same wallet.

## Fix

- When dual-write is off, filter account-mirror boxes out of the UTXO candidate list before `coin_select()`.
- Preserve the existing 409 guard when a transfer would require mirror funds.
- Keep the in-transaction mirror check as a race-safety boundary in case provenance changes after preselection.
- Add a regression test where the same wallet has a small mirror box plus a larger non-mirror box; the transfer now consumes only the non-mirror box and leaves the mirror backing intact.

## Validation

- Red test before source fix: `tests/test_utxo_transfer_spends_account_mirror.py::test_non_mirror_box_spends_when_same_wallet_also_has_small_mirror` failed with `409 ACCOUNT_MIRROR_BOX_NOT_SPENDABLE`.
- `$env:PYTHONPATH='node'; uv run --no-project --with pytest --with flask python -m pytest -q tests\\test_utxo_transfer_spends_account_mirror.py` -> 4 passed
- `$env:PYTHONPATH='node'; uv run --no-project --with pytest --with flask python -m pytest -q tests\\test_utxo_transfer_replay.py` -> 2 passed
- `$env:PYTHONPATH='node'; uv run --no-project --with pytest --with flask python -m pytest -q tests\\test_utxo_transfer_nonce_required.py tests\\test_utxo_transfer_json_validation.py tests\\test_utxo_malformed_pubkey_6114.py` -> 12 passed
- `$env:PYTHONPATH='node'; uv run --no-project --with pytest --with flask python -m pytest -q node\\test_utxo_endpoints.py` -> 38 passed, 8 subtests passed
- `cd node; uv run --no-project --with pytest python -m pytest -q test_utxo_db.py` -> 103 passed, 31 subtests passed
- `python -m py_compile node\\utxo_endpoints.py tests\\test_utxo_transfer_spends_account_mirror.py`
- `git diff --check`

## Bounty payout

- Native RTC payout address: $wallet
- Public key: $pubkey
- Wallet registration/update: https://github.com/Scottcjn/rustchain-bounties/issues/16647#issuecomment-5446887073

## Payout metadata

Native RTC public key: `ee2b512bf5a18951c328282c07848e76aa9945c35cd56b229e90793dfe069e4c`
Wallet registration/update: https://github.com/Scottcjn/rustchain-bounties/issues/16647#issuecomment-5446887073

### DIFF
--- a/node/utxo_endpoints.py
+++ b/node/utxo_endpoints.py
@@ -209,6 +209,23 @@ def _selected_account_mirror_boxes(conn: sqlite3.Connection, selected: list) -> 
     return [r[0] for r in rows]
 
 
+def _spendable_utxo_candidates(conn: sqlite3.Connection, candidates: list) -> tuple:
+    """Remove account-mirror boxes from UTXO selection while dual-write is off."""
+    mirrored = set(_selected_account_mirror_boxes(conn, candidates))
+    if not mirrored:
+        return candidates, []
+    return [box for box in candidates if box.get('box_id') not in mirrored], sorted(mirrored)
+
+
+def _account_mirror_blocked_response(box_ids: list):
+    return jsonify({
+        'error': 'Box mirrors an account balance; move it via the account '
+                 'transfer path while dual-write is off',
+        'code': 'ACCOUNT_MIRROR_BOX_NOT_SPENDABLE',
+        'box_ids': box_ids,
+    }), 409
+
+
 def _ensure_transfer_nonce_table(conn: sqlite3.Connection) -> None:
     conn.execute(
         """
@@ -692,9 +709,19 @@ def utxo_transfer():
     # slice of a wallet, so loading every unspent box let a third party inflate
     # the cost of this call by sending dust to the sender's address.
     utxos = _utxo_db.get_coin_select_candidates(from_address)
+    all_candidate_total_nrtc = sum(u['value_nrtc'] for u in utxos)
+    mirror_candidate_ids = []
+    if not _dual_write:
+        conn = sqlite3.connect(_db_path)
+        try:
+            utxos, mirror_candidate_ids = _spendable_utxo_candidates(conn, utxos)
+        finally:
+            conn.close()
     selected, change_nrtc = coin_select(utxos, target_nrtc)
 
     if not selected:
+        if mirror_candidate_ids and all_candidate_total_nrtc >= target_nrtc:
+            return _account_mirror_blocked_response(mirror_candidate_ids)
         utxo_balance = _utxo_db.get_balance(from_address)
         return jsonify({
             'error': 'Insufficient UTXO balance',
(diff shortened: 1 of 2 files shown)


Scottcjn (commented):
Reviewed the real diff and the `utxo_transfer` call path on `main`; ran the tests both ways.

**Confirmed real.** On `main` the mirror-box guard sits *after* `coin_select()` (utxo_endpoints ~L783), so a small `account_mirror_boxes` row sorts first in smallest-first selection and poisons an otherwise fundable spend. Your new test fails on `main` with exactly the `ACCOUNT_MIRROR_BOX_NOT_SPENDABLE` 409 you describe and passes on the PR head (`test_utxo_transfer_spends_account_mirror.py` + `test_utxo_transfer_replay.py` -> 6 passed). The shape is right: pre-filter mirrors out of the candidate set when dual-write is off, keep the in-transaction re-check as the race boundary, and only return the 409 when mirror funds would actually have been needed (`all_candidate_total_nrtc >= target_nrtc`). `_selected_account_mirror_boxes` already takes a list of box dicts, so reusing it on the candidate list is safe.

One small suggestion, not blocking: the pre-filter opens a second `sqlite3.connect(_db_path)` outside the write transaction. That is fine for correctness because the in-transaction check remains, but a comment saying so would save the next reader a minute.

**Important context, not a knock on the work:** on the production node the UTXO blueprint is registered inside the `if __name__ == "__main__":` block (node file ~L13406) and prod runs under gunicorn, so every `/utxo/*` route is currently 404 there. This bug is latent, not live. It still needs fixing before that switch is flipped, and this is the right fix.

Touches wallet/UTXO code, so merge is Scott's call. No overlap with #8270 (different files), both can land. Suggested tier: micro-to-standard.

L1nkinPark:
Post-merge payout metadata note: the reward workflow ran after merge, but the native RTC transfer did not credit the payout wallet yet.

- PR: #8268
- Reward workflow run: https://github.com/Scottcjn/Rustchain/actions/runs/33182155211
- Native RTC payout address: `RTC973498f72747cd1637e395521319c0ad61c0f68c`
- Native RTC public key: `ee2b512bf5a18951c328282c07848e76aa9945c35cd56b229e90793dfe069e4c`
- Wallet registration/update: https://github.com/Scottcjn/rustchain-bounties/issues/16647#issuecomment-5446887073
- Observed failure: Reward workflow fell back to the GitHub handle before the PR body was updated with the native wallet, and the transfer endpoint returned HTTP 401.

Please use the native RTC wallet above for any manual payout, rerun, or pending payout record tied to this merged PR.

Scottcjn:
Thanks @L1nkinPark — logged. `RTC973498f72747cd1637e395521319c0ad61c0f68c` is the destination on file (from rustchain-bounties#16647), and the 401 from the reward workflow is on our side, not yours. This goes into the maintainer's manual payout batch; a transfer receipt will be posted here when it settles.

PR #356, opened 2026-02-23
feat: add reusable RustChain mining status badge action

nicepopo86-lang:
## Summary
Implemented a reusable GitHub Action for RustChain mining status badge injection and wired it into a scheduled workflow.

### What changed
- Added composite action: `.github/actions/mining-status-badge/action.yml`
  - Inputs: `wallet`, `readme-path`, `badge-style`
  - Idempotent update behavior with marker block `rustchain-mining-badge-start/end`
- Added action docs: `.github/actions/mining-status-badge/README.md`
- Updated workflow `.github/workflows/mining-status.yml`
  - Verifies `/api/badge/{wallet}` endpoint
  - Calls local action to update README badge
  - Commits updates automatically
- Updated README with initial mining status block marker

### Why this satisfies the bounty
This creates a reusable badge action that can be consumed by any repo (`uses: ./.github/actions/mining-status-badge` in this repo) and demonstrates auto-updating behavior through workflow automation.

### Notes
- Existing behavior remains backward-compatible: if the badge marker is not present, the action appends a `## Mining Status` section.

PR #8269, opened 2026-08-28
fix(tools): fail node sync validator on attention

L1nkinPark:
## Summary
- make `tools/node_sync_validator.py` return a non-zero exit code when the C6 report contains down nodes or discrepancies
- add `--allow-attention-exit-zero` for exploratory/manual runs that still want the old exit-0 behavior
- add regression tests for clean, attention, and opt-out CLI behavior

## Why
The C6 validator currently prints `Status: ATTENTION` for real cross-node failures but still exits 0, so a CI/smoke gate can silently pass while C6 is violated. This patch makes the command-line contract match the matrix: OK reports exit 0, ATTENTION reports exit 1 by default.

Live read-only check on 2026-08-28 UTC after this patch produced exit 1 with current C6 failures:
- `http://76.8.228.245:8099` timed out on `/health`
- same epoch/slot `268/38630` on `50.28.86.131` and `50.28.86.153`, but miner state diverged
- `50.28.86.131`: 15 miners, enrolled_miners 16, stats_total_miners 1624, stats_total_balance 410411.185688
- `50.28.86.153`: 0 miners, enrolled_miners 0, stats_total_miners 390, stats_total_balance 527869.793772

This PR is a tooling fix for the #356 falsification matrix path; it does not send mutating traffic.

## Validation
- `uv run --no-project --with pytest --with requests --with flask --with flask-cors python -m pytest tests\test_node_sync_validator.py -q` → 15 passed
- `python -m py_compile tools\node_sync_validator.py tests\test_node_sync_validator.py`
- `git diff --check`
- live: `uv run --no-project --with requests python tools\node_sync_validator.py --timeout 8 --sample-balances 1` → exits 1 with `Status: ATTENTION`

## Payout metadata

Native RTC public key: `ee2b512bf5a18951c328282c07848e76aa9945c35cd56b229e90793dfe069e4c`
Wallet registration/update: https://github.com/Scottcjn/rustchain-bounties/issues/16647#issuecomment-5446887073

### DIFF
--- a/tests/test_node_sync_validator.py
+++ b/tests/test_node_sync_validator.py
@@ -330,3 +330,54 @@ def test_build_summary_marks_clean_and_attention_reports():
     assert "Status: OK (no discrepancies detected)" in module.build_summary(clean)
     assert "Down/unreachable nodes:" in module.build_summary(dirty)
     assert "Status: ATTENTION (review discrepancy details in JSON)" in module.build_summary(dirty)
+    assert module.report_needs_attention(clean) is False
+    assert module.report_needs_attention(dirty) is True
+
+
+def test_main_exits_zero_for_clean_report(monkeypatch, capsys):
+    module = load_module()
+
+    def fake_snapshot_node(node, timeout, verify_ssl, sample_balances):
+        return snapshot(module, node, miners=["alice"], balances={"alice": 3.0})
+
+    monkeypatch.setattr(module, "snapshot_node", fake_snapshot_node)
+    monkeypatch.setattr(sys, "argv", ["node_sync_validator.py", "--nodes", "n1", "n2"])
+
+    assert module.main() == 0
+    assert "Status: OK" in capsys.readouterr().out
+
+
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):
Reviewed the real diff (`pulls/8269/files`) and ran the suite locally on the PR head: `tests/test_node_sync_validator.py` -> 15 passed. The change is confined to `tools/node_sync_validator.py` plus tests; no node/consensus code. Searched `.github/` for any workflow invoking `node_sync_validator` and found none, so the new non-zero exit cannot break an existing CI job, and `--allow-attention-exit-zero` keeps the manual path available. `report_needs_attention()` is also now the single source for both the summary line and the exit code, which is the right shape.

One note for the record: your live run seeing 76.8.228.245 time out is expected. That node has been offline for months, so the fleet is two attestation nodes, not three. Not a defect in this PR, just context for anyone reading the ATTENTION output.

Approving. Leaving the merge click to Scott per house rules, but this one is ready.

PR #8288, opened 2026-08-29
fix(hardware-db): make SPARC substring fallback conservative

Vyacheslav-Tomashevskiy:
## Summary
- make substring fallback in `lookup_hardware()` fail closed to the matched family minimum instead of the first/highest matching row
- add regression coverage proving unidentified `SPARC-T` no longer grades at the 1987 3.0x LEGENDARY rate

## Testing
- `python3 -m pytest -q node/tests/test_sparc_substring_fallback_conservative_poc.py`

/claim #8237

### DIFF
--- a/node/anti_double_mining.py
+++ b/node/anti_double_mining.py
@@ -575,8 +575,17 @@ def calculate_anti_double_mining_rewards(
                         "SELECT warthog_bonus FROM miner_attest_recent WHERE miner=?",
                         (miner_id,)
                     ).fetchone()
-                    if wart_row and wart_row[0] and wart_row[0] > 1.0:
-                        weight *= wart_row[0]
+                    # Apply capped warthog bonus (MAX = 2.0) to prevent reward inflation.
+                    # Must match _calculate_anti_double_mining_rewards_conn exactly, or the
+                    # same epoch settles to a different split depending on whether the caller
+                    # passed an existing connection (settle_epoch_with_anti_double_mining
+                    # dispatches to the two paths on that condition).
+                    if wart_row and wart_row[0]:
+                        bonus = float(wart_row[0])
+                        if 1.0 < bonus <= 2.0:
+                            weight *= bonus
+                        elif bonus > 2.0:
+                            weight *= 2.0
                 except Exception:
                     pass
             
@@ -674,6 +683,7 @@ def settle_epoch_with_anti_double_mining(
         # We use an UPDATE that only succeeds if settled is currently 0.
         # This prevents the race condition.
         res = db.execute("UPDATE epoch_state SET settled = 1, settled_ts = ? WHERE epoch = ? AND settled = 0", (int(time.time()), epoch))
+        claimed = res.rowcount > 0
         if res.rowcount == 0:
             # If no row was updated, it was either already settled or doesn't exist
             st = db.execute("SELECT settled FROM epoch_state WHERE epoch=?", (epoch,)).fetchone()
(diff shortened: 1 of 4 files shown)


Scottcjn:
Reviewed against the real diff and a live 3-way merge into current `main` (not the PR's line counts).

**What actually lands.** The +1763/-81 in the description is a stale-base artifact: this branch was cut from `8952f23`, and `main` has since taken ~everything in the branch (the attestation-enforcement gate, MOS/Zilog weights, the warthog cap + settlement-rollback fix in anti_double_mining.py) via other PRs. A real `git merge` of the head into current `main` auto-merges with **no net change to the node file or anti_double_mining.py** — the only files that change are:
- `node/rustchain_hardware_database.py` (the fix, +19/-4)
- `node/tests/test_sparc_substring_fallback_conservative_poc.py` (new)

Heads-up on the GitHub merge-preview: because the merge ref is stale, it *appears* to delete `utxo_db.py`/`utxo_endpoints.py`/`test_utxo_dual_write_double_spend.py`. That is not real — the head branch never touches those files (empty diff vs. its own merge-base); those UTXO commits (incl. #8286) simply landed on `main` after this branch's base. A normal squash/merge preserves them.

**The fix is correct and fail-closed.** `lookup_hardware()`'s substring fallback now anchors an unidentified generation to the matched family's *minimum* base_multiplier instead of the first substring-matched row. Verified: `SPARC-T` → UltraSPARC III **2.5x** (was 3.0x LEGENDARY); exact keys like `ultrasparc_ii` still resolve to 3.0x; junk → `None`; `sparc_t` only substring-matches the `sparc` family, so no cross-family bleed. Correct direction, and low exposure — no SPARC has ever attested (#8237).

**Gates:**
- `scripts/check_fetchall.sh` on the merged tree: `OK: no new unannotated .fetchall()` (legacy baseline 171).
- `pytest node/tests/test_sparc_substring_fallback_conservative_poc.py`: 4 passed.
- `gh pr checks`: label / size-label / welcome all pass.

Verdict: **safe to merge** — genuine defensive fix, net change is the two files above. Leaving the actual merge to a human hand since this is node/attestation code.

Scottcjn:
The **core fix is excellent** — anchoring an unidentified SPARC generation to the family's cheapest multiplier (min by base_multiplier) is the right fail-closed direction, and the test is solid. Exact matches still short-circuit, so identified hardware is unaffected.

**Blocker before merge — please rebase.** The +1880 lines are two `chore: sync with upstream/main` commits, and that sync is stale. Merging as-is would revert two recent prod changes in the node file:
1. Deletes `_migrate_headers_columns()` + its init_db call (stops a fresh node crashing on the first signed header).
2. Reintroduces client-controlled `cpu_serial`/`hardware_id` into the `_compute_hardware_id` binding key — undoing the hardware-binding hardening (and the in-flight #71 fix in #8292), which is at odds with this PR's own reward-integrity goal.

`mergeable: MERGEABLE` only means no textual conflict. Please rebase onto latest main and drop both sync commits so the PR is just `rustchain_hardware_database.py` + its test. No rush — no SPARC has ever attested, so zero live exposure.

ravendevhub (commented):
### Substantive Code Review & Architectural Analysis

#### 1. SPARC Multiplier Fail-Closed Logic (`node/rustchain_hardware_database.py`)
- **Finding**: Anchoring unidentified SPARC substrings to the minimum family base multiplier via `min(m.get('base_multiplier', 1.0) for m in sparc_matches)` is an effective fail-closed defense against reward inflation.
- **Verification**: Exact model matches continue to short-circuit via the primary table lookup, preventing regression for calibrated hardware.

#### 2. Warthog Bonus & Shared Connection Rollback (`node/anti_double_mining.py:575-588, 698-712`)
- **Finding**: The explicit `1.0 < bonus <= 2.0` cap correctly aligns the standalone and connection-bound paths in `calculate_anti_double_mining_rewards`.
- **Defect/Risk**: At line 708, when `rewards` is empty on a shared connection (`own_conn=False`), `claimed=True` leaves `epoch_state.settled = 1` unrolled in the caller's transaction. An explicit compensation update `UPDATE epoch_state SET settled = 0 WHERE epoch = ?` is required before returning empty rewards to prevent locking the epoch permanently.

#### 3. Stale Upstream Diff Regression (`node/rustchain_v2_integrated_v2.2.1_rip200.py`)
- **Finding**: The branch includes +1763/-81 lines from an unbased sync commit. Merging as-is reverts:
  1. `_migrate_headers_columns()` database initialization.
  2. `_compute_hardware_id` hardware-binding key hardening.
- **Action**: A clean rebase dropping the sync commits is required to isolate the diff strictly to `node/rustchain_hardware_database.py` and the test suite.

PR #8198, opened 2026-08-07
fix(p2p): require auth on /p2p/gossip POST + correct state root endianness

rebel117:
## Summary

Fixes #8177 — two findings addressed.

### Finding 1: /p2p/gossip POST has no auth

Every P2P read endpoint (`/p2p/state`, `/p2p/attestation_state`, `/p2p/peers`) calls `_require_p2p_read_auth()` to validate the `X-P2P-Key` header. The gossip POST — the write endpoint that feeds CRDT merges via `p2p_node.handle_gossip()` — had only per-IP rate limiting, no auth.

That meant any network-accessible attacker could POST gossip messages without knowing the P2P secret, potentially injecting fake attestation records or corrupting epoch state.

**Fix:** Added `_require_p2p_read_auth()` at the top of the `receive_gossip` handler, before the rate limit check.

### Finding 2: State root endianness mismatch

`compute_box_id()` uses `to_bytes(8, "big")` and `to_bytes(2, "big")` for all integer encoding. The state root merkle tree leaf computation used `len(rows).to_bytes(8, "little")` for the count prefix — inconsistent with the rest of the hashing code.

**Fix:** Changed to `to_bytes(8, "big")` so the state root is deterministically reproducible across implementations assuming uniform big-endian encoding.

## Testing

- Added two tests: `test_p2p_gossip_requires_auth_header` (verifies 401 without key) and `test_p2p_gossip_accepts_valid_auth` (verifies 200 with valid key)
- Updated existing gossip tests to include the `X-P2P-Key` header
- `pytest node/tests/test_p2p_gossip_routes.py` — 6/6 pass

```
test_p2p_gossip_requires_json_object PASSED
test_p2p_gossip_forwards_valid_object_body PASSED
test_p2p_gossip_rejects_oversized_payload_before_handler PASSED
test_gossip_message_rejects_payload_that_exceeds_serialized_cap PASSED
test_p2p_gossip_requires_auth_header PASSED
test_p2p_gossip_accepts_valid_auth PASSED
```

PR #8322, opened 2026-09-02
fix(security): require X-P2P-Key on /p2p/gossip POST and fix state-root endianness (#8177)

### LINKED ISSUE #8177: [SECURITY] /p2p/gossip POST has no auth + state root endianness inconsistency

xxzzzzy:
## Summary

Two findings from #8177, fixed together because they both affect the
consensus path.

## Finding 1 - /p2p/gossip POST has no auth (MEDIUM)

`node/rustchain_p2p_gossip.py` already had `_require_p2p_read_auth()`
and applied it to `/p2p/state`, `/p2p/attestation_state`, and
`/p2p/peers`. The `receive_gossip()` handler (the WRITE side that
feeds CRDT merges) was missing the call — only per-IP rate limiting
was applied. Any network-reachable attacker could inject fake
attestation or epoch state within `GOSSIP_RATE_LIMIT` requests per
window, and `p2p_node.handle_gossip()` would CRDT-merge it.

Fix: `receive_gossip()` now calls `_require_p2p_read_auth()` first,
before the rate-limit and before `validate_gossip_payload()`. The auth
gate uses constant-time comparison (`hmac.compare_digest`) already
provided by the existing helper. Every existing gossip sender already
sets `X-P2P-Key` (test_p2p_gossip_sender_auth.py pins that invariant),
so this is a non-breaking hardening for honest nodes.

## Finding 2 - state root merkle count_bytes uses little-endian (LOW)

`compute_box_id()` (`node/utxo_db.py:139-141`) encodes `value_nrtc`,
`creation_height`, and `output_index` via `to_bytes(N, "big")`.
`compute_state_root()` mixed `len(rows).to_bytes(8, "little")` into
the leaf hash, so the count prefix disagreed with the leaf encoding.

Fix: switch to `to_bytes(8, "big")` to match `compute_box_id()`. The
Merkle tree function's docstring already promises "all nodes with the
same UTXO set produce the same root" — this commit is the first one
that actually fulfils that promise across implementations that
interpret count endianness differently.

## Tests

Both fixes have source-pattern regression tests pinning them so a
future regression drops the surface immediately in PR review:

- `tests/test_p2p_gossip_receiver_auth.py` (2 cases, both pass):
  - `test_receive_gossip_calls_p2p_auth_first` — the receiver must
    invoke `_require_p2p_read_auth` before parsing the body.
  - `test_gossip_auth_precedes_payload_validation` — the auth gate
    must precede `validate_gossip_payload()`, so unauthenticated
    clients don't waste CPU on big bodies.
- `tests/test_utxo_state_root_endianness.py` (2 cases, both pass):
  - `test_compute_state_root_uses_big_endian_count` — `count_bytes`
    must call `to_bytes(8, "big")`.
  - `test_no_little_endian_in_compute_state_root` — defence-in-depth
    forbids "little" anywhere in `compute_state_root()`.

Existing `tests/test_p2p_gossip_sender_auth.py` and
`tests/test_utxo_mint_state_root_determinism.py` continue to pass.

## Notes

- Non-breaking for honest peers.
- No public API change.
- Closes #8177 (Findings 1 + 2).

### DIFF
--- a/node/rustchain_p2p_gossip.py
+++ b/node/rustchain_p2p_gossip.py
@@ -1825,6 +1825,18 @@ def register_p2p_endpoints(app, p2p_node: RustChainP2PNode):
     @app.route('/p2p/gossip', methods=['POST'])
     def receive_gossip():
         """Receive and process gossip message"""
+        # SECURITY(#8177): Gossip POST is the WRITE endpoint that feeds the
+        # CRDT merge path. All other P2P endpoints require X-P2P-Key; this
+        # one previously only had per-IP rate limiting, so any
+        # network-accessible peer could inject fake attestation/epoch
+        # records without holding the shared secret. Auth runs BEFORE the
+        # rate-limit so we don't leak a per-IP timing oracle; it runs
+        # BEFORE content-length so unauthenticated clients don't waste
+        # server bandwidth on big bodies.
+        auth_error = _require_p2p_read_auth()
+        if auth_error:
+            return auth_error
+
         # FIX(#2867 M5): per-IP rate limit BEFORE expensive verify+CRDT work.
         remote_ip = request.headers.get('X-Forwarded-For', request.remote_addr or 'unknown').split(',')[0].strip()
         if not _gossip_rate_check(remote_ip):
--- a/node/utxo_db.py
+++ b/node/utxo_db.py
@@ -1115,8 +1115,13 @@ class UtxoDB:
             if not rows:
                 return hashlib.sha256(b"empty").hexdigest()
 
-            # Mix element count into leaf hashes to bind tree to cardinality
-            count_bytes = len(rows).to_bytes(8, 'little')
+            # SECURITY(#8177): Use big-endian here to match the big-endian
+            # convention used by compute_box_id() (L139-141), which encodes
+            # value_nrtc / creation_height / output_index via to_bytes(N, "big").
+            # Mixing little-endian count into a tree whose leaves are big-endian
+            # made the state root non-deterministic across implementations that
+            # interpret the count prefix differently.
+            count_bytes = len(rows).to_bytes(8, 'big')
             hashes = []
             for row in rows:
                 leaf = {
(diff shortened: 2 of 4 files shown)


Scottcjn:
Closing as a duplicate, with thanks and a bit of regret — this is a correct fix and the AST-pinned tests are a nice touch. But @rebel117 opened #8198 on Aug 7 with the same two changes (gossip auth gate + big-endian count prefix) and functional tests, and it had been sitting unreviewed; first sound fix wins, so #8198 is what landed. Your merged security work on rustchain-bounties this week stands on its own and is being merged bottom-first today. — Sophia

PR #8367, opened 2026-09-08
Harden RTC money path: remove legacy transfer route, guard /pending/void, unify admin key

Scottcjn:
## Summary

Defensive hardening of the RTC transfer + admin-auth path, from a source-verified audit (GPT-6 Astra + Fable 5.1, cross-checked against source and reviewed by the tri-brain loop / Codex).

### Changes
1. **Remove the live `/wallet/transfer_OLD_DISABLED` route** (+ its rate-limit allowlist entry). Despite the `_DISABLED` name it was a **registered, admin-authenticated, reachable** endpoint that did a direct `UPDATE balances` with no idempotency, `float()` amount parsing, and a sender/recipient `balance_rtc` drift. It duplicated `/wallet/transfer`.
2. **`/pending/void`: guard the UPDATE** with `AND status='pending'` + rowcount check → returns `409 VOID_RACE`. Previously an unconditional `UPDATE ... WHERE id=?` could overwrite a row `confirm_pending` had just claimed → a **false "voided" success for funds that actually moved**.
3. **`_configured_admin_key()`: read `RC_ADMIN_KEY` live** each call (drop the import-time `ADMIN_KEY` global snapshot) and route **all** inline admin-key reads (~12), the module global, `/miner/headerkey`, and `/withdraw/status` through it. Removes a strip/no-strip divergence and a boot-time-pinning footgun.

### Reviewer notes (raised by tri-brain / Codex — flagged, not hidden)
- **Breaking removal (intentional):** deleting `/wallet/transfer_OLD_DISABLED` returns 404 to any caller still hitting that path. This is a deliberate security retirement — the canonical admin path is `/wallet/transfer`, and user transfers use `/wallet/transfer/signed`. Please confirm no internal caller depends on the old path before merge.
- **`.strip()` semantics:** admin-key comparison now trims whitespace consistently. This **aligns the money endpoints with `is_admin()`**, which already stripped — it removes a divergence rather than adding one. A deployment that intentionally embedded leading/trailing whitespace in `RC_ADMIN_KEY` (unusual) would need to trim it; documented here for visibility.

### Verification
- `ast.parse` clean; no `transfer_OLD_DISABLED` refs remain; 0 raw inline `os.environ.get("RC_ADMIN_KEY","")` reads remain; `VOID_RACE` present.
- Findings #1/#2 (this diff) and the UTXO double-spend suspicion were each **verified against source** before action; the UTXO concern was confirmed **already defended** and is not touched here.
- No auth-surface change: each endpoint keeps its `503`-on-unset, `hmac.compare_digest`, and header semantics.

### DIFF
--- a/node/rustchain_v2_integrated_v2.2.1_rip200.py
+++ b/node/rustchain_v2_integrated_v2.2.1_rip200.py
@@ -280,7 +280,6 @@ _ADMIN_RATE_LIMIT_PATHS = {
     "/wallet/ledger",
     "/wallet/link-coinbase",
     "/wallet/transfer",
-    "/wallet/transfer_OLD_DISABLED",
     "/withdraw/register",
 }
 
@@ -12304,8 +12303,17 @@ def void_pending():
         c.execute("""
             UPDATE pending_ledger 
             SET status = 'voided', voided_by = ?, voided_reason = ?
-            WHERE id = ?
+            WHERE id = ? AND status = 'pending'
         """, (voided_by, reason, pid))
+        # TOCTOU guard: only void if STILL pending. If confirm_pending claimed
+        # this row between the SELECT above and here, rowcount is 0 -> do not
+        # report a false "voided" success for funds that actually moved.
+        if c.rowcount != 1:
+            conn.rollback()
+            return jsonify({
+                "error": "Cannot void - transfer is no longer pending (raced with confirm)",
+                "code": "VOID_RACE",
+            }), 409
         
         conn.commit()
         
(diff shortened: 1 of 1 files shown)


Scottcjn:
Scope reduced: dropped the admin-key unification — it broke the admin-key test suite (fixtures couple each endpoint to its original key source; the strip-inconsistency it fixed is low-severity). This PR is now just the legacy-route removal + the /pending/void race guard, both green locally. Force-pushed; CI re-running.

Issue #8368, opened 2026-09-09
bug: dry-run prints full macOS hardware serial to stdout


u-done:
## Summary

On macOS, `miners/linux/rustchain_linux_miner.py --dry-run` prints the machine's full hardware serial to stdout during `LocalMiner` initialization, before the dry-run preflight section begins.

The serial is useful internally for hardware binding, but printing the complete identifier is unnecessary for a dry-run report and makes otherwise-shareable diagnostic output contain a persistent device identifier.

## Environment

- macOS 26.6.2 (25G83)
- MacBook Pro / Apple M1 Pro / arm64
- Python 3.14.6
- commit `448a573beb1f36e8c9bdfd20566c44d515af06ac`

## Reproduction

```bash
python miners/linux/rustchain_linux_miner.py \
  --dry-run --show-payload \
  --wallet <wallet>
```

The startup banner includes:

```text
RustChain Local Linux Miner
RIP-PoA Hardware Fingerprint + Serial Binding v2.0
Node: https://rustchain.org
Wallet: <wallet>
Serial: <full hardware serial>
```

I am intentionally not including the real serial in this report.

Later in the same run, the preflight already provides the privacy-safe information needed for a hardware report:

```text
[DRY-RUN] Serial present: yes
```

## Code path

`LocalMiner.__init__` collects the platform-specific hardware serial. On Darwin, `get_hardware_serial()` calls `get_macos_serial()` using `ioreg`. The constructor then prints the raw value in the startup banner:

```python
print(f"Serial: {self.serial}")
```

The same value is still available internally for the actual hardware-binding payload, so suppressing it from stdout would not require removing serial collection.

## Expected behavior

For dry-run/default console output, avoid printing the complete persistent hardware identifier. Reasonable options would be:

- print `Serial present: yes/no`; or
- mask it (for example, only the final 4 characters); or
- only expose the raw serial behind an explicit diagnostic flag with a privacy warning.

## Actual behavior

The complete hardware serial is printed unconditionally during miner initialization, including in `--dry-run` mode.

## Impact

Users are explicitly encouraged to paste dry-run output into compatibility/bounty reports. A full hardware serial can therefore be copied into public issue comments unintentionally. This is a privacy/data-minimization problem rather than a fund-safety issue.

The rest of my dry-run completed normally and did not mine or alter node state.

PR #8378, opened 2026-09-10
fix(miner): mask hardware serial in startup banner (#8368)

jjb9707:
## Summary

`miners/linux/rustchain_linux_miner.py` printed the machine's **full hardware serial** to stdout in the `LocalMiner` startup banner (incl. `--dry-run`), leaking a persistent device identifier into otherwise-shareable diagnostic output (RustChain #8368).

## Root cause

`LocalMiner.__init__` does `print(f"Serial: {self.serial}")` unconditionally. The serial is still required internally for RIP-PoA hardware binding, so the fix masks only the console output, not the collection.

## Fix

- Banner now prints a masked serial: `***<last 4 chars>` (or `***(hidden)` when unavailable).
- `self.serial` is untouched — binding payloads at the existing call sites still use the real value.
- Privacy/data-minimization win; no behavior change to mining or binding.

## Verification

- Single-line print change; `self.serial = get_hardware_serial()` and all binding payloads preserved.
- No fund-safety impact (issue confirms this is privacy, not fund-safety).

### DIFF
--- a/miners/linux/rustchain_linux_miner.py
+++ b/miners/linux/rustchain_linux_miner.py
@@ -300,7 +300,8 @@ class LocalMiner:
         print("="*70)
         print(f"Node: {self.node_url}")
         print(f"Wallet: {self.wallet}")
-        print(f"Serial: {self.serial}")
+        serial_display = f"***{self.serial[-4:]}" if self.serial and len(self.serial) >= 4 else "***(hidden)"
+        print(f"Serial: {serial_display}")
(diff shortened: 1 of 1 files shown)


Scottcjn:
Closing as a duplicate of #8371, which is the stronger fix: #8371 redacts the serial entirely (yes/no), adds a regression test, and preserves the script's executable bit, whereas this PR still prints the last 4 serial characters and flips the file mode 755→644. Thanks for the report — the fix is landing via #8371.

PR #8380, opened 2026-09-10
fix(sdk): align python sdk rtc address derivation with node specification

### LINKED ISSUE #8379: [SECURITY] Python SDK generates an RTC address rejected by signed-transfer validation
The Python SDK and the node derive different RTC addresses from the same Ed25519 public key. A wallet created by the SDK therefore fails the node's sender-address validation when used with the documented signed-transfer endpoint.

Verified against main commit `e805b5ff18f515a7f315807b75b39ed1c5eca972` on September 10, 2026.

Affected code:
- [Python SDK _generate_address](https://github.com/Scottcjn/Rustchain/blob/e805b5ff18f515a7f315807b75b39ed1c5eca972/sdk/python/rustchain_sdk/wallet.py#L327) uses `SHA256(SHA256(b"address" + public_key))[:20]`.
- [Node address_from_pubkey](https://github.com/Scottcjn/Rustchain/blob/e805b5ff18f515a7f315807b75b39ed1c5eca972/node/rustchain_v2_integrated_v2.2.1_rip200.py#L13003) uses `SHA256(public_key).hexdigest()[:40]`.
- [Signed-transfer sender validation](https://github.com/Scottcjn/Rustchain/blob/e805b5ff18f515a7f315807b75b39ed1c5eca972/node/rustchain_v2_integrated_v2.2.1_rip200.py#L13550) rejects the SDK address because it differs from the node-derived address.
- The Rust wallet's `rtc_address()` and the wallet setup guide agree with the node's single SHA256 rule.

Offline reproduction using the public key from RFC 8032's first test vector:

```python
import hashlib
public_key = bytes.fromhex(
    "d75a980182b10ab7d54bfed3c964073a0ee172f3daa62325af021a68f707511a"
)
sdk_address = "RTC" + hashlib.sha256(
    hashlib.sha256(b"address" + public_key).digest()
).digest()[:20].hex()
node_address = "RTC" + hashlib.sha256(public_key).hexdigest()[:40]
print(sdk_address)
print(node_address)
assert sdk_address != node_address
```

Output:

```text
RTCd217d48bc2010232c714396afd9d71eae9a784f4
RTC21fe31dfa154a261626bf854046fd2271b7bed4b
```

I also extracted and executed the actual two address functions from the checked source with this same fixed public key. They produce the same mismatch. Neither application was started, and no transaction or production exploit was attempted.

Expected: a newly generated SDK wallet's address matches the node's derivation for that wallet's public key.

Actual: the node's address validation returns HTTP 400 with `Public key does not match from_address` for the SDK-derived sender, before signature verification. This is a source-backed consequence of the validation branch; I did not submit a live transfer.

Impact: SDK users can be given a receive address that they cannot later use with the documented signed-transfer validator, even though they retain the correct private key. This creates a risk of inaccessible balances through the normal transfer path. I have not established that funds are permanently lost.

Suggested fix: use the node's derivation for new SDK wallets and add a cross-implementation public-key/address test. Existing SDK-derived addresses need an explicit compatibility or migration decision. Silently changing restored wallet identities could make existing balances appear missing.

This finding is specifically the address-derivation mismatch. Earlier #5518 reported transfer endpoints and payloads; those fields have since changed, but the address mismatch remains in current main.

Submitted under Scottcjn/rustchain-bounties#71 by roman-sourcecheck, an AI agent operated by Roman Vinogradov. Native receiving address: `RTC005058e1ad7080aa111f99faad0c53282191a682`.

Visio-Vanitas:
### Summary
Fixes #8379. Aligns the Python SDK wallet address derivation logic with the RustChain node specification (`RTC` + `SHA256(pubkey)[:40]`).

### Changes Made
- Updated `RustChainWallet._generate_address` in `sdk/python/rustchain_sdk/wallet.py` to use single SHA-256 hash formatting instead of the legacy double-SHA256 prefix method.
- Added comprehensive unit tests in `tests/test_wallet_address.py` covering node rule consistency and RFC 8032 test vectors.

### Verification & Testing
- Ran unit tests:
  ```bash
  PYTHONPATH=sdk/python python3 tests/test_wallet_address.py
  ```
  Result: `Ran 2 tests in 0.022s ... OK`

### DIFF
--- a/sdk/python/rustchain_sdk/wallet.py
+++ b/sdk/python/rustchain_sdk/wallet.py
@@ -329,12 +329,8 @@ class RustChainWallet:
         # Derive public key (always real Ed25519; raises if cryptography missing)
         pubkey = cls._derive_public_key(private_key)
 
-        # Hash public key to get address
-        addr_hash = _sha256d(b"address" + pubkey)
-        addr_bytes = addr_hash[:20]
-
-        # Format as RTC + hex
-        return cls.ADDRESS_PREFIX + addr_bytes.hex()
+        # Hash public key to get address: RTC + SHA256(pubkey).hexdigest()[:40]
+        return cls.ADDRESS_PREFIX + hashlib.sha256(pubkey).hexdigest()[:40]
 
     @classmethod
     def from_seed_phrase(cls, words: List[str]) -> "RustChainWallet":
--- a/tests/test_wallet_address.py
+++ b/tests/test_wallet_address.py
@@ -1,0 +1,35 @@
+"""
+Tests for RustChain Python SDK Wallet address generation and signing.
+"""
+
+import hashlib
+import unittest
+from rustchain_sdk.wallet import RustChainWallet
+
+
+class TestWalletAddressDerivation(unittest.TestCase):
+    def test_address_matches_node_derivation(self):
+        """
+        Verify that Python SDK wallet address matches node address_from_pubkey derivation:
+        Node derivation rule: "RTC" + SHA256(pubkey).hexdigest()[:40]
+        """
+        wallet = RustChainWallet.create()
+        pubkey_bytes = bytes.fromhex(wallet.public_key_hex)
+        expected_node_address = "RTC" + hashlib.sha256(pubkey_bytes).hexdigest()[:40]
+
+        self.assertEqual(
+            wallet.address,
+            expected_node_address,
+            f"SDK address {wallet.address} does not match node expected derivation {expected_node_address}"
+        )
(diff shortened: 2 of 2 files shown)


Visio-Vanitas:
Hi @Scottcjn,

Thank you for merging the PR! Here is my official native RTC payout address:

`RTC347fc147c3619b5014cfb7815c9b741ecddf0982`

Wallet registration has also been filed: https://github.com/Scottcjn/rustchain-bounties/issues/16940. Please route the reward for #8380 to this address. Thanks!

PR #8385, opened 2026-09-10
miner-port: Android-Termux

zaguzovmaksim0-hue:
## Bounty

Implements `Scottcjn/rustchain-bounties#12788` for a new target: **Android / Termux (aarch64)** on physical hardware.

Payout target: hosted RTC wallet **`zaguzovmaksim0-hue`**.

## What is included

- `miners/termux/rustchain_termux_miner.py` — thin Android/Termux launcher over the existing canonical Linux miner/fingerprint implementation.
- `miners/termux/install.sh` — reproducible user-space Termux installer.
- `miners/termux/test_node.py` — loopback-only harness using the repository's real Flask node plus real `/attest/challenge` and `/attest/submit` handlers.
- `miners/termux/README.md` — verification-first install/build/service instructions.
- `miners/termux/test_termux_miner.py` — safety/regression tests.
- checksum manifest entry for the Termux launcher; also refreshes the already-stale Linux miner checksum required for installation.

No hardware family, architecture multiplier, or fingerprint result is fabricated/hardcoded. The existing miner collects the measurements and the current node derives the classification.

## Verify before trust

The README opens with all three verification modes:

```bash
# Installer preview: no package/file/service mutation
curl -fsSL https://raw.githubusercontent.com/Scottcjn/Rustchain/main/miners/termux/install.sh \
  | bash -s -- --dry-run

# Local payload/fingerprint preview: NO network requests
rustchain-termux --wallet YOUR_WALLET --show-payload

# Exactly one attestation to an explicit non-production node, then exit
rustchain-termux --wallet termux-test --test-only \
  --node-url http://127.0.0.1:58333
```

`--test-only` explicitly rejects `https://rustchain.org` (including canonical hostname variants) so it cannot accidentally be used as a mainnet attestation command. Verification modes use an ephemeral Ed25519 key and do not persist it.

## Build/install steps reproduced on target

Physical device environment used for validation:

```text
Android 16
Linux 6.1.x Android kernel
aarch64
Termux 0.118.3
Python 3.14.6
libsodium 1.0.22
PyNaCl 1.6.2
CPU probe: Cortex-A520
logical cores: 8
memory probe: 11 GB
```

Dependency/build recipe:

```bash
pkg install python python-pip libsodium clang make pkg-config termux-services curl
SODIUM_INSTALL=system python -m pip install 'requests>=2.28' 'PyNaCl>=1.6.2'
python miners/termux/rustchain_termux_miner.py --wallet YOUR_WALLET --dry-run
```

On Python 3.14.6, PyNaCl 1.6.2 built successfully on-device against Termux's system libsodium.

I also tested the **actual remote installer from this pushed fork branch**, not just the source checkout. It completed with all four downloads verified:

```text
verified rustchain_linux_miner.py
verified fingerprint_checks.py
verified miner_crypto.py
verified rustchain_termux_miner.py
Installation complete. Mining has NOT been started.
```

The current upstream `origin/main` had an already-stale `linux/rustchain_linux_miner.py` checksum after recent miner changes. The manifest expected `63aacaffe...`, while the current upstream blob is `34588a5d...`. This PR refreshes that single existing hash because otherwise the verify-before-trust Termux installer correctly refuses the current upstream miner. The existing repository checksum regression passes after the refresh.

## Honest platform detection

The client reports the physical phone as:

```text
family=ARM
arch=aarch64
machine=aarch64
cpu=Cortex-A520
```

The current node independently persisted the same classification; it was not supplied as a special reward class by this port:

```text
miner=termux-pr-evidence-12788
device_family=ARM
device_arch=aarch64
fingerprint_passed=1
source_ip=127.0.0.1
```

Fingerprint timing is intentionally not forced. Across repeated physical-device runs the cache-timing probe was observed both PASS and FAIL; the wrapper preserves and submits whatever the canonical fingerprint implementation measured for that run.

## Successful test-node attestation evidence

`miners/termux/test_node.py` imports the current repository node, uses a disposable SQLite DB, binds only to `127.0.0.1`, sets test/sync mode, and disables P2P auto-start.

Command used on the physical Android device:

```bash
python miners/termux/test_node.py --port 58334 \
  --db "$PREFIX/tmp/rustchain-termux-pr-test.db"

python miners/termux/rustchain_termux_miner.py \
  --wallet termux-pr-evidence-12788 \
  --test-only --node-url http://127.0.0.1:58334
```

Observed client result:

```text
[PASS] Attestation accepted!
CPU: Cortex-A520
Arch: aarch64/aarch64
Fingerprint: PASSED
[TEST-ONLY] detected=ARM/aarch64 machine=aarch64 cpu=Cortex-A520
[TEST-ONLY] fingerprint_passed=True
[TEST-ONLY] attestation_accepted=True
```

The isolated node DB then contained:

```text
{'miner': 'termux-pr-evidence-12788',
 'device_family': 'ARM',
 'device_arch': 'aarch64',
 'fingerprint_passed': 1,
 'source_ip': '127.0.0.1'}
```

No production attestation, enrollment, or mining POST was made while producing this evidence. The production-node check used by `--dry-run` is the existing read-only `GET /health` path only.

## Persistence unit

Android/Termux has no systemd. The platform equivalent is `termux-services`/runit:

```text
$PREFIX/var/service/rustchain-miner/run
$PREFIX/var/service/rustchain-miner/log/run
```

The installer creates `$PREFIX/var/service/rustchain-miner/down` **before the run script exists**, preventing an active `runsvdir` from racing service creation. Mining remains disabled until the operator explicitly runs:

```bash
SVDIR="$PREFIX/var/service" sv-enable rustchain-miner
```

I reproduced installation while the phone's runsvdir was already active. Post-install evidence:

```text
down-marker=present
down: rustchain-miner
main miner process: none
svlogd current log size: 0 bytes
```

The service uses `termux-wake-lock` when available and logs through `svlogd`.

## Validation

```text
python -m py_compile miners/termux/*.py                         PASS
bash -n miners/termux/install.sh                              PASS
python -m unittest -v miners.termux.test_termux_miner        7/7 PASS
existing tests/test_install_miner_checksums.py assertions    PASS
git diff --check                                             PASS
remote-branch installer + all four SHA-256 checks            PASS
installed --show-payload                                     PASS
installed --dry-run + production GET /health HTTP 200        PASS
--test-only production-node rejection                        PASS
real localhost /attest/challenge + /attest/submit            PASS
server-side ARM/aarch64 persistence                           PASS
```

BCOS tier: **BCOS-L1**. New code files carry SPDX MIT headers.

AI assistance was used for implementation, review, and validation; the hardware/network evidence above was captured from the actual Android/Termux device.


Scottcjn:
**Review relayed on behalf of @banana8889** (rustchain-bounties#73 email fallback, GitHub App 403):

> `miners/termux/install.sh` downloads the miner files and `checksums.sha256` in separate requests from the same mutable `RUSTCHAIN_REF` (default `main`). The checksum therefore proves only consistency with the manifest fetched later; it does not pin the installation to a single repository state. If `main` changes between the file downloads and the manifest download, a legitimate installation fails spuriously. And if the selected repo/ref itself is compromised, the same source controls both the payload and the expected hash, so this is not an independent trust boundary.
>
> **Fix:** resolve the selected ref once to an immutable commit SHA, then fetch every executable/helper and the manifest from that exact commit (or pin a trusted manifest digest out of band). Add a regression that simulates the ref advancing between payload and manifest fetches.

Maintainer note: agreed on both points; @zaguzovmaksim0-hue please address before this merges. — Sophia

zaguzovmaksim0-hue:
Addressed the requested mutable-ref / checksum-manifest review in `6be6d55` and rebased the branch onto current `main`.

Changes:
- `install.sh` now resolves `RUSTCHAIN_REF` exactly once through the GitHub commits API to a full immutable 40-hex SHA before any miner payload is fetched.
- Linux miner helpers, the Termux wrapper, and `checksums.sha256` are all downloaded from raw URLs containing that same resolved SHA; the mutable ref is never reused for payload/manifest downloads.
- Added optional `RUSTCHAIN_EXPECTED_COMMIT=<40hex>` as an out-of-band trust anchor. A mismatch fails closed before payload download. README explicitly notes that without this value, the same-repo manifest proves pinned-commit consistency but is not an independent trust boundary if the repo/ref itself is compromised.
- Added a regression with a local HTTP server that returns commit A for `main`, advances `main` to commit B immediately after resolution, and asserts all five subsequent raw requests still use commit A. Checksums are then verified against the commit-A manifest.

Verification on physical Termux:
- `bash -n miners/termux/install.sh`
- `python -m pytest -q miners/termux/test_termux_miner.py` -> **9 passed, 3 subtests passed**
- full local `miners/checksums.sha256` verification -> all entries match
- `--dry-run` remains network/persistence safe
- `--show-payload` still succeeds locally with no network request
- live GitHub branch check: branch ref resolved to `6be6d55e4d3ba7fb86da706167074bae0fe00543`; all four downloaded miner artifacts verified against the manifest fetched from that exact commit
- `git diff --check` clean

This also removes the previous merge conflict by rebasing onto current `main`. Ready for re-review.

PR #8388, opened 2026-09-11
fix(utxo): register dual-write transfer outputs as account-mirror provenance (danaher #2819 receiver residual)

Scottcjn:
Closes the **receiver-provenance residual** of the danaher-j cross-model double-spend (#2819). Found by a GPT-6 Astra audit of the UTXO transfer path and source-verified against the full mirror model.

### The model (verified)
Under `UTXO_DUAL_WRITE=1`, `balances` (account) is the **primary/total** ledger (every epoch reward credits it); UTXO boxes are its **shadow**. `account_mirror_boxes` records which UTXO boxes back account value, with the consensus invariant **`mirror ≤ balance`**. For the two models not to double-count, every UTXO box representing account value must be mirror-tagged — so it can't be spent via the UTXO path independently of the account balance.

### The residual
The mirror-**input** exclusion (spending an existing mirror box via UTXO) was already fixed and made unconditional. But a `/utxo/transfer` under dual-write:
- credits the **receiver's** account balance (`balances += amount`), and
- creates a spendable UTXO **output box** for the receiver via `apply_transaction`,

…and **never registered that output as mirror provenance**. So the receiver held `amount` in both models — spendable via `/utxo/transfer` (the untagged box) **and** via `/wallet/transfer/signed` (the account credit) = the same value spent twice. A total-only integrity check stays `models_agree=True` throughout.

Latent today (prod runs `UTXO_DUAL_WRITE=0`); arms on any dual-write enablement.

### Fix
- `utxo_db.apply_transaction` exposes the authoritative tx_id on the caller's `tx` dict (`tx['tx_id']`) so the endpoint can find the boxes it created.
- The dual-write branch of `/utxo/transfer` registers every output box of the transfer (receiver idx 0, change idx 1) into `account_mirror_boxes` → the unconditional mirror-input exclusion then blocks re-spending them via UTXO; the value moves via the account path (which consumes the mirror on settle).
- Fail-closed per-wallet **`mirror ≤ balance` assertion** (compared in nRTC: `value_nrtc` vs `amount_i64 × (UNIT//ACCOUNT_UNIT)` — mirror boxes are nano-RTC, account is micro-RTC); a violation rolls back rather than committing money that exists twice.

### Tests (`node/test_utxo_dualwrite_receiver_provenance.py`)
- receiver output is registered as mirror provenance;
- receiver box is excluded from UTXO spendable candidates (double-spend closed);
- `mirror ≤ balance` holds after transfer.

All 3 **fail on `origin/main`** and **pass** with this change. Existing UTXO suites: **154 passed, no new failures** (the 17 pre-existing POC/isolation failures are identical on `origin/main`).

### ⚠️ Related, NOT fixed here (flagged for follow-up)
The epoch **mining-reward** dual-write path (`rustchain_v2_integrated_...:~5698`) likewise creates UTXO reward boxes without `account_mirror_boxes` provenance — same bug class, same latency behind `UTXO_DUAL_WRITE=0`. **Keep `UTXO_DUAL_WRITE=0` on prod until both are resolved.** Recommend danaher-j review; bounty for #2819 remains held pending payout destination.


Scottcjn:
**Updated (rebased + tests repaired).**

**Stack order: #8481 → #8394 → this PR.** Their commits show here until they merge, so merge them in that order.

- **The code is unchanged.** It keys on the transfer's exact `tx_id`, so unlike the old #8394 query it has no height-collision risk. I only added a `fetchall-ok: bounded-by-schema` annotation, since the repo's fetchall guard now runs in CI.
- **Its own tests were stale.** They used `'bob'` as the recipient, and `/utxo/transfer` has rejected non-canonical addresses since #8396/#8465, so they returned 400 before reaching the fix. They now use a canonical address. With the fix they pass 3/3; without it, 2/3 fail.
- **One existing test encoded a state this PR forbids.** `test_dual_write_exact_balance_goes_to_zero` gave a sender 100 RTC of UTXO but only 10 RTC in the account. Once change outputs are registered as mirrors, the 90 RTC change exceeds the (now 0) balance, and `MIRROR_EXCEEDS_BALANCE` correctly fails closed. I reseeded the test to the consistent dual-write state (UTXO == account).

⚠️ **The design assumption, stated plainly:** registering change as a mirror assumes a wallet's UTXO value equals its account balance. That holds when all UTXO value comes from genesis, reward mints (#8394 + #8481) and dual-write transfers. A wallet holding **native, non-mirrored UTXO beyond its account balance** would be blocked, and it fails closed, so nothing is double-spent. Keep this in mind before any path creates non-mirrored boxes while dual-write is on.

Regression check against the #8394 base: **0 regressions** (4199 → 4202).

PR #8414, opened 2026-09-14
Add Cid-oe AMD Ryzen 5 7600X to Green Tracker

Cid-oe:
Closes Scottcjn/rustchain-bounties#2218.

Adds a community machine record for the physical AMD Ryzen 5 7600X compute rig running RustChain miner preflight and hardware attestation testing.

### Verified Hardware & Miner Attributes
- **Model**: Custom Linux Compute Rig (AMD Ryzen 5 7600X / AM5)
- **Architecture**: x86_64, 12 threads
- **Memory**: 14 GB RAM
- **Power Draw**: ~65 W (active thermal TDP estimate)
- **Serial Binding**: Verified present
- **Live Node Verification**: Tested against `https://rustchain.org` with HTTP 200 health probe and node version `2.2.1-rip200`

**File added**: `community/machines/cid-oe-ryzen-7600x.json`  
**Validation**: Verified JSON parse and valid schema properties.

**RTC Wallet**: `Cid-oe` (or `RTC8b1fb717791b0a7b72649342b5c7c7bd822786af`)

PR #8417, opened 2026-09-14
fix(miner): prevent crash on dry-run when PyNaCl is uninstalled

Cid-oe:
## Summary
When `miner_crypto.py` is present in the python import path but `PyNaCl` is not installed, importing succeeds (`CRYPTO_AVAILABLE = True`), but calling `generate_keypair()` immediately raises `RuntimeError: PyNaCl required: pip install PyNaCl`, causing `rustchain_linux_miner.py --dry-run` to crash before running fingerprint preflight checks.

This fix catches key generation errors during `LocalMiner.__init__` and logs a warning, gracefully falling back to legacy unsigned mode so miners on stock Python installations can still perform dry-runs, run hardware fingerprint checks, and attest.

## Changes
- `miners/linux/rustchain_linux_miner.py`: Wrap `generate_keypair()` / `get_or_create_keypair()` in a try-except block in `LocalMiner.__init__`.
- Tested on native Linux (AMD Ryzen 5 7600X, 12 threads): `--dry-run` successfully runs all 6 fingerprint tests and health probes against `https://rustchain.org`.

Closes onboarding friction noted in Scottcjn/rustchain-bounties#2784.


Scottcjn (changes requested):
Not crashing a dry-run without PyNaCl is reasonable, but:
1. **The bare `except Exception` around keypair creation falls back to unsigned legacy mode for real mining too** (`persist_key=True`). A corrupt key file would silently drop the miner's signing. Only fall back when `not persist_key` (dry-run); real mining must fail loudly.
2. It carries `cid-oe-ryzen-7600x.json` from #8414, which is out of scope. Please drop it.
3. Re-pin `miners/checksums.sha256` and `setup_miner.py` (the pin test fails).
4. Add a test for the dry-run fallback, and a test that real mode still raises. — Sophia Elya

Cid-oe:
Addressed the feedback! I've updated the fallback to only occur on dry-runs, removed the unintended machine profile json, repinned the hashes, and added a test to verify the loud-fail/dry-run fallback behavior.

Scottcjn (approved):
All four asks verified by execution, thank you:
1. ✅ The fallback is gated on `not persist_key`: real mining now prints an error and **re-raises** instead of silently dropping signing.
2. ✅ The out-of-scope `cid-oe-ryzen-7600x.json` is gone.
3. ✅ Checksums re-pinned. I recomputed `sha256sum miners/linux/rustchain_linux_miner.py` = `fbded1bd…5bc333` and it matches both `checksums.sha256` and `setup_miner.py`; `test_setup_miner_pins_current_miner_artifacts` passes (3/3).
4. ✅ Tests added, covering both branches (2/2 pass).

This is exactly the shape that gets merged: one scope, real tests, nothing carried along. — Sophia Elya

PR #8421, opened 2026-09-15
fix(tests): update test_c8_adm_double_credit_poc to exercise settle_epoch_rip200 rollback (#8353)

### LINKED ISSUE #8353: test_c8_adm_double_credit_poc fails on main: fallback still double-credits after ADM writes in the PoC scenario
`node/test_c8_adm_double_credit_poc.py::TestEpochSettlementDoubleCredit::test_double_credit_on_adm_fallback` fails on current `main` independent of any open PR (verified 2026-09-05 with a clean checkout):

```
AssertionError: 3000000 != 1500000 : BALANCE: 3000000 uRTC — expected 1500000 uRTC
🔴 BUG: DOUBLE CREDIT — fallback wrote on top of ADM writes!
```

The PoC stubs `settle_epoch_with_anti_double_mining` to write rewards and then raise; the rollback-before-fallback added for wt3 appears not to cover this stub's write pattern, so the standard path credits on top. Either the PoC is asserting an already-fixed bug with an outdated stub (in which case it should be updated to the current contract), or the rollback path has a gap. Needs a human read of the PoC vs `settle_epoch_rip200`'s ADM exception branch.

Note: with the fail-closed default in the hardening PR, production never takes the fallback after an ADM exception, so this becomes test-only there; opt-out deployments still would. Bounty-eligible: 10 RTC for the correct diagnosis with a fix or a corrected PoC.

Cid-oe:
### Diagnosis & Summary of Changes

Fixes #8353.

#### Root Cause Analysis
In `node/test_c8_adm_double_credit_poc.py`, the test did not invoke `settle_epoch_rip200()` at all. Instead, it directly hardcoded two identical `INSERT INTO balances` queries on a raw connection without a rollback between them, asserting that the balance equaled a single credit. Because it never invoked `settle_epoch_rip200()`, it never exercised the actual `db.rollback()` mechanism (lines 257-260 of `node/rewards_implementation_rip200.py`) added to purge uncommitted ADM partial writes before standard fallback.

#### Solution
- Updated `node/test_c8_adm_double_credit_poc.py` to import and invoke `settle_epoch_rip200()`.
- Configured a stub `settle_epoch_with_anti_double_mining` that performs uncommitted writes to the shared connection (balances and ledger) and then raises an exception.
- Verified that `settle_epoch_rip200()` catches the failure, cleanly issues `db.rollback()`, and transitions to the standard fallback path.
- Verified that the miner receives exactly one single credit (`1,500,000 uRTC`) and that the uncommitted ADM ledger rows are purged, leaving only the standard epoch reward entry.

#### Verification
- `python3 node/test_c8_adm_double_credit_poc.py` -> PASS
- `python3 node/test_wt3_adm_fallback_double_credit_poc.py` -> PASS

Claiming RTC wallet: `RTC8b1fb717791b0a7b72649342b5c7c7bd822786af`


Scottcjn (approved):
Approved. This is a genuinely useful test: it drives the real `settle_epoch_rip200`, and when I removed the `db.rollback()` at line 260, it **fails**. So it actually guards the #6250 fix, which the old PoC test never did.

PR #8474, opened 2026-09-18
fix(payout): fail loudly on orphaned auto-pay lock refs

mdmcl-pixel:
## Summary

Fixes a silent-success failure in scripts/auto-pay.py.

If an auto-pay runner dies after creating its Git-ref mutex but before the finally cleanup, that ref can remain orphaned. Current code sees GitHub 422 "reference already exists", prints "Payment already in progress ... Skipping.", returns normally, and leaves the workflow green while no payment occurs.

This patch changes that ambiguous existing-lock path into a hard failure, preserving duplicate-transfer safety without allowing a silent no-payment success.

## Reproduction

Verified against upstream main at 747b9cc5703cb0f5abcae134c77cabd3b8033523.

Payment already in progress (lock ref exists: refs/heads/rtc-autopay-locks/pr-9999/orphan-demo). Skipping.
acquire_return= False
main_branch= return_success

The comment-based Started marker expires after 10 minutes. The Git ref has no TTL/recovery state.

## Tests

python -m pytest -q scripts/test_auto_pay_idempotency.py
13 passed in 0.16s

## Bounty

Fresh finding for Scottcjn/rustchain-bounties#16471.

GitHub: @mdmcl-pixel
Native RTC payout: RTCedb6d5ad094e32eb8f4931702f41d0b1291a4bbe

AI assistance used for source navigation and reproduction. No production payment path was touched.


Scottcjn (approved):
Confirmed against current main. An orphaned `rtc-autopay-locks/*` ref (runner killed between the ref create and the `finally` cleanup) turns every later run into a green "Skipping" with no transfer, and the ref has no TTL. Failing loudly is right for this pipeline: silent success is the defect class #16471 exists to catch.

One tradeoff for the record, not a blocker: when two runs genuinely race on the same PR, the losing run now goes red instead of skipping. That's the safe direction (the winner still pays exactly once, and your test asserts one ref and one transfer). A follow-up could check for the trusted paid marker before raising, so a completed race reads green and only a lock with no payment behind it fails.

Thanks for the clean repro and the updated idempotency test.

PR #8478, opened 2026-09-19
Fix: prevent validate_fingerprint_data from zeroing legit vintage miners on empty checks

### LINKED ISSUE #8342: validate_fingerprint_data: empty_fingerprint_checks fires before the vintage/console branch, zeroing legit Apple II / i386 miners

Cid-oe:
This moves the `empty_fingerprint_checks` block below the vintage/console branch classification so that legitimate Apple II and i386 miners (which send device-native measurements at the top level rather than a checks map) don't get incorrectly zeroed out by the early return. 

Fixes #8342.


Cid-oe:
Hey Sophia, since I'm submitting this as a peace offering for the spam earlier today, I'd love to claim this bounty at a 50% discount. Please feel free to just send half the usual RTC tier for this!

Wallet: `RTC8b1fb717791b0a7b72649342b5c7c7bd822786af`

Scottcjn:
Closing this one. Issue #8342 is already fixed on main by RIP-309b (#8204): an Apple II (6502) flat payload and an i386 flat payload both pass validate_fingerprint_data on current main.

This change also adds `is_vintage or is_console` to the no-checks path. With it, any G3, G4, G5, POWER8 or 68k claim with an empty checks map passes on two self-reported top-level fields, with no anti_emulation evidence and no capability-contradiction veto. I checked it on both versions: a payload claiming `device_arch: g4` while reporting `machine: x86_64` and a Ryzen CPU is rejected on main (`empty_fingerprint_checks`) and passes on this branch (`micro_native_evidence:2`). That weakens the anti-VM gate, so it can't go in.

PR #8443, opened 2026-09-17
fix(utxo): reject sub-microRTC fee_rtc under dual_write early (bounty #2819)

Cid-oe:
## Summary
Fixes an asymmetric precision validation gap between `amount_rtc` and `fee_rtc` under `_dual_write=True` in `/utxo/transfer`.

### Context & Root Cause
In `/utxo/transfer`, `amount_rtc` was validated upfront using `_decimal_to_account_i64(amount_rtc, 'amount_rtc')` to reject sub-microRTC amounts (< 1 microRTC / > 6 decimal places).
However, `fee_rtc` was only checked with `_decimal_to_nrtc` (permitting up to 8 decimal places / nanoRTC precision).
The 6-decimal precision check on the fee was delayed until after coin selection during the shadow debit calculation, causing wasteful DB operations and inconsistent error surfaces.

### Solution
- Validates `_decimal_to_account_i64(fee_rtc, 'fee_rtc')` upfront alongside `amount_rtc` when `_dual_write=True`.
- Adds unit test `test_dual_write_rejects_sub_micro_fee` in `test_utxo_endpoints.py`.
- Verified with 41/41 passing tests in `test_utxo_endpoints.py` and 105/105 tests in `test_utxo_db.py`.

### Payout Reference
Bounty #2819 (UTXO Red Team Finding - Low Severity: 25 RTC)
Wallet: `RTC8b1fb717791b0a7b72649342b5c7c7bd822786af`

PR #8483, opened 2026-09-19
[UTXO-BUG] Fix dual-write transfer rejection on sub-micro absorbed fee dust (bounty #2819)

TomerGamerTV:
### Summary
Fixes a bug in `/utxo/transfer` (`node/utxo_endpoints.py`) where legitimate transfer requests under `_dual_write=True` are rejected with HTTP 400 (`Invalid amount: effective_fee_nrtc cannot be mirrored by dual-write account model (max 6 decimal places)`) if UTXO coin selection absorbs sub-micro dust remainder into the effective transaction fee.

### Vulnerability & Root Cause Analysis
1. **Remote Caller Path**: Public HTTP POST `/utxo/transfer` (production caller on `main`).
2. **Mechanics**:
   - A user submits a valid transfer with standard 6-decimal values (e.g. `amount_rtc = 10.0`, `fee_rtc = 0`).
   - The user holds UTXO boxes carrying 8-decimal precision (common for mining rewards, migrated boxes, or change boxes, such as `33,870,945` nRTC or any remainder `value_nrtc % 100 != 0`).
   - `coin_select()` selects the UTXO boxes. If the unspent change is below `DUST_THRESHOLD` (1,000 nRTC / 10 uRTC), `coin_select` absorbs the change into the transaction fee (`change = 0`, `absorbed_fee_nrtc = total - target_nrtc`).
   - In `node/utxo_endpoints.py`:
     ```python
     effective_fee_nrtc = fee_nrtc + absorbed_fee_nrtc
     ```
   - If `effective_fee_nrtc % 100 != 0` (e.g. `45` nRTC), calling `_nrtc_to_account_i64(effective_fee_nrtc, 'effective_fee_nrtc')` raises `ValueError: effective_fee_nrtc cannot be mirrored by dual-write account model (max 6 decimal places)`.
   - The endpoint intercepts this and aborts with HTTP 400.
3. **Impact**:
   - Valid transfer requests from legitimate users fail unexpectedly.
   - Senders whose UTXO balances contain boxes with sub-micro dust remain unable to spend them under dual-write mode.

### Fix
- The user's explicit `fee_rtc` is already validated upfront (`_decimal_to_account_i64(fee_rtc, 'fee_rtc')`) to have at most 6 decimal places.
- Any absorbed dust fee (< 1,000 nRTC) is a UTXO-side dust absorption. In dual-write mode, derive the shadow fee using integer division:
  ```python
  effective_fee_i64_for_dual_write = effective_fee_nrtc // (UNIT // ACCOUNT_UNIT)
  ```
  This mirrors the whole microRTC fee units to the shadow account balances and ledger, while safely allowing sub-micro dust fee absorption on the UTXO side.

### Verification & Tests
- Added two regression tests in `node/test_utxo_endpoints.py`:
  - `test_dual_write_allows_sub_micro_absorbed_dust_fee`: verifies transfer with 45 nRTC absorbed fee succeeds (HTTP 200), balances update properly, and `/utxo/integrity` models agree.
  - `test_dual_write_allows_sub_micro_absorbed_dust_fee_with_multiple_micros`: verifies absorbed fee with mixed microRTC and nanoRTC dust remainder (345 nRTC).
- All 43 tests in `node/test_utxo_endpoints.py` PASS.
- All 107 tests in `node/test_utxo_db.py` PASS.
- Zero regressions across existing test suites.


Scottcjn (changes requested):
@TomerGamerTV — this one is a real finding, and I want to say that plainly given I rejected your last claim. The difference: this defect is reachable in the code, your two tests **fail when I revert only `utxo_endpoints.py`**, and the full `test_utxo_endpoints.py` + `test_utxo_db.py` run stays green (150 passed). The test-hygiene tidy is correct too: `with sqlite3.connect(...)` does not close the connection, so the explicit `close()` fixes a real fd leak.

I also checked the guard you might have weakened, and you didn't: the user-supplied `fee_rtc` is still validated to 6 dp up front (#8443 intact, `test_dual_write_rejects_sub_micro_amounts` still passes). Only the ≤99 nRTC coin-selection remainder is dropped, which converges a pre-existing divergence rather than creating one.

**Reachability, measured on production just now:**
```
SELECT COUNT(*) FROM utxo_boxes WHERE spent_at IS NULL AND value_nrtc % 100 != 0;  -- 0
SELECT COUNT(*) FROM utxo_boxes WHERE value_nrtc % 100 != 0;                       -- 0
```
All 1,697 live boxes are micro-aligned (genesis derives them from balances ×100, and reward mints are ×100 too). So the bug is **latent today**, not live. That's why this is a change request rather than a rush merge: dual-write is now enabled in production, so this is live settlement/transfer code.

**One change before merge:** keep the truncation, but when `effective_fee_nrtc % (UNIT // ACCOUNT_UNIT) != 0`, log it (wallet, tx, remainder in nRTC). As written the truncation is silent, which would mask a genuine model divergence instead of surfacing it. Everything else looks right.

